PR #106, opened 2020-10-15
fix: use name under /dev/mapper as device for LVM volumes for disk plugin

haojhcwa:
# Description of the issue
gopsutil upgrade results in the "Device" dimension changed from the name
of the symbolic link under /dev/mapper to the actual device by resolving
the symbolic link. This is a breaking for customers thus we added the
original path under /dev/mapper back as "Source" in gopsutil in pull
request:
https://github.com/shirou/gopsutil/pull/957

And make use of this "Source" when it is available in the aws/Telegraf
fork:
https://github.com/aws/telegraf/commit/f7742938b37806a4b1d98f81f3fb5bf50b9bcb3f

# Description of changes
In this commit, we pick up the modified gopsutil and telegraf for the
fix, and would revert to later versions when the changes are upstreamed.

# License

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,7 +2,8 @@ module github.com/aws/amazon-cloudwatch-agent
 
 go 1.13
 
-replace github.com/influxdata/telegraf => github.com/aws/telegraf v0.10.2-0.20200902215110-5ec6811a19d9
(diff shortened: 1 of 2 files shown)


pingleig (approved):
LGTM

PR #110, opened 2020-10-23
Register prometheus config rule for Windows OS; Prometheus EMF metrics unit support

zhenweig:
# Description of the issue
_Register prometheus config rule for Windows OS; Prometheus EMF metrics unit support_

# Description of changes
_Register prometheus config rule for Windows OS; Prometheus EMF metrics unit support_

# License

# Tests
_make release; test local build rpm on linux instance, local exe builds on Windows instance_

### DIFF
--- a/plugins/processors/emfProcessor/emfProcessor.go
+++ b/plugins/processors/emfProcessor/emfProcessor.go
@@ -83,7 +83,7 @@ func (e *EmfProcessor) Apply(in ...telegraf.Metric) (result []telegraf.Metric) {
 			if _, ok := supportedUnits[v]; ok {
 				e.validMetricUnit[k] = v
 			} else {
-				log.Println("E! detect unsupported unit: ", v)
+				log.Println("W! detect unsupported unit: ", v)
 			}
 		}
 
--- a/plugins/processors/emfProcessor/emfProcessor.go
+++ b/plugins/processors/emfProcessor/emfProcessor.go
@@ -7,6 +7,7 @@ import (
 	"github.com/aws/amazon-cloudwatch-agent/internal/structuredlogscommon"
 	"github.com/influxdata/telegraf"
 	"github.com/influxdata/telegraf/plugins/processors"
+	"log"
 )
 
 type EmfProcessor struct {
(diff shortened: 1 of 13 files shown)


translator/translate/logs/metrics_collected/prometheus/emfprocessor/ruleMetricUnit.go
```diff
@@ -0,0 +1,22 @@
+package emfprocessor
```
pxaws, line 1: Please add license headers in new source code files like this: https://github.com/aws/amazon-cloudwatch-agent/blob/a2da7b1e4ab0791d9608b1ec8a46ed8849a901d6/translator/translate/logs/metrics_collected/prometheus/prometheus.go#L1-L3

zhenweig: Addressed in new commit.

plugins/processors/emfProcessor/emfProcessor_test.go
```diff
@@ -45,7 +53,7 @@ func buildEMFMetricRule() *structuredlogscommon.MetricRule {
 	return &structuredlogscommon.MetricRule{
 		Namespace:     "ContainerInsights/Prometheus",
 		DimensionSets: [][]string{{"tagA"}},
-		Metrics:       []structuredlogscommon.MetricAttr{{Name: "metric_a"}},
+		Metrics:       []structuredlogscommon.MetricAttr{{Name: "metric_a", Unit: "Count"}},
```
pxaws, line 56: You changed the metric rule to add the unit. This changes the existing test. I think it is probably better to keep the original test (without unit). You could add a new test with unit.

plugins/processors/emfProcessor/emfProcessor.go
```diff
@@ -40,6 +76,17 @@ func (e *EmfProcessor) Apply(in ...telegraf.Metric) (result []telegraf.Metric) {
...
 			declaration.init()
 		}
+
+		// Metric units validation
+		e.validMetricUnit = make(map[string]string)
+		for k, v := range e.MetricUnit {
+			if _, ok := supportedUnits[v]; ok {
+				e.validMetricUnit[k] = v
+			} else {
+				log.Println("E! detect unsupported unit: ", v)
```
haojhcwa, line 86: I think this should be a warning, and Telegraf logging normally would add the plugin name before the log message too. One easy way to do it is follow this pattern and let telegraf take care of the formatting:
https://github.com/influxdata/telegraf/wiki/Logging

haojhcwa (approved):

plugins/processors/emfProcessor/emfProcessor.go
zhenweig: Thanks Edward for reviewing! 
I changed this logging to be warning, but didn't implement Plugin logging, since currently emfProcessor plugin uses agent logging at other places, and it's a bit stretch for this commit.

plugins/processors/emfProcessor/emfProcessor_test.go
zhenweig: Good catch. I added back the original test without metric unit support.

pxaws (approved):

PR #113, opened 2020-10-26
chore: remove md5 from jitter caculation

yimuniao:
# Description of the issue
md5 may have security fisk.

# Description of changes
use random instead

# License

# Tests
make

### DIFF
--- a/plugins/outputs/cloudwatch/util.go
+++ b/plugins/outputs/cloudwatch/util.go
@@ -4,12 +4,9 @@
 package cloudwatch
 
 import (
-	"crypto/md5"
-	"encoding/hex"
 	"log"
-	"os"
+	"math/rand"
 	"sort"
-	"strconv"
 	"time"
 
 	"github.com/aws/amazon-cloudwatch-agent/metric/distribution"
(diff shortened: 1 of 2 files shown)


plugins/outputs/cloudwatch/util.go
```diff
@@ -50,19 +47,12 @@ const (
 )
 
 func publishJitter(publishInterval time.Duration) (publishJitter time.Duration) {
-	hostName, _ := os.Hostname()
-	jitter := computeMD5Hash(hostName, int64(publishInterval.Seconds()))
+	rand.Seed(time.Now().Unix())
```
pingleig, line 50: Using the package level `rand.Seed` changes the global default, which may have impact on some library we are using https://golang.org/pkg/math/rand/#Seed

Maybe we can create a package specific instance and use that instead?

```go
var r = rand.New() // can have a better longer name ..

func init() {
     r.Seed(time.Now().Unix())
}
```

pingleig (approved):
LGTM

karlw00t:
Why are you seeding your random number generation with the current timestamp. This is an anti-pattern and should be avoided. Let me see how GoLang recommends doing this.

plugins/outputs/cloudwatch/util.go
```diff
@@ -50,19 +47,12 @@ const (
 )
 
 func publishJitter(publishInterval time.Duration) (publishJitter time.Duration) {
-	hostName, _ := os.Hostname()
-	jitter := computeMD5Hash(hostName, int64(publishInterval.Seconds()))
+	r := rand.New(rand.NewSource(time.Now().Unix()))
+	jitter := r.Int63n(int64(publishInterval.Seconds()))
```
karlw00t, line 51: looking at this blog post - https://golang.cafe/blog/golang-random-number-generator.html

You should be able to just call rand.Int63n(...) and the correct thing should happen. Without seeding the RNG, you should get the system's default random number generation (/dev/urandom for most)

karlw00t: I'm still looking into this. I'm not sure if the above statement is true.

karlw00t:
It appears that using the current timestamp is the way in which this should be done for non-cryptographic random numbers. Thank you for getting this fixed quickly.

plugins/outputs/cloudwatch/util.go
haojhcwa: As stated in the articled you linked:
> however if you try to run it several times you will see always the same number in output
Without seeding, the same binary would always generate the same sequence of random numbers, which when a large fleet start, will deterministically create a request spike at the same time.

PR #116, opened 2020-11-02
Add formatted date as a variable

BVMiko:
# Description of the issue
We want to be able to truncate old data from our Cloudwatch logs; so we decided to break the Log Streams into individual chunks based on date.

# Description of changes
This PR adds a new placeholder variable `{date}` which can be put in the Log Stream template definition, and will be replaced with the current date in ISO-8601 format (such as "2020-11-02").

# License

# Tests
We've been using this commit in a custom build for over a month now, without issue.

### DIFF
--- a/translator/translate/util/placeholderUtil.go
+++ b/translator/translate/util/placeholderUtil.go
@@ -8,6 +8,7 @@ import (
 	"net"
 	"os"
 	"strings"
+	"time"
 
 	"github.com/aws/amazon-cloudwatch-agent/translator/translate/agent"
 	"github.com/aws/amazon-cloudwatch-agent/translator/util/ec2util"
@@ -19,6 +20,7 @@ const (
 	localHostnamePlaceholder = "{local_hostname}" //regardless of ec2 metadata
 	ipAddressPlaceholder     = "{ip_address}"
 	awsRegionPlaceholder     = "{aws_region}"
+	datePlaceholder          = "{date}"
 
 	unknownInstanceId = "i-UNKNOWN"
 	unknownHostname   = "UNKNOWN-HOST"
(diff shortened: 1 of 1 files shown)


translator/translate/util/placeholderUtil.go
```diff
@@ -35,6 +37,7 @@ func ResolvePlaceholder(placeholder string, metadata map[string]string) string {
 	for k, v := range metadata {
 		tmpString = strings.Replace(tmpString, k, v, -1)
 	}
+	tmpString = strings.Replace(tmpString, datePlaceholder, time.Now().Format("2006-01-02"), -1)
```
hdj630, line 40: It's not flexible to use placeholder {date} for a specific format, i.e. ISO-8601, because someone else may prefer another format for {date}. It will be more generic to support additional "customized suffix" for logstream.

BVMiko: I can understand the the desire to make the date format customizable, but would it make sense for ISO-8601 to at least be the default? Zero-padded, descending values is the ideal format when dealing with sorting by strings, such as browsing in the AWS console.

Anyway, for the formatting itself are you requesting that this be moved to a new property in the JSON config file, and the property's value being the desired format?
```json
{
    "file_path": "/var/log/apache2/access.log",
    "log_group_name": "/aws/ec2/{local_hostname}/logs",
    "log_stream_name": "access.log",
    "log_stream_name_suffix": "2006-01-02"
},
```
In this way, the separator may be an issue.  I'm not extremely familiar with the date formatting string used in Golang, but it seems like the format is somewhat limited; for example I don't see how we could use a numeral as a separator.

Another alternative that may be worth considering would be extending the template match string to include the date format string, like this:
```json
{
    "file_path": "/var/log/apache2/access.log",
    "log_group_name": "/aws/ec2/{local_hostname}/logs",
    "log_stream_name": "access.log-{date:2006-01-02}",
},
```
This is where the default mentioned above would come into play; the template match could select an optional date format string (selecting any characters after a `:` and continue until the next `}'), and use that in the date generation string.  Of the two options, I think this one is more flexible, as the user could decide if the date string should be a prefix or suffix.

hdj630 (approved):

Issue #118, opened 2020-11-11
Fargate instances incorrectly detected as OnPrem


rs-garrick:
My ECS tasks that use cloudwatch agent on EC2 instances work great. The exact same container run on Fargate detect themselves as OnPrem and then break instance role permissions by adding these lines to the toml config:
    profile = "AmazonCloudWatchAgent"
    shared_credential_file = "/root/.aws/credentials"

As a temporary measure, is there a way to tell cloudwatch agent _not_ to add those 2 lines? 

Thanks in advance!

ThisIsQasim:
Same issue happens when running Cloudwatch agent on EKS Fargate. The metadata service isn't available so the agent decides it is running On-prem and starts to look for credentials in `.aws/credentials` even though it has an IAM role attached via a serviceAccount. Running the same pod on EC2 with the same serviceAccount works fine.

ThisIsQasim:
@hdj630 bringing this to your notice

ThisIsQasim:
I have pushed an image to `thisisqasim/cloudwatch-agent` that allows an env `FORCE_MODE_EC2=True` which forces it to use ec2 mode and not look for credentials at `.aws/credetials`

rs-garrick:
@ThisIsQasim That's great! Your behaviour should be the default.

But really, the config system of amazon-cloudwatch-agent is just bizarre. Why must it act differently than every other AWS client app?  It's trying to be "smart" and just being inflexible. It should just use the auth defaults of the aws library and be done with it.

ThisIsQasim:
I think it’s being done to validate the config for EC2 specific features e.g.  `{instance_id}`. It should however let the sdk handle the credentials chain and not manually handle it.

ThisIsQasim:
I have changed the flag for forcing EC2 mode from `FORCE_MODE_EC2=True` to `RUN_IN_AWS=True` as per instructions from the maintainers on #122

rs-garrick:
Just to be clear, your patch does not change the default behaviour, correct? Fargate instances will continue to be detected as OnPrem.

While I appreciate having a work-around, the patch seems strange to me. You've added an undocumented env var to fix behaviour that should be a straight-forward command-line option or automatic. Why not look at one of the existing variables like `AWS_EXECUTION_ENV`? It can be either AWS_ECS_FARGATE or AWS_ECS_EC2`

ThisIsQasim:
[...]

PR #122, opened 2020-11-19
enable flag to force ec2 mode

ThisIsQasim:
# Description of the issue
When running CloudWatch Agent on Fargate the agent container determines the environment as `OnPrem` since the metadata service is unavailable on Fargate. Afterwards the generated `toml` expects a credentials file at `.aws/credentials` instead of using the attached role. #118 

# Description of changes
This PR adds an ENV flag `RUN_IN_AWS=True` that would force the agent to skip metadata check allowing it to fallback to the usual credentials procedure. The downside is that the agent is unable to automatically determine the region and other variables that are dependent on the metadata service e.g. `{instance_id}` don’t work.

# License

# Tests

### DIFF
--- a/cmd/start-amazon-cloudwatch-agent/start-amazon-cloudwatch-agent.go
+++ b/cmd/start-amazon-cloudwatch-agent/start-amazon-cloudwatch-agent.go
@@ -37,13 +37,21 @@ var (
 
 	translatorBinaryPath string
 	agentBinaryPath      string
+	agentMode            string
 )
 
 // We use an environment variable here because we need this condition before the translator reads agent config json file.
+var forceModeEc2 = os.Getenv(config.FORCE_MODE_EC2)
 var runInContainer = os.Getenv(config.RUN_IN_CONTAINER)
 
 func translateConfig() error {
-	args := []string{"--output", tomlConfigPath, "--mode", "auto"}
+	if forceModeEc2 == config.FORCE_MODE_EC2_TRUE {
+		agentMode = "ec2"
+	} else {
+		agentMode = "auto"
+	}
+
+	args := []string{"--output", tomlConfigPath, "--mode", agentMode}
 	if runInContainer == config.RUN_IN_CONTAINER_TRUE {
 		args = append(args, "--input-dir", CONFIG_DIR_IN_CONTAINE)
 	} else {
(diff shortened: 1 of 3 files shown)


cmd/start-amazon-cloudwatch-agent/start-amazon-cloudwatch-agent.go
```diff
@@ -37,13 +37,21 @@ var (
 
 	translatorBinaryPath string
 	agentBinaryPath      string
+	agentMode            string
```
pxaws, line 40: I prefer to have all the changes located inside config-translator. For example, you could change this function for config-translator: https://github.com/aws/amazon-cloudwatch-agent/blob/master/translator/util/sdkutil.go#L29-L45. You can check the environment variable here and then return ec2 mode.

translator/config/envconst.go
```diff
@@ -8,6 +8,8 @@ const (
 	RUN_IN_CONTAINER_TRUE   = "True"
 	USE_DEFAULT_CONFIG      = "USE_DEFAULT_CONFIG"
 	USE_DEFAULT_CONFIG_TRUE = "True"
+	FORCE_MODE_EC2          = "FORCE_MODE_EC2"
```
pxaws, line 11: Could you change the name to something like "RUN_IN_AWS"? Currently the ec2 mode is overloaded with many meanings (e.g. ec2 instance, container running in ec2 instance in ecs, container running in ec2 instance in eks, container running in ecs fargate). We would like to do some refactoring about it. So I guess "RUN_IN_AWS" is probably more meaningful in you case.

cmd/start-amazon-cloudwatch-agent/start-amazon-cloudwatch-agent.go
ThisIsQasim: Moved the env check to config-translator

translator/config/envconst.go
ThisIsQasim: renamed the env name to `RUN_IN_AWS`

pxaws (commented):
Please update the PR message:

>This PR adds an ENV flag FORCE_MODE_EC2=True that would force the agent to skip metadata check allowing it to fallback to the usual credentials procedure. The only downside is that the agent is unable to automatically determine the region.

ThisIsQasim:
PR description updated.

yimuniao (approved):

PR #132, opened 2020-11-29
Fix incorrect systemd detection

### LINKED ISSUE #131: Weak detection of the systemd in the amazon-cloudwatch-agent-ctl
The amazon-cloudwatch-agent-ctl contains the next piece of the code for detecting of systemd ` [ "$(systemctl | grep -c '\-\.mount')" = 1 ]`. 

```
    # detect which init system is in use
    if [ "$(/sbin/init --version 2>/dev/null | grep -c upstart)" = 1 ]; then
        SYSTEMD='false'
    elif [ "$(systemctl | grep -c '\-\.mount')" = 1 ]; then
        SYSTEMD='true'
    elif [ -f /etc/init.d/cron ] && [ ! -h /etc/init.d/cron ]; then
        echo "sysv-init is not supported" >&2
        exit 1
    else
        echo "unknown init system" >&2
        exit 1
    fi
```

Unfortunately, it didn't work on my laptop; when I run the command `systemctl | grep '\-\.mount'` it shows at least two entities in the list

```
$ systemctl | grep '\-\.mount'
  -.mount                                                                                  loaded active     mounted   Root Mount                                                                
  tmp-.mount_jetbraP6HI4I.mount                                                            loaded active     mounted   /tmp/.mount_jetbraP6HI4I
```

There is a solution for systemd detection here https://superuser.com/a/1183819/457737 `ps --no-headers -o comm 1`, let's use it?

asaushkin:
Closes #131

# Description of the issue
https://github.com/aws/amazon-cloudwatch-agent/issues/131

# Description of changes
It is possible that in some systems the command `systemctl | grep -c '\-\.mount'` will return a value greater than zero. In these circumstances the `amazon-cloudwatch-agent-ctl` script doesn't detect `systemd` correctly. This patch is intended to fix this behavior.

# License

# Tests
Given that

```
$ systemctl | grep '\-\.mount'
  -.mount                                                                                  loaded active     mounted   Root Mount                                                                
  tmp-.mount_jetbraP6HI4I.mount                                                            loaded active     mounted   /tmp/.mount_jetbraP6HI4I
```

Old behavior 

```
$ [ "$(systemctl | grep -c '\-\.mount')" = 1 ]; echo $?
1
```

New behavior:
```
$ [ "$(systemctl | grep -c -E '\-\.mount\s')" = 1 ]; echo $?
0
```

### DIFF
--- a/packaging/dependencies/amazon-cloudwatch-agent-ctl
+++ b/packaging/dependencies/amazon-cloudwatch-agent-ctl
@@ -249,7 +249,7 @@ main() {
     # detect which init system is in use
     if [ "$(/sbin/init --version 2>/dev/null | grep -c upstart)" = 1 ]; then
 	SYSTEMD='false'
-    elif [ "$(systemctl | grep -c -E '\s\-\.mount\s')" = 1 ]; then
+    elif [ "$(systemctl | grep -c '\-\.mount')" -gt 0 ]; then
 	SYSTEMD='true'
     elif [ -f /etc/init.d/cron ] && [ ! -h /etc/init.d/cron ]; then
 	echo "sysv-init is not supported" >&2
--- a/packaging/dependencies/amazon-cloudwatch-agent-ctl
+++ b/packaging/dependencies/amazon-cloudwatch-agent-ctl
@@ -249,7 +249,7 @@ main() {
     # detect which init system is in use
     if [ "$(/sbin/init --version 2>/dev/null | grep -c upstart)" = 1 ]; then
 	SYSTEMD='false'
-    elif [ "$(systemctl | grep -c '\-\.mount')" -gt 0 ]; then
+    elif [ "$(systemctl | grep -c -E '\s\-\.mount\s')" = 1 ]; then
 	SYSTEMD='true'
     elif [ -f /etc/init.d/cron ] && [ ! -h /etc/init.d/cron ]; then
 	echo "sysv-init is not supported" >&2
--- a/packaging/dependencies/amazon-cloudwatch-agent-ctl
+++ b/packaging/dependencies/amazon-cloudwatch-agent-ctl
@@ -249,7 +249,7 @@ main() {
     # detect which init system is in use
     if [ "$(/sbin/init --version 2>/dev/null | grep -c upstart)" = 1 ]; then
 	SYSTEMD='false'
-    elif [ "$(systemctl | grep -c '\-\.mount')" = 1 ]; then
+    elif [ "$(systemctl | grep -c -E '\s\-\.mount\s')" = 1 ]; then
 	SYSTEMD='true'
     elif [ -f /etc/init.d/cron ] && [ ! -h /etc/init.d/cron ]; then
 	echo "sysv-init is not supported" >&2
--- a/packaging/dependencies/amazon-cloudwatch-agent-ctl
+++ b/packaging/dependencies/amazon-cloudwatch-agent-ctl
@@ -249,7 +249,7 @@ main() {
     # detect which init system is in use
     if [ "$(/sbin/init --version 2>/dev/null | grep -c upstart)" = 1 ]; then
 	SYSTEMD='false'
-    elif [ "$(systemctl | grep -c -E '\s\-\.mount\s')" = 1 ]; then
+    elif [ "$(systemctl | grep -c -E '\-\.mount\s')" = 1 ]; then
 	SYSTEMD='true'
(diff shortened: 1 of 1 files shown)


yimuniao (commented):
The fix should be ok. Just be curious how following line appears?
`tmp-.mount_jetbraP6HI4I.mount   `

asaushkin:
> The fix should be ok. Just be curious how following line appears?
> `tmp-.mount_jetbraP6HI4I.mount `

As I see this mount point was created by IntelliJ Jetbrains Toolbox:

```
$ cat /proc/mounts | grep jet
jetbrains-toolbox /tmp/.mount_jetbraP6HI4I fuse.jetbrains-toolbox ro,nosuid,nodev,relatime,user_id=1000,group_id=1000 0 0
```

asaushkin:
Judging by IntelliJ documentation they "Linux: JetBrains Toolbox App is packaged in [AppImage](http://appimage.org/) and requires FUSE to run. See [AppImage Wiki](https://github.com/probonopd/AppImageKit/wiki/FUSE) for details."

yimuniao:
> > The fix should be ok. Just be curious how following line appears?
> > `tmp-.mount_jetbraP6HI4I.mount `
> 
> As I see this mount point was created by IntelliJ Jetbrains Toolbox:
> 
> ```
> $ cat /proc/mounts | grep jet
> jetbrains-toolbox /tmp/.mount_jetbraP6HI4I fuse.jetbrains-toolbox ro,nosuid,nodev,relatime,user_id=1000,group_id=1000 0 0
> ```

In this case, will intellij generate the same temp .mount file on upstart system? If yes, it will be a problem, that exactly match "-.mount" maybe a good choice.

asaushkin:
> 
> In this case, will intellij generate the same temp .mount file on upstart system? If yes, it will be a problem, that exactly match "-.mount" maybe a good choice.

Agree with you. Changed to the exact pattern.

packaging/dependencies/amazon-cloudwatch-agent-ctl
```diff
@@ -249,7 +249,7 @@ main() {
     # detect which init system is in use
     if [ "$(/sbin/init --version 2>/dev/null | grep -c upstart)" = 1 ]; then
 	SYSTEMD='false'
-    elif [ "$(systemctl | grep -c '\-\.mount')" = 1 ]; then
+    elif [ "$(systemctl | grep -c -E '\s\-\.mount\s')" = 1 ]; then
```
yimuniao, line 252: shall we change `\s` to `\s*`? not sure if the number of whitespace will change in the future.

Or other solution is to exclude `tmp-.mount_jetbra`,  the I am not sure what the exactly output systemctl should be, but looks like current fix has some backward compatible issue.

asaushkin: > shall we change `\s` to `\s*`? not sure if the number of whitespace will change in the future.
> 

I don't think so, `'\s\-\.mount\s'` will match to any extra symbols at the start or at the end because it doesn't contain line start/end markers "^$".

> Or other solution is to exclude `tmp-.mount_jetbra`, the 

Also, I don't think that we should rely on any 3rd party apps behavior in this script

> I am not sure what the exactly output systemctl should be, but looks like current fix has some backward compatible issue.

We can omit the first space detector in the pattern and use this one `'\-\.mount\s'` - it will mitigate the risk of backward incompatibility.

yimuniao: > > shall we change `\s` to `\s*`? not sure if the number of whitespace will change in the future.
> 
> I don't think so, `'\s\-\.mount\s'` will match to any extra symbols at the start or at the end because it doesn't contain line start/end markers "^$".
> 
> > Or other solution is to exclude `tmp-.mount_jetbra`, the
> 
> Also, I don't think that we should rely on any 3rd party apps behavior in this script
> 
> > I am not sure what the exactly output systemctl should be, but looks like current fix has some backward compatible issue.
> 
> We can omit the first space detector in the pattern and use this one `'\-\.mount\s'` - it will mitigate the risk of backward incompatibility.

I tested `'\s\-\.mount\s'` does not work.   you can use '\-\.mount\s'

yimuniao (approved):

packaging/dependencies/amazon-cloudwatch-agent-ctl
asaushkin: Changed it

PR #157, opened 2021-01-07
Avoid cwagent-otel-collector to be auto started from the "stop" state…

hdj630:
# Description of the changes
Update ctl scripts to avoid auto-restart during system reboot when cwagent-otel-collector is in 'stop' state. Note that it only work for cwagent-otel-collector managed by systemd or windows service manager. cwagent-otel-collector managed by upstart is TBD as old version of upstart doesn't support stanza 'Manual'. 

If cwagent-otel-collector managed by Upstart is started during system reboot (in such case, configuration is possibly not provided), it will Exit with return code 0, thus won't be restarted by Upstart.

# License


yimuniao (approved):

jedwards1211:
What does it mean if I get this when starting the cloudwatch agent?  I don't get why this isn't documented anywhere.

```
****** processing cwagent-otel-collector ******
cwagent-otel-collector will not be started as it has not been configured yet.
```

PR #159, opened 2021-01-13
Split EMF metric batches into 100s

roystchiang:
# Description of the issue
Cloudwatch EMF supports up to 100 metrics in a single batch, https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Embedded_Metric_Format_Specification.html.

Any EMF that exceeds 100 metrics are dropped.

# Description of changes
I made a change to Prometheus metric handler to add metrics in batches of 100. Please let me know if there's a better place to add this logic. 

# License

# Tests
Built and tested in our EKS cluster, and verified that EMF metrics that exceeded 100 are now split into multiple EMFs, and we are no longer losing metrics.

### DIFF
--- a/plugins/inputs/prometheus_scraper/metrics_handler.go
+++ b/plugins/inputs/prometheus_scraper/metrics_handler.go
@@ -13,6 +13,11 @@ import (
 	"github.com/influxdata/telegraf"
 )
 
+
+const (
+	emfMetricsLimit		= 100
+)
+
 // Use metricMaterial instead of mbMetric to avoid unnecessary tags&fields copy
 type metricMaterial struct {
 	tags     map[string]string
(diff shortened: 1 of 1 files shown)


roystchiang:
Now that I think about it, a cleaner way to do batching is probably under the `cloudwatchlogs` plugin. This way, input is independent from the output.

I'll refactor.

haojhcwa:
Hi, thanks for your contribution,

Yes, you are right, I totally agree the input plugin might not be the best place to do the metric splitting.
However, before we dive into where it should be split, let me clarify one apparent misunderstanding of the EMF limitation we are trying to deal with here. Below is a direct quote from the documentation page linked above:

> Metrics— An array of MetricDefinition objects. This array MUST NOT contain more than **100 MetricDefinition objects**.

The limit is on MetricDefinition objects, NOT on number of metrics (it is possible to contain many more metrics in the log entry to be saved as logs in cloudwatch logs).

At the moment I think the best place to add this splitting logic is probably in the [emfProcessor](https://github.com/aws/amazon-cloudwatch-agent/blob/master/plugins/processors/emfProcessor/emfProcessor.go#L74-L117). The processor interface it uses allow you to return multiple metrics from a single metric, and it was designed to create the MetricDefinitions later used by EMF, which has the 100 limitation.

I've discussed with the team about the possibility of moving the emfProcessor logic into cloudwatchlogs plugin in the future, but we cannot see a very convincing reason to do it now.

Please let us know what do you think

roystchiang:
Thanks for the feedback. I'll dive deeper into EMF Processor.

haojhcwa:
closing for now. Until we have the better solution implemented as a processor.

PR #163, opened 2021-01-26
Allow ignoring pods with the aws.amazon.com/cloudwatch-agent: ignore annotation

darkhelmet:
# Description of the issue

Having metrics come in for each CI job with unique names is polluting our metrics and costs more money. I wanted to ignore some pods.

# Description of changes

In the `podstore.Decorate` method of the k8s processor, after we're confident we've got a pod, check the annotations and return false so as to not handle that pod.

# License

# Tests

I have this running in our devops cluster where Jenkins is creating pods with these annotations. Automated tests pass, but nothing explicitly tests the method I changed, and I couldn't in a reasonable time figure out how to write a good test for it. If somebody wants to help out let me know.

### DIFF
--- a/plugins/processors/k8sdecorator/stores/podstore.go
+++ b/plugins/processors/k8sdecorator/stores/podstore.go
@@ -142,7 +142,7 @@ func (p *PodStore) Decorate(metric telegraf.Metric, kubernetesBlob map[string]in
 		}
 
 		// Ignore if we're told to ignore
-		if "ignore" == entry.pod.ObjectMeta.Annotations[ignoreAnnotation] {
+		if entry.pod.ObjectMeta.Annotations[ignoreAnnotation] == "ignore" {
 			return false
 		}
 
--- a/plugins/processors/k8sdecorator/stores/podstore.go
+++ b/plugins/processors/k8sdecorator/stores/podstore.go
@@ -28,6 +28,7 @@ const (
 	cpuKey             = "cpu"
 	splitRegexStr      = "\\.|-"
 	kubeProxy          = "kube-proxy"
+	ignoreAnnotation   = "aws.amazon.com/cloudwatch-agent"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


haojhcwa:
Re-running the unit tests

darkhelmet:
@haojhcwa Anything I need to do to move this along?

pingleig:
@darkhelmet Sorry for the late reply. We are already in the (internal progress) of current release so I tag this PR to next release.

The implementation is correct, all the pod and container metrics (from cadvisor) will have a pod name,  pod info is fetched from k8s apiserver and if there is a matched annotation, the processor drops the metrics (pod cpu, container mem etc.)

There are a few issues

- we might want to change the annotation name to something like prometheus's `prometheus.io/scrape` to something like `aws.amazon.com/cloudwatch-agent/ignore` and do a `if labelVal == 'true'` because `aws.amazon.com/cloudwatch-agent` is too broad, and we may want to use it for other purposes (or just make it a prefix for other keys).
- we can make this annotation a configurable value so user can reuse their existing annotations, but the logic for configuration (translation) is tricky and not well documented for both user and dev. I think it's fine to only support a well known default.

For test

- if you think the test logic is a bit messy to figure out you can leave it to us.
- podstore does have its unit test, e.g. you can copy the large json and add annotation like https://github.com/aws/amazon-cloudwatch-agent/blob/master/plugins/processors/k8sdecorator/stores/podstore_test.go#L54 , then write some assertion using the new json payload.
- we will to update our internal integration test (the github repo only has unit test)

darkhelmet:
I thought about using `aws.amazon.com/cloudwatch-agent/ignore` first but turns out that's not a valid annotation. You can't even deploy a pod with it. It doesn't allow multiple slashes, it's just `prefix/name`, per https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/#syntax-and-character-set

I also considered allowing configuration, but supporting the known default was so easy and straightforward it didn't seem necessary.

I'll see if I can mess around with tests this weekend to get something going.

darkhelmet:
Also, I went with the `ignore` route instead of `enabled = true` or something because then it maintains the current experience of tracking everything by default. This drops in and nobody has to change anything.

darkhelmet:
@pingleig I've got a test, it works fine locally, not sure why things failed.

pingleig (dismissed):
LGTM (retried test and it worked ...)

Though before I merge it,  I need to discuss w/ the team next Monday to see if there is any problems.

pingleig:
I checked with the team today, can we change the annotation from `aws.amazon.com/cloudwatch-agent:ignore` to `aws.amazon.com/cloudwatch-agent-ignore:true`?

pingleig (approved):
Thanks! LGTM. Will merge after the CI is done.

PR #166, opened 2021-02-06
1. Expanding the translator capability by exposing an API to override…

arlnav:
1. Expanding the translator capability by exposing an API to override the schema 
2. Allowing translator plugins to specify and override the metricPath by setting its own return value.
# Description of the issue
Issue 1. Lack of capability of extending current schema since schema value is hardcoded.
Issue 2. If input plugin metricPath has its own value it should not get different metricPath value as you see below:

`
[inputs.myPlugin.tags]
`
`
           metricPath = "metrics"
`
Should not get logs metricPath.
` 
[inputs.myPlugin.tags]
`
`
           metricPath = "logs"
`
# Description of changes
Issue 1. **OverwriteSchema** method will allow developers to override current schema with flexibility of expanding the translator schema. (adding new plugins, update existing plugins)
Issue 2. If translator plugins already have their own metricPath value, we do not want to override the value.

# License

# Tests
Ran **make test** all unit tests passed.

### DIFF
--- a/translator/config/schema.go
+++ b/translator/config/schema.go
@@ -983,6 +983,10 @@ func GetJsonSchema() string {
 	return schema
 }
 
+func OverwriteSchema(newSchema string) {
+	schema = newSchema
+}
+
 // Translate Sample:
 // (root).agent.metrics_collection_interval -> /agent/metrics_collection_interval
 // (root).metrics.metrics_collected.cpu.resources.1 -> /metrics/metrics_collected/cpu/resources/1
--- a/translator/setMetricPath.go
+++ b/translator/setMetricPath.go
@@ -28,7 +28,10 @@ func SetMetricPath(result map[string]interface{}, sectionKey string) {
 				input := inputIntf.(map[string]interface{})
 				if val, ok := input[tagKey]; ok {
 					tags := val.(map[string]interface{})
-					tags[routingTagKey] = sectionKey
+					// If a plugin already has its own metricPath value, we do not want to override its value.
+					if _, ok := tags[routingTagKey]; !ok {
+						tags[routingTagKey] = sectionKey
+					}
 				} else {
 					input[tagKey] = map[string]interface{}{routingTagKey: sectionKey}
 				}
--- a/translator/setMetricPath_test.go
+++ b/translator/setMetricPath_test.go
@@ -92,6 +92,48 @@ func getExpectedSectionWithTags() map[string]interface{} {
 	}
 }
 
+func getSectionWithTagsMetricPath() map[string]interface{} {
+	input := map[string]interface{}{}
+	input["tags"] = map[string]interface{}{routingTagKey: "metrics"}
+	input["fieldpass"] = "fp1"
+
+	output := map[string]interface{}{}
+	output["namespace"] = "ns1"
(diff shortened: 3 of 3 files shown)


haojhcwa (dismissed):
LGTM

translator/config/schema.go
```diff
@@ -983,6 +983,10 @@ func GetJsonSchema() string {
 	return schema
 }
 
+func OverwriteSchema(newSchema string) {
```
yimuniao, line 986: I do not see where calls this method and there is no description how to use it.

Could you add some description about how to use it?

translator/setMetricPath.go
```diff
@@ -28,7 +28,9 @@ func SetMetricPath(result map[string]interface{}, sectionKey string) {
 				input := inputIntf.(map[string]interface{})
 				if val, ok := input[tagKey]; ok {
 					tags := val.(map[string]interface{})
-					tags[routingTagKey] = sectionKey
+					if _, ok := tags[routingTagKey]; !ok {
+						tags[routingTagKey] = sectionKey
+					}
```
yimuniao, line 33: I am not fully understand this case. Could you explain more about why  routingTagKey will conflict.

translator/config/schema.go
arlnav:
```
import (
	parent "github.com/aws/amazon-cloudwatch-agent/translator/config"
)

func applyMyPluginSchema(oldSchema string) string {
    // Update oldSchema with new changes
}

func init() {
	//We interfere external schema by adding updated schema changes
	parent.OverwriteSchema(applyMyPluginSchema(parent.GetJsonSchema()))
}
```
We are working on a new input plugin which needed some changes on existing schema.
The above **OverwriteSchema** method allow us to add our schema changes to existing one.

translator/setMetricPath.go
arlnav: For new plugin to work, we have translator changes under logs which gets **logs** metricPath by default see below:
``` 
[inputs.myPlugin.tags]
    metricPath = "logs"
```
Since we are emitting metrics using **outputs.cloudwatchlogs**, we need tags metrics path to set as **metrics**.
```
[inputs.myPlugin.tags]
   metricPath = "metrics"
```

translator/config/schema.go
yimuniao: Is your plugin name fixed?

If yes, shall we just enhance the existing schema?

translator/setMetricPath.go
yimuniao: Could you add annotation here?

translator/config/schema.go
yimuniao: Confirmed with arlnav, they have a private plugin, it is fine.

arlnav: OverwriteSchema will allow anyone to override existing schema with their own schema when they roll out their own version of plugin instead of hardcoding existing schema, specifically when new schema changes are not for public use.

translator/setMetricPath.go
arlnav: I added the comment.

yimuniao (approved):

PR #169, opened 2021-02-09
chore: fix timestamp parse error if timestamp_format starts with space

yimuniao:
# Description of the issue
 if the format startswith "%-m" or "%-d", there is an "\\s{0,1}" at the beginning of timestamp_regex.
like "timestamp_format": "%-m %-d %H:%M:%S" will be converted into following layout and regex
timestamp_layout = "1 2 15:04:05"
timestamp_regex = "(\\s{0,1}\\d{1,2} \\s{0,1}\\d{1,2} \\d{2}:\\d{2}:\\d{2})"
following timestamp string " 2 1 07:10:06" matches the regex, but it can not match the layout.
		
# Description of changes
Remove the \\s{0,1} prefix 

# License

# Tests
Unittest


haojhcwa (approved):
LGTM

PR #189, opened 2021-03-22
cadvisor: Fix pod detection for containerd runtime on k8s

### LINKED ISSUE #188: [k8s] Pod metrics is gone when using containerd as runtime
This is exported from internal ticket

## TL;DR

The latest image is released, if you were using temp image from this comment https://github.com/aws/amazon-cloudwatch-agent/issues/188#issuecomment-803764697 please update to the latest tag.

If the error message `W! No pod metric collected, metrics count is still 7 is containerd socket mounted? https://github.com/aws/amazon-cloudwatch-agent/issues/188`leads you to this issue 

-  make sure you have updated the yaml to mount the containerd socket into the cloudwatch-agent pod
-  the path for containerd socket may not be in the standard location. e.g. bottlerocket uses `/run/dockershim.sock` instead of `/run/containerd/containerd.sock`

## Background

We were relying on pause container to have `POD` for detecting pod, which is the case for docker but not for containerd https://github.com/containerd/cri/issues/922#issuecomment-423729537

User will not see pod metrics in container insight dashboard and they will find the following log which is introduced in #171 

https://github.com/aws/amazon-cloudwatch-agent/blob/fbdd619269be7a00172e06992a8d40b22be1a6d7/plugins/inputs/cadvisor/container_info_processor.go#L72-L72

The root cause is we are expecting `containerName == 'POD'` to mark a path as pod

https://github.com/aws/amazon-cloudwatch-agent/blob/fbdd619269be7a00172e06992a8d40b22be1a6d7/plugins/inputs/cadvisor/container_info_processor.go#L119-L126

## Fix

- code should be similar to https://github.com/DataDog/integrations-core/pull/2283 
- manifest need to mount containerd sock into cwagent daemonset container

## Release

The fix will be included in next release, the release date is not determined (yet).

pingleig:
# Description of the issue

Fix #188 k8s cluster using containerd runtime was not able to get pod metrics because

- contained socket is not mounted
- pod detection logic relies on pause container with name 'POD`, which is only the case for docker

# Description of changes

- [x] Define a new constant `TypeInfraContainer` and remove most usage on `POD` container name.
- [x] update all the manifest to mount containerd
- [x] deal with `kubernetes>docker>container_id`  not sure if it is used by any place (it comes from k8s api server though, not cadvisor), just use containerd in json path when it is containerd
- [x] container file system metric like `container_filesystem_utilization` is missing in structured log https://github.com/aws/amazon-cloudwatch-agent/issues/192, it's known issue in cadviisor.

# License

# Tests

- [x] containerd cluster using kops w/ k8s version 1.20.4
- [x] docker cluster using kops/eks

### DIFF
--- a/plugins/inputs/cadvisor/container_info_processor.go
+++ b/plugins/inputs/cadvisor/container_info_processor.go
@@ -167,6 +167,7 @@ func processContainer(info *cinfo.ContainerInfo, detailMode bool, containerOrche
 func processPod(info *cinfo.ContainerInfo, podKeys map[string]podKey) []*extractors.CAdvisorMetric {
 	var result []*extractors.CAdvisorMetric
 	if isContainerInContainer(info.Name) {
+		log.Printf("D! drop metric because it's nested container, name %s", info.Name)
 		return result
 	}
 
--- a/plugins/inputs/cadvisor/container_info_processor.go
+++ b/plugins/inputs/cadvisor/container_info_processor.go
@@ -133,6 +133,11 @@ func processContainer(info *cinfo.ContainerInfo, detailMode bool, containerOrche
 			tags[ContainerNamekey] = containerName
 			tags[ContainerIdkey] = path.Base(info.Name)
 			containerType = TypeContainer
+
+			// TODO(pingleig): wait for upstream fix https://github.com/aws/amazon-cloudwatch-agent/issues/192
+			if !info.Spec.HasFilesystem {
+				log.Printf("D! containerd does not have container filesystem metrics from cadvisor, See https://github.com/aws/amazon-cloudwatch-agent/issues/192")
+			}
 		}
 	} else {
 		containerType = TypeNode
--- a/plugins/inputs/cadvisor/container_info_processor.go
+++ b/plugins/inputs/cadvisor/container_info_processor.go
@@ -98,17 +98,23 @@ func processContainer(info *cinfo.ContainerInfo, detailMode bool, containerOrche
 		if !detailMode {
 			return result, pKey
 		}
+
+		if len(info.Spec.Labels) == 0 {
+			log.Printf("W! no label found from container spec, is containerd socket mounted? https://github.com/aws/amazon-cloudwatch-agent/issues/188")
+		}
 		// Only a container has all these three labels set.
 		containerName := info.Spec.Labels[containerNameLable]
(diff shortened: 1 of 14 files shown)


plugins/inputs/cadvisor/extractors/mem_extractor.go
```diff
@@ -25,11 +25,12 @@ func (m *MemMetricExtractor) HasValue(info *cinfo.ContainerInfo) bool {
 
 func (m *MemMetricExtractor) GetValue(info *cinfo.ContainerInfo, containerType string) []*CAdvisorMetric {
 	var metrics []*CAdvisorMetric
-	if info.Spec.Labels[containerNameLable] == infraContainerName {
+	if containerType == TypeInfraContainer {
 		return metrics
 	}
 
 	metric := newCadvisorMetric(containerType)
+	metric.cgroupPath = info.Name
```
pxaws, line 33: why we only do it for mem metrics?

plugins/inputs/cadvisor/container_info_processor.go
```diff
@@ -98,32 +100,39 @@ func processContainer(info *cinfo.ContainerInfo, detailMode bool, containerOrche
...
 
 		tags[PodIdKey] = podId
 		tags[K8sPodNameKey] = podName
 		tags[K8sNamespace] = namespace
-		if containerName != infraContainerName {
+
+		switch containerName {
+		// For docker, pause container name is set to POD while containerd does not set it.
+		// See https://github.com/aws/amazon-cloudwatch-agent/issues/188
+		case "", infraContainerName:
```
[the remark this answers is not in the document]
pingleig: I think based on the spec only pod and namespace are required. container name is set to whatever user specified in yaml.  The `pause` container is hidden from user so the name relies on runtime, which is `""` for containerd and `POD` for docker. Even if docker changed to `""`, this logic of detecting `pause` container should still work.

plugins/inputs/cadvisor/extractors/mem_extractor.go
pingleig: (because only one of them is enough for debug, I will add others

pingleig: done

pxaws (approved):

PR #196, opened 2021-04-02
Work around direct proxy version issue with golang.zx2c4.com/wireguard.

straussb:
proxy.golang.org has versions of golang.zx2c4.com/wireguard with leading v's, whereas the git repo has tags without
leading v's: https://git.zx2c4.com/wireguard-go/refs/tags

So, fetching this module with version v0.0.20200121 (as done by the transitive dependency
https://github.com/WireGuard/wgctrl-go/blob/e35592f146e40ce8057113d14aafcc3da231fbac/go.mod#L12 ) was not working when
using GOPROXY=direct.

Replacing with the pseudo-version works around this.

# Description of the issue

# Description of changes

# License

# Tests


pingleig (approved):
LGTM

I will merge it after all the CI run has finished (2/3 already passed).

PR #208, opened 2021-05-09
start: Skip pidfile when running in a container.

nairb774:
# Description of the issue

When the agent is running in a container, the use and management of a
pidfile shouldn't be necessary. By making this change, it moves the
agent container one step closer to being able to run entirely with a read
only root file system. While the pid file functionality doesn't prevent
running with a read only file system, it generates an unnecessary error
line in the logs:

2021-05-09T17:03:00Z E! Unable to create pidfile: open /opt/aws/amazon-cloudwatch-agent/var/amazon-cloudwatch-agent.pid: read-only file system

While it is possible to provide a small scratch filesystem for the pid
file, this seems entirely unnecessary.

# License

Sure.

# Tests

None. Only visual inspection of the code in question, and what I believe to be impacted functionality (none).

### DIFF
--- a/cmd/start-amazon-cloudwatch-agent/path.go
+++ b/cmd/start-amazon-cloudwatch-agent/path.go
@@ -33,7 +33,6 @@ func startAgent(writer io.WriteCloser) error {
 		execArgs := []string{
 			agentBinaryPath, // when using syscall.Exec, must pass binary name as args[0]
 			"-config", tomlConfigPath, "-envconfig", envConfigPath,
-			"-pidfile", AGENT_DIR_LINUX + "/var/amazon-cloudwatch-agent.pid",
 		}
 		if err := syscall.Exec(agentBinaryPath, execArgs, os.Environ()); err != nil {
 			return fmt.Errorf("error exec as agent binary: %w", err)


cmd/start-amazon-cloudwatch-agent/path.go
```diff
@@ -33,7 +33,6 @@ func startAgent(writer io.WriteCloser) error {
 		execArgs := []string{
 			agentBinaryPath, // when using syscall.Exec, must pass binary name as args[0]
 			"-config", tomlConfigPath, "-envconfig", envConfigPath,
-			"-pidfile", AGENT_DIR_LINUX + "/var/amazon-cloudwatch-agent.pid",
```
pxaws, line 36: Though this seems to be safe, it still changes the existing behavior. If customers build their own image with cloudwatch agent and want to run it with other applications in one container, they might still want the pid file. Could you introduce an option to disable the generation of pid file?

pxaws:
>By making this change, it moves the agent container one step closer to being able to run entirely with a read only root file system.

Could you explain what's your plan with the read only file system? What changes are required? Thanks!

nairb774:
> Could you explain what's your plan with the read only file system?

The way we have the container setup has it running as root, and with part of the node (EKS) filesystem mounted in. This is being done to ship node level logs to CloudWatch. Given the simplicity of the tool, and the privileged user it runs as (root), it makes sense to simplify and limit as much of the environment as possible. One easy step is to try running containers with a read only file system. This forces any local state to be well documented and correctly managed. In the case of the CloudWatch Agent, there are currently 4 uses of the file system - at least that we've observed:

1. Reading logs/removing old log files.
2. Maintaining the state pointers written to `/opt/aws/amazon-cloudwatch-agent/logs/state`
3. The json->toml translated config file.
4. This pid file.

The first 2 need to exist. The other two, could potentially be removed. The json->toml translator thing, at least using disk, seemed like more work to get rid of than any sort of value would be gained from removing. The pid file seemed like easy pickings - and in looking through the history of the code, potentially only exists due to a long ago copy/paste oversight when adding the `if container` related code.

jhnlsn:
Closing this PR and moving to an issue so we can hash out the feature request.  We would need to introduce this in a non-breaking way.

PR #212, opened 2021-05-18
chore: Enable build mode override via env variable

haojhcwa:
# Description of the issue
Go compiler has switched to internal linking mode when building in PIE
mode since version 1.15 (https://golang.org/doc/go1.15#linker). This had
a side effect of not able to compile a purely statically linked PIE
binary with CGO disabled, see: https://github.com/golang/go/issues/40711

As a result, the agent binaries build with PIE mode will fail to run in
a container that is built from SCRATCH.

# Description of changes
In this CR we allow the users to
override the build mode with environment variable `CWAGENT_BUILD_MODE`,
allowing a non-PIE binary to be built to be used in a container without
libc.
For example, a non-pie agent can be built with:
```
CWAGENT_BUILD_MODE=default make
```

# License

# Tests
Built locally and tested with simple docker file below:
```
# syntax=docker/dockerfile:1
FROM scratch
COPY amazon-cloudwatch-agent /
CMD ["/amazon-cloudwatch-agent", "-version"]

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,13 +9,19 @@ VERSION := $(shell echo ${VERSION} | sed -e "s/^v//")
 ifeq ($(VERSION),)
 VERSION := `cat CWAGENT_VERSION`
 endif
+
+# Determine agent build mode, default to PIE mode
+ifndef CWAGENT_BUILD_MODE
+CWAGENT_BUILD_MODE=pie
+endif
+
 BUILD := $(shell date -u +"%Y-%m-%dT%H:%M:%SZ")
 LDFLAGS = -s -w
 LDFLAGS +=  -X github.com/aws/amazon-cloudwatch-agent/cfg/agentinfo.VersionStr=${VERSION}
 LDFLAGS +=  -X github.com/aws/amazon-cloudwatch-agent/cfg/agentinfo.BuildStr=${BUILD}
-LINUX_AMD64_BUILD = CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -buildmode=pie -ldflags="${LDFLAGS}" -o $(BUILD_SPACE)/bin/linux_amd64
-LINUX_ARM64_BUILD = GOOS=linux GOARCH=arm64 go build -buildmode=pie -ldflags="${LDFLAGS}" -o $(BUILD_SPACE)/bin/linux_arm64
-WIN_BUILD = GOOS=windows GOARCH=amd64 go build -buildmode=pie -ldflags="${LDFLAGS}" -o $(BUILD_SPACE)/bin/windows_amd64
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -9,13 +9,19 @@ VERSION := $(shell echo ${VERSION} | sed -e "s/^v//")
 ifeq ($(VERSION),)
 VERSION := `cat CWAGENT_VERSION`
 endif
+
+# Determine agent build mode, default to PIE mode
+ifndef CWAGENT_BUILD_MODE
```
pingleig, line 14: I think maybe we should default to non pie build mode, also can you update `Dockerfile` like https://github.com/aws/amazon-cloudwatch-agent/compare/master...pingleig:k8s-skip-metric?expand=1 to see if build from source using docker works (btw: you might need to switch network to solve go mod getting blocked problem)

Makefile
```diff
@@ -9,13 +9,19 @@ VERSION := $(shell echo ${VERSION} | sed -e "s/^v//")
...
+endif
+
 BUILD := $(shell date -u +"%Y-%m-%dT%H:%M:%SZ")
 LDFLAGS = -s -w
 LDFLAGS +=  -X github.com/aws/amazon-cloudwatch-agent/cfg/agentinfo.VersionStr=${VERSION}
 LDFLAGS +=  -X github.com/aws/amazon-cloudwatch-agent/cfg/agentinfo.BuildStr=${BUILD}
-LINUX_AMD64_BUILD = CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -buildmode=pie -ldflags="${LDFLAGS}" -o $(BUILD_SPACE)/bin/linux_amd64
-LINUX_ARM64_BUILD = GOOS=linux GOARCH=arm64 go build -buildmode=pie -ldflags="${LDFLAGS}" -o $(BUILD_SPACE)/bin/linux_arm64
-WIN_BUILD = GOOS=windows GOARCH=amd64 go build -buildmode=pie -ldflags="${LDFLAGS}" -o $(BUILD_SPACE)/bin/windows_amd64
+LINUX_AMD64_BUILD = CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -buildmode=${CWAGENT_BUILD_MODE} -ldflags="${LDFLAGS}" -o $(BUILD_SPACE)/bin/linux_amd64
```
pingleig, line 22: If build mode is not set I will get error like the following, I think we should ignore `-buildmode` entirely in that case.

```
go build -buildmode= a.go                                                                                                                                                                                                                   
buildmode= not supported
go build -buildmode=pie a.go # ok
```

There is a long list of supported build mode, I guess `default` is the default one we were using?

```
go help buildmode
```

Makefile
haojhcwa: Updated the default build mode to `default` instead of `pie`

Makefile
haojhcwa: The default build mode is `default`, which has been updated to be the default build mode.

pingleig (approved):
LGTM

PR #213, opened 2021-05-20
prom: Add labels for the task and cluster ARNs

arlyon:
# Description of the issue

Currently the only identifier of the source of an event is the `instance` field from prometheus. This allows you to use the TaskARN and ClusterARN as well by exposing them as labels. This means I can select them as dimensions for example.

# Description of changes

Add two labels for task arn and task cluster arn.

# License

- [x] yes

# Tests

I tried to test this on my AWS infra but for some reason the image outputted from `make dockerized-build` crashes

```
2021-05-20T17:03:09Z E! [telegraf] Error running agent: No config file specified
```

 while `amazon/cloudwatch-agent` doesn't. Seems completely unrelated to my changes, so any pointers here welcome.

### DIFF
--- a/internal/ecsservicediscovery/decoratedtask.go
+++ b/internal/ecsservicediscovery/decoratedtask.go
@@ -18,6 +18,8 @@ const (
 	taskFamilyLabel      = "TaskDefinitionFamily"
 	taskRevisionLabel    = "TaskRevision"
 	taskGroupLabel       = "TaskGroup"
+	taskARNLabel         = "TaskARN"
+	taskClusterARNLabel  = "TaskClusterARN"
 	taskStartedbyLabel   = "StartedBy"
 	taskLaunchTypeLabel  = "LaunchType"
 	taskJobNameLabel     = "job"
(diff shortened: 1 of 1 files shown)


pingleig:
The not found error is caused by https://github.com/aws/amazon-cloudwatch-agent/pull/212 we enabled PIE build mode by default and it will crash on scratch container because it is looking for dynamic linked libraries (i.e. CGO_ENABLED=0 does not produce a static binary when `buildmode=pie`)

You can rebase and the updated Makefile should work.

arlyon:
> The not found error is caused by #212 we enabled PIE build mode by default and it will crash on scratch container because it is looking for dynamic linked libraries (i.e. CGO_ENABLED=0 does not produce a static binary when `buildmode=pie`)
> 
> You can rebase and the updated Makefile should work.

Thanks for the quick reply, will make some time to test in a live environment today 👍

pingleig:
@arlyon any update?

arlyon:
Hi, sorry we've been busy and with a release and I haven't had the go-ahead to put this in prod until now but I am testing now.

PR #228, opened 2021-06-14
build: Build docker without open telemetry collector

pingleig:
# Description of the issue

We were using deb when building image from source, i.e. source -> deb -> docker image. But we have composite agent, which is not used when running as image.

# Description of changes

- Add two new targets (x86_64 and arm64) to only build binary required for docker image
- Make amd64 the default target
- Use the new target in `Dockerfile`

# License

# Tests

Run the new commands (in a place that does not block go mod proxy)

```
docker buildx build --platform linux/amd64,linux/arm64 -f amazon-cloudwatch-container-insights/cloudwatch-agent-dockerfile/Dockerfile .
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,10 +102,17 @@ config-downloader: copy-version-file
 	$(DARWIN_BUILD)/config-downloader github.com/aws/amazon-cloudwatch-agent/cmd/config-downloader
 
 # A fast build that only builds amd64, we don't need wizard and config downloader
-build-for-docker:
+build-for-docker: build-for-docker-amd64
+
+build-for-docker-amd64:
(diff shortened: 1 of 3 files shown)


epf:
Note that this produces dynamically linked arm64 binaries (which ultimately fail to execute because the container doesn't contain the required libraries).

Adding `CGO_ENABLED=0` to `LINUX_ARM64_BUILD` in the Makefile fixes this.

pingleig:
Actually things has been working fine (for our internal build) because we are not using dockerized build . We build arm64 binary on x86_64 host using cross compile, and [cgo is is disabled by default when cross-compiling](https://golang.org/cmd/cgo/#:~:text=The%20cgo%20tool%20is%20enabled,to%200%20to%20disable%20it.) But we should set it explicitly.

pingleig:
Fixed the arm cgo problem in https://github.com/aws/amazon-cloudwatch-agent/pull/235

pxaws (approved):

PR #233, opened 2021-06-15
Default config metrics exporter

jefchien:
# Description of the issue
In order to support EC2 metrics published by the CW Agent on Metrics Explorer, the dimensions of the metrics should be set as "Instance Id". This is something we achieve for the existing customers using some guideline docs. But for new customers who will onboard for the first time, we would like to offer this support by default by arranging these settings in the CW Agent config file.

# Description of changes
Added a yes/no prompt in `defaultConfig.go` as part of the config wizard to add the `aggregation_dimensions` field to the `metrics` config. By default, the `aggregation_dimensions` will have a single dimension of "Instance Id". Moved pricing disclaimer from prompt to welcome message.

# License

# Tests
Updated `defaultConfig_test.go` to account for the additional prompt and added the `aggregation_dimensions` field to the expected.
Tested the validity of the generated config and performed end-to-end test.

### DIFF
--- a/tool/data/config/metric/aggregationDimensions.go
+++ b/tool/data/config/metric/aggregationDimensions.go
@@ -1,0 +1,19 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+package metric
+
+import "github.com/aws/amazon-cloudwatch-agent/tool/runtime"
+
+type AggregationDimensions struct {
+	Dimensions [][]string
+}
+
+func (config *AggregationDimensions) ToMap(ctx *runtime.Context) (string, [][]string) {
+	config.SetDefaultDimensions()
+	return "aggregation_dimensions", config.Dimensions
+}
+
+func (config *AggregationDimensions) SetDefaultDimensions() {
+	config.Dimensions = [][]string{{"InstanceId"}}
+}
--- a/tool/data/config/metrics.go
+++ b/tool/data/config/metrics.go
@@ -11,6 +11,8 @@ import (
 type Metrics struct {
 	AppendDimensions *metric.AppendDimensions
 
+	AggregationDimensions *metric.AggregationDimensions
+
 	MetricsCollect *metric.Collection
 }
 
@@ -24,6 +26,17 @@ func (config *Metrics) ToMap(ctx *runtime.Context) (string, map[string]interface
 
 	if config.AppendDimensions != nil {
 		mapKey, mapValue := config.AppendDimensions.ToMap(ctx)
+		if mapValue != nil {
+			resultMap[mapKey] = mapValue
+		}
+	}
+
+	if config.AggregationDimensions == nil && ctx.WantAggregateDimensions {
+		config.AggregationDimensions = new(metric.AggregationDimensions)
+	}
+
+	if config.AggregationDimensions != nil {
+		mapKey, mapValue := config.AggregationDimensions.ToMap(ctx)
 		if mapValue != nil {
 			resultMap[mapKey] = mapValue
 		}
--- a/tool/data/config/metrics_test.go
+++ b/tool/data/config/metrics_test.go
@@ -16,7 +16,8 @@ import (
 func TestMetrics_ToMap(t *testing.T) {
 	expectedKey := "metrics"
 	expectedValue := map[string]interface{}{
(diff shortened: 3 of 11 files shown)


tool/data/config/metric/aggregationDimensions.go
```diff
@@ -0,0 +1,19 @@
...
+type AggregationDimensions struct {
+	Dimensions [][]string
+}
+
+func (config *AggregationDimensions) ToMap(ctx *runtime.Context) (string, [][]string) {
+	config.SetDefaultDimensions()
+	return "aggregation_dimensions", config.Dimensions
+}
+
+func (config *AggregationDimensions) SetDefaultDimensions() {
```
pingleig, line 17: It seems you only call `SetDefault...` in this file? Maybe just inline the func manually. Also we might have a constant string for `InstanceId` defined somewhere.

tool/data/config/metrics_test.go
```diff
@@ -16,7 +16,8 @@ import (
 func TestMetrics_ToMap(t *testing.T) {
 	expectedKey := "metrics"
 	expectedValue := map[string]interface{}{
-		"append_dimensions": map[string]interface{}{"ImageId": "${aws:ImageId}", "InstanceId": "${aws:InstanceId}", "InstanceType": "${aws:InstanceType}", "AutoScalingGroupName": "${aws:AutoScalingGroupName}"},
+		"aggregation_dimensions": [][]string{{"InstanceId"}},
```
pingleig, line 19: I not sure if this part is properly formatted.

pingleig (commented):
There are a bunch of format related diffs, I guess we didn't run go fmt on all files when we export them to github in first release? @haojhcwa

tool/data/config/metrics_test.go
jefchien: I agree that it looks off, but when I run `make fmt`, this is the way it's formatted.

tool/data/config/metric/aggregationDimensions.go
jefchien: It is only used in a single place, so inlining does make sense. I was following `appendDimensions.go`, so it'd probably make sense to change it in both places.

There's a constant string for `InstanceId` defined in `ruleInstanceId.go`, but it doesn't seem like it's used anywhere else except for within that file.

pingleig: I think hard code the string here is fine, you can add a constant within this file/package if you want to.

tool/data/config/metrics_test.go
pingleig: ok, there is no way to tweak gofmt anyway 😄

pingleig (approved):
LGTM but still need to wait @pxaws on the config translator logic before merging it.

Also @jefchien please update the description after you have validated the config generated by new wizard is working end to end. i.e. run the wizard and see right cw metrics dimension in console.

tool/data/config/metrics.go
```diff
@@ -29,6 +31,17 @@ func (config *Metrics) ToMap(ctx *runtime.Context) (string, map[string]interface
 		}
 	}
 
+	if config.AggregationDimensions == nil && ctx.WantAggregateDimensions {
```
pxaws, line 34: So this assume that the new added option should not coexist with other option like `WantEC2TagDimensions`? Do we ensure that in the wizard?

jefchien: This doesn't prevent `WantEC2TagDimensions` from coexisting. It just adds a separate prompt to add the `AggregationDimensions`. I believe that based on the documentation for setting up the config for metric explorer (https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Metrics-Explorer.html), both the `WantEC2TagDimensions` and `WantAggregateDimensions` options are required during setup, but I felt that keeping the two options separate made sense if they weren't configuring with that purpose in mind. The default would include both.

pxaws:
As @pingleig suggested, once you confirm that this works from end to end, we can merge the PR.

pingleig:
@jefchien any update on this?

jefchien:
@pingleig Sorry for the delay. I've validated the configuration and behavior. Feel free to merge it when ready.

pingleig:
@jefchien btw: remember to add milestone for this PR

pingleig:
@jefchien it seems the test are failing, though we do have flaky tests ...  https://github.com/aws/amazon-cloudwatch-agent/actions/runs/984367255

Issue #234, opened 2021-06-18
PodName of a pod from a job is incorrectly truncated


victornoel:
I have jobs named like so with dashes in it:
- `some-prefix.some-middle-name.a-name`
- `some-prefix.some-middle-name.another-name`
- `some-prefix.some-other-name.a-name`
- `other-prefix.some-name.name`
- etc

And it creates pods as expected with the following names `some-prefix.some-middle-name.a-name-vmrpg` and so on.

Obviously we would like that the inferred name for the pod to be the name of the job itself, but in practice, it gets to be `some` or `other`. As you can imagine, this is quite useless to be able to use effectively the information exported by cloudwatch agent :)

Is that desired? It looks to be a bug for me.

pxaws:
Could you share the deployment yaml file so that we understand how the pod name is composed? The logic for parsing the pod name from job is here: https://github.com/aws/amazon-cloudwatch-agent/blob/ae7730ceeea112b2fc6cda79d45b32a7ea6dadaa/plugins/processors/k8sdecorator/stores/podstore.go#L555-L561. I would expect `some-prefix.some-middle-name.a-name-vmrpg` gives a pod name of `some-prefix.some-middle-name.a-name` instead based on code.

victornoel:
@pxaws thank you. Some more context: we are using the images `amazon/cloudwatch-agent:1.247346.1b249759` and `amazon/aws-for-fluent-bit:latest`.

here is the job description (I dropped some parts related to the container themselves):
```yaml
kind: Job
apiVersion: batch/v1
metadata:
  name: staging-eu.demo-poc.execution-engine
  namespace: prod
spec:
  parallelism: 1
  completions: 1
  backoffLimit: 0
  selector:
    matchLabels:
      controller-uid: d9b5511f-2c17-4c85-b4dd-32bb97e5c54f
  template:
    metadata:
      labels:
        controller-uid: d9b5511f-2c17-4c85-b4dd-32bb97e5c54f
        job-name: staging-eu.demo-poc.execution-engine
      annotations:
        cluster-autoscaler.kubernetes.io/safe-to-evict: 'false'
    spec:
      volumes:
        - name: secrets
          secret:
            secretName: staging-eu.demo-poc.execution-engine
            defaultMode: 420
      containers: {}
      restartPolicy: Never
      terminationGracePeriodSeconds: 30
      dnsPolicy: ClusterFirst
      securityContext: {}
      schedulerName: default-scheduler
```

[...]

PR #236, opened 2021-06-23
fix: properly generate PodName for Pods owned by a Job

victornoel:
# Description of the issue

See #234 

# Description of changes

Two changes:
- fix the bug
- remove tests with values that are not possible (see https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names)

# License

# Tests

Updated existing tests to add more examples.

### DIFF
--- a/plugins/processors/k8sdecorator/stores/podstore_test.go
+++ b/plugins/processors/k8sdecorator/stores/podstore_test.go
@@ -426,19 +426,6 @@ func (client *MockClient2) Shutdown() {
 // Mock client 2 end
 //
 
-func TestGetJobNamePrefix(t *testing.T) {
-	assert.Equal(t, "abcd", getJobNamePrefix("abcd-efg"))
-	assert.Equal(t, "abcd", getJobNamePrefix("abcd.efg"))
-	assert.Equal(t, "abcd", getJobNamePrefix("abcd-e.fg"))
-	assert.Equal(t, "abc", getJobNamePrefix("abc.d-efg"))
-	assert.Equal(t, "abcd", getJobNamePrefix("abcd-.efg"))
-	assert.Equal(t, "abcd", getJobNamePrefix("abcd.-efg"))
-	assert.Equal(t, "abcdefg", getJobNamePrefix("abcdefg"))
-	assert.Equal(t, "abcdefg", getJobNamePrefix("abcdefg-"))
-	assert.Equal(t, "", getJobNamePrefix(".abcd-efg"))
-	assert.Equal(t, "", getJobNamePrefix(""))
-}
-
 func TestPodStore_addPodOwnersAndPodNameFallback(t *testing.T) {
 	k8sclient.Get = mockGet2
 	mockClient2.On("JobToCronJob").Return(map[string]string{})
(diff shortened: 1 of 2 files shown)


victornoel:
Extra note: I'm not 100% sure this is really the way to go… maybe instead we should just take the job name and not bother with some regexp magic?

plugins/processors/k8sdecorator/stores/podstore_test.go
```diff
@@ -429,13 +429,12 @@ func (client *MockClient2) Shutdown() {
 func TestGetJobNamePrefix(t *testing.T) {
 	assert.Equal(t, "abcd", getJobNamePrefix("abcd-efg"))
 	assert.Equal(t, "abcd", getJobNamePrefix("abcd.efg"))
-	assert.Equal(t, "abcd", getJobNamePrefix("abcd-e.fg"))
-	assert.Equal(t, "abc", getJobNamePrefix("abc.d-efg"))
-	assert.Equal(t, "abcd", getJobNamePrefix("abcd-.efg"))
-	assert.Equal(t, "abcd", getJobNamePrefix("abcd.-efg"))
```
pxaws, line 435: Could you explain why those test cases are wrong?

pxaws:
> Extra note: I'm not 100% sure this is really the way to go… maybe instead we should just take the job name and not bother with some regexp magic?

I am not sure how to get the job name based on the pod name. As far as I know, you need to call k8s api server to get the mapping. This is what we are trying to avoid in order not to overwhelm the api server especially in a big cluster with thousands of nodes.

plugins/processors/k8sdecorator/stores/podstore_test.go
victornoel: according to what I read, it's not possible to have . and - next to each other in DNS subdomain name, the regexp given by k8s when creating such a resource (in the error message) is: `[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*`.

So I made a mistake I think for the first two apparently, I will fix it.

victornoel:
> > Extra note: I'm not 100% sure this is really the way to go… maybe instead we should just take the job name and not bother with some regexp magic?
> 
> I am not sure how to get the job name based on the pod name. As far as I know, you need to call k8s api server to get the mapping. This is what we are trying to avoid in order not to overwhelm the api server especially in a big cluster with thousands of nodes.

Ah of course, I better understand. So from what I read, it's indeed quite hard to have a deterministic way of finding the name. There are two cases:
- `generateName` is used in the Job description and a suffix is added (apparently 5 random chars: https://github.com/kubernetes/kubernetes/blob/c9fb3c8a1b3f407a5e84562843780aa3047d7d06/staging/src/k8s.io/apiserver/pkg/storage/names/generate.go#L53
- `generateName` is not used, and the name of the job is taken as a prefix, followed by `-` then the 5 random chars as above

So I'm wondering:
- if by convention people do use `-` as a separator before the 5 random chars
- if we don't want to rely on the knowledge it's 5 random chars at the end
- then maybe it would be enough to split based on the last `-` found and ignore `.`?

What do you think of this solution?

Another alternative is to rely on the label added to pod named `job-name`, but I don't know if it is a stable enough API though... to be investigated.

Thank you for the review!

plugins/processors/k8sdecorator/stores/podstore_test.go
victornoel: actually it's ok, I didn't delete the first two, only changed the expected string

victornoel:
@pxaws actuall, the ownerReference of a pod contains the name of the owner, shouldn't we be using that?

victornoel:
@pxaws IMHO there are a few incorrect assumption in the tests of podstore:
- The `Pod`s of a `Job` named `jobname` are named `jobname-7xsin` and their `ownerReference` name is `jobname` itself
- This is different from `CronJob` that creates `Job`s with name `cronjobname-87849302928` and pods with name `cronjobname-87849302928-7xsin`

victornoel:
@pxaws I've updated the PR in the light of my above comments.

There is just one open question to me: what is the use of this `prefFullPodName` option? I have trouble to see why it's really needed… I'm not even sure I properly implemented the corresponding behaviour.

victornoel:
Hi @pxaws, any news about this? Maybe @pingleig or @darkhelmet that worked on this file could help too?

pxaws:
This behavior is by design and I suggest to set `prefer_full_pod_name` to true to get the correct podname.

PR #237, opened 2021-06-25
Support SSM SecureString parameters.

jefchien:
# Description of the issue
Currently, if the configuration file is stored in a `ParameterStore` as a `SecureString`, the `config-downloader` will successfully retrieve the file, but will not decrypt it, so the translator is unable to successfully parse it.
```
Reading json config file path: /opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.d/ssm_SecureStringConfig.tmp ...
unable to scan config dir /opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.d with error: unable to parse json, error: invalid character 'A' looking for beginning of value
```

# Description of changes
Updated `ssm.GetParameter` to include the `WithDecryption` flag to support `SecureString` parameters. According to the SSM API (https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_GetParameter.html), the flag is ignored for `String` and `StringList` parameter types, so it can always be set to true.

# License

# Tests
Created a `SecureString` parameter in SSM along with a `String` parameter and tested both against the existing setup to confirm the issue. Built and replaced `config-downloader` on the EC2 instance and tested the download/translation again to validate the fix.

### DIFF
--- a/cmd/config-downloader/downloader.go
+++ b/cmd/config-downloader/downloader.go
@@ -66,9 +66,9 @@ func downloadFromSSM(region, parameterStoreName, mode string, credsConfig map[st
 
 	ssmClient := ssm.New(ses)
 	input := ssm.GetParameterInput{
+		Name:           aws.String(parameterStoreName),
 		WithDecryption: aws.Bool(true),
 	}
-	input.SetName(parameterStoreName)
 	output, err := ssmClient.GetParameter(&input)
 	if err != nil {
 		fmt.Printf("Error in retrieving parameter store content: %v\n", err)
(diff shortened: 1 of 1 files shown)


cmd/config-downloader/downloader.go
```diff
@@ -65,7 +65,9 @@ func downloadFromSSM(region, parameterStoreName, mode string, credsConfig map[st
 	}
 
 	ssmClient := ssm.New(ses)
-	input := ssm.GetParameterInput{}
+	input := ssm.GetParameterInput{
+		WithDecryption: aws.Bool(true),
```
pingleig, line 69: Actually can we do and avoid the `SetName` call. Init struct in literal way is more common in go.

```go
input := ssm.GetParameterInput{
		Name: aws.String(parameterStoreName),
		WithDecryption: aws.Bool(true),
	}
```

pingleig (changes requested):

pingleig (approved):
LGTM

PR #241, opened 2021-07-15
Change ownership of files excluding executable files and Suid Sgid files.

ZhenyuTan-amz:
# Description of the issue
The Cloudwatch agent change ownership of directories to the user the service is running as.

# Description of changes
Excluding executable file and Suid Sgid when change file ownership with chown. Prevent possible privilege escalation. Files with o+w permission are also skipped

# License

# Tests
Added unit test that creates regular files, executable files, symbol links and verified the behavior of ownership change does not affect executable files and suid/sgid files.

### DIFF
--- a/translator/cmdutil/userutil.go
+++ b/translator/cmdutil/userutil.go
@@ -8,17 +8,22 @@ package cmdutil
 import (
 	"fmt"
 	"log"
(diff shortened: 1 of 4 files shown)


jefchien (approved):
LGTM

PR #242, opened 2021-07-15
Added run_as_user to UserAgent

jefchien:
# Description of the issue
Indicate if run_as_user is being used in the agent configuration.

# Description of changes
Added a run_as_user tag to the input plugins in the UserAgent if the agent isn't running as root.

# License

# Tests
Updated `info_test.go` unit test to account for run_as_user in the input plugins.

### DIFF
--- a/cfg/agentinfo/info.go
+++ b/cfg/agentinfo/info.go
@@ -19,6 +19,10 @@ const versionFilename = "CWAGENT_VERSION"
 // We will fall back to a major version if no valid version file is found
 const fallbackVersion = "1"
 
+var isRunningAsRoot = func() bool {
+	return os.Getuid() == 0
+}
+
 var (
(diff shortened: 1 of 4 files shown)


cfg/agentinfo/info.go
```diff
@@ -19,6 +19,10 @@ const versionFilename = "CWAGENT_VERSION"
 // We will fall back to a major version if no valid version file is found
 const fallbackVersion = "1"
 
+var isRunningAsRoot = func() bool {
+	return os.Getuid() == 0
```
jefchien, line 23: Should this account for the OS? I don't believe run_as_user is supported on windows hosts and `isRunningAsRoot` will always be false.

ZhenyuTan-amz (approved):

Issue #217, opened 2021-06-01
[Feature Request] Possibility to specify retention period in Cloud Watch log group names


volphy:
It would be nice to have a new configuration option added to the logs section:
https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Agent-Configuration-File-Details.html

It could be called:
log_group_retention
It should allow specifying a non-default retention period setting.

I have observed that using Amazon Cloud Watch Agent (amazon-cloudwatch-agent-1.247347.4-1.amzn2.x86_64 RPM) creates a new Cloud Watch log group without any retention period set ('Never expire').

jefchien:
I'll take a look at this.

dacodas:
Would love this, and am pumped to see the MR go through

PR #250, opened 2021-07-29
Put retention policy when writing or creating a log group

aateeqi:
# Add functionality for editing log retention policy

Users wish to edit log group retention policy via CWAgent config 

#217

# Description of changes

- Add ability to specify `retention_in_days` config inside `amazon-cloudwatch-agent.json` config file.
- Calls [PutRetentionPolicy](https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_PutRetentionPolicy.html) after creating or updating a log group

# License

# Tests
- Created Unit tests for retentionInDays rule
- Added unit tests in pusher for putRetentionPolicy function


plugins/outputs/cloudwatchlogs/pusher.go
```diff
@@ -224,6 +228,7 @@ func (p *pusher) send() {
 	for {
 		input.SequenceToken = p.sequenceToken
 		output, err := p.Service.PutLogEvents(input)
+		p.Log.Infof("calling put log events %v", p.Group)
```
ZhenyuTan-amz, line 231: Considering the volume of this call, maybe not a good idea to log every put log events? I'm afraid this could flood the agent log, or maybe make it debug level?

plugins/outputs/cloudwatchlogs/pusher.go
```diff
@@ -53,21 +55,23 @@ type pusher struct {
...
+		FlushTimeout:    flushTimeout,
+		RetryDuration:   retryDuration,
+		Log:             logger,
+		Retention:       retention,
 		events:          make([]*cloudwatchlogs.InputLogEvent, 0, 10),
 		eventsCh:        make(chan logs.LogEvent, 100),
 		flushTimer:      time.NewTimer(flushTimeout),
 		stop:            make(chan struct{}),
 		startNonBlockCh: make(chan struct{}),
+		retentionSet:    map[string]int{},
```
ZhenyuTan-amz, line 74: Should this be rename to retentionMap or maybe more descriptive LogGroupRetentionMap?

translator/isValid.go
```diff
@@ -63,3 +63,12 @@ func ResetMessages() {
 	ErrorMessages = make([]string, 0)
 	InfoMessages = make([]string, 0)
 }
+
+// ValidDays represents the valid possible values for retentionInDays
+var ValidDays = map[int]bool{1: true, 3: true, 5: true, 7: true, 14: true,
```
ZhenyuTan-amz, line 68: Should -1 included in the list or not? Also if every entry in this map has a value true, why not using a Set instead?

logs/logs.go
```diff
@@ -115,6 +118,19 @@ func (l *LogAgent) Run(ctx context.Context) {
 	}
 }
 
+// func adjustRetentions(srcs []LogSrc) []LogSrc{
```
ZhenyuTan-amz, line 121: nit: why we are keeping the commented code here?

plugins/inputs/logfile/logfile.go
```diff
@@ -394,6 +395,25 @@ func (t *LogFile) cleanUpStoppedTailerSrc() {
...
+		if fileconfig.LogGroupName != "" {
+			configSet[fileconfig.LogGroupName] += 1
+		}
+
+	}
+	for i := range t.FileConfig {
+		fileconfig := &t.FileConfig[i]
+
+		if fileconfig.LogGroupName != "" && configSet[fileconfig.LogGroupName] > 1 {
+			fileconfig.RetentionInDays = -1
```
ZhenyuTan-amz, line 412: Can you explain a little why we choose to use -1 to mark Retention to be ignored over 0. Guess that don't make too much of difference but I'm curious if there's any particular reason.

ZhenyuTan-amz (commented):
Mostly nit picking comments, overall logic looks good to me.

plugins/outputs/cloudwatchlogs/pusher.go
aateeqi: thanks, I put these as i was debugging, deleting them now

plugins/inputs/logfile/logfile.go
aateeqi: I felt like -1 would be more indicative of `never expire` than 0 would. Also I felt like -1 would be a bit less bug prone incase we have any unset values when Retention would be undefined or 0

plugins/outputs/cloudwatchlogs/pusher_test.go
```diff
@@ -44,10 +45,16 @@ func (s *svcMock) CreateLogStream(in *cloudwatchlogs.CreateLogStreamInput) (*clo
 	}
 	return nil, nil
 }
+func (s *svcMock) PutRetentionPolicy(in *cloudwatchlogs.PutRetentionPolicyInput) (*cloudwatchlogs.PutRetentionPolicyOutput, error) {
```
jefchien, line 48: Should probably add a test for the put retention policy aside from just setting up the mock.

tool/data/config/logs/config.go
```diff
@@ -34,5 +35,8 @@ func (config *Config) ToMap(ctx *runtime.Context) (string, map[string]interface{
 	if config.Encoding != "" {
 		resultMap["encoding"] = config.Encoding
 	}
+	if config.Retention != 0 {
```
jefchien, line 38: Should this be <= 0 to account for negative values or is that checked beforehand?

tool/data/config/logs/eventConfig_test.go
```diff
@@ -23,6 +25,6 @@ func TestEventConfig_ToMap(t *testing.T) {
 		"event_name":      "System",
 		"event_levels":    []string{"INFORMATION", "WARNING", "ERROR", "SUCCESS"},
 		"log_group_name":  "SystemGroup",
-		"log_stream_name": "SystemStream"},
+		"log_stream_name": "SystemStream", "retention_in_days": 1},
```
jefchien, line 28: This should be on its own line. I thought the formatter would catch this.

tool/processors/migration/linux/knownConfigKeys.go
```diff
@@ -81,5 +81,16 @@ func addLogConfig(logsConfig *config.Logs, filePath, section string, p *configpa
...
 		}
 	}
-	logsConfig.AddLogFile(logFilePath, logGroupName, logStreamName, timestampFormat, timezone, multiLineStartPattern, encoding)
+	retentionInDays, _ := p.Get(section, "retention_in_days")
+	retention := -1
+	if retentionInDays != "" {
+		if i, err := strconv.Atoi(retentionInDays); err == nil {
+			fmt.Printf("Retention is valid from config and value of %v", i)
+			retention = i
+		} else if err != nil {
```
jefchien, line 90: Do you need the else if here? Can't it just be an else since at this point you know that err is not nil?

tool/processors/question/events/events.go
```diff
@@ -82,8 +83,14 @@ func monitorEvents(ctx *runtime.Context, config *data.Config) {
 			eventFormat = EventFormatPlainText
 			eventFormatDefaultOption = 2
 		}
+		retentionInDays := util.Choice("Log Group Retention in days", 1, []string{"-1", "1", "3", "5", "7", "14", "30", "60", "90", "120", "150", "180", "365", "400", "545", "731", "1827", "3653"})
```
jefchien, line 86: Is it really necessary to restrict the choices in this way?

translator/totomlconfig/sampleConfig/complete_linux_config.conf
```diff
@@ -57,6 +57,7 @@
       log_group_name = "amazon-cloudwatch-agent.log"
       log_stream_name = "amazon-cloudwatch-agent.log"
       pipe = false
+      retention_in_days = 1
```
jefchien, line 60: Is this meant to be -1? Why are some 1 and some -1?

translator/translate/logs/logs_collected/files/collect_list/ruleRetentionInDays.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+const RetentionInDaysSectionKey = "retention_in_days"
+
+type RetentionInDays struct {
+}
+
+func (f *RetentionInDays) ApplyRule(input interface{}) (returnKey string, returnVal interface{}) {
+	_, returnVal = translator.DefaultIntegralCase(RetentionInDaysSectionKey, float64(-1), input)
+	if returnVal == "" {
```
jefchien, line 17: Will this ever be an empty string? Isn't the default value the -1 you pass in?

logs/logs.go
```diff
@@ -74,6 +75,7 @@ func (l *LogAgent) Run(ctx context.Context) {
 		if !ok {
 			continue
 		}
+		log.Printf("output is %v", output.Config)
```
jefchien, line 78: Might not want to log the entire config.

jefchien (changes requested):
Overall, it looks good. I think some additional tests would be useful to validate your changes.

Also, update your description.

translator/isValid.go
jefchien: It seems like Golang doesn't have builtin Sets, but some of our modules have defined Sets (e.g. https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#String) that we can use.

tool/processors/question/events/events.go
aateeqi: This is based on whats allowed on the [ cw logs api ](https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_PutRetentionPolicy.html) but I believe that the user should be allowed more flexibility and their looking into it

jefchien: Right, but validation of these values isn't done anywhere else in the code.

aateeqi: I had written the `isValidRetention` function for that purpose but I'm realizing I forgot to use it in the `knownConfigKeys.go`

sethAmazon:
Is there a default so that for people upgrading do not have to change their config?

aateeqi:
> Is there a default so that for people upgrading do not have to change their config?

@sethAmazon Yes, If you dont include the retention_In_days param its automatically -1 (no retention)

plugins/outputs/cloudwatchlogs/pusher.go
```diff
@@ -336,6 +341,20 @@ func (p *pusher) createLogGroupAndStream() error {
...
 
+func (p *pusher) putRetentionPolicy() {
+	if p.Retention > 0 {
+		i := aws.Int64(int64(p.Retention))
+		putRetentionInput := &cloudwatchlogs.PutRetentionPolicyInput{
+			LogGroupName:    &p.Group,
+			RetentionInDays: i,
+		}
+		_, err := p.Service.PutRetentionPolicy(putRetentionInput)
+		if err != nil {
```
SaxyPandaBear, line 352: So we proceed with initializing the agent even if we fail to create the retention policy?

plugins/inputs/logfile/logfile.go
```diff
@@ -394,6 +398,26 @@ func (t *LogFile) cleanUpStoppedTailerSrc() {
 	}
 }
 
+func (t *LogFile) unsetRetentionsForDuplicateLogGroups() error {
```
SaxyPandaBear, line 401: I _think_ I understand what this does, but can you add tests for this?

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -79,22 +82,25 @@ func (c *CloudWatchLogs) Write(metrics []telegraf.Metric) error {
...
+	if retention <= 0 {
+		retention = -1
+	}
 
 	t := Target{
 		Group:  group,
 		Stream: stream,
 	}
-	return c.getDest(t)
+	return c.getDest(t, retention)
```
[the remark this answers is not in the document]
SaxyPandaBear: Agreed. I don't see any reason you wouldn't just put the retention int in the Target struct. If the log group and log stream were separate parameters for the `getDest()` function, I could see why you wouldn't, but since they're being wrapped in the struct, the retention period should as well.

tool/processors/question/events/events.go
SaxyPandaBear: I'd prefer abstracting away this so that we don't parrot what the "correct" values are here, since they are subject to change. If there's another place where we actually perform the validation, this logic should be in there.

tool/processors/question/logs/logs.go
```diff
@@ -51,7 +53,14 @@ func monitorLogs(ctx *runtime.Context, config *data.Config) {
 			logStreamNameHint = "{hostname}"
 		}
 		logStreamName := util.AskWithDefault("Log stream name:", logStreamNameHint)
-		logsConf.AddLogFile(logFilePath, logGroupName, logStreamName, "", "", "", "")
+		retentionInDays := util.Choice("Log Group Retention in days", 1, []string{"-1", "1", "3", "5", "7", "14", "30", "60", "90", "120", "150", "180", "365", "400", "545", "731", "1827", "3653"})
```
SaxyPandaBear, line 56: Same conversation as the other thread about where/how this gets validated

plugins/outputs/cloudwatchlogs/pusher.go
aateeqi: right, we don't want to stop the agent from writing logs even though the retention policy setting was not successful

plugins/inputs/logfile/logfile.go
aateeqi: Changed the function to fail based on recent discussions. Also added test, thanks

translator/defaultKeyCase.go
```diff
@@ -62,3 +62,20 @@ func DefaultStringArrayCase(key string, defaultVal, input interface{}) (returnKe
...
+		if IsValidRetentionDays(intVal) {
+			returnVal = intVal
+		} else {
+			returnVal = -1
+		}
+	} else {
+		returnVal = -1
+		AddErrorMessages(
+			fmt.Sprintf("Retention in Days key: %s", key),
+			fmt.Sprintf("%s value (%v) in is not valid retention in days.", key, returnVal))
```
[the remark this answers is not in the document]
aateeqi: right, I think this was causing an issue with an edge case but i removed it now.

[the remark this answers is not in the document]
SaxyPandaBear: Ignore me. I just realized the `&&` is there.

translator/isValid.go
```diff
@@ -63,3 +65,22 @@ func ResetMessages() {
...
+	for i := 0; i < len(ValidRetentionInDays); i++ {
+		val, err := strconv.Atoi(ValidRetentionInDays[i])
+		if err == nil {
+			ValidDays[val] = true
+		}
+	}
+}
+
+func IsValidRetentionDays(days int) bool {
+	initializeValidMap()
```
SaxyPandaBear, line 84: Do we have to initialize this every time? How often does this get called?

plugins/inputs/logfile/logfile_test.go
```diff
@@ -1063,3 +1063,22 @@ func TestGenerateLogGroupName(t *testing.T) {
...
+	tt := NewLogFile()
+	logGroupName := "DuplicateLogGroupName"
+	tt.FileConfig = []FileConfig{{
+		FilePath:        "SampleFilePath",
+		FromBeginning:   true,
+		LogGroupName:    logGroupName,
+		RetentionInDays: 1,
+	},
+		{
+			FilePath:        "SampleFilePath",
```
SaxyPandaBear, line 1077: One thing I just thought about - do we have to account for case sensitivity for file paths? IIRC Linux/Unix are case sensitive but Windows is case insensitive. Given that, I think it's possible for this to not work on Windows. We build a map of log group names to count their occurrences, but I _know_ that the internal map keys are case sensitive.

aateeqi: This is a good consideration specifically for log group names. I changed it to normalize strings as lowercase for logGroupNames.

jefchien (approved):

PR #268, opened 2021-09-13
Test Deployment Of Cloud Watch Agent

sethAmazon:
Brake Up Deployment Into Steps

Add Update Intgration Test Job

Run End To End Tests

Add Needs To Run Integration Test

Add Log Output Of Tests

Have Deployment Generate Keys

Force Recreate

Add Terraform Support In Github Actions

Add Terraform Destroy

Always Run Terraform Destroy After RunIntegrationTest

Add Vars To Terraform

Echo The Host Name

Add Sleep For Host Initialize:

Use Raw Output

Save Terraform State To Variable

Wrap Cat Command

Destroy All Resources

# Description of the issue
Need integration test proving providing a ca bundle works

# Description of changes
Add Integration test proving ca bundle works

# License

# Tests
Automated test using github actions


sethAmazon:
Other option instead of forking the localstack is to use secrets to place a conf file inside local stack and docker-compose.

integration/test/ca_bundle/ca_bundle_test.go
```diff
@@ -0,0 +1,114 @@
...
+	for _, parameter := range parameters {
+		//before test run
+		log.Printf("resource file location %s find target %t", parameter.dataInput, parameter.findTarget)
+		clearLogFile()
+		t.Run(fmt.Sprintf("resource file location %s find target %t", parameter.dataInput, parameter.findTarget), func(t *testing.T) {
+			copyFile(parameter.dataInput + configJSON, configOutputPath)
+			copyFile(parameter.dataInput + commonConfigTOML, commonConfigOutputPath)
+			startTheAgent();
+			time.Sleep(agentRuntime * time.Minute);
+			log.Printf("Agent has been running for : %d", agentRuntime);
```
SaxyPandaBear, line 47: I initially had a gripe that we're storing a plain int in `agentRuntime`, but got over it because I think it's a typical use case to do `num * time.Unit` for a duration, however here we're logging that the agent has been running for n... without the unit of measure. And I don't like the idea of hardcoding the string `"minute(s)"` into the log if we don't have to (could accidentally miss that if we change the sleep time later). 

Would you be opposed to initializing a duration before the sleep so we can reference it in the print log after?
```
d := agentRuntime * time.Minute
time.Sleep(d);
log.Printf("Agent has been running for %s, d.String());
```
or something like that - though, I think it might just be simpler to change `agentRuntime` to be the duration value itself.

integration/test/ca_bundle/ca_bundle_test.go
```diff
@@ -0,0 +1,114 @@
...
+	log.Printf("Agent is stopped");
+}
+
+func readTheOutputLog() bool {
+	logFile, err := ioutil.ReadFile(outputLog)
+	if err != nil {
+		log.Fatal(err)
+	}
+	log.Printf("Log file %s", string(logFile))
+	log.Printf("Finished reading log file")
```
SaxyPandaBear, line 110: Do we need this log? wouldn't the log above it already prove that we finished reading the log file? Separately, I don't think this needs to be a `Printf()` because there's nothing to format in there.

integration/test/empty/empty_test.go
```diff
@@ -0,0 +1,7 @@
+package empty
+
+import "testing"
+
+func TestEmpty(t *testing.T) {
```
SaxyPandaBear, line 5: What's this for?

.github/workflows/integrationTest.yml
```diff
@@ -0,0 +1,195 @@
+name: Deploy app to ubuntu server
+
+on: push
+
+jobs:
+
+  Terraform:
+    name: 'Terraform'
+    runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
sethAmazon: The reason why I chose to run on ubuntu over other linux distro is for github actions it seems ubuntu is the default linux. This will support more distro in the future such as al2.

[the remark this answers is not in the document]
sethAmazon: The runner can be forever ubuntu since we only care about the instances on terraform. By support al2 I mean add al2 to the terraform list to run on.

integration/localstack/docker-compose.yml
```diff
@@ -0,0 +1,25 @@
...
+  localstack:
+    container_name: "${LOCALSTACK_DOCKER_NAME-localstack_main}"
+    image: localstack/localstack
+    network_mode: bridge
+    ports:
+      - "127.0.0.1:53:53"
+      - "127.0.0.1:53:53/udp"
+      - "127.0.0.1:443:443"
+      - "127.0.0.1:4566:4566"
+      - "127.0.0.1:4571:4571"
```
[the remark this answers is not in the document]
sethAmazon: That is ripped from their repo https://github.com/localstack/localstack/blob/master/docker-compose.yml

integration/test/ca_bundle/ca_bundle_test.go
```diff
@@ -0,0 +1,114 @@
...
+		t.Run(fmt.Sprintf("resource file location %s find target %t", parameter.dataInput, parameter.findTarget), func(t *testing.T) {
+			copyFile(parameter.dataInput + configJSON, configOutputPath)
+			copyFile(parameter.dataInput + commonConfigTOML, commonConfigOutputPath)
+			startTheAgent();
+			time.Sleep(agentRuntime * time.Minute);
+			log.Printf("Agent has been running for : %d", agentRuntime);
+			stopTheAgent();
+			containsTarget := readTheOutputLog()
+			if (parameter.findTarget && !containsTarget) || (!parameter.findTarget && containsTarget) {
+				t.Errorf("Find target is %t contains target is %t", parameter.findTarget, containsTarget)
```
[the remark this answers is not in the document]
sethAmazon: const targetString = "x509: certificate signed by unknown authority"

This is the target for reading from the log. When I pass the proper ssl key it should connect thus this will not be in the log. Maybe I can leave a comment in the code

integration/test/empty/empty_test.go
sethAmazon: I like to have an empty test as a sanity check to make sure there is no issue with starting / running tests

sethAmazon: Will add comment

sethAmazon: Maybe this is just a bad habit of mine from my spring days where application context was a bigger deal for integration tests. As go does not have bean wiring.

.github/workflows/integrationTest.yml
```diff
@@ -0,0 +1,195 @@
+name: Deploy app to ubuntu server
+
+on: push
```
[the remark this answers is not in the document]
sethAmazon: The repo should block all not authorized users from pushing to a branch. This will allow testing on forked repo using only the forked repo's secrets.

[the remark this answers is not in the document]
sethAmazon: I have not looked at the security blocks for branch of this repo that is just my assumption based on experience with repos.

integration/test/empty/empty_test.go
SaxyPandaBear: You're free from the clutches of Spring now. There's nothing to `@Autowire` anymore. Let go. 😃 

I'm not opposed to having it if we include a comment why it's here. It just popped out to me as odd when I saw it during review.

.github/workflows/integrationTest.yml
```diff
@@ -0,0 +1,200 @@
...
+          PRIVATE_KEY: ${{ secrets.AWS_PRIVATE_KEY  }}
+          HOSTNAME: ${{ needs.Terraform.outputs.hostname }}
+          USER_NAME : ${{ secrets.USER_NAME  }}
+
+        run: |
+          echo "$PRIVATE_KEY" > private_key && chmod 600 private_key
+          ssh -o StrictHostKeyChecking=no -i private_key ${USER_NAME}@${HOSTNAME} '
+            cd /home/ubuntu &&
+            chmod +x clone-repo.sh &&
+            sh clone-repo.sh &&
```
jhnlsn, line 80: Where is this file: `clone-repo.sh`?

.github/workflows/integrationTest.yml
```diff
@@ -0,0 +1,200 @@
...
+        env:
+          PRIVATE_KEY: ${{ secrets.AWS_PRIVATE_KEY  }}
+          HOSTNAME: ${{ needs.Terraform.outputs.hostname }}
+          USER_NAME : ${{ secrets.USER_NAME  }}
+
+        run: |
+          echo "$PRIVATE_KEY" > private_key && chmod 600 private_key &&
+          echo git clone $GITHUB_SERVER_URL/$GITHUB_REPOSITORY.git > clone-repo.sh &&
+          echo $GITHUB_SHA > GITHUB_SHA.txt &&
+          scp -o StrictHostKeyChecking=no -i private_key clone-repo.sh ${USER_NAME}@${HOSTNAME}:/home/ubuntu/clone-repo.sh &&
```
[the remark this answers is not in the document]
sethAmazon: I need a way pass in the git hash to the environment.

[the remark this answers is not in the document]
sethAmazon: If I set up ssm I can I think I can use the run command instead. This will require giving the iam user admin privileges. Thoughts?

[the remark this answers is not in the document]
sethAmazon: You can't pass variable values directly through ssh unless you allow on the server side.

integration/test/ca_bundle/ca_bundle_test.go
```diff
@@ -0,0 +1,117 @@
+package ca_bundle
```
adam-mateen, line 1: We should probably add
```
// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
// SPDX-License-Identifier: MIT

// +build linux
```
at the top of this go file.
Based on all the paths and the calls to `exec.Command()` I'd say it won't run on a windows host.

.github/workflows/integrationTest.yml
```diff
@@ -0,0 +1,200 @@
+name: Deploy app to ubuntu server
+
```
jhnlsn, line 2: see `build.yml` for event structure.  Should be push branch master and pull_request with specific types.

sethAmazon: We can't run this flow on pr since this is a public repo. Github secrets can't be shared. Each person will need to provide their own secrets to have this work on pushes.

sethAmazon: Made the change

.github/workflows/integrationTest.yml
```diff
@@ -0,0 +1,156 @@
...
+        run: |
+          echo "${STATE}" &&
+          echo "${STATE}" > terraform.tfstate
+
+      - name: Terraform init
+        env:
+          TERRAFORM_AWS_ACCESS_KEY_ID: ${{ secrets.TERRAFORM_AWS_ACCESS_KEY_ID }}
+          TERRAFORM_AWS_SECRET_ACCESS_KEY: ${{ secrets.TERRAFORM_AWS_SECRET_ACCESS_KEY }}
+        run: terraform init -var="aws_access_key=${TERRAFORM_AWS_ACCESS_KEY_ID}" -var="aws_secret_access_key=${TERRAFORM_AWS_SECRET_ACCESS_KEY}"
```
jefchien, line 150: Do you need to do an init before running destroy? Is the state file not enough?
> The terraform destroy command destroys all of the resources being managed by the current working directory and workspace, **using state data to determine which real world objects correspond to managed resources.**

https://www.terraform.io/docs/cli/run/index.html#destroying

integration/test/ca_bundle/resources/integration/ssl/without/bundle/common-config.toml
```diff
@@ -0,0 +1,24 @@
+# This common-config is used to configure items used for both ssm and cloudwatch access
```
jefchien, line 1: This is just a stub file right? Can't it just be empty?

integration/test/ca_bundle/ca_bundle_test.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	cmd := exec.Command("bash", "-c", "sudo cp " + pathInAbs + " " + pathOut)
+	err = cmd.Run()
+	if err != nil {
+		log.Fatal(err)
+	}
+	log.Printf("File : %s copied to : %s", pathIn, pathOut);
+}
+
+func startAgent() {
```
jefchien, line 92: A lot of these functions will be reused for future integration tests, so it might be worth pulling them out of this file into a util.

integration/test/ca_bundle/ca_bundle_test.go
```diff
@@ -0,0 +1,123 @@
...
+			containsTarget := readOutputLog()
+			if (parameter.findTarget && !containsTarget) || (!parameter.findTarget && containsTarget) {
+				t.Errorf("Find target is %t contains target is %t", parameter.findTarget, containsTarget)
+			}
+		})
+	}
+}
+
+func clearLogFile() {
+	cmd := exec.Command("bash", "-c", "sudo rm " + outputLog + " && sudo touch " + outputLog)
```
jefchien, line 68: You can configure the agent to print to stderr and use a buffer to capture all of that instead of rotating log files.
> `logfile` – Specifies the location where the CloudWatch agent writes log messages. If you specify an empty string, the log goes to stderr.

https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Agent-Configuration-File-Details.html#CloudWatch-Agent-Configuration-File-Agentsection

```
cmd := exec.Command(bin, cmdArgs...)
var buf bytes.Buffer
cmd.Stderr = &buf
```

integration/test/ca_bundle/ca_bundle_test.go
```diff
@@ -0,0 +1,123 @@
...
+	cmd := exec.Command("bash", "-c", "sudo /opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent-ctl -a stop")
+	err := cmd.Run()
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	log.Printf("Agent is stopped");
+}
+
+func readOutputLog() bool {
```
jefchien, line 114: Consider renaming this function to make the purpose clearer. Wouldn't expect `readOutputLog` to return a boolean. Maybe `outputLogContainsTarget`.

integration/localstack/ls_tmp/snakeoil.conf
```diff
@@ -0,0 +1,18 @@
+[req]
```
jefchien, line 1: Should this be included in the repo? Could it be made a Github secret as well?

integration/test/ca_bundle/resources/integration/ssl/with/combine/bundle/config.json
```diff
@@ -0,0 +1,27 @@
+{
+  "agent": {
```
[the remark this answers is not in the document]
sethAmazon: No, I guess I can add a test using http with no addition to the bundle. That will make the config files different and harden the tests to even more specifically target https.

integration/test/ca_bundle/ca_bundle_test.go
```diff
@@ -89,16 +79,20 @@ func copyFile(pathIn string, pathOut string) {
...
+	cmd.Stderr = &buff
+
 	err := cmd.Run()
 	if err != nil {
 		log.Fatal(err)
 	}
 
-	log.Printf("Agent has started");
+	log.Printf("Agent has started")
+	return buff.String()
```
jefchien, line 95: I think you need to return a pointer to the buffer instead of the string. `buff.String` will only give you the contents of the buffer up to the return statement. Then you can dump the buffer in your target check or when you stop the agent.

.github/workflows/integrationTest.yml
sethAmazon: Jobs run on separate runners thus you still  need to run terraform init before destroy.

PR #282, opened 2021-10-01
Removed runc dependency

jefchien:
# Description of the issue
The agent has a dependency on `github.com/opencontainers/runc v1.0.0-rc10`, which requires an upgrade. While the version could be bumped up to `v1.0.0-rc95`, that would affect the dependency on `github.com/google/cadvisor v0.36.0`, which depends on `v1.0.0-rc10`. Without updating the cadvisor dependency, which would require more evaluation and incremental changes, the next best step would be to remove the local dependency.

See https://github.com/aws/amazon-cloudwatch-agent/pull/254 for more details.

# Description of changes
Removed dependency on runc in userutil_linux.

Replaced with golang os/user lookup and version/arch specific setUid/setGid.

# License

# Tests
Tested on EC2 linux instance. Added unit test for `getGroupIds`.

### DIFF
--- a/translator/cmdutil/set_uid_gid_1_15_x32.go
+++ b/translator/cmdutil/set_uid_gid_1_15_x32.go
@@ -1,0 +1,30 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+// +build linux
+// +build 386 arm
+// +build !go1.16
+
+package cmdutil
+
+import (
+	"golang.org/x/sys/unix"
+)
+
+// go1.15 and before: use unix raw syscall. Can be removed once minimum go version has been bumped.
+
+func setUid(uid int) (err error) {
+	_, _, e1 := unix.RawSyscall(unix.SYS_SETUID32, uintptr(uid), 0, 0)
+	if e1 != 0 {
+		err = e1
+	}
+	return
+}
+
+func setGid(gid int) (err error) {
+	_, _, e1 := unix.RawSyscall(unix.SYS_SETGID32, uintptr(gid), 0, 0)
+	if e1 != 0 {
+		err = e1
+	}
+	return
+}
--- a/translator/cmdutil/userutil_linux.go
+++ b/translator/cmdutil/userutil_linux.go
@@ -25,8 +25,17 @@ type ExecUser struct {
 	Gids []int
 }
 
-func getGroupIds(u *user.User) ([]int, error) {
-	file, err := os.Open("/etc/group")
+func containsUser(users []string, match string) bool {
+	for _, user := range users {
+		if user == match {
+			return true
+		}
+	}
+	return false
+}
+
+func getGroupIds(user, filePath string) ([]int, error) {
+	file, err := os.Open(filePath)
 	if err != nil {
 		log.Printf("E! Failed to open group file: %v", err)
 		return nil, err
@@ -65,8 +74,8 @@ func getGroupIds(u *user.User) ([]int, error) {
 			continue
 		}
 		parts := strings.SplitN(string(wholeLine), ":", 4)
-		users := strings.TrimSpace(parts[3])
-		if len(users) == 0 || !strings.Contains(users, u.Username) {
+		users := strings.Split(parts[3], ",")
+		if len(users) == 0 || !containsUser(users, user) {
 			continue
 		}
 		groupId, err := strconv.Atoi(parts[2])
@@ -88,7 +97,7 @@ func toExecUser(u *user.User) (*ExecUser, error) {
 		log.Printf("E! Failed to convert gid to int: %v", err)
 		return nil, err
 	}
-	gids, err := getGroupIds(u)
+	gids, err := getGroupIds(u.Username, "/etc/group")
 	if err != nil {
 		log.Printf("E! Failed to get group IDs: %v", err)
 		return nil, err
--- a/translator/cmdutil/userutil_linux.go
+++ b/translator/cmdutil/userutil_linux.go
@@ -7,28 +7,68 @@ package cmdutil
 
 import (
 	"fmt"
+	"golang.org/x/sys/unix"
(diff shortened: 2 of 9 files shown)


translator/cmdutil/userutil_linux.go
```diff
@@ -6,29 +6,108 @@
...
+	Uid  int
+	Gid  int
+	Home string
+	Gids []int
+}
 
-	"github.com/opencontainers/runc/libcontainer/system"
-	"github.com/opencontainers/runc/libcontainer/user"
-)
+func getGroupIds(u *user.User) ([]int, error) {
```
SaxyPandaBear, line 28: I skimmed it and it looks good but can how hard would it be to include a test for this code?

translator/cmdutil/userutil_linux.go
```diff
@@ -6,29 +6,108 @@
...
+		log.Printf("E! Failed to open group file: %v", err)
+		return nil, err
+	}
+	defer file.Close()
+	reader := bufio.NewReader(file)
+	groupIds := []int{}
+	for {
+		var wholeLine []byte
+		for {
+			line, isPrefix, err := reader.ReadLine()
```
SaxyPandaBear, line 40: My guess is we don't want to do anything super bloated, but I noticed this from the `bufio` docs:
> ReadLine is a low-level line-reading primitive. Most callers should use ReadBytes('\n') or ReadString('\n') instead or use a Scanner.

Should we change the call to something else? It does say _most_, and I'm not really opposed to how it's written currently. Just figured I'd bring it up while I was looking at the docs.

translator/cmdutil/set_uid_gid_1_15_x32.go
```diff
@@ -0,0 +1,30 @@
...
+
+package cmdutil
+
+import (
+	"golang.org/x/sys/unix"
+)
+
+// go1.15 and before: use unix raw syscall. Can be removed once minimum go version has been bumped.
+
+func setUid(uid int) (err error) {
```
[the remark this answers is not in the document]
jhnlsn: Looking around, it seems this is the general go semantics: `If the function name starts with a lowercase letter, it won't be exported to other packages, but you can call this function within the same package.`

translator/cmdutil/userutil_linux.go
```diff
@@ -6,29 +6,108 @@
...
+	if err != nil {
+		log.Printf("E! Failed to convert uid to int: %v", err)
+		return nil, err
+	}
+	gid, err := strconv.Atoi(u.Gid)
+	if err != nil {
+		log.Printf("E! Failed to convert gid to int: %v", err)
+		return nil, err
+	}
+	gids, err := getGroupIds(u)
```
jhnlsn, line 91: Was there a reason why you couldn't use `u.GroupIds()` in this case?

translator/cmdutil/set_uid_gid_1_15_x32.go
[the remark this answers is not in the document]
jefchien: Yep it's just like John said. Also, by using the build tag (`// +build go1.16`), the file will only be included for all go versions including and past 1.16. This allows us to call the same function and have it implemented different ways based on platform, architecture, and version.

translator/cmdutil/userutil_linux.go
jefchien: The `GroupIds` function introduces a dependency on cgo, which can be enabled with `CGO_ENABLED=1`, but adds a lot of unnecessary build complexity for a single function.

https://pkg.go.dev/cmd/cgo
>When cross-compiling, you must specify a C cross-compiler for cgo to use. You can do this by setting the generic CC_FOR_TARGET or the more specific CC_FOR_${GOOS}_${GOARCH} (for example, CC_FOR_linux_arm) environment variable when building the toolchain using make.bash, or you can set the CC environment variable any time you run the go tool.

translator/cmdutil/userutil_linux.go
jefchien: I saw that, but mainly based it on the golang implementation: https://golang.org/src/os/user/lookup_unix.go

sethAmazon:
Should we wait on this until we update to 1.17 when it is released?

translator/cmdutil/userutil_linux.go
```diff
@@ -6,29 +6,117 @@
...
 	}
 
-	if err := system.Setgid(execUser.Gid); err != nil {
+	if err := setGid(execUser.Gid); err != nil {
 		log.Printf("E! Failed to set gid: %v", err)
 		return err
 	}
 
-	if err := system.Setuid(execUser.Uid); err != nil {
+	if err := setUid(execUser.Uid); err != nil {
```
sethAmazon, line 119: So this will take the setUid method from inside this package based on the go version number? Sorry my go knowledge is not vast.

jefchien: That's right. Only the files with the correct go version number, platform, and architecture will be included in the build.

jefchien:
I don't really think there's a need to wait. The only difference would be the inclusion of the two 1.15 specific files, which can easily be removed once 1.15 is no longer supported.

translator/cmdutil/userutil_linux_test.go
```diff
@@ -0,0 +1,35 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+// +build linux
```
sethAmazon, line 4: Why will this test not run on mac? Can you just post the error message you get out here. I'm assuming windows file permission system is different. I do not know much about windows.

jefchien: To run the test on mac, you would have to include darwin in the build tag. https://pkg.go.dev/cmd/go#hdr-Build_constraints

jefchien: There's no error message. It just won't get picked up.

PR #289, opened 2021-10-06
Add al2 support

sethAmazon:
# Description of the issue
Support running integration tests on more than 1 linux ami 

# Description of changes
Add run integration test to al2 and ubutnu

# Tests
Integration tests where on my test account.

### DIFF
--- a/.github/workflows/integrationTest.yml
+++ b/.github/workflows/integrationTest.yml
@@ -1,0 +1,67 @@
+name: Run Integration Tests
+env:
+  PRIVATE_KEY: ${{ secrets.AWS_PRIVATE_KEY  }}
+  TERRAFORM_AWS_ACCESS_KEY_ID: ${{ secrets.TERRAFORM_AWS_ACCESS_KEY_ID }}
+  TERRAFORM_AWS_SECRET_ACCESS_KEY: ${{ secrets.TERRAFORM_AWS_SECRET_ACCESS_KEY }}
+
+on:
+  push:
+    branches:
+      - master
+  workflow_dispatch:
+
+jobs:
+  EC2LinuxIntegrationTest:
+    name: 'EC2LinuxIntegrationTest'
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: integration/terraform/ec2/linux
+    strategy:
+      fail-fast: false
+      matrix:
+        arrays: [
+          { os: "ubuntu", package: "package-deb",  username: "ubuntu",
+            installAgentCommand: "dpkg -i -E ./amazon-cloudwatch-agent.deb",
+            ami: "cloudwatch-agent-integration-test-ubuntu", caCertPath: "/etc/ssl/certs/ca-certificates.crt",
+            arc: "amd64" },
+          { os: "al2", package: "package-rpm",  username: "ec2-user",
+            installAgentCommand: "rpm -U ./amazon-cloudwatch-agent.rpm",
+            ami: "cloudwatch-agent-integration-test-al2", caCertPath: "/etc/ssl/certs/ca-bundle.crt",
+            arc: "amd64" }
+        ]
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Echo OS
+        run: echo run on ec2 instance os ${{ matrix.arrays.os }}
+
+      - name: Verify Terraform version
+        run: terraform --version
+
+      - name: Terraform init
+        run: terraform init -var="aws_access_key=${TERRAFORM_AWS_ACCESS_KEY_ID}" -var="aws_secret_access_key=${TERRAFORM_AWS_ACCESS_KEY_ID}"
+
+      - name: Terraform apply
+        run: >
+          echo run terraform and execute test code &&
+          terraform apply --auto-approve
+          -var="ssh_key=${PRIVATE_KEY}"
+          -var="aws_access_key=${TERRAFORM_AWS_ACCESS_KEY_ID}"
+          -var="aws_secret_access_key=${TERRAFORM_AWS_SECRET_ACCESS_KEY}"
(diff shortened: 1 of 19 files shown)


.github/workflows/integrationTest.yml
```diff
@@ -0,0 +1,60 @@
...
+    defaults:
+      run:
+        working-directory: integration/terraform/ec2/linux
+    strategy:
+      fail-fast: false
+      matrix:
+        arrays: [
+          { os: "ubuntu", package: "package-deb",  username: "ubuntu",
+            installAgentCommand: "dpkg -i -E ./amazon-cloudwatch-agent.deb",
+            ami: "ami-0b284e8646fbf58a8", caCertPath: "/etc/ssl/certs/ca-certificates.crt" },
```
jhnlsn, line 26: Can you use terraform to look up the latest version of an AMI

integration/localstack/ls_tmp/snakeoil.conf
```diff
@@ -0,0 +1,18 @@
...
+prompt = no
+default_md = sha256
+x509_extensions = v3_req
+distinguished_name = dn
+[dn]
+C = US
+ST = VA
+L = Virginia
+O = CloudwatchAgent
+emailAddress = sethlevi@amazon.com
```
SaxyPandaBear, line 12: Do we need this? I'm not convinced we need an email address for a signed SSL cert

integration/terraform/ec2/linux/main.tf
```diff
@@ -0,0 +1,50 @@
...
+  iam_instance_profile   = var.iam_instance_profile
+  vpc_security_group_ids = var.vpc_security_group_ids
+  provisioner "remote-exec" {
+    inline = [
+      "cloud-init status --wait",
+      "echo clone, build, and install agent",
+      "git clone ${var.github_repo}",
+      "cd amazon-cloudwatch-agent",
+      "git reset --hard ${var.github_sha}",
+      "make clean build ${var.package}",
```
SaxyPandaBear, line 14: On the one hand, I feel like the usage of the `package` var is obvious, but on the other hand, it's not - even though I read the action first, when I got to this, I thought that `package` referred to like, the RPM or package name, and not the Makefile action for packaging. I can't think of a much better name for it though. I'll have to think about that one.

integration/terraform/ec2/linux/main.tf
```diff
@@ -0,0 +1,50 @@
...
+      "cd build/bin/linux/amd64",
```
SaxyPandaBear, line 15: This may feel like an obvious question, but will this always work for all Linux installs we do in the GitHub actions?

integration/terraform/ec2/linux/main.tf
```diff
@@ -0,0 +1,50 @@
...
+      "cat snakeoil.pem > server.test.pem.crt",
+      "cat ${var.ca_cert_path} > original.pem",
+      "cat original.pem snakeoil.pem > combine.pem",
+      "sudo cp original.pem /opt/aws/amazon-cloudwatch-agent/original.pem",
+      "sudo cp combine.pem /opt/aws/amazon-cloudwatch-agent/combine.pem",
+      "cd ~/amazon-cloudwatch-agent/integration/localstack",
+      "docker-compose up -d --force-recreate",
+      "echo run tests with the tag integration, one at a time, and verbose",
+      "cd ~/amazon-cloudwatch-agent/integration/test",
+      "sudo go test ./... -p 1 -v --tags=integration"
```
SaxyPandaBear, line 31: Do we want to run `go test` or use a Makefile abstraction on it? Does it require `sudo`?

integration/terraform/ec2/linux/variables.tf
```diff
@@ -0,0 +1,72 @@
...
+}
+
+variable "iam_instance_profile" {
+  type = string
+  default = "CloudWatchAgentServerRole"
+}
+
+variable "vpc_security_group_ids" {
+  type = list(string)
+  default = ["sg-013585129c1f92bf0"]
```
SaxyPandaBear, line 18: This smells fishy. This is a specific security group in an individual AWS account, right?

integration/terraform/ec2/linux/variables.tf
```diff
@@ -0,0 +1,72 @@
...
+  type = string
+  default = ""
+}
+
+variable "github_repo" {
+  type = string
+  default = ""
+}
+
+variable "package" {
```
SaxyPandaBear, line 59: This looks really ambiguous as just `package` to me (made a point in another file about it already). Can we add a `description` value here to explain what this is used for?

integration/terraform/ec2/linux/variables.tf
```diff
@@ -0,0 +1,72 @@
...
+  type = string
+  default = ""
+}
+
+variable "install_agent" {
```
SaxyPandaBear, line 64: Similar to `package, `install_agent` feels ambiguous for what the value should be here. I think this warrants a `description` value in the variable declaration.

integration/localstack/ls_tmp/snakeoil.conf
sethAmazon: duplicate

integration/terraform/ec2/linux/main.tf
sethAmazon: duplicate

integration/terraform/ec2/linux/main.tf
sethAmazon: Yes it requires sudo. I can make a make file abstraction

integration/terraform/ec2/linux/variables.tf
SaxyPandaBear: Don't think it's necessary to make it a secret.

SaxyPandaBear (dismissed):
LGTM

PR #291, opened 2021-10-06
Upload Nightly Builds

sethAmazon:
# Description of the issue
Upload nightly builds to s3

# Description of changes
Add git hub action to upload binaries to s3 nightly

# Tests
Github action set up on my fork

### DIFF
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -1,0 +1,44 @@
+name: Upload Nightly Binaries
+
+on:
+  schedule:
+    - cron: '0 0 * * *' # run at mid night
+  workflow_dispatch:
+
+jobs:
+
+  build:
+    name: Upload Nightly Binaries
+    runs-on: ubuntu-latest
+    steps:
+
+      - name: Set up Go 1.x
+        uses: actions/setup-go@v2
+        with:
+          go-version: ^1.15
+        id: go
+
+      #Need to install rpm so ubuntu can make rpm by default ubuntu can make deb
+      - name: Install rpm
+        run: sudo apt install rpm
+
+      - name: Check out code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+          submodules: 'true'
+
+      - name: Release
+        run: make nightly-release
+
+      - name: Upload binaries to latest
+        uses: jakejarvis/s3-sync-action@master
+        with:
+          args: --acl public-read
+        env:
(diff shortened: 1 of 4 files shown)


jhnlsn:
I thought -failfast would fail the build when the first test fails.  It seems to be running all tests even if one fails, making it really difficult to find the FAIL.

.github/workflows/nightly-build.yml
```diff
@@ -0,0 +1,69 @@
...
+          go-version: ^1.17
+        id: go
+
+      - name: Install zip
+        run: sudo apt install zip
+
+      - name: Install rpm
+        run: sudo apt install rpm
+
+      # Must clone a fresh repo because of how the .git folder is used in the versioning of the agent for packaging
```
jhnlsn, line 33: I'm not exactly sure what your trying to fix, but the way I understand it, github created this: https://github.com/actions/checkout

jhnlsn:
First of all, this is awesome!  Glad we are taking this direction!

I've seen a lot of projects that release nightly builds, most recently I saw that telegraf releases nightlies: https://github.com/influxdata/telegraf/blob/master/docs/NIGHTLIES.md

- [ ] A couple of things need to be decided before we roll out something like where we are going to store these builds.
- [ ] A readme page that fully outlines that nightly builds are not production ready and should be used at own risk
- [ ] Break out artifacts into their own consumables.  Separate files for downloading deb/rpm/raw binaries.
- [ ] Use `tar.gz` instead of `zip` for everything but windows artifact :)

.github/workflows/nightly-build.yml
sethAmazon: It has to do with how git tags work.  When using checkout we lose these tags thus we need to clone a fresh repo. What you have posted will not work. https://github.com/actions/checkout explains by default checkout uses the event sha

.github/workflows/nightly-build.yml
```diff
@@ -0,0 +1,43 @@
...
+    steps:
+
+      - name: Set up Go 1.x
+        uses: actions/setup-go@v2
+        with:
+          go-version: ^1.15
+        id: go
+
+      - name: Install rpm
+        run: sudo apt install rpm
```
adam-mateen, line 22: I suggest adding a comment that `sudo apt install rpm` is done so the Ubuntu host can create both the DEB and RPM install packages.

Also, can you confirm that the RPM that gets built by this Ubuntu machine can be installed on a AL2 (or RedHat or CentOS) host?

adam-mateen (dismissed):

Issue #280, opened 2021-09-30
Change runtime log level


jhnlsn:
Enable the ability to change the logging level without restarting the agent

PR #293, opened 2021-10-12
Change runtime log level

jefchien:
# Description of the issue
For this issue: https://github.com/aws/amazon-cloudwatch-agent/issues/280

The CloudWatch agent uses the telegraf logger to manage its logging. Currently, to change the log level, the agent configuration file needs to be updated to include `debug: true` in the agent section and the process needs to be restarted for the change to take effect.

# Description of changes
Added a goroutine created in the `reload_loop` that will check the `env_config.json` file every 30 seconds to see if it has been changed. If the file has been changed, it will re-read the file and update the known environment variables. Normally, this is only done once on start up. It then tries to set the log level to the `CWAGENT_LOG_LEVEL` environment variable.

The user is able to manually update the `env-config.json` file if desired, but a `setenv` flag has been added to the `amazon-cloudwatch-agent` binary that will update the log level in the file. This can be accessed by running:

**on linux**: `sudo /opt/aws/amazon-cloudwatch/bin/amazon-cloudwatch-agent-ctl -a set-log-level -l DEBUG`
or
**on windows**: `& $Env:ProgramFiles\Amazon\AmazonCloudWatchAgent\amazon-cloudwatch-agent-ctl.ps1 -a set-log-level -l DEBUG`

Note: If a user has `debug:true` present in their configuration file, `CWAGENT_LOG_LEVEL` will be set to "DEBUG" after translation of the configuration file has completed.

# License

# Tests
Tested on linux and windows ec2 instances.


cmd/amazon-cloudwatch-agent/amazon-cloudwatch-agent.go
```diff
@@ -139,6 +140,36 @@ func reloadLoop(
...
+			go func(ctx context.Context, envConfigPath string) {
+				var previousModTime time.Time
+				ticker := time.NewTicker(30 * time.Second)
+				defer ticker.Stop()
+				for {
+					select {
+					case <-ticker.C:
+						if info, err := os.Stat(envConfigPath); err == nil && info.ModTime().After(previousModTime) {
+							if err := loadEnvironmentVariables(envConfigPath); err != nil {
+								log.Printf("E! Unable to load env variables: %v", err)
```
SaxyPandaBear, line 154: Are we only going to log when we are unable to load environment variables? I guess it would make sense that if we encountered an error here, we wouldn't modify the log level.

packaging/windows/amazon-cloudwatch-agent-ctl.ps1
```diff
@@ -500,8 +511,24 @@ Function CWATestEC2() {
...
+            Write-Output "Invalid log level: ${LogLevel}`n${UsageString}"
+            Exit 1
+        }
+    }
+
+    & cmd /c "`"${CWAProgramFiles}\amazon-cloudwatch-agent.exe`" --setenv CWAGENT_LOG_LEVEL=${LogLevel} --envconfig ${ENV_CONFIG} 2>&1"
+    CheckCMDResult "" "Set CWAGENT_LOG_LEVEL to ${LogLevel}"
+}
 
+Function main() {
```
SaxyPandaBear, line 531: I'm out of practice with powershell. If we wrap this in a main function, does that maintain the existing run functionality? Pulling from knowledge about Python, just declaring a main function doesn't necessarily run it when you call the script.

translator/totomlconfig/sampleConfig/complete_windows_config.conf
```diff
@@ -1,6 +1,6 @@
 [agent]
   collection_jitter = "0s"
-  debug = false
+  debug = true
```
SaxyPandaBear, line 3: Do we have any tests where debug is false? Is there a way we can test to validate that the log level gets set to WARN or OFF, etc.?

packaging/windows/amazon-cloudwatch-agent-ctl.ps1
jefchien: This was actually already here, but yeah the existing run functionality is preserved.

translator/totomlconfig/sampleConfig/complete_windows_config.conf
jefchien: I wanted to have minimal changes to the behavior with this change. If debug isn't present in the agent section of the JSON config, then it will be set to false in the TOML config. This can be seen in most of the other tests.

When debug is set to true, the env-config.json will have "CWAGENT_LOG_LEVEL": "DEBUG" present. If debug is set to false, then the "CWAGENT_LOG_LEVEL" key isn't added. If the environment variable is not present or is empty, the log level defaults to "INFO". This is consistent with the current behavior.

cmd/amazon-cloudwatch-agent/amazon-cloudwatch-agent.go
```diff
@@ -139,6 +140,36 @@ func reloadLoop(
 			}
 		}(ctx)
 
+		if envConfigPath, err := getEnvConfigPath(*fConfig, *fEnvConfig); err == nil {
+			// Reloads environment variables when file is changed
+			go func(ctx context.Context, envConfigPath string) {
+				var previousModTime time.Time
+				ticker := time.NewTicker(30 * time.Second)
```
adam-mateen, line 147: It is hard to know if every 30 seconds is the right polling frequency.
I saw this and wondered if we could or should poll every 60 seconds.
Then I wondered if it matters.
I am assuming calling `os.Stat()` is pretty cheap so 30 seconds is fine.
I just wanted to bring some attention to it in case others had opinions...

adam-mateen (approved):

packaging/windows/amazon-cloudwatch-agent-ctl.ps1
```diff
@@ -500,8 +511,24 @@ Function CWATestEC2() {
...
+Function SetLogLevelAll() {
+    switch -exact ($LogLevel) {
+        INFO { }
+        DEBUG { }
+        ERROR { }
+        WARN { }
+        OFF { }
+        default {
+            Write-Output "Invalid log level: ${LogLevel}`n${UsageString}"
+            Exit 1
```
adam-mateen, line 523: PowerShell syntax, I am familiar with:
```
Write-Output "Invalid log level: $LogLevel`n$UsageString"
```
(without extra braces around the vars).
What you have seems to work though:
https://code.labstack.com/powershell

jefchien: That's good to know. I just based it on other existing code within the script.

PR #283, opened 2021-10-03
Remove empty files/windows_events field at config migration

kyoicito:
# Description of the issue
Config migration wizard for Windows includes empty files/windows_events field when old SSM config includes only one type of these log types at present. This field cause the agent fails to run, so users must remove it from config by themselves beforehand.

# Description of changes
I changed `LogsCollectedEntry` struct type to use pointer for removing empty *files*/*windows_events* field by *omitempty* option.
I also changed transformation function `MapOldWindowsConfigToNewConfig` to keep the consistency for this change.

# License
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

# Tests
1. Run `make release`
2. Put *amazon-cloudwatch-agent-config-wizard.exe* on Windows Server and run it with [input2.json](https://github.com/kyoicito/amazon-cloudwatch-agent/blob/improve-config-wizard/tool/processors/migration/windows/testData/input2.json) put at `C:\Program Files\Amazon\SSM\Plugins\awsCloudWatch\AWS.EC2.Windows.CloudWatch.json`

```
PS C:\Program Files\Amazon\AmazonCloudWatchAgent> .\amazon-cloudwatch-agent-config-wizard.exe
================================================================
= Welcome to the Amazon CloudWatch Agent Configuration Manager =
=                                                              =
= CloudWatch Agent allows you to collect metrics and logs from =
= your host and send them to CloudWatch. Additional CloudWatch =
= charges may apply.                                           =
================================================================
On which OS are you planning to use the agent?
1. linux
2. windows
3. darwin
default choice: [2]:

Trying to fetch the default region based on ec2 metadata...
Are you using EC2 or On-Premises hosts?
1. EC2
2. On-Premises
default choice: [1]:

Do you want to turn on StatsD daemon?
1. yes
2. no
default choice: [1]:
2
Do you have any existing CloudWatch Log Agent configuration file to import for migration?
1. yes
2. no
default choice: [2]:
1
What is the file path for the existing Windows CloudWatch log agent configuration file?
default choice: [C:\Program Files\Amazon\SSM\Plugins\awsCloudWatch\AWS.EC2.Windows.CloudWatch.json]

Saved config file to config.json successfully.
Do you want to store the config in the SSM parameter store?
1. yes
2. no
default choice: [1]:
2
Please press Enter to exit...
```

Output with **NEW** migration (this PR):
```
{
    "agent": {
        "region": "us-west-2"
    },
    "logs": {
[...]

PR #294, opened 2021-10-14
use json comparison for windows migration tests

### LINKED ISSUE #285: Investigate falst positive test
`windows_migration_core_test.go` has a potential for having a false positive test with input2.json and output2.json.

it was found that an empty files object is included when it is expected that one should not be.

See PR: #283

SaxyPandaBear:
Closes #285 

# Description of the issue
#285 describes how the test json included an empty `files` object when there shouldn't be one. This was observed in #283. 
Debugging the issue showed that because the struct included the files object, even though the JSON blob itself did not contain it, when unmarshalling in Go, the files key was included in the test struct, which still passed assertions because both the input and output had an empty files object.

# Description of changes
Changed the testing logic for the `windows_migration_core_test.go` that specifically target the matrix of input and output files to perform JSON string assertions instead of unmarshalling the JSON files as typed Go structs and comparing those, to avoid injecting an unintended, empty files object into the interface.

# License

# Tests
Added logs to print out the resulting JSON strings, and confirmed that `files` does not show up in the output when running the test suite. See my comment [here](https://github.com/aws/amazon-cloudwatch-agent/issues/285#issuecomment-943464393) for the output of my findings.

### DIFF
--- a/tool/processors/migration/windows/windows_migration_core_test.go
+++ b/tool/processors/migration/windows/windows_migration_core_test.go
@@ -4,7 +4,9 @@
 package windows
 
 import (
+	"encoding/json"
 	"fmt"
+	"github.com/stretchr/testify/assert"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


tool/processors/migration/windows/testData/output2.json
```diff
@@ -4,6 +4,7 @@
   },
   "logs": {
     "logs_collected": {
+      "files": {},
```
SaxyPandaBear, line 7: Added the `files` struct because this is what currently gets output. The ask in #283 would be to remove these since that PR addresses the fact that these shouldn't be in the output to begin with.

jhnlsn (approved):

PR #301, opened 2021-10-29
add cache step to build action

SaxyPandaBear:
# Description of the issue
GitHub CI build times are long because we always have to download dependencies for each build, fresh.

# Description of changes
Uses the [cache](https://github.com/actions/cache/blob/main/examples.md#go---modules) action to cache build output from Go. 

# License

# Tests
Pushed this to my fork and confirmed that (1), the new step ran, and (2) that the matrix picked the correct paths for the different OS types. See [my build](https://github.com/SaxyPandaBear/amazon-cloudwatch-agent/runs/4041624140?check_suite_focus=true).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -16,6 +16,19 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
+        include:
+          - os: ubuntu-latest
+            cachepath: |
+              ~/.cache/go-build
+              ~/go/pkg/mod
+          - os: macos-latest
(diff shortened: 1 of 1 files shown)


SaxyPandaBear:
Further, performing a subsequent build action restored the cache successfully, reducing macos and ubuntu build times significantly. Windows still took a while. My assumption based on previous experience is that it just takes longer to bootstrap/tear down the windows box that is running the build. I checked the output and the windows build did not redownload build dependencies so I'm not sure what attributes to the slow build time for it.

sethAmazon (approved):

PR #310, opened 2021-11-10
Fix TestLogsFileRecreate.

adam-mateen:
# Description of the issue
`TestLogsFileRecreate()` intermittently fails on Windows OS.
example: https://github.com/aws/amazon-cloudwatch-agent/runs/4156147013?check_suite_focus=true

There are actually 2 issues.
First is that there is a `for` loop that will does not end until Go does a `panic`.
Second, the delay between deleting the monitored temp file and recreating it is a fixed 100 ms.
These 2 changes are minimally required.

# Description of changes
The loop will timeout after 10 seconds if the LogSrc is not found, instead of `Panic` after 10 minutes.
The fixed delay between deleting and recreating the temp file has been increased from 100ms to 1 second.

I added some info messages to describe what the test is doing.
I was just going to add code comments, but logging what the test is doing makes it easier to correlate with what the functions being tested are doing and correlate with if/when errors occur.

# License

# Tests
```
make release
```

### DIFF
--- a/plugins/inputs/logfile/logfile_test.go
+++ b/plugins/inputs/logfile/logfile_test.go
@@ -781,7 +781,9 @@ func TestLogsFileRecreate(t *testing.T) {
 		}
 		time.Sleep(1 * time.Second)
 	}
-
+	if len(lsrcs) != 1 {
+		t.Fatalf("%v log src was returned when 1 should be available", len(lsrcs))
+	}
 	lsrc = lsrcs[0]
 	lsrc.SetOutput(func(e logs.LogEvent) {
 		if e != nil {
--- a/plugins/inputs/logfile/logfile_test.go
+++ b/plugins/inputs/logfile/logfile_test.go
@@ -705,6 +705,10 @@ func TestLogsFileWithInvalidOffset(t *testing.T) {
 	tt.Stop()
 }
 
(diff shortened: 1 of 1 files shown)


plugins/inputs/logfile/logfile_test.go
```diff
@@ -768,7 +773,8 @@ func TestLogsFileRecreate(t *testing.T) {
 	}
 	defer lsrc.Stop()
 
-	for {
+	// Waiting 10 seconds for the recreated temp file to be detected is plenty sufficient on any OS.
+	for start := time.Now(); time.Since(start) < 10 * time.Second; {
```
SaxyPandaBear, line 777: Should we have an assertion after the for-loop to ensure that `len(lsrcs) > 0`? Previously, this would run indefinitely until the test runner timed out. So we would never actually get to the point where we try to access the element in the slice below on line 785:
```
lsrc = lsrcs[0]
```
But now that we're forcing this loop to complete in at most 10 seconds, we can reach a state where there aren't any sources. 

I guess if it's empty and we try to access it, that'll be an indirect failure so it's not like it's uncaught.. but I think we should include an explicit assertion after the loop to be plain and obvious about what we expect.

adam-mateen: good idea

SaxyPandaBear (approved):
LGTM

PR #327, opened 2021-12-13
Agent-side log filtering

SaxyPandaBear:
# Description of the issue
Customers want the ability to apply agent-side filtering of log messages in order to reduce the number of logs that get published to CloudWatch, and key in on specific types of logs to track.

# Description of changes
Allow per-log-file configuration for the CloudWatch agent to perform regex-based filtering, in order to evaluate whether a given log message should be published to CloudWatch logs.

## Caveat - Windows Event Logs
One thing I wasn't sure about, and is worth discussion, is whether or not to duplicate the changes I made here for the Windows event logs. I don't think it would be a heavy lift to port the code over, but it feels like tacking on more changes than necessary for this. I did have to modify the LogSrc impl there for compilation, though. 

# License

# Tests
Added unit tests where the change proliferated, as well as rudimentary benchmark tests for the regex filtering


plugins/inputs/logfile/fileconfig.go
```diff
@@ -144,6 +174,24 @@ func (config *FileConfig) init() error {
 		config.RetentionInDays = -1
 	}
 
+	if config.Filters != nil {
+		filters := make([]LogFilter, len(config.Filters))
+		for i, f := range config.Filters {
+			err = f.init()
+			if err != nil {
+				return err
```
adam-mateen, line 182: `return` on error instead of `continue` means that the entire `logfile` plugin will not get started.
```
	// Initialize all the file configs
	for i := range t.FileConfig {
		if err := t.FileConfig[i].init(); err != nil {
			return err
		}
	}

	t.started = true
```
So no files will be monitored.

I think you made the right choice.
CWA should fail quickly and loudly so the user knows there is a problem and can fix their configuration.

Just thought I'd point this out in case others think CWA should just log the bad filter and continue to monitor this file using any other valid filters.

plugins/inputs/logfile/fileconfig.go
```diff
@@ -203,3 +251,55 @@ func (config *FileConfig) isMultilineStart(logValue string) bool {
...
+
+func (config *FileConfig) shouldFilterLog(event logs.LogEvent) bool {
+	if config.Filters == nil || len(config.Filters) < 1 {
+		return false
+	}
+	config.sampleCount += 1
+	start := time.Now()
+	for _, filter := range config.Filters {
+		if filter.ExpressionP == nil {
+			filter.ExpressionP = regexp.MustCompile(filter.Expression)
```
adam-mateen, line 263: `LogFilter.init()` already does `regexp.Compile()`. So why is it necessary to do `regexp.MustCompile()` here?
`MustCompile()` could `panic` - https://pkg.go.dev/regexp#MustCompile

Why not `continue` or `return false`?

plugins/inputs/logfile/fileconfig.go
```diff
@@ -85,6 +96,25 @@ type FileConfig struct {
 	BlacklistRegexP *regexp.Regexp
 	//Decoder object
 	Enc encoding.Encoding
+	sampleCount int
+}
+
+type LogFilter struct {
+	Type string `toml:"type"`
+	Expression string `toml:"expression"`
+	ExpressionP *regexp.Regexp
```
adam-mateen, line 105: If there is no intent to use `ExpressionP` outside the `logfile` package, then you can rename it starting with a lowercase.

It looks like the initialization and use of the `Regexp` struct is confined to this file.

plugins/inputs/logfile/fileconfig.go
```diff
@@ -203,3 +251,55 @@ func (config *FileConfig) isMultilineStart(logValue string) bool {
 	}
 	return config.MultiLineStartPatternP.MatchString(logValue)
 }
+
+func (config *FileConfig) shouldFilterLog(event logs.LogEvent) bool {
+	if config.Filters == nil || len(config.Filters) < 1 {
+		return false
+	}
+	config.sampleCount += 1
+	start := time.Now()
```
adam-mateen, line 260: Optional / suggestion - you could do:
```
start := time.Now()
defer func() {
    elapsed := time.Since(start)
    if config.sampleCount >= sampleThreshold {
        log.Printf("foo")
        config.sampleCount = 0
    }
}()
```
Then remove it from the 3 places below.

SaxyPandaBear: ahhh that's a good point. I can make that change.

plugins/inputs/logfile/fileconfig.go
SaxyPandaBear: > CWA should fail quickly and loudly so the user knows there is a problem

Yup, we're on the same page for this. There's a difference between writing an improper expression (something that doesn't do what you expect) versus writing an invalid one (one that doesn't compile). If the customer chooses to use the expression and write an expression that doesn't filter out what they want, that's not something we can control. But we CAN control the behavior if they supply a regular expression that has invalid syntax.

plugins/inputs/logfile/fileconfig.go
SaxyPandaBear: I _think_ I had this in here from testing where the `ExpressionP` was uninitialized. The reason I chose `MustCompile` here was because the expression should have been validated in the `init()` function, so there shouldn't be any risk of it panicking here. I will have to look over this again to see if I can take this out, since, like you said, the init function should already compile the expression.

plugins/inputs/logfile/fileconfig.go
SaxyPandaBear: Good call out. I can change that. Was really just following what was already there for the timestamp expression

plugins/inputs/logfile/fileconfig.go
```diff
@@ -203,3 +251,55 @@ func (config *FileConfig) isMultilineStart(logValue string) bool {
...
+		matched := filter.ExpressionP.MatchString(event.Message())
+		switch filter.Type {
+		case includeType:
+			if matched {
+				elapsed := time.Since(start)
+				if config.sampleCount >= sampleThreshold {
+					log.Printf("D! [inputs.logfile.%s.%s] Matched log message with expression {%s} in %dns", config.LogGroupName, config.LogStreamName, filter.Expression, elapsed.Nanoseconds())
+					config.sampleCount = 0
+				}
+				return false
```
adam-mateen, line 274: If there are 5 LogFilters, all of which are INCLUDE type, then this code would check the first LogFilter and if it matches it returns false.
Is it allowable for a user to configure 2 or more INCLUDE type filters?
If so, then you cannot return before checking all filters.

SaxyPandaBear: I'm not sure I follow. 

Yes, it is allowable for a user to configure more than one INCLUDE type filter. I believe that the easiest implementation for the log filter order of operations is to short circuit when we find a first match. What is the use case where we would match one include filter, and still need to check a subsequent include filter? By design, the filter configuration is an ordered list. The order of the list is used to determine the order in which the agent processes filters.

adam-mateen: I confused myself too.

Example config:
```
{
            "file_path": "/var/log/foo.log",
            "log_group_name": "bar",
            "log_stream_name": "baz",
            "filters": [
              {
                "type": "include",
                "expression": "(ERROR|error)"
              },
              {
                "type": "include",
                "expression": "(CRITICAL|critical)"
              }
            ]
          },
```

In this case I want to upload log lines that match filter_1 *OR*** filter_2.

At first I was thinking I would only want upload log lines that match filter_1 *AND*** filter_2.

So I am fine with short circuit `return false` on this line.

... Hopefully nobody else gets confused like I did :)

SaxyPandaBear: Yeahhhhh I think that semantically I could have inverted the boolean so that it returns `true` when it's matched and we want to return early, but I don't think the structure would be changed much. It does what I want it to do; the return value is just kind of unintuitive. I'm open to flipping the condition and changing it from `shouldFilterLog` to something like `shouldPublishLog`, if you think that could help. Otherwise I think I'll keep it as is.

plugins/inputs/logfile/fileconfig.go
```diff
@@ -203,3 +251,55 @@ func (config *FileConfig) isMultilineStart(logValue string) bool {
...
+				elapsed := time.Since(start)
+				if config.sampleCount >= sampleThreshold {
+					log.Printf("D! [inputs.logfile.%s.%s] Matched log message with expression {%s} in %dns", config.LogGroupName, config.LogStreamName, filter.Expression, elapsed.Nanoseconds())
+					config.sampleCount = 0
+				}
+				profiler.Profiler.AddStats(
+					[]string{"logfile", config.LogGroupName, config.LogStreamName, "messages", "dropped"},
+					1,
+				)
+				return true
```
adam-mateen, line 287: Another use case to consider:
config: `[include_FOO, exclude_BAR]` will upload log lines with "FOO BAR".
config: `[exlclude_BAR, include_FOO]` will NOT upload log lines with "FOO BAR".

Did I get that right?
Is this by design?

plugins/inputs/logfile/tailersrc.go
```diff
@@ -227,7 +243,13 @@ func (ts *tailerSrc) runTail() {
 					offset: *fo,
 					src:    ts,
 				}
-				ts.outputFn(e)
+				if ts.filterFn != nil {
```
straussb, line 246: (I wanted to leave this comment on line 230 above, but GitHub won't let me comment on unchanged lines...)

This is tricky, but what should we do about lines that are truncated because they go over the max event size?  I suppose we have no choice but to apply the regex to the truncated line?  If that's what we decide, then just leave a comment so future readers at least know we thought about it.

logs/logs.go
```diff
@@ -29,6 +29,7 @@ type LogEvent interface {
 // e.g. a single log file
 type LogSrc interface {
 	SetOutput(func(LogEvent))
+	SetFilter(func(LogEvent) bool)
```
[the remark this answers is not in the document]
SaxyPandaBear: This is in the interface, so it gets overridden in the tailer log source. Not sure if you wanted it to go somewhere else

plugins/inputs/logfile/fileconfig.go
```diff
@@ -20,8 +22,12 @@ import (
 const (
 	defaultMaxEventSize   = 1024 * 256 //256KB
 	defaultTruncateSuffix = "[Truncated...]"
+	includeType = "include"
+	excludeType = "exclude"
```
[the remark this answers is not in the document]
SaxyPandaBear: Good point. I can make the change

plugins/inputs/logfile/fileconfig.go
```diff
@@ -85,6 +96,25 @@ type FileConfig struct {
...
+type LogFilter struct {
+	Type string `toml:"type"`
+	Expression string `toml:"expression"`
+	ExpressionP *regexp.Regexp
+}
+
+func (filter *LogFilter) init() error {
+	var err error
+	if filter.Type != includeType && filter.Type != excludeType {
+		return fmt.Errorf("filter type is incorrect, regexp: Compile( %v )", filter.Expression)
```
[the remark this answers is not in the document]
SaxyPandaBear: Yup good catch. I'll see to trying to define a set of valid values like you mentioned in the above comment and then just reference it here. I think that was the issue I had trying to output a proper error here and just never circled back to it.

plugins/inputs/logfile/fileconfig.go
```diff
@@ -203,3 +251,55 @@ func (config *FileConfig) isMultilineStart(logValue string) bool {
 	}
 	return config.MultiLineStartPatternP.MatchString(logValue)
 }
+
+func (config *FileConfig) shouldFilterLog(event logs.LogEvent) bool {
```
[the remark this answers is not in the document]
SaxyPandaBear: Funnily enough, your suggestion about negative naming was something I was mulling over myself, so I appreciate getting confirmation that it's something I should pull the trigger on. 

As far as emitting the matched lines, I think I'm doing something _close_ already for matching the number of dropped logs, and I just put it into the profiler. When you flip the agent to debug mode, you can see the profiler dump the dropped log count per log file.

plugins/inputs/logfile/fileconfig.go
```diff
@@ -75,6 +81,11 @@ type FileConfig struct {
 
 	//Indicate retention in days for log group
 	RetentionInDays int `toml:"retention_in_days"`
+
+	Filters []LogFilter `toml:"filters"`
```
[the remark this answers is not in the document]
SaxyPandaBear: I'm not sure about the difference for doing this, so I'll have to dig into that to understand what the change would do.

plugins/inputs/logfile/fileconfig.go
```diff
@@ -85,6 +96,25 @@ type FileConfig struct {
...
+
+type LogFilter struct {
+	Type string `toml:"type"`
+	Expression string `toml:"expression"`
+	ExpressionP *regexp.Regexp
+}
+
+func (filter *LogFilter) init() error {
+	var err error
+	if filter.Type != includeType && filter.Type != excludeType {
```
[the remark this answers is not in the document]
SaxyPandaBear: Interesting way to make a set in Go. Didn't think of that. I can make this change.

plugins/inputs/logfile/fileconfig.go
SaxyPandaBear: Based on Ben's review, I think the suggestion is to go to what you were thinking originally where it's an AND condition. So, a lot of this conversation is moot. But I will make an effort to document the desired behavior.

plugins/inputs/logfile/fileconfig.go
SaxyPandaBear: That was my intent, but noting Ben's comment on the implementation, may not be true going forward.

plugins/inputs/logfile/logfile.go
```diff
@@ -235,6 +240,7 @@ func (t *LogFile) FindLogSrc() []logs.LogSrc {
 				tailer,
 				fileconfig.AutoRemoval,
 				mlCheck,
+				filterFn,
```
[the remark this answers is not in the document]
SaxyPandaBear: Noted. I'll see how it goes

plugins/inputs/logfile/tailersrc.go
SaxyPandaBear: Ahh I didn't even notice that this didn't account for truncation - assumed the truncation happened after. I'm of the opinion that trying to do the filtering on the log line before truncation is just going to end up biting us down the road more than only processing the filters on the truncated log line.

Will also try to cook up a unit test that illustrates behavior for truncated log lines

plugins/inputs/logfile/tailersrc_test.go
```diff
@@ -296,6 +300,261 @@ func TestOffsetDoneCallBack(t *testing.T) {
 	}
 }
 
+func TestTailerSrcFiltersSingleLineLogs(t *testing.T) {
```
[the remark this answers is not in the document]
SaxyPandaBear: Noted

translator/translate/logs/logs_collected/files/collect_list/ruleLogFilters.go
```diff
@@ -0,0 +1,51 @@
...
+	if val, ok := im[FiltersSectionKey]; ok {
+		filterArr := val.([]interface{})
+		for _, filter := range filterArr {
+			filterMap := map[string]interface{}{}
+
+			_, filterVal := translator.DefaultCase(FiltersTypeSectionKey, "", filter)
+			if filterVal == "" {
+				translator.AddErrorMessages(GetCurPath()+FiltersSectionKey, fmt.Sprintf("Filter %s is invalid", filter))
+				continue
+			}
```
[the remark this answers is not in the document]
SaxyPandaBear: How early were you thinking? In another part of the code, we check if the type is include/exclude and fail there, which stops the agent. This is in the translator, so I think we could duplicate the validation here, but would still need it in the other location I noted.

logs/logs.go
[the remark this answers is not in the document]
straussb: I mean I don't see the `tailerSrc` implementation being called anywhere either.

translator/translate/logs/logs_collected/files/collect_list/ruleLogFilters.go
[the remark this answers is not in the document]
straussb: Fair enough that we would still need to keep the check in the agent itself, so probably not worth duplicating the logic here.

plugins/inputs/logfile/fileconfig_test.go
```diff
@@ -173,3 +176,650 @@ func TestMultiLineStartPattern(t *testing.T) {
...
+	}
+
+	fileConfig := &FileConfig{
+		FilePath:              "/tmp/logfile.log",
+		LogGroupName:          "logfile.log",
+		TimestampRegex:        "(\\d{2} \\w{3} \\d{4} \\d{2}:\\d{2}:\\d{2})",
+		TimestampLayout:       "02 Jan 2006 15:04:05",
+		Timezone:              "UTC",
+		MultiLineStartPattern: "{timestamp_regex}",
+		Filters:               []LogFilter{filter1, filter2},
```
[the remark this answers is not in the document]
SaxyPandaBear: Good call out. I can trim it down

plugins/inputs/logfile/fileconfig.go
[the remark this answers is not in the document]
SaxyPandaBear: Ahh duh it's just pass by reference. Makes sense to me.

logs/logs.go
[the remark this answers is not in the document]
SaxyPandaBear: You're right. The function gets passed in with the construction in `NewTailerSrc`. With the refactoring, I think I'll just rip this out since it's probably not necessary anymore.

plugins/inputs/logfile/fileconfig.go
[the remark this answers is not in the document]
SaxyPandaBear: WRT the debug lines, I think a concern with using this functionality would be troubleshooting and supporting it. What are the consequences of being slow with processing the filters? How do we determine the filter processes are bottlenecking log publication? 

It was not entirely useful in my development because I wrote the benchmark unit tests which illustrate the information to me. I'm not opposed to dropping the elapsed time debug line. As far as the count metric for log messages, I am partial to tracking dropped log messages because technically, people who use CloudWatch today can already see the number of log events that get submitted to a log group.

plugins/inputs/logfile/fileconfig_test.go
```diff
@@ -173,3 +176,650 @@ func TestMultiLineStartPattern(t *testing.T) {
...
+		}},
+	}
+
+	err := fileConfig.init()
+	assert.NoError(t, err)
+
+	res := fileConfig.shouldFilterLog(LogEvent{
+		msg: "API responded with [StatusCode: 500] for call to /foo/bar",
+	})
+	assert.True(t, res)
```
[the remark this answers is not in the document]
SaxyPandaBear: I remember there was _some_ quirkiness when I was writing benchmark tests before, but don't recall if it pertained to assigning the result to `res` and asserting it, or if it was something else. I'll run benchmark tests with just the refactor, and then run again removing the assignment/assertion and see if there's obvious skew

plugins/inputs/logfile/fileconfig.go
SaxyPandaBear: As of right now I'm dropping these debug logs. Will keep this in mind if I reintroduce it though

plugins/inputs/logfile/tailersrc_test.go
```diff
@@ -321,3 +542,76 @@ func logLine(s string, l int, t time.Time) string {
...
+
+	ts.SetOutput(func(evt logs.LogEvent) {
+		if evt == nil {
+			close(done)
+			return
+		}
+		atomic.AddInt32(consumed, 1)
+		evt.Done()
+		// extra sanity check
+		assert.True(t, ShouldPublish(t.Name(), t.Name(), config.Filters, evt))
```
SaxyPandaBear, line 604: I took out another assertion here for the multiline check. Before introducing the truncated logs scenario, I had an assertion in the single line test to verify the log message did not contain a newline, and the opposite assertion in the multiline test. 

**Not sure why, but truncation only occurs for multiline logs**. As such, that broke my logic for this, so I had to remove that assertion from here. I considered adding another flag to check whether or not there is a newline in the log event, but figured that didn't add enough value to be worthwhile.

plugins/inputs/logfile/tailersrc_test.go
```diff
@@ -321,3 +542,76 @@ func logLine(s string, l int, t time.Time) string {
...
+	})
+}
+
+func assertExpectedLogsPublished(t *testing.T, total, numConsumed, expectedDropped int) {
+	assert.Equal(t, total - expectedDropped, numConsumed)
+	stats := profiler.Profiler.GetStats()
+	statKey := fmt.Sprintf("logfile_%s_%s_messages_dropped", t.Name(), t.Name())
```
SaxyPandaBear, line 611: Took out an assertion here about the size of the stats map because that just causes rigidity for no real value.

translator/config/schema.go
```diff
@@ -802,6 +808,25 @@ var schema = `{
             1827,
             3653
           ]
+        },
+        "filterDefinition": {
+          "type": "object",
+          "descriptions": "Define filters to apply to the log messages in this log file to determine whether to publish the message or not",
```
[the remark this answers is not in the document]
SaxyPandaBear: Can circle back with a PM about these

plugins/inputs/logfile/logfilter.go
```diff
@@ -0,0 +1,49 @@
...
+	expressionP *regexp.Regexp
+}
+
+func (filter *LogFilter) init() error {
+	if validFilterTypes == nil {
+		validFilterTypes = []string{includeFilterType, excludeFilterType}
+		validFilterTypesSet = make(map[string]bool)
+		for _, f := range validFilterTypes {
+			validFilterTypesSet[f] = true
+		}
```
straussb, line 32: You could do this right where `validFilterTypes` and `validFilterTypesSet` are declared, doesn't have to be inside this method.

translator/totomlconfig/toTomlConfig_test.go
```diff
@@ -165,6 +165,15 @@ func TestECSNodeMetricConfig(t *testing.T) {
 	os.Setenv("HOST_IP", "127.0.0.1")
 	checkIfTranslateSucceed(t, ReadFromFile("./sampleConfig/log_ecs_metric_only.json"), "./sampleConfig/log_ecs_metric_only.conf", "linux")
 	checkIfTranslateSucceed(t, ReadFromFile("./sampleConfig/log_ecs_metric_only.json"), "./sampleConfig/log_ecs_metric_only.conf", "darwin")
+	os.Unsetenv("RUN_IN_CONTAINER")
```
SaxyPandaBear, line 168: Had to add this because it gets set in the test but not cleaned up.

plugins/inputs/logfile/tailersrc_test.go
```diff
@@ -296,6 +305,65 @@ func TestOffsetDoneCallBack(t *testing.T) {
...
+	publishLogsToFile(file, matchedLog, unmatchedLog, n, 100)
+
+	// Removal of log file should stop tailersrc
+	if err := os.Remove(file.Name()); err != nil {
+		t.Errorf("failed to remove log file '%v': %v", file.Name(), err)
+	}
+	<-done
+	assertExpectedLogsPublished(t, n, int(consumed))
+}
+
```
straussb, line 366: By the way, about your comment "Not sure why, but truncation only occurs for multiline logs."  I believe that's because the buffered reader itself is [declared with a max line size](https://github.com/aws/amazon-cloudwatch-agent/blob/abdd1653cdcdf2cba7a91fa9615ca8deb683c7b2/plugins/inputs/logfile/tail/tail.go#L475-L475).  So if you have a single line that is longer than the max size, it will be split into multiple log events anyway.  (But then I would expect you would see more than `n` log events in `consumed`.)

straussb (commented):
I was doing my review last week before you pushed a bunch more commits, so the line numbers might be a little off.  But the content of the comments is still applicable.

plugins/inputs/logfile/fileconfig_test.go
```diff
@@ -173,3 +175,350 @@ func TestMultiLineStartPattern(t *testing.T) {
...
+
+func initializeLogFilters(t *testing.T, filters []*LogFilter) []*LogFilter {
+	for _, f := range filters {
+		err := f.init()
+		assert.NoError(t, err)
+	}
+	return filters
+}
+
+func initializeLogFiltersForBenchmarks(b *testing.B, filters []*LogFilter) []*LogFilter {
```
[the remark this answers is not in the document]
SaxyPandaBear: Yeah I think the `ResetTimer` for `testing.B` is what makes it annoying. Figured it wasn't worth trying to get around that.

plugins/inputs/logfile/tailersrc_test.go
```diff
@@ -296,6 +301,222 @@ func TestOffsetDoneCallBack(t *testing.T) {
...
+			Expression: "(ERROR|WARN)",
+		},
+		{
+			Type:       excludeFilterType,
+			Expression: "search_(\\w+)",
+		},
+	}
+	done := make(chan struct{})
+	var consumed int32
+	setupTailer(t, filters, nil, file, statefile, done, &consumed)
```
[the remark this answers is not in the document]
SaxyPandaBear: Good call out. I can change that.

plugins/inputs/logfile/tailersrc_test.go
SaxyPandaBear: That makes sense and explains the weird numbers I saw when I was writing a test for truncated logs. Wasn't aware of that. Noted.

plugins/inputs/logfile/logfilter.go
SaxyPandaBear: It didn't like me trying to loop through the splice to populate the map outside of a function definition. Not sure if what I have now is what you had in mind, but just copy pasting that code block out did not compile for me.

translator/config/schema.go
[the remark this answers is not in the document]
SaxyPandaBear: Got a green light on this. Also, I'm not sure this constitutes "customer-facing" in the same vein as the AWS public doc, since the schema is just used internally in the agent, as far as I know.

plugins/inputs/logfile/logfilter.go
straussb: Ah, sorry for recommending that without trying it out myself.  I like what you came up with.

plugins/inputs/logfile/tailersrc_test.go
straussb: Did we decide against having a truncated logs test at all then?  I think it would be good to verify the logic that filters are applied on the truncated (multi-line) message.  (You can choose a different max event size so you don't have to allocate 256KB of memory to run the test.)

translator/config/schema.go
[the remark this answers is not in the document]
straussb: schema.json is deployed in the RPM for customers to reference (I guess?).  So it's public-ish.

plugins/inputs/logfile/tailersrc_test.go
SaxyPandaBear: That was the conclusion from yesterday. I did tweak the max event size to be smaller, but it still fails consistently in CI, but works without issue locally. Also, the flakiness doesn't seem related to the sleep timing like the multiline log test.

straussb: Argh, that's annoying.  OK I will approve then.

straussb (approved):

PR #332, opened 2021-12-27
Remove some unnecessary Kubernetes API calls.

straussb:
The job code is never called anywhere else in the code.

The extra replicaset call was always intended to be removed before Container Insights went GA.  It was added during
Container Insights beta because CWAgent didn't always query K8S for replicaset information, so customers might not have
granted CWAgent permission to do so.  This line was supposed to fail fast in case permissions were lacking.  Now that
Container Insights is GA, customers should always be using the latest YAML file granting CWAgent the appropriate
permissions, so we don't need this fail-fast behavior.


SaxyPandaBear (approved):
LGTM

adam-mateen (approved):

Issue #199, opened 2021-04-12
Missing Windows Event logs in CloudWatch Logs `windows_events` plugin


mxiamxia:
### Description
This is a known issue when uploading Windows Event Logs to CloudWatch by using `windows_events` in CloueWatch Agent. Some of Windows Events will be failed to send  in the following condition.

When an operation is performed in Windows OS. A bunch of event logs will appear in `Event Viewer` with `Timestamp`, `event ID`, `Source service`.  For example,  Chrome application update events could have several related source service registered, (e.g. gupdate, gupdatem). If any of them is not registered itself correctly, it will block the whole event logs being retrieved from CWAgent and upload to CloudWatch.

adam-mateen:
I reproduced the issue on Windows Server 2019 (EC2 instance).
I ran this PowerShell script:
```
$ename = "Application"
$src = "Hermy"
$msg = "Hello World! "

if ([System.Diagnostics.EventLog]::SourceExists($src) -eq $False) {
	New-EventLog -LogName $ename -Source $src
}

$i = 0
while (1) {
	$cur_msg = $msg + "$i"
	write-output $cur_msg
	Write-EventLog -LogName $ename -Source $src -EventID 3001 -EntryType Error -Message $cur_msg
	sleep(1)
	$i += 1
}
```
Which outputs a new ERROR level event every second.
While that was running I installed Chrome, uninstalled Chrome, repeat.
I looked at CloudWatch console and I saw a dropped event.
I expected to see a messages like:
```
Hello World! 1
Hello World! 2
Hello World! 3
Hello World! 4
Hello World! 5
```
Instead 1 of the events was not reported.

The issue is that events may be retrieved in a batch.
And if the Agent encounters an issue with any of the events in the batch, it drops the whole batch.
Refer to `getRecords()`:
https://github.com/aws/amazon-cloudwatch-agent/blob/c3894b985db74b51a8dc07f087e1a2da698dff2b/plugins/inputs/windows_event_log/wineventlog/wineventlog.go#L327

I will open a PR to continue processing the batch of events even when 1 or more events cannot be rendered.

PR #333, opened 2021-12-29
Fix missing Windows events by changing getRecords().

adam-mateen:
# Description of the issue
Closes https://github.com/aws/amazon-cloudwatch-agent/issues/199
CloudWatch Agent reads from the Windows event logs in batches.
Currently, if the Agent has issues reading or rendering _any_ event in the batch, it will drop all the events in the batch.
In this specific case there was an event related to the install of Google Chrome for which `EvtOpenPublisherMetadata()` failed.

# Description of changes
Change `getRecords()` to `continue` when an error is encountered, instead of `return`.
Added unit tests.
Fixed another bug in `windowsEventLog.Open()` so it returns an error when `EvtSubscribe()` fails.

# License

# Tests
Manual test:
1. Start an EC2 instance running Windows Server 2019 with the latest CloudWatch Agent installed and configured to monitor the Application Event log for ERROR, INFO, VERBOSE.
2. Run a PowerShell script which repeatedly writes an ERROR level event to the Application event log with a specific message and an increasing counter.
3. Install and uninstall Google Chrome.
4. Expect CloudWatch console to show all events from the PowerShell script, but actual result shows missing events.

Added unit test:
```
PS C:\Users\Administrator\go\src\amazon-cloudwatch-agent> go test  .\plugins\inputs\windows_event_log\wineventlog\...
ok      github.com/aws/amazon-cloudwatch-agent/plugins/inputs/windows_event_log/wineventlog     6.594s
```

### DIFF
--- a/plugins/inputs/windows_event_log/wineventlog/sys_call.go
+++ b/plugins/inputs/windows_event_log/wineventlog/sys_call.go
@@ -1,3 +1,4 @@
+//go:build windows
 // +build windows
 
 // Portions Licensed under the Apache License, Version 2.0, Copyright (c) 2012–2017 Elastic <http://www.elastic.co>
--- a/plugins/inputs/windows_event_log/wineventlog/wineventlog.go
+++ b/plugins/inputs/windows_event_log/wineventlog/wineventlog.go
@@ -114,7 +114,7 @@ func (l *windowsEventLog) run() {
 			for _, record := range records {
 				value, err := record.Value()
 				if err != nil {
-					log.Printf("E! [windows_event_log] Error happened when collecting windows events : %v", err)
+					log.Printf("E! [wineventlog] Error happened when collecting windows events : %v", err)
 					continue
 				}
 				recordNumber, _ := strconv.ParseUint(record.System.EventRecordID, 10, 64)
@@ -159,7 +159,7 @@ func (l *windowsEventLog) Open() error {
 	// Subscribe for events
 	eventHandle, err := EvtSubscribe(0, uintptr(signalEvent), channelPath, query, bookmark, 0, 0, EvtSubscribeStartAfterBookmark)
 	if err != nil {
-		log.Printf("EvtSubscribe(), name %v, err %v", l.name, err)
+		log.Printf("E! [wineventlog] EvtSubscribe(), name %v, err %v", l.name, err)
 		return err
 	}
 
@@ -210,14 +210,14 @@ func (l *windowsEventLog) runSaveState() {
 			}
 			err := l.saveState(offset)
 			if err != nil {
-				log.Printf("E! [windows_event_log] Error happened when saving file state %s to file state folder %s: %v", l.logGroupName, l.stateFilePath, err)
+				log.Printf("E! [wineventlog] Error happened when saving file state %s to file state folder %s: %v", l.logGroupName, l.stateFilePath, err)
 				continue
 			}
 			lastSavedOffset = offset
 		case <-l.done:
 			err := l.saveState(offset)
 			if err != nil {
-				log.Printf("E! [windows_event_log] Error happened during final file state saving of logfile %s to file state folder %s, duplicate log maybe sent at next start: %v", l.logGroupName, l.stateFilePath, err)
+				log.Printf("E! [wineventlog] Error happened during final file state saving of logfile %s to file state folder %s, duplicate log maybe sent at next start: %v", l.logGroupName, l.stateFilePath, err)
 			}
 			break
 		}
@@ -250,10 +250,10 @@ func (l *windowsEventLog) read() []*windowsEventLogRecord {
 		// Handle special case when events size is too large - retry with smaller size
 		if err == RPC_S_INVALID_BOUND {
 			if maxToRead == 1 {
-				log.Printf("E! [windows_event_log] Out of bounds error due to large events size. Will skip the event as we cannot process it. Details: %v\n", err)
+				log.Printf("E! [wineventlog] Out of bounds error due to large events size. Will skip the event as we cannot process it. Details: %v\n", err)
 				return nil
 			}
-			log.Printf("W! [windows_event_log] Out of bounds error due to large events size. Retrying with half of the read batch size (%d). Details: %v\n", maxToRead/2, err)
+			log.Printf("W! [wineventlog] Out of bounds error due to large events size. Retrying with half of the read batch size (%d). Details: %v\n", maxToRead/2, err)
 			maxToRead /= 2
 			for _, h := range eventHandles {
 				EvtClose(h)
@@ -270,7 +270,7 @@ func (l *windowsEventLog) read() []*windowsEventLogRecord {
 
 	records := l.getRecords(eventHandles[:numRead])
 	if len(records) != int(numRead) {
-		log.Printf("E! [windows_event_log] requested events (%d) != returned events (%d)\n", numRead, len(records))
+		log.Printf("D! [wineventlog] requested events (%d) != returned events (%d)\n", numRead, len(records))
 	}
 	return records
 }
@@ -309,7 +309,7 @@ func (l *windowsEventLog) getRecords(handles []EvtHandle) (records []*windowsEve
 		// for rendering the event and getting a readable XML format that contains the log message.
 		// - We can later do more research on comparing other methods to get the publisher details such as EvtCreateRenderContext
 		if outputBuf, err = RenderEventXML(evtHandle, renderBuf); err != nil {
-			log.Printf("I! [windows_event_log] RenderEventXML() err %v", err)
+			log.Printf("I! [wineventlog] RenderEventXML() err %v", err)
 			continue
 		}
 
@@ -320,21 +320,21 @@ func (l *windowsEventLog) getRecords(handles []EvtHandle) (records []*windowsEve
 
 		var publisherMetadataEvtHandle EvtHandle
 		if publisherMetadataEvtHandle, err = EvtOpenPublisherMetadata(0, publisher, nil, 0, 0); err != nil {
-			log.Printf("I! [windows_event_log] EvtOpenPublisherMetadata() err %v, publisher %v", err, newRecord.System.Provider.Name)
+			log.Printf("I! [wineventlog] EvtOpenPublisherMetadata() err %v, publisher %v", err, newRecord.System.Provider.Name)
 			continue
 		}
 
 		var bufferUsed uint32
 		if err = EvtFormatMessage(publisherMetadataEvtHandle, evtHandle, 0, 0, 0, EvtFormatMessageXml, uint32(bufferSize), &renderBuf[0], &bufferUsed); err != nil {
 			EvtClose(publisherMetadataEvtHandle)
-			log.Printf("I! [windows_event_log] EvtFormatMessage() err %v, publisher %v", err, newRecord.System.Provider.Name)
+			log.Printf("I! [wineventlog] EvtFormatMessage() err %v, publisher %v", err, newRecord.System.Provider.Name)
 			continue
 		}
 		EvtClose(publisherMetadataEvtHandle)
 
 		var descriptionBytes []byte
 		if descriptionBytes, err = UTF16ToUTF8Bytes(renderBuf, bufferUsed); err != nil {
-			log.Printf("I! [windows_event_log] UTF16ToUTF8Bytes() err %v", err)
+			log.Printf("I! [wineventlog] UTF16ToUTF8Bytes() err %v", err)
 			continue
 		}
 
@@ -346,13 +346,13 @@ func (l *windowsEventLog) getRecords(handles []EvtHandle) (records []*windowsEve
 			//old SSM agent Windows format
 			var recordMessage eventMessage
 			if err = xml.Unmarshal(descriptionBytes, &recordMessage); err != nil {
-				log.Printf("I! [windows_event_log] UTF16ToUTF8Bytes() err %v", err)
+				log.Printf("I! [wineventlog] UTF16ToUTF8Bytes() err %v", err)
 				continue
 			}
 
 			newRecord.System.Description = recordMessage.Message
 		default:
-			log.Printf("I! [windows_event_log] l.renderFormat is not recognized, %s", l.renderFormat)
+			log.Printf("I! [wineventlog] l.renderFormat is not recognized, %s", l.renderFormat)
 			continue
 		}
 
@@ -364,22 +364,22 @@ func (l *windowsEventLog) getRecords(handles []EvtHandle) (records []*windowsEve
 
 func (l *windowsEventLog) loadState() {
 	if _, err := os.Stat(l.stateFilePath); err != nil {
-		log.Printf("I! [windows_event_log] The state file for %s does not exist: %v", l.stateFilePath, err)
+		log.Printf("I! [wineventlog] The state file for %s does not exist: %v", l.stateFilePath, err)
 		return
 	}
 
 	byteArray, err := ioutil.ReadFile(l.stateFilePath)
 	if err != nil {
-		log.Printf("W! [windows_event_log] Issue encountered when reading offset from file %s: %v", l.stateFilePath, err)
+		log.Printf("W! [wineventlog] Issue encountered when reading offset from file %s: %v", l.stateFilePath, err)
 		return
 	}
 
 	offset, err := strconv.ParseInt(strings.Split(string(byteArray), "\n")[0], 10, 64)
 	if err != nil {
-		log.Printf("W! [windows_event_log] Issue encountered when parsing offset value %v: %v", byteArray, err)
+		log.Printf("W! [wineventlog] Issue encountered when parsing offset value %v: %v", byteArray, err)
 		return
 	}
 
-	log.Printf("I! [windows_event_log] Reading from offset %v in %s", offset, l.stateFilePath)
+	log.Printf("I! [wineventlog] Reading from offset %v in %s", offset, l.stateFilePath)
 	l.eventOffset = uint64(offset)
 }
--- a/plugins/inputs/windows_event_log/wineventlog/wineventlog.go
+++ b/plugins/inputs/windows_event_log/wineventlog/wineventlog.go
@@ -67,50 +67,50 @@ func NewEventLog(name string, levels []string, logGroupName, logStreamName, rend
 	return eventLog
 }
 
-func (l *windowsEventLog) Init() error {
-	go l.runSaveState()
-	l.loadState()
-	return l.Open()
+func (w *windowsEventLog) Init() error {
+	go w.runSaveState()
+	w.loadState()
+	return w.Open()
 }
 
-func (l *windowsEventLog) SetOutput(fn func(logs.LogEvent)) {
+func (w *windowsEventLog) SetOutput(fn func(logs.LogEvent)) {
 	if fn == nil {
 		return
 	}
-	l.outputFn = fn
-	l.startOnce.Do(func() { go l.run() })
+	w.outputFn = fn
+	w.startOnce.Do(func() { go w.run() })
 }
 
-func (l *windowsEventLog) Group() string {
-	return l.logGroupName
+func (w *windowsEventLog) Group() string {
+	return w.logGroupName
 }
 
-func (l *windowsEventLog) Stream() string {
-	return l.logStreamName
+func (w *windowsEventLog) Stream() string {
+	return w.logStreamName
 }
 
-func (l *windowsEventLog) Description() string {
-	return fmt.Sprintf("%v%v", l.name, l.levels)
+func (w *windowsEventLog) Description() string {
+	return fmt.Sprintf("%v%v", w.name, w.levels)
 }
 
-func (l *windowsEventLog) Destination() string {
-	return l.destination
+func (w *windowsEventLog) Destination() string {
+	return w.destination
 }
 
-func (l *windowsEventLog) Retention() int {
-	return l.retention
+func (w *windowsEventLog) Retention() int {
+	return w.retention
 }
-func (l *windowsEventLog) Stop() {
-	close(l.done)
+func (w *windowsEventLog) Stop() {
+	close(w.done)
 }
 
-func (l *windowsEventLog) run() {
+func (w *windowsEventLog) run() {
 	ticker := time.NewTicker(1 * time.Second)
 	defer ticker.Stop()
 	for {
 		select {
 		case <-ticker.C:
-			records := l.read()
+			records := w.read()
 			for _, record := range records {
 				value, err := record.Value()
 				if err != nil {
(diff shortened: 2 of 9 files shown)


plugins/inputs/windows_event_log/wineventlog/sys_call.go
```diff
@@ -1,3 +1,4 @@
+//go:build windows
```
adam-mateen, line 1: This was added by:
```
 go fmt ./plugins/inputs/windows_event_log/wineventlog/... 
```

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
```diff
@@ -104,32 +105,19 @@ func (l *windowsEventLog) Stop() {
...
-				log.Printf("E! [windows_event_log] Due to corrupted/large event, skipping event log with record number %d, and log group name %s", recordNumber, l.logGroupName)
-				recordNumber = recordNumber + 1 // Advance to the next event to avoid being stuck
-				continue
-			}
-			if err != nil {
-				log.Printf("E! [windows_event_log] Failed to read Windows event logs for log group name %s. Details: %v\n", l.logGroupName, err)
-				recordNumber = recordNumber + 1
-				continue
-			}
-
```
adam-mateen, line 124: This code to track `recordNumber` is unecessary.
It gets parsed below and set in the `LogEvent` struct.
I moved printing the error inside the `read()` function.

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
```diff
@@ -171,7 +159,8 @@ func (l *windowsEventLog) Open() error {
 	// Subscribe for events
 	eventHandle, err := EvtSubscribe(0, uintptr(signalEvent), channelPath, query, bookmark, 0, 0, EvtSubscribeStartAfterBookmark)
 	if err != nil {
-		fmt.Errorf("error when subscribing for events. Details: %v", err)
+		log.Printf("EvtSubscribe(), name %v, err %v", l.name, err)
+		return err
```
adam-mateen, line 163: This bug was found by the new unit test for the `Open()` function.

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
```diff
@@ -244,7 +233,7 @@ func (l *windowsEventLog) saveState(offset uint64) error {
 	return ioutil.WriteFile(l.stateFilePath, content, 0644)
 }
 
-func (l *windowsEventLog) read() ([]*windowsEventLogRecord, error) {
+func (l *windowsEventLog) read() []*windowsEventLogRecord {
```
adam-mateen, line 236: The `run()` function was just logging the error, nothing more.
If that is all that is going to be done, then this `read()` can log the error itself and not return it.
If anyone feels strongly, we can return the first error encountered or last error encountered (if any).
But we'd just be guessing about how someone might want to use this function in the future.

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
```diff
@@ -277,10 +265,14 @@ func (l *windowsEventLog) read() ([]*windowsEventLogRecord, error) {
...
 	// Decode the events into objects
 	if numRead == 0 {
-		return nil, nil
+		return nil
 	}
 
-	return l.getRecords(eventHandles[:numRead])
+	records := l.getRecords(eventHandles[:numRead])
+	if len(records) != int(numRead) {
+		log.Printf("E! [windows_event_log] requested events (%d) != returned events (%d)\n", numRead, len(records))
```
adam-mateen, line 273: I am leaning towards making this a DEBUG level message.
Please share your thoughts.

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
```diff
@@ -324,19 +320,22 @@ func (l *windowsEventLog) getRecords(handles []EvtHandle) (records []*windowsEve
 
 		var publisherMetadataEvtHandle EvtHandle
 		if publisherMetadataEvtHandle, err = EvtOpenPublisherMetadata(0, publisher, nil, 0, 0); err != nil {
-			return nil, err
+			log.Printf("I! [windows_event_log] EvtOpenPublisherMetadata() err %v, publisher %v", err, newRecord.System.Provider.Name)
+			continue
```
adam-mateen, line 324: This is where the user's issue was coming from.
I didn't think these failures warranted an ERROR.
It is expected that some events in the event log cannot be rendered, which explains why the user had this issue in the first place.

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
```diff
@@ -347,18 +346,20 @@ func (l *windowsEventLog) getRecords(handles []EvtHandle) (records []*windowsEve
...
 			if err = xml.Unmarshal(descriptionBytes, &recordMessage); err != nil {
-				return nil, err
+				log.Printf("I! [windows_event_log] UTF16ToUTF8Bytes() err %v", err)
+				continue
 			}
 
 			newRecord.System.Description = recordMessage.Message
 		default:
-			return nil, fmt.Errorf("format %s is not recognized", l.renderFormat)
+			log.Printf("I! [windows_event_log] l.renderFormat is not recognized, %s", l.renderFormat)
```
adam-mateen, line 355: This file is part of the `package wineventlog`.
It makes sense to me to change all the logging messages to have `[wineventlog]` instead of `[windows_event_log]`.
I didn't want to clutter this PR anymore though.

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
```diff
@@ -171,7 +159,8 @@ func (l *windowsEventLog) Open() error {
 	// Subscribe for events
 	eventHandle, err := EvtSubscribe(0, uintptr(signalEvent), channelPath, query, bookmark, 0, 0, EvtSubscribeStartAfterBookmark)
 	if err != nil {
-		fmt.Errorf("error when subscribing for events. Details: %v", err)
+		log.Printf("EvtSubscribe(), name %v, err %v", l.name, err)
```
SaxyPandaBear, line 162: Do you need to prefix it with the E! in order to make it an error level log?

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
SaxyPandaBear: I'm not opposed to the change.

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
SaxyPandaBear: I agree with debug

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
```diff
@@ -347,18 +346,20 @@ func (l *windowsEventLog) getRecords(handles []EvtHandle) (records []*windowsEve
 			//old SSM agent Windows format
 			var recordMessage eventMessage
 			if err = xml.Unmarshal(descriptionBytes, &recordMessage); err != nil {
-				return nil, err
+				log.Printf("I! [windows_event_log] UTF16ToUTF8Bytes() err %v", err)
```
SaxyPandaBear, line 349: Is this log message correct? It looks like a copy paste error

plugins/inputs/windows_event_log/wineventlog/wineventlog_test.go
```diff
@@ -0,0 +1,163 @@
...
+	// common inputs for creating an EventLog.
+	NAME = "Application"
+	// 2 is ERROR
+	LEVELS          = []string{"2"}
+	GROUP_NAME      = "fake"
+	STREAM_NAME     = "fake"
+	RENDER_FMT      = FormatPlainText
+	DEST            = "fake"
+	STATE_FILE_PATH = "fake"
+	BATCH_SIZE      = 99
```
SaxyPandaBear, line 29: There's nothing that would require us to test different batch size scenarios, right? At least, not related to this change - just wanted to ask to be sure.

plugins/inputs/windows_event_log/wineventlog/wineventlog_test.go
```diff
@@ -0,0 +1,163 @@
...
+	assert.Equal(t, nil, elog.Close())
+}
+
+// seekToEnd skips past all current events in the event log.
+func seekToEnd(t *testing.T, elog *windowsEventLog) {
+	// loop until we stop getting records.
+	numRetrieved := 0
+	for {
+		records := elog.read()
+		//assert.Equal(t, nil, err)
```
SaxyPandaBear, line 108: Don't need this comment.

plugins/inputs/windows_event_log/wineventlog/wineventlog_test.go
```diff
@@ -0,0 +1,163 @@
...
+		t.Logf("seekToEnd() current count %v", len(records))
+		numRetrieved += len(records)
+		if len(records) == 0 {
+			break
+		}
+	}
+	t.Logf("seekToEnd() total %v", numRetrieved)
```
SaxyPandaBear, line 115: Question - do you think there's any added value for having an assertion/check for the number of records that get ingested when we seek to the end? I'm not opposed to leaving it as is, but if it is an easily derivable number we can check against, I think it could help prevent regression down the road.

plugins/inputs/windows_event_log/wineventlog/wineventlog_test.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+// writeEvents writes msgCount number of events to the Application event log.
+// Optionally register the given logSrc.
+// Fail the test if an error occurs.
+func writeEvents(t *testing.T, msgCount int, doRegister bool, logSrc string, eventId uint32) {
+	if doRegister {
+		// Expected to fail if unit test previously ran and installed the event src.
+		_ = eventlog.InstallAsEventCreate(logSrc, eventlog.Info|eventlog.Warning|eventlog.Error)
```
SaxyPandaBear, line 124: Do we not want to check what gets returned here?

plugins/inputs/windows_event_log/wineventlog/wineventlog_test.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+	wlog, err := eventlog.Open(logSrc)
+	assert.NoError(t, err)
+	for i := 0; i < msgCount; i++ {
+		msg := fmt.Sprintf("CWA_UnitTest event msg %v", i)
```
SaxyPandaBear, line 129: If `msg` just gets used in the next line, I think we could just consolidate and put the `fmt.Sprintf()` call in the log call

plugins/inputs/windows_event_log/wineventlog/wineventlog_test.go
```diff
@@ -0,0 +1,163 @@
...
+	err = wlog.Close()
+	assert.NoError(t, err)
+}
+
+// checkEvents reads all events (since last read) then checks each one for the
+// given substr. Verify count == msgCount.
+func checkEvents(t *testing.T, elog *windowsEventLog, msgCount int, substr string) {
+	// Get all new records.
+	var records []*windowsEventLogRecord
+	for {
```
SaxyPandaBear, line 141: Is there a possibility for this to run indefinitely? I see there's the `break` when the records we read is empty, but I feel like this is an easy thing to catch that could save a testing headache later on.

plugins/inputs/windows_event_log/wineventlog/wineventlog_test.go
```diff
@@ -0,0 +1,163 @@
...
+// registered event log source, even if the batch contains events from an
+// unregistered source too.
+func TestReadWithBothSources(t *testing.T) {
+	elog := NewEventLog(NAME, LEVELS, GROUP_NAME, STREAM_NAME, RENDER_FMT, DEST,
+		STATE_FILE_PATH, BATCH_SIZE, RETENTION)
+	assert.Equal(t, nil, elog.Open())
+	seekToEnd(t, elog)
+	writeEvents(t, 10, true, "CWA_UnitTest111", 777)
+	writeEvents(t, 10, false, "CWA_UnitTest222", 888)
+	time.Sleep(1 * time.Second)
```
SaxyPandaBear, line 97: What do we need the sleep for? Just going to leave this one comment, but I'm applying it to all of the sleep statements in the tests. I did not have fun with the tailersrc tests and their flaky timing. Don't want a repeat of that here if we can avoid it.

plugins/inputs/windows_event_log/wineventlog/wineventlog_test.go
```diff
@@ -0,0 +1,163 @@
...
+	elog := NewEventLog(NAME, LEVELS, GROUP_NAME, STREAM_NAME, RENDER_FMT, DEST,
+		STATE_FILE_PATH, BATCH_SIZE, RETENTION)
+	assert.NoError(t, elog.Open())
+	assert.NotNil(t, elog.eventHandle)
+	assert.NoError(t, elog.Close())
+	// bad name
+	elog = NewEventLog("FakeBadElogName", LEVELS, GROUP_NAME, STREAM_NAME,
+		RENDER_FMT, DEST, STATE_FILE_PATH, BATCH_SIZE, RETENTION)
+	assert.Error(t, elog.Open())
+	assert.Zero(t, elog.eventHandle)
```
SaxyPandaBear, line 55: Question - above, we assert in the happy case that `elog.eventHandle` is not nil, but here we assert that it is zero. For parity, should this be a nil check instead of a zero check?

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
SaxyPandaBear: Frankly I'm fine with including it in this change. That or a followup PR that does that change. I agree that the qualifier of the log message should match the package

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
adam-mateen: Yes, good catch.

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
adam-mateen: No, good catch.
Will fix.

plugins/inputs/windows_event_log/wineventlog/wineventlog_test.go
adam-mateen: Correct, this PR is not changing the hard coded batch size, `BatchReadSizeValue = 170`.

There is another open issue related to choosing the right batch size:
https://github.com/aws/amazon-cloudwatch-agent/issues/231

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
adam-mateen: Fixed.

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
adam-mateen: Changed.

plugins/inputs/windows_event_log/wineventlog/wineventlog_test.go
adam-mateen: In this case _I thought_ 2 lines looked more readable.
Either way is pretty simple though, so I am not opposed to changing it.

In general I am not shy about using more lines to make the code more readable.

plugins/inputs/windows_event_log/wineventlog/wineventlog_test.go
adam-mateen: I don't think so.
The call should succeed the first time the test is run.
Subsequent test runs should fail and that is okay.
I don't think we need to bother with "cleaning up" the registered event source either.

For context I expect this test to be run by the GitHub actions runner which uses `windows-latest` aka Windows 2019:
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources
And it is an actual VM, not just a docker container - so I expect this test to to work.

plugins/inputs/windows_event_log/wineventlog/wineventlog_test.go
adam-mateen: If the VM had a never ending stream of events getting created and it outpaced our ability to read events then yes it could be infinite.
I will add a max loop count to it just to be safe.

plugins/inputs/windows_event_log/wineventlog/wineventlog_test.go
adam-mateen: Because of a bad behavior in this package:
`"golang.org/x/sys/windows/svc/eventlog"`
https://cs.opensource.google/go/x/sys/+/1d35b9e2:windows/svc/eventlog/log.go;l=52

I added this comment in 1 place, but not all.
```
	// Without a sleep we do not read any events!!!
	// To work around the bad behavior of writeEvents() we must sleep.
	time.Sleep(1 * time.Second)
```
Let me know if you prefer I add that comment before ALL sleep() calls.

I definitely tested without the sleep and it failed.
I considered polling until we get the new events we expect or reach a time limit.
A single sleep call is simpler.

plugins/inputs/windows_event_log/wineventlog/wineventlog_test.go
adam-mateen: We should use `assert.Zero()` and `assert.NotZero()` because the type is:
```
type EvtHandle uintptr
```

I tried using `assert.Nil()` and `assert.NotNil()`, but got:
```
    wineventlog_test.go:55:
                Error Trace:    wineventlog_test.go:55
                Error:          Expected nil, but got: 0x0
                Test:           TestOpen
--- FAIL: TestOpen (0.03s)
```

plugins/inputs/windows_event_log/wineventlog/wineventlog_test.go
adam-mateen: We cannot.
We want the test to run on many different hosts.
We do not know the past history of events written to each hosts event log.
And even during our test we cannot guarantee that our test is the ONLY thing writing new events.
Which is why we get all events and just see if it at least contains the events our test wrote.

plugins/inputs/windows_event_log/wineventlog/wineventlog_test.go
adam-mateen: Removed.

plugins/inputs/windows_event_log/wineventlog/wineventlog_test.go
SaxyPandaBear: I would prefer adding the comment on all of them, but that's just me. That way, as things move around/change, we still have reference to this issue in every test that has the sleep to account for it.

plugins/inputs/windows_event_log/wineventlog/wineventlog_test.go
SaxyPandaBear: Generally I agree about readability, but in this case, it only gets used in the next line so there's not really any benefit on that front in my opinion.

plugins/inputs/windows_event_log/wineventlog/wineventlog_test.go
adam-mateen: Fixed.

plugins/inputs/windows_event_log/wineventlog/wineventlog_test.go
```diff
@@ -0,0 +1,160 @@
...
+// registered event log source, even if the batch contains events from an
+// unregistered source too.
+func TestReadWithBothSources(t *testing.T) {
+	elog := NewEventLog(NAME, LEVELS, GROUP_NAME, STREAM_NAME, RENDER_FMT, DEST,
+		STATE_FILE_PATH, BATCH_SIZE, RETENTION)
+	assert.Equal(t, nil, elog.Open())
+	seekToEnd(t, elog)
+	writeEvents(t, 10, true, "CWA_UnitTest111", 777)
+	writeEvents(t, 10, false, "CWA_UnitTest222", 888)
+	checkEvents(t, elog, 10, "[Application] [ERROR] [777] [CWA_UnitTest111] ")
```
straussb, line 94: Also include a `checkEvents(0, 888)` check here?

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
```diff
@@ -277,10 +265,14 @@ func (l *windowsEventLog) read() ([]*windowsEventLogRecord, error) {
...
 	// Decode the events into objects
 	if numRead == 0 {
-		return nil, nil
+		return nil
 	}
 
-	return l.getRecords(eventHandles[:numRead])
+	records := l.getRecords(eventHandles[:numRead])
+	if len(records) != int(numRead) {
+		log.Printf("D! [wineventlog] requested events (%d) != returned events (%d)\n", numRead, len(records))
```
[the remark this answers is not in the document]
adam-mateen: If a user was debugging why they aren't getting the windows events they expect, then this line would give them a an estimate of how many events have been dropped.

Each failure to read/render is also logged in the `getRecords()` method.
So logging a summary is just for convenience.
It definitely is not necessary.
I will remove it.

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
```diff
@@ -302,9 +294,12 @@ func (le LogEvent) Done() {
...
 }
 
-func (l *windowsEventLog) getRecords(handles []EvtHandle) (records []*windowsEventLogRecord, err error) {
+// getRecords attempts to render and format each of the given handles.
+// if one handle has an error, continue on because something is better than nothing.
+func (l *windowsEventLog) getRecords(handles []EvtHandle) (records []*windowsEventLogRecord) {
 	//Windows event message supports 31839 characters. https://msdn.microsoft.com/EN-US/library/windows/desktop/aa363679.aspx
 	bufferSize := 1 << 17
+	var err error
 	for _, evtHandle := range handles {
```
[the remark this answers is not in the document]
adam-mateen: I like that.
I will update it.

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
adam-mateen: In `windows_event_log.go` the `Start()` method for this plugin will fail if any of the monitored event logs fails `Init()`.

For new users, if they configure the CWA to monitor 5 event logs and 1 of them is invalid, I think it is better to fail like this so the user is more likely to notice and fix the configuration.

Although existing users may already have _bad_ event logs listed, and this would be a breaking change in behavior. So I will undo this change.

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
adam-mateen: Windows event log publishers typically register before writing events to a specific event log (i.e. "application", "system", "security"). It is possible to write an event without doing so though.

PowerShell won't let you do it:
```
PS C:\Users\Administrator\go\src\amazon-cloudwatch-agent> Write-EventLog -logname "Application" -Source "AdamSource" -EventId 123 -EntryType "Error" -Message "hello world"
Write-EventLog : The source name "AdamSource" does not exist on computer "localhost".
At line:1 char:1
+ Write-EventLog -logname "Application" -Source "AdamSource" -EventId 1 ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [Write-EventLog], InvalidOperationException
    + FullyQualifiedErrorId : Microsoft.PowerShell.Commands.WriteEventLogCommand
```

You must first call:
```
New-EventLog -LogName "Application" -Source "AdamSource"
```

You can do it with Python though:
```
"""
This is writing to the Windows application event log.
However the LOGSOURCE (applicationName) has not been created.
So the event will not display correctly in windows event viewer.
"""
import win32api
import win32con
import win32evtlog
import win32security
import win32evtlogutil

proc_handle = win32api.GetCurrentProcess()
proc_token = win32security.OpenProcessToken(proc_handle, win32con.TOKEN_READ)
my_sid = win32security.GetTokenInformation(proc_token, win32security.TokenUser)[0]

applicationName = "BarryPotter"
eventID = 321
category = 5    # Shell
myType = win32evtlog.EVENTLOG_ERROR_TYPE
descr = ["An error", "you know what it is"]
data = "Snape kills Dumbledore".encode("ascii")

win32evtlogutil.ReportEvent(applicationName, eventID, eventCategory=category,
    eventType=myType, strings=descr, data=data, sid=my_sid)
```

Note that even the windows event viewer will display the event, but the message is not displayed - which is expected, right?

So without digging into the how or why the Chrome install also causes this behavior, let's just leave it at - _it is possible_.

plugins/inputs/windows_event_log/wineventlog/wineventlog_test.go
```diff
@@ -0,0 +1,160 @@
...
+	"fmt"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"golang.org/x/sys/windows/svc/eventlog"
+)
+
+var (
```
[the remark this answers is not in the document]
adam-mateen: Tried that first, but it is not allowed:
```
plugins\inputs\windows_event_log\wineventlog\wineventlog_test.go:23:2: const initializer []string{...} is not a constant
```

plugins/inputs/windows_event_log/wineventlog/wineventlog_test.go
adam-mateen: Nope, the 2nd source was "bad".

Before my fix we would not get ANY events returned.
At least now we get some events from the valid source.

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
adam-mateen: Let me know if I am crazy, but in the current `master` branch there is only 1 place where the `recordNumber` variable is used:
https://github.com/aws/amazon-cloudwatch-agent/blob/master/plugins/inputs/windows_event_log/wineventlog/wineventlog.go#L136

And just before it is used it is set:
https://github.com/aws/amazon-cloudwatch-agent/blob/master/plugins/inputs/windows_event_log/wineventlog/wineventlog.go#L131

There is no point in setting a local variable or incrementing it if it is not used.
I was confused by this code too :)

adam-mateen: `EvtNext()` is a stateful call, meaning each call will return a different  list of EvtHandles, or nothing.
Since the `windowsEventLog.read()` method calls `EvtNext()` it is also stateful.

MSDN:
https://docs.microsoft.com/en-us/windows/win32/wes/querying-for-events#reading-events-from-the-result-set
```
he events in the result set are not static; new events that are written to the channel will be included in the result set until ERROR_NO_MORE_ITEMS is set.
```

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
adam-mateen: MSDN:
https://docs.microsoft.com/en-us/windows/win32/eventlog/event-sources
```
An application can use the Application log without adding a new event source to the registry. If the application calls RegisterEventSource and passes a source name that cannot be found in the registry, the event-logging service uses the Application log by default. However, because there are no message files, the Event Viewer cannot map any event identifiers or event categories to a description string, and will display an error. For this reason, you should add a unique event source to the registry for your application and specify a message file.
```

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
straussb: Ah I see.  This almost seems like a bug then, if the following happened:

1. Read corrupted/large event from event log, get RPC_S_INVALID_BOUND error.  No more events available in the log yet.
2. Agent stops.  No updated offset is persisted in the state file.
3. Agent starts again, starts reading at the bad event again.
4. Repeat.

I guess that's an edge case though and would be resolved as soon as a new valid event is available in the log.

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
straussb: Ah great point that this would be a breaking change.  That's unfortunate but I think you're right that we can't change it now.

plugins/inputs/windows_event_log/wineventlog/wineventlog_test.go
straussb: Yeah I mean, should we verify that we DON'T get any events from the bad source.

adam-mateen: Yeah, I didn't do that because the way I wrote `checkEvents()` doesn't allow it.
And `TestReadBadSource()` already verifies we DON't get any events from the bad source.
I will make it work though.

adam-mateen: Fixed.

plugins/inputs/windows_event_log/wineventlog/wineventlog_test.go
```diff
@@ -0,0 +1,184 @@
...
+	elog := NewEventLog(NAME, LEVELS, GROUP_NAME, STREAM_NAME, RENDER_FMT, DEST,
+		STATE_FILE_PATH, BATCH_SIZE, RETENTION)
+	assert.NoError(t, elog.Open())
+	seekToEnd(t, elog)
+	writeEvents(t, 10, true, "CWA_UnitTest111", 777)
+	writeEvents(t, 10, false, "CWA_UnitTest222", 888)
+	ee := []expectedEvent{
+		{"[Application] [ERROR] [777] [CWA_UnitTest111] ", 10},
+		{"CWA_UnitTest222", 0},
+	}
```
straussb, line 115: Ah, I see you had to do it this way because `checkEvents` statefully reads from the event log.  I might have done it like this instead then, for slightly better readability:

```
writeEvents(...)
records := readEvents(t, elog)
checkEvents(t, records, "CWA_UnitTest111", 10)
checkEvents(t, records, "CWA_UnitTest222", 0)
```

But not a big deal.

straussb (dismissed):

plugins/inputs/windows_event_log/wineventlog/wineventlog_test.go
adam-mateen: Fixed.

straussb (approved):

PR #334, opened 2022-01-04
Export Task ARN resource via Prometheus

leonsodhi-lf:
# Description of the issue
Metrics exported to Prometheus don't include the cluster name or task ID, making it harder to map the metric back to the particular task that generated it using the AWS web UI.

# Description of changes
The task ID doesn't appear to be directly accessible, but the last part of the task's ARN (the resource) includes the cluster name and ID. This PR extracts this part of the ARN and exports it.

# License

# Tests
Automated tests have been updated and pass, and the changes have also been manually tested using a Fargate cluster.

### DIFF
--- a/internal/ecsservicediscovery/decoratedtask.go
+++ b/internal/ecsservicediscovery/decoratedtask.go
@@ -9,6 +9,7 @@ import (
 	"strconv"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/arn"
 	"github.com/aws/aws-sdk-go/service/ecs"
 )
 
@@ -22,6 +23,7 @@ const (
 	taskLaunchTypeLabel  = "LaunchType"
 	taskJobNameLabel     = "job"
 	taskMetricsPathLabel = "__metrics_path__"
+	taskArnResourceLabel = "TaskArnResource"
 	ec2InstanceTypeLabel = "InstanceType"
 	ec2VpcIdLabel        = "VpcId"
 	ec2SubnetIdLabel     = "SubnetId"
@@ -138,6 +140,9 @@ func (t *DecoratedTask) generatePrometheusTarget(
 	addExporterLabels(labels, taskGroupLabel, t.Task.Group)
 	addExporterLabels(labels, taskStartedbyLabel, t.Task.StartedBy)
 	addExporterLabels(labels, taskLaunchTypeLabel, t.Task.LaunchType)
+	if arn, err := arn.Parse(*t.Task.TaskArn); err == nil {
+		addExporterLabels(labels, taskArnResourceLabel, &arn.Resource)
+	}
 	if t.EC2Info != nil {
 		addExporterLabels(labels, ec2InstanceTypeLabel, &t.EC2Info.InstanceType)
 		addExporterLabels(labels, ec2VpcIdLabel, &t.EC2Info.VpcId)
--- a/internal/ecsservicediscovery/decoratedtask.go
+++ b/internal/ecsservicediscovery/decoratedtask.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"regexp"
 	"strconv"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/arn"
@@ -23,7 +24,8 @@ const (
 	taskLaunchTypeLabel  = "LaunchType"
 	taskJobNameLabel     = "job"
 	taskMetricsPathLabel = "__metrics_path__"
-	taskArnResourceLabel = "TaskArnResource"
+	taskClusterNameLabel = "TaskClusterName"
+	taskIdLabel          = "TaskId"
 	ec2InstanceTypeLabel = "InstanceType"
 	ec2VpcIdLabel        = "VpcId"
 	ec2SubnetIdLabel     = "SubnetId"
@@ -140,9 +142,23 @@ func (t *DecoratedTask) generatePrometheusTarget(
 	addExporterLabels(labels, taskGroupLabel, t.Task.Group)
 	addExporterLabels(labels, taskStartedbyLabel, t.Task.StartedBy)
 	addExporterLabels(labels, taskLaunchTypeLabel, t.Task.LaunchType)
+
 	if arn, err := arn.Parse(*t.Task.TaskArn); err == nil {
-		addExporterLabels(labels, taskArnResourceLabel, &arn.Resource)
+		// ARN formats: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-account-settings.html#ecs-resource-ids
+		splitResource := strings.Split(arn.Resource, "/")[1:]
+		if len(splitResource) == 1 {
+			// Old ARN format
+			taskId := splitResource[0]
+			addExporterLabels(labels, taskIdLabel, &taskId)
+		} else if len(splitResource) == 2 {
+			// New ARN format
+			clusterName := splitResource[0]
+			taskId := splitResource[1]
+			addExporterLabels(labels, taskClusterNameLabel, &clusterName)
+			addExporterLabels(labels, taskIdLabel, &taskId)
+		}
 	}
+
 	if t.EC2Info != nil {
 		addExporterLabels(labels, ec2InstanceTypeLabel, &t.EC2Info.InstanceType)
 		addExporterLabels(labels, ec2VpcIdLabel, &t.EC2Info.VpcId)
--- a/plugins/inputs/logfile/tailersrc_test.go
+++ b/plugins/inputs/logfile/tailersrc_test.go
@@ -321,7 +321,7 @@ func TestTailerSrcFiltersSingleLineLogs(t *testing.T) {
 	n := 100
 	matchedLog := "ERROR: this has an error in it."
 	unmatchedLog := "Some other log message"
-	publishLogsToFile(resources.file, matchedLog, unmatchedLog, n, 0)
+	publishLogsToFile(resources.file, matchedLog, unmatchedLog, n, 100)
 
 	// Removal of log file should stop tailersrc
 	if err := os.Remove(resources.file.Name()); err != nil {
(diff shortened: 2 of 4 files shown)


leonsodhi-lf:
cc @SaxyPandaBear @adam-mateen as you've recently reviewed one or more other PRs

leonsodhi-lf:
@adam-mateen thanks for kicking off the previous test run. I don't _think_ the error is related to my changes and I also can't reproduce it locally using the same Go build (`1.15.5`). Guessing it's a race condition and that perhaps a fix is to increase the sleep period.

Would you mind giving the tests another go?

plugins/outputs/cloudwatch/aggregator_test.go
```diff
@@ -110,13 +110,13 @@ func TestAggregator_MultipleAggregationPeriods(t *testing.T) {
 	aggregator.AddMetric(m)
 
 	assertNoMetricsInChan(t, metricChan)
-	time.Sleep(2 * aggregationInterval)
+	time.Sleep(4 * aggregationInterval)
```
SaxyPandaBear, line 113: I disagree with these changes. We shouldn't increase sleep times in tests just to make the test pass. If it's a flaky test because of timing, then we should figure out what needs to be done in order to make the test more consistent, not sleep longer. This masks the issue, and makes the tests take longer to run.

leonsodhi-lf: Completely agree, but it would be good to see if increasing the sleep time even makes a difference - perhaps the race condition is elsewhere or it's something else entirely.

@SaxyPandaBear would you mind rerunning the tests with this change in place to see what happens?

adam-mateen: Sorry for the delay.
I just started the tests.

leonsodhi-lf: I've pushed a slightly better fix. Still not ideal, but there's only so much time I can devote to this so let's see how it pans out.

Would someone mind kicking off another test run?

leonsodhi-lf: Making progress. I've pushed a potential ~hack~ fix for another unrelated test

adam-mateen: Kicked off tests.

plugins/inputs/logfile/tailersrc_test.go
```diff
@@ -321,7 +321,7 @@ func TestTailerSrcFiltersSingleLineLogs(t *testing.T) {
 	n := 100
 	matchedLog := "ERROR: this has an error in it."
 	unmatchedLog := "Some other log message"
-	publishLogsToFile(resources.file, matchedLog, unmatchedLog, n, 0)
+	publishLogsToFile(resources.file, matchedLog, unmatchedLog, n, 100)
```
SaxyPandaBear, line 324: 1, the sleep timer that is in the test + the multilineWaitPeriod interval in the tailersrc implementation should not impact ingestion of log lines when there is no multiline function (the tailer created in this test passes in `nil` for that in line 318). The wait period is used to determine how long the agent waits with no new logs ingested before it pushes what is in memory and flushes. 

2, I already have my eye on these tests to figure out why they are flaky - I also wrote them.

leonsodhi-lf: Makes sense on 1, and I now realise my most recent change is pointless because leaving it at `0` won't update `multilineWaitPeriod` anyway and thus they'll never be a race condition there.

On 2, many thanks for looking into it! I suspect you'll have much more luck figuring this out than I will

SaxyPandaBear: #340 should hopefully address the flakiness in these tests. Can you revert this change that you made?

leonsodhi-lf: Done

leonsodhi-lf:
@SaxyPandaBear do you have any ideas on how to fix `TestAggregator_MultipleAggregationPeriods` in `github.com/aws/amazon-cloudwatch-agent/plugins/outputs/cloudwatch`? I gave it a shot but it's still a bit flakey

SaxyPandaBear:
@leonsodhi-lf I'd have to spend some time digging into how the source code and the tests work. I was intimately familiar with the flaky tests from the `tailersrc_test` suite because I wrote them and was modifying functionality in the tailer. 

I've seen that test suite be flaky in other PRs as well, so my take on it is that we should just retry the CI on this PR (rather than try to fix them in this PR) until it succeeds, and look at the flaky test(s) concurrently.

leonsodhi-lf:
Works for me. In that case, I think I'm all done here. Ready for review.

jhnlsn:
Just wanting to call out that we need to consider ARN format changes that happened recently.  It's not immediately clear that this code path is affected.  older services that have been around a while could follow the old ARN format that does not contain cluster-id

internal/ecsservicediscovery/decoratedtask.go
```diff
@@ -138,6 +140,9 @@ func (t *DecoratedTask) generatePrometheusTarget(
 	addExporterLabels(labels, taskGroupLabel, t.Task.Group)
 	addExporterLabels(labels, taskStartedbyLabel, t.Task.StartedBy)
 	addExporterLabels(labels, taskLaunchTypeLabel, t.Task.LaunchType)
+	if arn, err := arn.Parse(*t.Task.TaskArn); err == nil {
+		addExporterLabels(labels, taskArnResourceLabel, &arn.Resource)
```
jhnlsn, line 144: is there any value to be gained by attempting to break out the parts of the resource such as cluster-name and cluster-instance-id?

leonsodhi-lf: Personally, I'd really like that, but AFAIK the AWS SDK doesn't provide a way to do it so I settled for this. We could do a string split but that felt a bit dirty, and I was also concerned about ARN format changes. Definitely open to ideas on this one, though

SaxyPandaBear: I think it could be possible to take the ARN string and then use a regex to capture those values

jhnlsn: ECS task ARNs are deterministic and AFAIK there is just 2 formats for the resource.  String split on the / should do the trick.  I'm just wondering if you can put a / in the name. I'm going to guess it's not allowed.

SaxyPandaBear: Quickly perused on AWS console and don't think they allow you to put `/` in any of the names. The only issue is if a user created an ECS resource prior to the ARN change, I don't think their ARN would be updated (opt-out). Based on [docs](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-account-settings.html#ecs-resource-ids), there would only be one `/` in the old ARN format so if splitting the string gives us 2 tokens, then we don't have the cluster name in the resource.

SaxyPandaBear: I think rather than exporting the ARN, we should export the task ID and cluster name (if available) since that's the root issue you're facing. If the cluster name isn't in the ARN, then there's no difference between exporting just the task ID versus exporting the whole ARN. Personally, I like having the `arn.Parse()` call here just as a sanity check to make sure that the ARN passed in is a valid ARN, and would really just want to add additional work inside of the if-block here to extract the task ID and cluster name

leonsodhi-lf: Sgtm, and I think @SaxyPandaBear particularly nailed it with:

> we should export the task ID and cluster name (if available) since that's the root issue you're facing

Leave it with me and I'll get back to this next week or the week after.

jhnlsn: Based on the docs here there is an old and new format.  Looks like 2 or 3 parts of a string split by `/`.  Both old and new formats should be supported.

https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-account-settings.html#ecs-resource-ids

leonsodhi-lf: Done, and both formats are supported. Ready for another review.

SaxyPandaBear: I'll look through the changes tonight

internal/ecsservicediscovery/decoratedtask.go
```diff
@@ -138,6 +142,23 @@ func (t *DecoratedTask) generatePrometheusTarget(
 	addExporterLabels(labels, taskGroupLabel, t.Task.Group)
 	addExporterLabels(labels, taskStartedbyLabel, t.Task.StartedBy)
 	addExporterLabels(labels, taskLaunchTypeLabel, t.Task.LaunchType)
+
+	if arn, err := arn.Parse(*t.Task.TaskArn); err == nil {
+		// ARN formats: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-account-settings.html#ecs-resource-ids
+		splitResource := strings.Split(arn.Resource, "/")[1:]
```
SaxyPandaBear, line 148: My _only_ hesitation with this is if there were some drastic change and ARNs were changed to have zero `/` characters, this would panic, IIRC. But that feels like an impossible situation to me. And if it were to happen, I imagine such a breaking change would have a long runway for adoption, so we would have (hopefully) ample time to accommodate for that. Cross that bridge when we get there.

SaxyPandaBear (approved):

internal/ecsservicediscovery/decoratedtask.go
SaxyPandaBear: I'll still sleep at night even if this doesn't get changed, but I would feel better if there was a check like
```go
splitResource := strings.Split(arn.Resource, "/")
numTokens = len(splitResource)
if numTokens >= 3 {
  // assume index 1 and 2 are the task ID and cluster name
  clusterName := splitResource[1]
  taskId := splitResource[2]
} else if numTokens == 2 {
  taskId := splitResource[1]
}
```

leonsodhi-lf: I had the same concern, but Go handles it nicely: https://go.dev/play/p/EFjwYRYb7tZ

SaxyPandaBear: Ah okay. That works for me. My bad for assuming it would panic instead of just giving an empty splice.

Issue #315, opened 2021-11-16
[logfile plugin] agent start broken since "Put retention policy when writing or creating a log group (#250) "


cloud4t0r:
hi,

Since your using the function checkForDuplicateRetentionSettings() with
the merge of  "Put retention policy when writing or creating a log group (#250) " in master
my (valid) configuration which has been working for a long time don't works any more.
the cloudwatch-agent failed with the (false) error:
panic: error: retention for the same log group set in multiple places. Log Group Name: aphrodite_log

here is an excerpt from the amazon-cloudwatch-agent.json file :
        "logs": {
                "logs_collected": {
                        "files": {
                                "collect_list": [
                                        {
                                                "file_path": "/var/log/**.log",
                                                "log_group_name": "aphrodite_log",
                                                "log_stream_name": "{instance_id}_syslog",
                                                "timezone": "Local"
                                        },
                                        {
                                                "file_path": "/var/log/messages",
                                                "log_group_name": "aphrodite_log",
                                                "log_stream_name": "{instance_id}_messages",
                                                "timezone": "Local"
                                        },
                                        {
                                                "file_path": "/var/log/secure",
                                                "log_group_name": "aphrodite_log",
                                                "log_stream_name": "{instance_id}_secure",
                                                "timezone": "Local"
                                        },
                                        {
                                                "file_path": "/var/log/cron",
                                                "log_group_name": "aphrodite_log",
                                                "log_stream_name": "{instance_id}_cron",
                                                "timezone": "Local"
                                        }
                                ]
                        }
                }
        },
before this change, all my logs were grouped into a single log group for a given ec2 (here aphrodite_log),and that's exactly what i want ...
Now, it seems that if I don't specify the retention or if I only specify it once in the first block, the generated toml file has for each file_config block a "retention_in_days = -1" by default, what triggers the error in the checkForDuplicateRetentionSettings() function.

I do not understand the logic of this control which prevents us from now on using the same log-group to centralize and group our logs by machines in cloudwatch, which for me is a regression ... 
Please can you investigate this problem?
 thank you in advance for your feedback.

cloud4t0r.

jhnlsn:
Hello @cloud4t0r,

Thanks for reporting this issue.  First, Master has not been released yet.  How are you currently using the code?

This does look like a regression.  Current configuration should not be affected and retention should be an optional configuration.  We will take a look at this issue.

cloud4t0r:
Thanks John,
[...]

PR #339, opened 2022-01-18
fix duplicate retention check

aateeqi:
# Description of the issue
There was a customer issue where when retention in unset (and automatically set to -1), checkForDuplicateRetention would assume that there was multiple retention sets
https://github.com/aws/amazon-cloudwatch-agent/issues/315

# Description of changes
Add a check that if retention is set to -1, do not assume that retention is considered duplicate since its a default value and considered unset

# License

# Tests
Added additonal Unit tests `checkForDuplicateRetention`

### DIFF
--- a/plugins/inputs/logfile/logfile.go
+++ b/plugins/inputs/logfile/logfile.go
@@ -403,7 +403,7 @@ func (t *LogFile) checkForDuplicateRetentionSettings() {
 
 	for i := range t.FileConfig {
 		fileconfig := &t.FileConfig[i]
-		if fileconfig.LogGroupName != "" {
+		if fileconfig.LogGroupName != "" && fileconfig.RetentionInDays !=-1 {
 			logGroup := strings.ToLower(fileconfig.LogGroupName)
 			configMap[logGroup] += 1
 		}
--- a/plugins/inputs/logfile/logfile.go
+++ b/plugins/inputs/logfile/logfile.go
@@ -149,7 +149,7 @@ func (t *LogFile) FindLogSrc() []logs.LogSrc {
 
 	t.cleanUpStoppedTailerSrc()
 
-	// If a log group has retention settings defined in more than one place, stop the agent
+	// If a log group has differing retentionInDays values defined in multiple places, stop the agent
 	t.checkForDuplicateRetentionSettings()
 	// Create a "tailer" for each file
 	for i := range t.FileConfig {
@@ -410,11 +410,11 @@ func (t *LogFile) checkForDuplicateRetentionSettings() {
 		}
 		// if the configMap[logGroup] exists, retention has been set for the same logGroup somewhere
 		if configMap[logGroup] != 0 {
-			// 2 different retentions has been set for the same log group, panic and stop the agent
+			// different retentions has been set for the same log group, panic and stop the agent
 			if configMap[logGroup] != fileConfig.RetentionInDays {
-				panic(fmt.Sprintf("error: retention for the same log group set in multiple places. Log Group Name: %v", fileConfig.LogGroupName))
+				panic(fmt.Sprintf("error: The Log Group has differing retentionInDays values defined in two or more places. Log Group Name: %v", fileConfig.LogGroupName))
 			}
-			// The same retention for a log group has been configured in multiple places, so unset it so that the retention api is only called once
+			// The same retention for a log group has been configured in multiple places. Unset it so that the retention api is only called once
 			fileConfig.RetentionInDays = -1
 		} else {
 			configMap[logGroup] = fileConfig.RetentionInDays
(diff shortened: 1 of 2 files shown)


plugins/inputs/logfile/logfile_test.go
```diff
@@ -1090,3 +1090,65 @@ func TestCheckForDuplicateRetentionSettingsPanics(t *testing.T) {
...
+		RetentionInDays: 5,
+	},
+		{
+			FilePath:        "SampleFilePath",
+			FromBeginning:   true,
+			LogGroupName:    "logGroupName2",
+			RetentionInDays: 3,
+		},
+	}
+	assert.NotPanics(t, func() { tt.checkForDuplicateRetentionSettings() }, "Did not panic after finding duplicate log group")
```
SaxyPandaBear, line 1128: I think there's a copy paste issue with the error message for the assertion. Shouldn't this be "Did panic"? Based on the `assert.NotPanics()`

plugins/inputs/logfile/logfile.go
```diff
@@ -398,22 +398,17 @@ func (t *LogFile) cleanUpStoppedTailerSrc() {
...
+// checkForDuplicateRetentionSettings: Checks if a log group has retention set in multiple places and stops the agent if found
 func (t *LogFile) checkForDuplicateRetentionSettings() {
 	configMap := make(map[string]int)
-
 	for i := range t.FileConfig {
-		fileconfig := &t.FileConfig[i]
-		if fileconfig.LogGroupName != "" {
-			logGroup := strings.ToLower(fileconfig.LogGroupName)
+		fileConfig := &t.FileConfig[i]
+		if fileConfig.LogGroupName != "" && fileConfig.RetentionInDays !=-1 {
```
SaxyPandaBear, line 406: Thinking about this now, I think this (the feature, not this fix specifically) introduces an issue where if the user provides the same non-default value for their retention, we panic. I think that's confusing behavior.

plugins/inputs/logfile/logfile.go
```diff
@@ -398,22 +398,17 @@ func (t *LogFile) cleanUpStoppedTailerSrc() {
...
 			configMap[logGroup] += 1
 		}
-
-	}
-	for i := range t.FileConfig {
-		fileconfig := &t.FileConfig[i]
-		// log group has Retention settings in multiple places: throw an error
-		if fileconfig.LogGroupName != "" && configMap[strings.ToLower(fileconfig.LogGroupName)] > 1 {
-			panic(fmt.Sprintf("error: retention for the same log group set in multiple places. Log Group Name: %v", fileconfig.LogGroupName))
+		if fileConfig.LogGroupName != "" && configMap[strings.ToLower(fileConfig.LogGroupName)] > 1 {
```
SaxyPandaBear, line 410: If we intend to panic whenever we have more than one non-default value for the retention period for any given log group, I think we can restructure this check rather than just dropping the two existing if-statements into the loop. I feel like having the count is unnecessary if we're just concerned about the existence of at least two conflicting configurations. 

1. Define a `map[string]bool` for marking LogGroupName -> HasNonDefaultRetention
2. Iterate over each of the file configurations
3. If the retention policy is non-default (>= -1), then check if the log group name is in the map. If it is not, put it in the map, else panic

```go
configMap := make(map[string]bool)
for i := range t.FileConfig {
    fileConfig := &t.FileConfig[i]
    if fileConfig.RetentionInDays != -1 {
        if _, ok := configMap[fileConfig.LogGroupName]; !ok {
            configMap[fileConfig.LogGroupName] = true
        } else {
             panic()
        }
    }
}
```
or something like that. Obviously, if we factor in the other review comment I had about two equivalent non-default retention periods in the configuration, this would change.

plugins/inputs/logfile/logfile_test.go
```diff
@@ -1087,6 +1087,105 @@ func TestCheckForDuplicateRetentionSettingsPanics(t *testing.T) {
...
+			FilePath:        "SampleFilePath",
+			FromBeginning:   true,
+			LogGroupName:    logGroupName,
+			RetentionInDays: 3,
+		},
 	}
 	assert.Panics(t, func() { tt.checkForDuplicateRetentionSettings() }, "Did not panic after finding duplicate log group")
 }
+
+func TestCheckForDuplicateRetentionSettingsWithDefaultRetention(t *testing.T) {
```
SaxyPandaBear, line 1100: I think a test scenario I don't see (maybe it's already there but I don't see it in the diff) is a case where there's one config with a non-default value, and one config with the default -1 value, and asserting that it doesn't panic.

SaxyPandaBear (approved):
LGTM

adam-mateen (approved):

PR #340, opened 2022-01-18
Address flaky tests in GitHub CI

SaxyPandaBear:
# Description of the issue
`tailersrc_test` tests for log filtering added in #327 are flaky in GitHub.

# Description of changes
1. Noticed that older tailer tests had a 10x difference between the multilineWaitPeriod and the sleep timer (100ms vs 1s), so tweaked the sleep time in the log filter tests to match that. 
2. The single line log filter test sometimes would fail and not consume any logs. Added a sleep to allow the tailer to keep up with the log publication to (hopefully) keep it consistent

# License

# Tests
Kept PR in draft mode and reran the GitHub Actions to manually test several times to see if the tests passed more consistently.


adam-mateen (approved):
approved

PR #342, opened 2022-01-19
Add a feature flag to user agent for container insights

taohungyang:
# Description of the issue
We don't have the ability to see how much usage comes from `Container Insights` right now. Adding this so that we can have an idea what plugins/features are in use and potentially make users more willing to use these features.

# Description of changes
- As discussed, the outputs plugin name should be `container_insights`
- Update the output plugin list (in the user agent info) when the log group name matches the pattern of `Container Insights`

# License

# Tests
Added a new unit test `TestUserAgent`

### DIFF
--- a/cfg/agentinfo/info.go
+++ b/cfg/agentinfo/info.go
@@ -50,8 +50,8 @@ func Build() string {
 	return BuildStr
 }
 
-func Plugins() string {
-	outputs := strings.Join(OutputPlugins, " ")
+func Plugins(extraOutputs ...string) string {
+	outputs := strings.Join(append(OutputPlugins, extraOutputs...), " ")
 	inputs := strings.Join(InputPlugins, " ")
 
 	if !isRunningAsRoot() {
@@ -61,11 +61,11 @@ func Plugins() string {
 	return fmt.Sprintf("inputs:(%s) outputs:(%s)", inputs, outputs)
 }
 
-func UserAgent() string {
+func UserAgent(extraOutputs ...string) string {
 	if userAgent == "" {
 		ua := os.Getenv(envconfig.CWAGENT_USER_AGENT)
 		if ua == "" {
-			ua = fmt.Sprintf("%s %s", FullVersion(), Plugins())
+			ua = fmt.Sprintf("%s %s", FullVersion(), Plugins(extraOutputs))
 		}
 		userAgent = ua
 	}
--- a/cfg/agentinfo/info.go
+++ b/cfg/agentinfo/info.go
@@ -8,16 +8,19 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strings"
 
 	"github.com/aws/amazon-cloudwatch-agent/cfg/envconfig"
 )
 
-const versionFilename = "CWAGENT_VERSION"
-
-// We will fall back to a major version if no valid version file is found
-const fallbackVersion = "1"
+const (
+	containerInsightRegexp = "^/aws/.*containerinsights/.*/(performance|prometheus)$"
+	versionFilename        = "CWAGENT_VERSION"
+	// We will fall back to a major version if no valid version file is found
+	fallbackVersion = "1"
+)
 
 var isRunningAsRoot = func() bool {
 	return os.Getuid() == 0
@@ -29,7 +32,8 @@ var (
 	InputPlugins  []string
 	OutputPlugins []string
 
-	userAgent string
+	userAgentMap        = make(map[string]string)
+	ciCompiledRegexp, _ = regexp.Compile(containerInsightRegexp)
 )
 
 func Version() string {
@@ -50,30 +54,30 @@ func Build() string {
 	return BuildStr
 }
 
-func Plugins() string {
+func Plugins(groupName string) string {
 	outputs := strings.Join(OutputPlugins, " ")
 	inputs := strings.Join(InputPlugins, " ")
 
 	if !isRunningAsRoot() {
 		inputs += " run_as_user" // `inputs` is never empty, or agent will not start
 	}
+	if ciCompiledRegexp.MatchString(groupName) && !strings.Contains(outputs, "container_insights") {
+		outputs += "container_insights"
+	}
 
 	return fmt.Sprintf("inputs:(%s) outputs:(%s)", inputs, outputs)
 }
 
-func UserAgent() string {
-	if userAgent == "" {
-		ua := os.Getenv(envconfig.CWAGENT_USER_AGENT)
+func UserAgent(groupName string) string {
+	ua, found := userAgentMap[groupName]
+	if !found {
+		ua = os.Getenv(envconfig.CWAGENT_USER_AGENT)
 		if ua == "" {
-			ua = fmt.Sprintf("%s %s", FullVersion(), Plugins())
+			ua = fmt.Sprintf("%s %s", FullVersion(), Plugins(groupName))
 		}
-		userAgent = ua
+		userAgentMap[groupName] = ua
 	}
-	return userAgent
-}
-
-func ResetUserAgent() {
-	userAgent = ""
+	return ua
 }
 
 func FullVersion() string {
--- a/cfg/agentinfo/info.go
+++ b/cfg/agentinfo/info.go
@@ -50,8 +50,8 @@ func Build() string {
 	return BuildStr
 }
 
-func Plugins(extraOutputs ...string) string {
-	outputs := strings.Join(append(OutputPlugins, extraOutputs...), " ")
+func Plugins() string {
+	outputs := strings.Join(OutputPlugins, " ")
 	inputs := strings.Join(InputPlugins, " ")
 
 	if !isRunningAsRoot() {
@@ -61,11 +61,11 @@ func Plugins(extraOutputs ...string) string {
 	return fmt.Sprintf("inputs:(%s) outputs:(%s)", inputs, outputs)
 }
 
-func UserAgent(extraOutputs ...string) string {
+func UserAgent() string {
 	if userAgent == "" {
 		ua := os.Getenv(envconfig.CWAGENT_USER_AGENT)
 		if ua == "" {
-			ua = fmt.Sprintf("%s %s", FullVersion(), Plugins(extraOutputs))
+			ua = fmt.Sprintf("%s %s", FullVersion(), Plugins())
 		}
 		userAgent = ua
 	}
--- a/cfg/agentinfo/info.go
+++ b/cfg/agentinfo/info.go
@@ -72,6 +72,10 @@ func UserAgent() string {
 	return userAgent
 }
 
+func ResetUserAgent() {
+	userAgent = ""
+}
+
 func FullVersion() string {
 	return fmt.Sprintf("CWAgent/%s (%s; %s; %s) %s", Version(), runtime.Version(), runtime.GOOS, runtime.GOARCH, Build())
 }
--- a/cfg/agentinfo/info.go
+++ b/cfg/agentinfo/info.go
@@ -62,7 +62,7 @@ func Plugins(groupName string) string {
 		inputs += " run_as_user" // `inputs` is never empty, or agent will not start
 	}
 	if ciCompiledRegexp.MatchString(groupName) && !strings.Contains(outputs, "container_insights") {
-		outputs += "container_insights"
+		outputs += " container_insights"
 	}
 
 	return fmt.Sprintf("inputs:(%s) outputs:(%s)", inputs, outputs)
--- a/cfg/agentinfo/info_test.go
+++ b/cfg/agentinfo/info_test.go
@@ -82,20 +82,20 @@ func TestPlugins(t *testing.T) {
 
 	isRunningAsRoot = func() bool { return true }
 	plugins := Plugins("")
-	expected := fmt.Sprintf("inputs:(a b c) outputs:(x y z)")
+	expected := "inputs:(a b c) outputs:(x y z)"
+	if plugins != expected {
+		t.Errorf("wrong plugins string constructed '%v', expecting '%v'", plugins, expected)
+	}
+
+	plugins = Plugins("/aws/ecs/containerinsights/ecs-cluster-name/performance")
+	expected = "inputs:(a b c) outputs:(x y z container_insights)"
 	if plugins != expected {
 		t.Errorf("wrong plugins string constructed '%v', expecting '%v'", plugins, expected)
 	}
 
 	isRunningAsRoot = func() bool { return false }
 	plugins = Plugins("")
-	expected = fmt.Sprintf("inputs:(a b c run_as_user) outputs:(x y z)")
-	if plugins != expected {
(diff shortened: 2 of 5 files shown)


SaxyPandaBear:
Looks like a test failure:
```
--- FAIL: TestUserAgent (0.00s)
    cloudwatchlogs_test.go:69: The "container_insights" flag should be in the outputs plugin list for log groups from container insights
```

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -127,6 +128,12 @@ func (c *CloudWatchLogs) getDest(t Target) *cwDest {
 		},
 	)
 	client.Handlers.Build.PushBackNamed(handlers.NewRequestCompressionHandler([]string{"PutLogEvents"}))
+
+	containerInsightRegexp := "^/aws/.*containerinsights/.*/(performance|prometheus)$"
```
SaxyPandaBear, line 132: This looks like a constant. I know that we call `getDest` more than once (I believe once per log file?). We should move the regular expression outside of this function so that we don't have to compile it every time. Not sure exactly what the syntax would look like, but I think it'd be best as a top-level `var` or `const` for the Regexp struct like:
```go
const (
    containerInsightExpression := "^/aws/.*containerinsights/.*/(performance|prometheus)$"
)

var (
    containerInsightExpressionP := regexp.MustCompile(containerInsightExpression)
)
```

Or something like that. It might not work exactly like that, but the point is I think we can refactor this so that we don't have to compile it every time the `getDest` function is called.

plugins/outputs/cloudwatchlogs/cloudwatchlogs_test.go
```diff
@@ -44,3 +49,23 @@ func TestCreateDest(t *testing.T) {
...
+	}
+	d0 := c.getDest(target)
+	logClient := d0.pusher.Service.(*cloudwatchlogs.CloudWatchLogs)
+	req := request.New(
+		aws.Config{}, metadata.ClientInfo{}, logClient.Handlers, nil, &request.Operation{
+			HTTPMethod: "GET",
+			HTTPPath:   "/",
+		}, nil, nil)
+	logClient.Handlers.Build.Run(req)
+	if !strings.Contains(req.HTTPRequest.UserAgent(), "container_insights") {
```
SaxyPandaBear, line 68: Looking at the code change above in the implementation, I don't see it injecting the `container_insights` string into the user agent header. Is this the assertion we want to be making here? Currently the code appends to the `agentinfo.OutputPlugins` array, so I think the most direct assertion we could make is checking that the array contains `container_insights`. Can you explain why we're checking the HTTP request UserAgent value for `container_insights` here? It's not obvious to me how that is being set. 

Further, I think it would be good to add an assertion to an existing test (if we don't want to write another test for this sole purpose) to verify that the `container_insights` value is NOT appended when the group does not match the Regexp that you added.

taohungyang: The `container_insights` string is added into the HTTP request header from this line:
```
client.Handlers.Build.PushBackNamed(handlers.NewCustomHeaderHandler("User-Agent", agentinfo.UserAgent()))
```
where `agentinfo.UserAgent()` builds string from `agentinfo.OutputPlugins` and other variable.

But I agree this is not an obvious/direct way for an unit test. @jhnlsn also pointed out we could just checked `agentinfo.UserAgent()`/`agentinfo.OutputPlugins` for the expected values. I've updated the test code and added some more test cases to cover more

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
taohungyang: Good call! I've moved the regex to the top level and reuse it. Thanks!

taohungyang:
> Looks like a test failure:
> 
> ```
> --- FAIL: TestUserAgent (0.00s)
>     cloudwatchlogs_test.go:69: The "container_insights" flag should be in the outputs plugin list for log groups from container insights
> ```

yeah...I pushed this after I ran `go test -run TestUserAgent` successfully and didn't expect `make test` would act differently.
Turned out that because `agentinfo` stores the user agent string only when building at the 1st time and there are multiple test cases, the string was never updated by the test cases after the 1st one. Hence I added a function to reset the user agent string when dealing with the `container_insights` flag.

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -127,6 +131,20 @@ func (c *CloudWatchLogs) getDest(t Target) *cwDest {
...
+		for _, plugin := range agentinfo.OutputPlugins {
+			if plugin == "container_insights" {
+				hasCIFlag = true
+				break
+			}
+		}
+		if !hasCIFlag {
+			agentinfo.OutputPlugins = append(agentinfo.OutputPlugins, "container_insights")
+		}
+		agentinfo.ResetUserAgent()
```
jhnlsn, line 146: I think you want to only reset the UserAgent if you have to set the value, otherwise you're resetting it everytime there is a match.

SaxyPandaBear: +1. I don't think we want this here. The impetus for adding the `ResetUserAgent()` function is to reset for the tests. It should be used there.

cfg/agentinfo/info.go
```diff
@@ -72,6 +72,10 @@ func UserAgent() string {
 	return userAgent
 }
 
+func ResetUserAgent() {
```
SaxyPandaBear, line 75: This was added to reset the user agent during the tests. Should add a comment so that others can know that it's not meant to be used normally.

plugins/outputs/cloudwatchlogs/cloudwatchlogs_test.go
```diff
@@ -44,3 +47,51 @@ func TestCreateDest(t *testing.T) {
 		t.Errorf("Empty create dest should return dest to default group and stream, %v/%v found", d.pusher.Group, d.pusher.Stream)
 	}
 }
+
+func TestUserAgent(t *testing.T) {
+	c := outputs.Outputs["cloudwatchlogs"]().(*CloudWatchLogs)
```
SaxyPandaBear, line 52: Should reset the agent info in the test here.

cfg/agentinfo/info.go
jhnlsn: Thinking through this a little more in the concurrency model, is there a possibility of unsetting this as it's being passed to the handler?

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
jhnlsn: I'm going to take this back actually.  Before this change, useragent was a value that stayed static for the entire execution of the agent.  Now we have a situation where sometimes container_insights might be included in the plugins list for user agent and sometimes it might not.  The agent is able to both run container_insights to that log group as well as be configured for other log groups where we do not want that useragent string to have `container_insights`

taohungyang: Agreed. @SaxyPandaBear also mentioned `getDest` gets called multiple times. We don't need it to be reset every time there's a match

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -38,8 +39,11 @@ const (
 	metricRetryTimeout = 2 * time.Minute
 
 	attributesInFields = "attributesInFields"
+	containerInsightRegexp = "^/aws/.*containerinsights/.*/(performance|prometheus)$"
 )
 
+var ciCompiledRegexp, _ = regexp.Compile(containerInsightRegexp)
```
adam-mateen, line 45: I see there is a `prometheus_config_linux.json` file with this:
```
{
  "agent": {
    "region": "us-east-1"
  },
  "logs": {
    "metrics_collected": {
      "prometheus": {
        "cluster_name": "TestCluster",
        "log_group_name": "/aws/ecs/containerinsights/TestCluster/prometheus",
        "prometheus_config_path": "file:/tmp/prometheus.yaml",
        "ecs_service_discovery": {
          "docker_label": {
            "sd_job_name_label": "ECS_PROMETHEUS_JOB_NAME_1",
            "sd_metrics_path_label": "ECS_PROMETHEUS_METRICS_PATH",
            "sd_port_label": "ECS_PROMETHEUS_EXPORTER_PORT_SUBSET"
          },
```
But I also see docs say the `log_group_name` is optional:
>log_group_name— specifies the log group name for the scraped Prometheus metrics. This field is optional. If you omit it, CloudWatch uses /aws/containerinsights/cluster_name/prometheus for logs from Amazon EKS and Kubernetes clusters.

adam-mateen:
I do not see a linked issue.
Is there one?

`prometheus_scraper` is an input plugin.
So why is it necessary to also specify `container_insights` as an output plugin?
If it is still necessary, would it make more sense to add `container_insights` to the input plugin list?
https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/ContainerInsights.html

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
adam-mateen: +1 for removing the `agentinfo.ResetUserAgent()` call.
Even if the Agent was configured for uploading multiple log groups and only 1 of which matched the regex, `"^/aws/.*containerinsights/.*/(performance|prometheus)$"` It would still be okay for all uploads from this agent to have a `userAgent` string with `container_insights`.

My interpretation is that the `userAgent` is letting us know which features the customer is using, not which specific feature is associated with this specific upload.

cfg/agentinfo/info.go
taohungyang: As discussed, this has been removed. Also added a map to store the user agent string for each log group.

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
taohungyang: As discussed, this has been removed.

ZhenyuTan-amz (approved):

cfg/agentinfo/info_test.go
```diff
@@ -80,42 +81,79 @@ func TestPlugins(t *testing.T) {
...
+	tests := []struct {
+		name         string
+		logGroupName string
+		expected     string
+		errorMessage string
+	}{
+		{
+			"non container insights",
+			"test-group",
+			fmt.Sprintf("CWAgent/VSTR (%v; %v; %v) BSTR inputs:(a b c) outputs:(x y z)", runtime.Version(), runtime.GOOS, runtime.GOARCH),
```
ZhenyuTan-amz, line 121: Not sure if those print statement is necessary since this is just tests.

SaxyPandaBear: It's not a print statement that goes to the console. It's a format string for the `expected` attribute in the test struct. I'm fine with it as is, since the value for the UserAgent output is derivable.

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -145,7 +145,7 @@ func (c *CloudWatch) Connect() error {
 		})
 
 	svc.Handlers.Build.PushBackNamed(handlers.NewRequestCompressionHandler([]string{opPutLogEvents, opPutMetricData}))
-	svc.Handlers.Build.PushBackNamed(handlers.NewCustomHeaderHandler("User-Agent", agentinfo.UserAgent()))
+	svc.Handlers.Build.PushBackNamed(handlers.NewCustomHeaderHandler("User-Agent", agentinfo.UserAgent("")))
```
SaxyPandaBear, line 148: What are the repercussions of using an empty string for the group? What plugins are used in the empty string group?

cfg/agentinfo/info_test.go
```diff
@@ -80,42 +81,79 @@ func TestPlugins(t *testing.T) {
...
+			"/aws/containerinsights/cluster-name/prometheus",
+			fmt.Sprintf("CWAgent/VSTR (%v; %v; %v) BSTR inputs:(a b c) outputs:(x y z container_insights)", runtime.Version(), runtime.GOOS, runtime.GOARCH),
+			"\"container_insights\" flag shoould be in the outputs plugin list in container insights mode",
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			assert.Equal(t, tc.expected, UserAgent(tc.logGroupName), tc.errorMessage)
+
```
SaxyPandaBear, line 147: extra line here

cfg/agentinfo/info_test.go
```diff
@@ -80,42 +81,79 @@ func TestPlugins(t *testing.T) {
...
 
 	isRunningAsRoot = func() bool { return false }
-	plugins = Plugins()
+	plugins = Plugins("")
 	expected = fmt.Sprintf("inputs:(a b c run_as_user) outputs:(x y z)")
 	if plugins != expected {
 		t.Errorf("wrong plugins string constructed '%v', expecting '%v'", plugins, expected)
 	}
+
+	plugins = Plugins("/aws/ecs/containerinsights/ecs-cluster-name/performance")
```
SaxyPandaBear, line 97: Personally, I don't like that this test has multiple things checked in it, and this includes the `run_as_user` as a byproduct of the above scenario. Can we add an isolated test for just the change introduced?

taohungyang: I see what you are referring to. Tbh, I was debating if I need to add this new test here as the `container_insights` flag is also tested/verified in the `TestUserAgent` below. The tests below just make this one a bit redundant. But if we still want to add test cases to `Plugins()`, do you think it's ok if we move it to between the 1st and 2nd test case in this function (so that the expected string will not contain `run_as_user`)

plugins/outputs/cloudwatch/cloudwatch.go
taohungyang: This is the `cloudwatch` output plugin instead of `cloudwatchlogs`. When I check this function/package, I don't see a variable for log group name, so I just put an empty string here. Is there a better variable I can use here?

cfg/agentinfo/info_test.go
```diff
@@ -80,42 +81,79 @@ func TestPlugins(t *testing.T) {
...
+	expected = fmt.Sprintf("inputs:(a b c run_as_user) outputs:(x y z container_insights)")
```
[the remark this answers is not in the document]
taohungyang: good catch...I should have noticed this.

plugins/outputs/cloudwatch/cloudwatch.go
taohungyang: I don't think we need to worry about different goroutines for now as for `cloudwatch` plugin, the user agent will always be the same before and after this change:
- before: there was no group name as input so that it was always `agentinfo.userAgent`
- after: every thread/gorutine is using `""` as key to find `userAgentMap[""]`

SaxyPandaBear (approved):
LGTM

Issue #325, opened 2021-12-06
PodName of CronJob running on EKS 1.21 is not correct


yu-shiba:
The default controller for CronJob in Kubernetes v1.21 has been changed to CronJobControllerV2.
At that time, the minimum execution unit of CronJob was changed to 1 minute.

Therefore, the generation of Cron Job name is changed from 11 characters of `"-" + UnixTime`. (Probably from 9 to 10 characters).
https://github.com/kubernetes/kubernetes/blob/v1.20.1/pkg/controller/cronjob/utils.go#L245

But CloudWatch Agent is not able to capture the correct PodName because the code assumes 11 characters (10 characters without "-" in the code).
https://github.com/aws/amazon-cloudwatch-agent/blob/master/plugins/processors/k8sdecorator/stores/utils.go#L71

So, even if define a CronJob like "iam-waiting-for-EKS-1-22", the PodName will be "iam".

This issue can be confirmed by running CloudWatch Agent with EKS 1.21 in the standard state (not disabled by feature-gate).

khanhntd:
Hey @yu-shiba. Thanks for letting us know the issues. After deep dive into the CronJobControllerV2, the [getJobName](https://github.com/kubernetes/kubernetes/blob/c4d752765b3bbac2237bf87cf0b1c2e307844666/pkg/controller/cronjob/cronjob_controllerv2.go#L594-L596) has indeed used Unix Time in minutes instead of only Unix Time. Therefore, changing UnixTime from 10 characters to 9 characters currently and as a result, making the [assumption of 10 characters](https://github.com/aws/amazon-cloudwatch-agent/blob/master/plugins/processors/k8sdecorator/stores/utils.go#L71-L73) wrong though. What I am asking myself was is there really a need to check 10 runes or only Unix Time is fine? Will ask my teammates about this before making any decision and fixed this ASAP

yu-shiba:
Hi @khanhntd and team. Thank you for providing us with a useful product.  
I noticed this issue because the CronJob pod name shown in "CloudWatch Container Insights" was wrong.  
I hope my findings will help you to resolve the issue as soon as possible.

> What I am asking myself was is there really a need to check 10 runes or only Unix Time is fine?

I thought the same thing about this. And I couldn't make this decision, so I just let you know the results of the survey instead of PR.

PR #345, opened 2022-01-27
Address naming rules with CronJobControllerV2 for k8s v1.21

khanhntd:
# Description of the issue
Currently, K8s v1.21 has CronJobControllerV2 which uses Unix Time in Minutes to [define their CronJob name](https://github.com/kubernetes/kubernetes/blob/c4d752765b3bbac2237bf87cf0b1c2e307844666/pkg/controller/cronjob/cronjob_controllerv2.go#L594-L596) instead of using Unix Time (which previous version does use).  Therefore, it breaks our assumption about Unix Time (which has 10 characters). After the changes, the Unix Time in Minutes can be 7 - 9 characters (testing with the test case to figure it out).
**Link to track issues:** https://github.com/aws/amazon-cloudwatch-agent/issues/325

# Description of changes
There are two changes:
* Changing the assumption by convert the Unix Time In Minutes to Unix Time and checking it as a same condition with Unix Time.
* Instead of checking for every rune set, only need to use RegEx to reduce time complexity and adding test case for new CronJobControllerV2 from K8s.

# License

# Tests
By using ```go run utils_test.go``` or ```make test```, these units tests have been passed.


plugins/processors/k8sdecorator/stores/utils.go
```diff
@@ -51,42 +53,47 @@ func parseDeploymentFromReplicaSet(name string) string {
...
 	suffix := name[lastDash+1:]
-	if len(suffix) != 10 {
-		// Invalid suffix if it is not 10 rune
+	suffixInt, _ := strconv.ParseInt(suffix, 10, 64)
+	//Convert Unix Time In Minutes to Unix Time
+	suffixStringMultiply := strconv.FormatInt(suffixInt*60, 10)
+
+	//Checking if the suffix is a unix time by checking: the length and contains character
+	// - Checking for the length: CronJobControllerV2 is Unix Time in Minutes (7-9 characters) while CronJob is Unix Time (10 characters).
+	//   However, multiply by 60 to convert the Unix Time In Minutes back to Unix Time in order to have the same condition as Unix Time
```
aateeqi, line 83: nitpick but these could you format these comments like you did in line 64 to 66 for consistency?

plugins/processors/k8sdecorator/stores/utils.go
```diff
@@ -51,42 +53,47 @@ func parseDeploymentFromReplicaSet(name string) string {
...
 
-func stringInRuneset(name, subset string) bool {
-	for _, r := range name {
-		if !strings.ContainsRune(subset, r) {
-			// Found an unexpected rune in suffix
-			return false
-		}
-	}
-	return true
+func containUnexpectedRuneSet(name string, unexpectedRegEx string) bool {
```
SaxyPandaBear, line 96: Both variables here are a string, so the argument list should be collapsed like it was originally: `name, unexpectedRegEx string)`

plugins/processors/k8sdecorator/stores/utils.go
```diff
@@ -51,42 +53,47 @@ func parseDeploymentFromReplicaSet(name string) string {
...
+	var validUnixTimeRegExp = regexp.MustCompile(unexpectedRegEx)
```
SaxyPandaBear, line 97: I don't like having a `MustCompile` in here every time that this function is called, on two fronts.

1. We assume that the regex always compiles here. Otherwise, we get a hard panic if the string does not compile. 
2. We take a computational hit for the compilation every time this is called. Rather than doing this, I think you could pass in a `*regexp.Regexp` pointer, and at the top level, create and compile the regular expressions once.

plugins/processors/k8sdecorator/stores/utils.go
```diff
@@ -51,42 +53,47 @@ func parseDeploymentFromReplicaSet(name string) string {
...
-		// No dash
 		return ""
 	}
+
 	suffix := name[lastDash+1:]
-	if len(suffix) != 10 {
-		// Invalid suffix if it is not 10 rune
+	suffixInt, _ := strconv.ParseInt(suffix, 10, 64)
+	//Convert Unix Time In Minutes to Unix Time
+	suffixStringMultiply := strconv.FormatInt(suffixInt*60, 10)
```
SaxyPandaBear, line 79: Do we want to always do this check? I thought this was a breaking change in only some clusters?

plugins/processors/k8sdecorator/stores/utils.go
khanhntd: Reasonable. New things to learn in golang. Fixed it in newest commit

plugins/processors/k8sdecorator/stores/utils.go
khanhntd: Yes we always want to. Since AWS [has supported version 1.21 ](https://docs.aws.amazon.com/eks/latest/userguide/kubernetes-versions.html). Moreover, I haven't able to found a way in checking whether the CronJob is a CronJob or a ControllerV2. Therefore, leading to this solution though.

plugins/processors/k8sdecorator/stores/utils.go
khanhntd: Reasonable. Fixed it for newest commit. I thought this format would be easier for everyone to read though

plugins/processors/k8sdecorator/stores/utils.go
khanhntd: Reasonable for reason 1. However, I am having a hard time to understand the reason 2 though.

SaxyPandaBear: There is a cost associated with compiling the regex. If we only call this once or twice when the agent first starts up, I think it's fine. But if this gets called during the lifecycle of the agent, then there is a CPU cost that we incur for compiling the regular expression every time. In this case, we only have two static regular expressions to choose from, so there's no need to dynamically compile the input regex string every time. We can compile the regular expressions once at the top level in a `var ( )` block, and then refer to it as needed.

Take a look at https://github.com/aws/amazon-cloudwatch-agent/blob/master/plugins/processors/k8sdecorator/stores/podstore.go#L23-L36 as an example

plugins/processors/k8sdecorator/stores/utils.go
```diff
@@ -30,13 +31,23 @@ func createContainerKeyFromMetric(tags map[string]string) string {
...
 	// https://github.com/kubernetes/apimachinery/blob/master/pkg/util/rand/rand.go#L83
-	kubeAllowedStringAlphaNums = "bcdfghjklmnpqrstvwxz2456789"
-	cronJobAllowedString       = "0123456789"
+	deploymentUnexpectedRegEx = regexp.MustCompile(`[^b-hj-np-tv-xz-z2-24-9]+`)
+	// cronJobUnexpectedRegex ensures the characters in cron job name are only numbers.
+	cronJobUnexpectedRegex = regexp.MustCompile(`^[^0-9]+$`)
 )
 
+//Defining Embedding type to access to the RegExp methods (https://stackoverflow.com/a/43507669) since we cannot define new methods to non-local type
+type Regexp struct {
```
SaxyPandaBear: I don't think we need this. Will explain in below comment

plugins/processors/k8sdecorator/stores/utils.go
khanhntd: Interesting. I thought it would be better to use dynamic compiling. However, you are right about if we only have two then we only need to refer when it is needed instead of complexing it though. Also interesting about agent lifecycle though.

plugins/processors/k8sdecorator/stores/utils.go
```diff
@@ -30,13 +31,23 @@ func createContainerKeyFromMetric(tags map[string]string) string {
...
+	*regexp.Regexp
+}
+
+func (unexpectedRegEx *Regexp) containUnexpectedRuneSet(name string) bool {
```
SaxyPandaBear: Rather than defining a function of a struct like this, we could have just changed the `containUnexpectedRuneSet()` function to accept a regex as a parameter.
```suggestion
func containsUnexpectedRuneSet(name string, r *regexp.Regexp) bool {
```

Then further down, we can just use the regular expressions compiled as vars for the check:
```go
if containsUnexpectedRuneSet(suffix, cronJobUnexpectedRegex) {
```

khanhntd: Reasonable. Fixed it in the newest commit though. Thought it would be better to defining new structure type and thought of get ridding of the function at a later point. Seeing this make more sense to me though

plugins/processors/k8sdecorator/stores/utils.go
```diff
@@ -51,42 +53,52 @@ func parseDeploymentFromReplicaSet(name string) string {
 		return ""
 	}
 
-	if !stringInRuneset(suffix, kubeAllowedStringAlphaNums) {
+	if containUnexpectedRuneSet(suffix,deploymentUnexpectedRegExp) {
```
SaxyPandaBear, line 56: Nitpick: spacing between the arguments
```suggestion
	if containUnexpectedRuneSet(suffix, deploymentUnexpectedRegExp) {
```

plugins/processors/k8sdecorator/stores/utils.go
```diff
@@ -51,42 +53,52 @@ func parseDeploymentFromReplicaSet(name string) string {
...
+	//Checking if the suffix is a unix time by checking: the length and contains character
+
+	//Checking for the length: CronJobControllerV2 is Unix Time in Minutes (7-9 characters) while CronJob is Unix Time (10 characters).
+	//However, multiply by 60 to convert the Unix Time In Minutes back to Unix Time in order to have the same condition as Unix Time
+	if len(suffix) != 10 && len(suffixStringMultiply) != 10 {
+		return ""
+	}
+	
+	//Checking for unexpected character such as having characters others than numbers
+	if containUnexpectedRuneSet(suffix,cronJobUnexpectedRegexp) || containUnexpectedRuneSet(suffixStringMultiply,cronJobUnexpectedRegexp) {
```
SaxyPandaBear, line 95: Nitpick:
```suggestion
	if containUnexpectedRuneSet(suffix, cronJobUnexpectedRegexp) || containUnexpectedRuneSet(suffixStringMultiply, cronJobUnexpectedRegexp) {
```

plugins/processors/k8sdecorator/stores/utils.go
```diff
@@ -51,42 +53,52 @@ func parseDeploymentFromReplicaSet(name string) string {
...
 
-func stringInRuneset(name, subset string) bool {
-	for _, r := range name {
-		if !strings.ContainsRune(subset, r) {
-			// Found an unexpected rune in suffix
-			return false
-		}
-	}
-	return true
+func containUnexpectedRuneSet(name string, unexpectedRegexp *regexp.Regexp) bool {
```
SaxyPandaBear, line 102: Looking at this again, now at the entire function, I think that this is unnecessary. I should have thought about that before. 

We don't need this function. We can just call `MatchString()` where this is used and remove the extra function call. If there is other logic that needs to be done, having another function like this would make sense, but the way it's written right now, it doesn't provide any added value than just calling the `MatchString` function on the regular expressions directly.

So the above call would change from
```go
if containUnexpectedRuneSet(suffix, cronJobUnexpectedRegexp)
```
to:
```go
if cronJobUnexpectedRegexp.MatchString(suffix)
```

plugins/processors/k8sdecorator/stores/utils.go
```diff
@@ -30,11 +31,12 @@ func createContainerKeyFromMetric(tags map[string]string) string {
...
 }
 
-const (
-	// kubeAllowedStringAlphaNums holds the characters allowed in replicaset names from as parent deployment
+var (
+	// deploymentUnexpectedRegEx holds the characters allowed in replicaset names from as parent deployment
 	// https://github.com/kubernetes/apimachinery/blob/master/pkg/util/rand/rand.go#L83
-	kubeAllowedStringAlphaNums = "bcdfghjklmnpqrstvwxz2456789"
-	cronJobAllowedString       = "0123456789"
+	deploymentUnexpectedRegExp = regexp.MustCompile(`[^b-hj-np-tv-xz-z2-24-9]+`)
```
jhnlsn, line 37: I think this naming is throwing me off a bit.  The variable name makes it seem like the characters in the regex are not allowed and all other characters are allowed, however, I think the regex is defining only the allowed characters.

plugins/processors/k8sdecorator/stores/utils.go
khanhntd: [Same for me at some point](https://github.com/aws/amazon-cloudwatch-agent/pull/345/commits/df6456ccd5d4ae4aaf28ec9917b6c87e9cb0448d) and convinced by the function you provided

plugins/processors/k8sdecorator/stores/utils.go
```diff
@@ -51,42 +53,49 @@ func parseDeploymentFromReplicaSet(name string) string {
...
-		// Invalid suffix
+	//Convert Unix Time In Minutes to Unix Time
+	suffixStringMultiply := strconv.FormatInt(suffixInt*60, 10)
+
+	//Checking if the suffix is a unix time by checking: the length and contains character
+
+	//Checking for the length: CronJobControllerV2 is Unix Time in Minutes (7-9 characters) while CronJob is Unix Time (10 characters).
+	//However, multiply by 60 to convert the Unix Time In Minutes back to Unix Time in order to have the same condition as Unix Time
+	if len(suffix) != 10 && len(suffixStringMultiply) != 10 {
+		return ""
```
jhnlsn, line 91: Add tests for this return case

plugins/processors/k8sdecorator/stores/utils.go
```diff
@@ -51,42 +53,49 @@ func parseDeploymentFromReplicaSet(name string) string {
...
+	}
+	
+	//Checking for unallowed character such as having characters others than numbers
+	if cronJobUnallowedRegexp.MatchString(suffix) || cronJobUnallowedRegexp.MatchString(suffixStringMultiply) {
 		return ""
```
jhnlsn, line 96: Also add tests for this return case.  See cov report for reduction in test coverage for more details.

plugins/processors/k8sdecorator/stores/utils.go
khanhntd: Yes, the regexp is for checking un-allowed character in their name. In this way it would be easier to check for allowed character though.

plugins/processors/k8sdecorator/stores/utils.go
```diff
@@ -30,11 +31,12 @@ func createContainerKeyFromMetric(tags map[string]string) string {
...
 }
 
-const (
-	// kubeAllowedStringAlphaNums holds the characters allowed in replicaset names from as parent deployment
+var (
+	// deploymentUnallowedRegExp holds the characters allowed in replicaset names from as parent deployment
 	// https://github.com/kubernetes/apimachinery/blob/master/pkg/util/rand/rand.go#L83
-	kubeAllowedStringAlphaNums = "bcdfghjklmnpqrstvwxz2456789"
-	cronJobAllowedString       = "0123456789"
+	deploymentUnallowedRegExp = regexp.MustCompile(`[^b-hj-np-tv-xz-z2-24-9]+`)
```
jhnlsn, line 37: Can you create a test table test that sanity checks different string values that can be matched.

plugins/processors/k8sdecorator/stores/utils.go
khanhntd: These threes are for this return case https://github.com/aws/amazon-cloudwatch-agent/pull/345/files#diff-56aec85db8ce21fd23ac105fdbd3e19b8e7c93ec59777091a68deb2b726e6c59R25-R26, https://github.com/aws/amazon-cloudwatch-agent/pull/345/files#diff-56aec85db8ce21fd23ac105fdbd3e19b8e7c93ec59777091a68deb2b726e6c59R21

plugins/processors/k8sdecorator/stores/utils.go
khanhntd: [This test](https://github.com/aws/amazon-cloudwatch-agent/pull/345/files#diff-56aec85db8ce21fd23ac105fdbd3e19b8e7c93ec59777091a68deb2b726e6c59R21) is for this return case.

khanhntd: However, add some more unit test to ensure for both cases.

plugins/processors/k8sdecorator/stores/utils.go
jhnlsn: Can you check again, if you look at the coverage report this return point is never executed.

plugins/processors/k8sdecorator/stores/utils.go
SaxyPandaBear: I agree with @jhnlsn. The name of the variable is contradictory. It inverts the original one (this applies to both variables).

I think the name of the variable should be changed to reflect it's usage. Even in the updated `deploymentUnallowed`, it is directly contradictory to the original variable name, `kubeAllowed`, but matches the same characters.

My suggestion is to reuse what was already there:
```go
var (
    kubeAllowedChars = regexp.MustCompile()
    cronJobAllowedChars = regexp.MustCompile()
)
```

SaxyPandaBear: Looked over the old code, and I still stand by my opinion on this. The way that the strings were used previously, the code iterated over the input string, and checked if there were any chars that were not contained in the strings above. Switching from that to a regular expression makes sense. 

That being said, the naming is confusing because what we want to ensure is that the name of the cron job **matches** the two patterns.

plugins/processors/k8sdecorator/stores/utils.go
SaxyPandaBear: All the unit tests for this now use Unix time. Can you pass a string literal with a suffix that doesn't have a length of 10 instead of depending on the output of the Unix time? That would be a simple way to confirm hitting this return in test coverage.

plugins/processors/k8sdecorator/stores/utils_test.go
```diff
@@ -14,7 +16,14 @@ func TestUtils_parseDeploymentFromReplicaSet(t *testing.T) {
...
-	assert.Equal(t, "", parseCronJobFromJob("hello-123"))
-	assert.Equal(t, "hello", parseCronJobFromJob("hello-1234567890"))
-	assert.Equal(t, "", parseCronJobFromJob("hello-123456789a"))
+	//For CronJobControllV2 which is after K8s v1.21
+	assert.Equal(t, "hello", parseCronJobFromJob("hello-"+strconv.FormatInt(time.Now().Unix()/60, 10)))
+	assert.Equal(t, "", parseCronJobFromJob("hello-"+strconv.FormatInt(time.Now().Unix()/60, 10)+"abc"))
+	assert.Equal(t, "", parseCronJobFromJob("hello-"+strconv.FormatInt(time.Now().Unix()/60, 10)+"-name"))
+
+	//For CronJob which is before K8s v1.21
+	assert.Equal(t, "hello", parseCronJobFromJob("hello-"+strconv.FormatInt(time.Now().Unix(), 10)))
```
SaxyPandaBear, line 25: Do the original test assertions not work? If not, then this looks like a breaking change.

plugins/processors/k8sdecorator/stores/utils_test.go
```diff
@@ -14,7 +16,14 @@ func TestUtils_parseDeploymentFromReplicaSet(t *testing.T) {
 }
 
 func TestUtils_parseCronJobFromJob(t *testing.T) {
-	assert.Equal(t, "", parseCronJobFromJob("hello-123"))
-	assert.Equal(t, "hello", parseCronJobFromJob("hello-1234567890"))
-	assert.Equal(t, "", parseCronJobFromJob("hello-123456789a"))
+	//For CronJobControllV2 which is after K8s v1.21
```
SaxyPandaBear, line 19: I think iterating over a set of inputs/outputs would make this test clearer. See https://github.com/aws/amazon-cloudwatch-agent/blob/master/cfg/agentinfo/info_test.go#L104 for an example.

khanhntd: Yes. Changed to use table test driven and just committed. Takes long time to cover the coverage test case and goes beyond it.

plugins/processors/k8sdecorator/stores/utils.go
khanhntd: Reasonable. Fixed it in the newest commit.

plugins/processors/k8sdecorator/stores/utils_test.go
khanhntd: The original test assertions only covers the Cron Job but not CronJobControllerV2. Moreover, it does not cover some test case such as parsing Unix Time to integer with special characters though.

plugins/processors/k8sdecorator/stores/utils.go
khanhntd: Yes. I agrees that it needs to ```matches``` the two patterns. However, the two patterns are for unallowed character (meaning name does not allow these characters). Therefore, I am still unconvincing that we should use ```allowedChars```.

SaxyPandaBear: I think what we should do is flip the regular expression. I see why I'm confused now.

The original code would check:
```go
if !stringInRuneset(suffix, "bcdfghjklmnpqrstvwxz2456789")
```

the `stringInRuneset()` function iterates over `suffix`, and then if a character in `suffix` is NOT in the string `"bcdfghjklmnpqrstvwxz2456789"`, then it returns false, indicating that there was a character that should not have been there. For example, `stringInRuneset("bghk2789?", "bcdfghjklmnpqrstvwxz2456789")` would return `false` because ? is not an allowable character.

---
What you're changing it to is using a regular expression.
With the regex you defined:
```go
deploymentUnallowedRegExp = regexp.MustCompile(`[^b-hj-np-tv-xz-z2-24-9]+`)
```
I finally noticed the `^` for negativing it. 

This is what makes the variable and usage confusing. We should avoid using negative wording as it can be confusing. We should change it to affirmative wording, like what I suggested, and then flip the regular expression like:
```go
kubeAllowedChars = regexp.MustCompile(`[b-hj-np-tv-xz-z2-24-9]+`)
```

plugins/processors/k8sdecorator/stores/utils.go
khanhntd: Yes. Added only number test case with suffix smaller than 10 numbers though.

plugins/processors/k8sdecorator/stores/utils.go
khanhntd: For my opinion, flipping with ! outside is different from flipping with ! inside the regex though. For example, the `2a` would return false; however, it expected to be return true.

SaxyPandaBear: Why is it expected to return true? We control whether it returns true or false. 

The original code was as follows, and expected to return `false`, which gets negated.
```go
if !stringInRuneset(suffix, kubeAllowedStringAlphaNums) {
```
Your change inverts the if statement, so it drops the negation:
```go
if deploymentUnallowedRegExp.MatchString(suffix) {
```

If we change the regex to `regexp.MustCompile(`[b-hj-np-tv-xz-z2-24-9]+`)`, this checks if the input string contains only those characters.

So for an input suffix like `bjjk24789`, doing `MatchString()` on this with the above regex will return `true`. This indicates that the suffix is **valid**. As such, if we update the conditionals to expect false (which is what it was before your change), we maintain the correct functionality.

```go
if kubeAllowedChars.MatchString(suffix) {
```

plugins/processors/k8sdecorator/stores/utils.go
```diff
@@ -46,47 +48,46 @@ func parseDeploymentFromReplicaSet(name string) string {
...
+
+	if err != nil {
 		return ""
 	}
 
-	if !stringInRuneset(suffix, cronJobAllowedString) {
-		// Invalid suffix
-		return ""
+	//Convert Unix Time In Minutes to Unix Time
+	suffixStringMultiply := strconv.FormatInt(suffixInt*60, 10)
```
SaxyPandaBear, line 79: Technically, we only need this if the first if-statement evaluates to `false`.

plugins/processors/k8sdecorator/stores/utils.go
```diff
@@ -46,47 +48,46 @@ func parseDeploymentFromReplicaSet(name string) string {
...
+	//Checking if the suffix is a unix time by checking: the length and contains character
+	//Checking for the length: CronJobControllerV2 is Unix Time in Minutes (7-9 characters) while CronJob is Unix Time (10 characters).
+	//However, multiply by 60 to convert the Unix Time In Minutes back to Unix Time in order to have the same condition as Unix Time
+	if len(suffix) == 10 && !cronJobUnallowedRegexp.MatchString(suffix)  { //Condition for CronJob before k8s v1.21
```
SaxyPandaBear, line 83: I think that having happy paths wrapped in these if-statements and having the default invalid case return at the end is unintuitive. And this gets more confusing when we use the regular expression and check if the string does NOT match the regex. I understand why you would combine the two conditions, but what was the reason for flipping them? 

I think that 
```go
if len(suffix) != 10 || !cronJobUnallowedRegexp.MatchString(suffix) {
  return ""
}
```
is just as readable, and very easy to determine what conditions cause a failure case.

plugins/processors/k8sdecorator/stores/utils.go
```diff
@@ -30,11 +31,12 @@ func createContainerKeyFromMetric(tags map[string]string) string {
...
 
-const (
-	// kubeAllowedStringAlphaNums holds the characters allowed in replicaset names from as parent deployment
+var (
+	// deploymentUnallowedRegExp holds the characters allowed in replicaset names from as parent deployment
 	// https://github.com/kubernetes/apimachinery/blob/master/pkg/util/rand/rand.go#L83
-	kubeAllowedStringAlphaNums = "bcdfghjklmnpqrstvwxz2456789"
-	cronJobAllowedString       = "0123456789"
+	deploymentUnallowedRegExp = regexp.MustCompile(`[^b-hj-np-tv-xz-z2-24-9]+`)
+	// cronJobUnallowedRegexp ensures the characters in cron job name are only numbers.
```
SaxyPandaBear, line 38: This is what makes the variable names so confusing. The comment above the regex indicates that the regex is used to "ensure the characters [...] are **only numbers**", but the regular expression itself uses the `^` character to negate that. The statement about the regular expression contradicts (1) the name of the variable, and (2) the actual check the regex performs.

Looking at what is provided by Go's regexp library, there isn't a high level method for matching the regex against the entire string, but that's not necessary because we can construct a regular expression that means the same thing. We can use anchors in the regular expression to denote where the start and end of the string is, for the matching we want. In this case, we want to match `^\d+$`, so that from the beginning to the end of the string, there are only digit characters, and there's at least 1. 

I wrote some quick benchmark tests, too. 
```go
var cronJobUnallowedRegexp = regexp.MustCompile(`[^0-9]+`)
var anchored = regexp.MustCompile(`^\d+$`)

func BenchmarkDoNotMatchAnchored(b *testing.B) {
	s := "123abc" // should NOT match because of "abc"
	for i := 0; i < b.N; i++ {
		anchored.MatchString(s)
	}
	assert.False(b, anchored.MatchString(s))
}

func BenchmarkMatchAnchored(b *testing.B) {
	s := "123456" // should match because only digit chars
	for i := 0; i < b.N; i++ {
		anchored.MatchString(s)
	}
	assert.True(b, anchored.MatchString(s))
}

func BenchmarkMatchRegex(b *testing.B) {
	s := "123abc"
	for i := 0; i < b.N; i++ {
		cronJobUnallowedRegexp.MatchString(s)
	}
	assert.True(b, cronJobUnallowedRegexp.MatchString(s))
}

func BenchmarkDoNotMatchRegex(b *testing.B) {
	s := "123456"
	for i := 0; i < b.N; i++ {
		cronJobUnallowedRegexp.MatchString(s)
	}
	assert.False(b, cronJobUnallowedRegexp.MatchString(s))
}
```

```
goos: darwin
goarch: amd64
pkg: LoadTest
cpu: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
BenchmarkDoNotMatchAnchored
BenchmarkDoNotMatchAnchored-12    	12695895	        89.58 ns/op
BenchmarkMatchAnchored
BenchmarkMatchAnchored-12         	 8242377	       138.4 ns/op
BenchmarkMatchRegex
BenchmarkMatchRegex-12            	 8494279	       142.9 ns/op
BenchmarkDoNotMatchRegex
BenchmarkDoNotMatchRegex-12       	 7746750	       155.3 ns/op
PASS
```

plugins/processors/k8sdecorator/stores/utils.go
```diff
@@ -30,11 +31,12 @@ func createContainerKeyFromMetric(tags map[string]string) string {
...
+	cronJobUnallowedRegexp = regexp.MustCompile(`[^0-9]+`)
```
SaxyPandaBear, line 39: BTW, this regex could be simplified to:
```go
cronJobUnallowedRegexp = regexp.MustCompile(`\D+`)
```

plugins/processors/k8sdecorator/stores/utils.go
khanhntd: Using && would fail faster than using or though.  Since if the first condition is failed, then and would fail the results. However, if for the readable reason, it would be fine by me though.

plugins/processors/k8sdecorator/stores/utils.go
khanhntd: Nice catch. Fixed it in the newest commit.

plugins/processors/k8sdecorator/stores/utils.go
```diff
@@ -30,11 +32,12 @@ func createContainerKeyFromMetric(tags map[string]string) string {
...
 }
 
-const (
-	// kubeAllowedStringAlphaNums holds the characters allowed in replicaset names from as parent deployment
+var (
+	// deploymentUnallowedRegExp holds the characters allowed in replicaset names from as parent deployment
 	// https://github.com/kubernetes/apimachinery/blob/master/pkg/util/rand/rand.go#L83
-	kubeAllowedStringAlphaNums = "bcdfghjklmnpqrstvwxz2456789"
-	cronJobAllowedString       = "0123456789"
+	deploymentUnallowedRegExp = regexp.MustCompile(`[^b-hj-np-tv-xz-z2-24-9]+`)
```
SaxyPandaBear, line 38: Also don't need to do a range between z and 2. 
```go
deploymentUnallowedRegExp = regexp.MustCompile(`[^b-hj-np-tv-xz24-9]+`)
```

plugins/processors/k8sdecorator/stores/utils.go
khanhntd: Honestly, I do not think this would work though. These conditions would override the ```suffixStringMultiply``` conditions though.

plugins/processors/k8sdecorator/stores/utils.go
khanhntd: Nice catch. A little bad with RegEx.

plugins/processors/k8sdecorator/stores/utils.go
khanhntd: Yes we can control whether it returns true or false. However, my main point is the conditions with the RegEx. For example, checking whether not the number and whether it has the character besides number is different. 
To be more specific about that, 
```
if !kubeAllowedChars.MatchString(suffix) {
   return ""
}
```
and 
```
if kubeUnallowedChars.MatchString(suffix){
  return ""
}
```
is different though. Does the first case pass the "2ab", simply no . However, the second case passes.

SaxyPandaBear: See my discussion on this [here](https://github.com/aws/amazon-cloudwatch-agent/pull/345#discussion_r795713916)

plugins/processors/k8sdecorator/stores/utils.go
khanhntd: Interesting. You are right about anchor though. I need to gather my thoughts a little bit

SaxyPandaBear: Nitpick - the comments should either drop the variable name reference entirely, or update them to match the new names of the variables.

plugins/processors/k8sdecorator/stores/utils_test.go
```diff
@@ -5,16 +5,94 @@ package stores
...
+			expected: "",
+		},
+		{
+			name: "Get CronJobControllerV2 or CronJob's Name with Unix Time not equal to 10 letters",
+			inputString: "hello-238",
+			expected: "",
+		},
+		{
+			name: "Get CronJobControllerV2's Name after k8s v1.21 with correct Unix Time",
+			inputString: "hello-"+strconv.FormatInt(time.Now().Unix()/60, 10),
```
SaxyPandaBear, line 66: Nitpick - we only really need to get `time.Now()` once. The actual value isn't super important in these tests. We don't need to check the exact time multiple times. We only need one valid Unix timestamp that we can then call `strconv.FormatInt()` on it.

So something like:
```go
now := time.Now().Unix()
unixTime := strconv.FormatInt(now, 10)
unixMinutes := strconv.FormatInt(now/60, 10)

inputString1 := "hello-"+unixTime
inputString2 := "hello-"+unixMinutes+"foo"
```

plugins/processors/k8sdecorator/stores/utils.go
khanhntd: True. Forget to change it after the commit thought. Nice catch and thanks for pointing it out.

plugins/processors/k8sdecorator/stores/utils_test.go
khanhntd: Reasonable. Fixed it in the newest commit.

SaxyPandaBear (approved):

plugins/processors/k8sdecorator/stores/utils.go
```diff
@@ -30,11 +32,12 @@ func createContainerKeyFromMetric(tags map[string]string) string {
...
 }
 
-const (
-	// kubeAllowedStringAlphaNums holds the characters allowed in replicaset names from as parent deployment
+var (
+	// deploymentAllowedRegExp holds the characters allowed in replicaset names from as parent deployment
 	// https://github.com/kubernetes/apimachinery/blob/master/pkg/util/rand/rand.go#L83
-	kubeAllowedStringAlphaNums = "bcdfghjklmnpqrstvwxz2456789"
-	cronJobAllowedString       = "0123456789"
+	deploymentAllowedRegExp = regexp.MustCompile(`^[b-hj-np-tv-xz24-9]+$`)
```
jhnlsn, line 38: I didn't realize until now that the series of strings came from `rand.go`.  Would it be more future proof if we utilized `rand.SafeEncodeString()` instead of maintaining our own logic that could break if apimachinery changes their implementation?

something like `suffix == rand.SafeEncodeString(suffix)`?

SaxyPandaBear: I don't want to get bogged down by the design choices of k8s for their naming convention in apimachinery, which they could change at any point. As an aside, though, that code referenced hasn't been touched in any commits for over 4 years.

What is the agent's usage of the string? We've already proven that we're fragile to changes. In hindsight, I should have thought about this more to begin with, and I also didn't notice that this is pulled directly from a constant internally in k8s. If we're going to tackle this with the angle of future proofing from any updates that come to how apimachinery is creating identifiers, I think we have to take a step back and look at how we're using it. I'll have to read through the code later to see what it's used for. 

---

the `SafeEncodeString()` function is **not** idempotent. Although deterministic, and I think technically reversible, it is not idempotent, so I don't think that there is ever a scenario (at least not a reasonable scenario) where `suffix == rand.SafeEncodeString(suffix)` holds true.

https://go.dev/play/p/Jkbn-ceoaEf

khanhntd: Same thought. At least for rand.SafeEncodeString I believe. For now, we can merge this PR and create an issue for the independent between packages to track it. The only thing that I can offer is to track the milestone and see any updates relating to ```naming convention``` though. If we have any dependencies with K8s then I guess we can utilize ```dependabot alert``` to keep track automatically for us though but I guess we not

SaxyPandaBear: Rather than trying to guess at what k8s could change their naming convention to in the future, I think what we can do is the following:
1. Add two integration tests.
    1. One integration test that brings up an EKS cluster that uses k8s < v1.21, and verifies that the pod name is accurate
    2. One integration test that brings up an EKS cluster **on the latest version** and verifies that the pod name is accurate
2. Update the GitHub workflow to run integration tests on a weekly or bi-weekly schedule, so that we are consistently running the EKS integration tests, and not waiting for merges to the main branch to execute them

This way, we (1) provide further confidence that the change works, (2) bolster our GitHub integration test count, (3) continuously monitor integration test regression on GitHub, and (4) have a better chance of finding out about changes in k8s name changes in the future. Unless k8s provides a public method to extract the pod name themselves, then we are stuck in a situation where we have to infer the ruleset for the name generation regardless. Could be worthwhile to check if they already have an issue for that.

plugins/processors/k8sdecorator/stores/utils_test.go
```diff
@@ -4,17 +4,99 @@
...
 func TestUtils_parseDeploymentFromReplicaSet(t *testing.T) {
-	assert.Equal(t, "", parseDeploymentFromReplicaSet("cloudwatch-agent"))
-	assert.Equal(t, "cloudwatch-agent", parseDeploymentFromReplicaSet("cloudwatch-agent-42kcz"))
+	testcases := []struct {
+		name        string
+		inputString string
+		expected    string
+	}{
+		{
+			name:        "Get ReplicaSet Name with unallowed characters",
```
jhnlsn, line 21: One additional test for allowed characters but less than 3

khanhntd: Reasonable. Added it and it passed. Currently I am working on integration test with eks to ensure the naming convention.

plugins/processors/k8sdecorator/stores/utils.go
khanhntd: Yes. @SaxyPandaBear  has offered the idea of integration test for me and I have created it in a separate PR so no worries about holding up this PR.

SaxyPandaBear (approved):

PR #346, opened 2022-01-27
Add concurrency for pr

khanhntd:
# Description of the issue
Currently, there are two issues I want to address: 
- When we add changes to readme doc, create a PR out of it, merge into main; we would trigger the workflow build; however, that's not needed since relating to docs are not needed to trigger the build. 
- When creating a PR with multiple commits or merging into main, the Github Runners will scale based on the commits and the PRs have been merged into main. Therefore, to reduce the needed Github Runners, we can use [concurrency](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency) to avoid running simultaneously.

# Description of changes
Add paths ignore when merging into master and add concurrency to reduce GitHub Runners running a PR for a single branch.

# License

# Tests
**Concurrency:** https://github.com/khanhntd/aws-otel-collector/blob/main/.github/workflows/CI.yml#L47-L49
**Ignore:** https://github.com/khanhntd/aws-otel-collector/blob/main/.github/workflows/CI.yml#L22-L25

### DIFF
--- a/.github/workflows/build-test-windows.yml
+++ b/.github/workflows/build-test-windows.yml
@@ -6,9 +6,10 @@ on:
       - master
 
     paths-ignore:
+      - '**.md'
       - '.github/**'
       - '!.github/workflows/**'
-      - '**.md'
+
 
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
--- a/.github/workflows/build-test-windows.yml
+++ b/.github/workflows/build-test-windows.yml
@@ -7,8 +7,7 @@ on:
 
     paths-ignore:
       - '.github/**'
-      - '!.github/workflows/build.yml'
-      - '!.github/workflows/build-test-windows.yml'
+      - '!.github/workflows/**'
       - '**.md'
 
   pull_request:
--- a/.github/workflows/build-test-windows.yml
+++ b/.github/workflows/build-test-windows.yml
@@ -16,7 +16,7 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: build-and-test-windows-${{ github.ref_name }}
+  group: ${{ github.workflow }}-${{ github.ref_name }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/build-test-windows.yml
+++ b/.github/workflows/build-test-windows.yml
@@ -8,7 +8,7 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: build-and-test-windows-${{ github.event.pull_request.number }}
+  group: build-and-test-windows-${{ github.ref_name }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/build-test-windows.yml
+++ b/.github/workflows/build-test-windows.yml
@@ -2,7 +2,15 @@
 name: build-and-test-windows
 on:
   push:
-    branches: [ main, master ]
+    branches:
+      - master
+
+    paths-ignore:
+      - '.github/**'
+      - '!.github/workflows/build.yml'
+      - '!.github/workflows/build-test-windows.yml'
+      - '**.md'
+
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
   workflow_dispatch:
--- a/.github/workflows/build-test-windows.yml
+++ b/.github/workflows/build-test-windows.yml
@@ -10,7 +10,6 @@ on:
       - '.github/**'
       - '!.github/workflows/**'
 
-
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
   workflow_dispatch:
--- a/.github/workflows/build-test-windows.yml
+++ b/.github/workflows/build-test-windows.yml
@@ -7,6 +7,10 @@ on:
     types: [opened, synchronize, reopened, ready_for_review]
   workflow_dispatch:
 
+concurrency:
+  group: build-and-test-windows-${{ github.ref_name }}
+  cancel-in-progress: true
+
 jobs:
   windows-unittest:
     runs-on: windows-latest
--- a/.github/workflows/build-test-windows.yml
+++ b/.github/workflows/build-test-windows.yml
@@ -8,7 +8,7 @@ on:
   workflow_dispatch:
 
 concurrency:
-  group: build-and-test-windows-${{ github.ref_name }}
+  group: build-and-test-windows-${{ github.event.pull_request.number }}
   cancel-in-progress: true
 
 jobs:
--- a/.github/workflows/build-test-windows.yml
+++ b/.github/workflows/build-test-windows.yml
@@ -8,7 +8,7 @@ on:
     paths-ignore:
       - '**.md'
       - '.github/**'
-      - '!.github/workflows/**'
+      - '!.github/workflows/build-*'
 
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,8 +7,7 @@ on:
 
     paths-ignore:
       - '.github/**'
-      - '!.github/workflows/build.yml'
-      - '!.github/workflows/build-test-windows.yml'
+      - '!.github/workflows/**'
       - '**.md'
 
   pull_request:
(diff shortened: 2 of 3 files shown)


.github/workflows/build-test-windows.yml
```diff
@@ -2,11 +2,23 @@
...
+      - '.github/**'
+      - '!.github/workflows/build.yml'
+      - '!.github/workflows/build-test-windows.yml'
+      - '**.md'
+
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
   workflow_dispatch:
 
+concurrency:
```
SaxyPandaBear, line 18: > When creating a PR with multiple commits or merging into main, the Github Runners will scale based on the commits and the PRs have been merged into main

It took me a bit, but I think I understand what you're trying to do in this change. Essentially, you want to be able to cancel an ongoing GitHub workflow if/when you push a new commit to the PR, right?

.github/workflows/build-test-windows.yml
```diff
@@ -2,11 +2,23 @@
...
+  group: build-and-test-windows-${{ github.ref_name }}
```
SaxyPandaBear, line 19: I don't think that we need to write the literal string `build-and-test-windows` here. We should be able to use a placeholder like you did with `${{ github.ref_name}}`.

I think this should be
```suggestion
  group: ${{ github.workflow }}-${{ github.ref_name }}
```

.github/workflows/build.yml
```diff
@@ -4,12 +4,22 @@ on:
...
+      - '!.github/workflows/build-test-windows.yml'
+      - '**.md'
+
   pull_request:
     types: [opened, synchronize, reopened, ready_for_review]
   workflow_dispatch:
 
-jobs:
+concurrency:
+  group: build-and-test-${{ github.ref_name }}
```
SaxyPandaBear, line 19: same comment as above

.github/workflows/integrationTest.yml
```diff
@@ -10,6 +10,10 @@ on:
       - master
   workflow_dispatch:
 
+concurrency:
+  group: integration-tests-${{ github.ref_name }}
```
SaxyPandaBear, line 14: same comment as above

.github/workflows/build-test-windows.yml
khanhntd: Yes. That's correct though.

.github/workflows/build-test-windows.yml
khanhntd: Reasonable. Fixed it with the newest commit. Nice catch on this one though.

.github/workflows/integrationTest.yml
```diff
@@ -10,6 +10,10 @@ on:
       - master
   workflow_dispatch:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref_name }}
+  cancel-in-progress: true
```
SaxyPandaBear, line 15: I don't have a lot of insight on the integration tests. If we cancel it in progress, will the test still tear down the resources created during the integration test?

.github/workflows/build-test-windows.yml
```diff
@@ -2,11 +2,23 @@
...
 on:
   push:
-    branches: [ main, master ]
+    branches:
+      - master
+
+    paths-ignore:
+      - '.github/**'
+      - '!.github/workflows/build.yml'
+      - '!.github/workflows/build-test-windows.yml'
```
SaxyPandaBear, line 11: Instead of this can't we just put `- '!.github/workflows/**'`?

.github/workflows/integrationTest.yml
khanhntd: Yes since the integration test use [always as a condition](https://github.com/aws/amazon-cloudwatch-agent/blob/master/.github/workflows/integrationTest.yml#L62), it also covers the ```canceled``` condition too. Therefore, it would cancel the progress and run to the ```terraform destroy``` command before starting another workflow

.github/workflows/build-test-windows.yml
khanhntd: For ```nightly-build, integration-test or close-stalled```, I believe it does not affect to the main branch until it has been triggered by ```dispatch``` or ```schedule``` though. However, for build and build-test-windows it would affect immediately when we do a merge so I don't see any reasons to include ```nightly-build, integration-test or close-stalled```

jhnlsn: We will have more workflows in the future and a `/**` would future proof us instead of having to add absolute paths here each time a new file is added.

khanhntd: Reasonable for me at that solution but still, I find it weird to trigger the workflow is unrelated to branch master which will be triggered later  though

.github/workflows/build-test-windows.yml
```diff
@@ -2,11 +2,22 @@
...
 on:
   push:
-    branches: [ main, master ]
+    branches:
+      - master
+
+    paths-ignore:
+      - '.github/**'
+      - '!.github/workflows/**'
+      - '**.md'
```
SaxyPandaBear, line 11: Order matters in this configuration: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-and-excluding-paths

So shouldn't `**.md` go first, so that it future-proofs us if we ever put a markdown file in `./.github/workflows/`?

.github/workflows/build-test-windows.yml
SaxyPandaBear: If this is an issue, we could leverage the file names for the workflows to differentiate between what needs to be checked and what doesn't. Maybe something like
```yaml
  paths-ignore:
    - '.github/**'
    - '!.github/workflows/build-*.yml'
```
This would continue to ignore `nightly-build`, `integration-test`, and `close-stalled`

khanhntd: Make sense. Fixed it in the newest commit and nice catch though

.github/workflows/build-test-windows.yml
khanhntd: Make sense to me. Fixed it in the newest commit though.

.github/workflows/build-test-windows.yml
```diff
@@ -2,11 +2,22 @@
...
 on:
   push:
-    branches: [ main, master ]
+    branches:
+      - master
+
+    paths-ignore:
+      - '**.md'
+      - '.github/**'
+      - '!.github/workflows/**'
```
SaxyPandaBear, line 11: This should be updated like the other one, right? 
```suggestion
      - '!.github/workflows/build-*'
```

khanhntd: Weird. Thought I updated it too. Let's me fixed it again

SaxyPandaBear (approved):

.github/workflows/build.yml
```diff
@@ -4,16 +4,27 @@ on:
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref_name }}
+  cancel-in-progress: true
 
+jobs:
   build:
     name: Build
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
```
SaxyPandaBear, line 27: Why do we not want to fail fast? In the case where we have a flaky test, are we expecting to let the workflow run on both ubuntu and macos all the way through?

khanhntd: Yes. Based on these reasons though:
1. Fail-fast is not good for retry workflow since it will disapprove the success jobs. 
2. Even if we have a flaky test, it does not mean that it would fail based on the OS. Therefore, it should show the corresponding result with corresponding OS. Moreover, it would reduce the attempt to address the flakey test with the only failed OS.

SaxyPandaBear (approved):

.github/workflows/build-test-windows.yml
```diff
@@ -2,11 +2,22 @@
 name: build-and-test-windows
 on:
   push:
-    branches: [ main, master ]
+    branches:
+      - master
```
jhnlsn, line 6: Keep main, it's there for future transition to "main" instead of "master"

SaxyPandaBear: I missed that in my review. I did see that the other build yaml doesn't have `main` either though, so they'll both need to be updated.

PR #347, opened 2022-01-29
Delete runner cache and test cache if success

khanhntd:
# Description of the issue
Currently, there are two issues with time we got with:
* Should we build again even though the first retry has success? 
* The time for build is exactly the same without cache and with cache

# Description of changes
By changing these two,
* Adding cache to avoid building again for the second retry would reduce time and increase more resources spending on the remaining jobs
* Delete this path ```runner/Library/Caches/go-build``` when cache to reduce the time (testing it out with the PR) 

# License

# Tests
Test with a PR and found unexpected results without caching this path ```runner/Library/Caches/go-build```:
https://github.com/aws/amazon-cloudwatch-agent/runs/4990690788?check_suite_focus=true
https://github.com/aws/amazon-cloudwatch-agent/runs/4990745483?check_suite_focus=true

Compared when cached this path ```runner/Library/Caches/go-build```:
https://github.com/aws/amazon-cloudwatch-agent/runs/4990443776?check_suite_focus=true

### DIFF
--- a/.github/workflows/build-test-windows.yml
+++ b/.github/workflows/build-test-windows.yml
@@ -15,7 +15,7 @@ jobs:
         id: build-test
         uses: actions/cache@v2
         with:
-          path: AOC_GIT_SHA
+          path: LICENSE
           key: ${{ github.workflow }}-${{ github.run_id }}
 
       - name: Checkout Repo
--- a/.github/workflows/build-test-windows.yml
+++ b/.github/workflows/build-test-windows.yml
@@ -17,15 +17,19 @@ jobs:
         with:
           path: AOC_GIT_SHA
           key: ${{ github.workflow }}-${{ github.run_id }}
+
       - name: Checkout Repo
+        if: steps.build-test.outputs.cache-hit != 'true'
         uses: actions/checkout@v2
 
       - name: Setup Go
+        if: steps.build-test.outputs.cache-hit != 'true'
         uses: actions/setup-go@v2
         with:
           go-version: ~1.15.15
 
       - name: Cache Go
+        if: steps.build-test.outputs.cache-hit != 'true'
         uses: actions/cache@v2
         with:
           path: |
@@ -36,4 +40,5 @@ jobs:
             ${{ runner.os }}-go-
 
       - name: Run Unit tests
+        if: steps.build-test.outputs.cache-hit != 'true'
         run: go test ./...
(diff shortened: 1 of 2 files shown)


SaxyPandaBear:
Did you confirm that removing the macos cache path did not impact the build and test time?

SaxyPandaBear:
BTW if you are running verifications on changes to the GitHub workflow, I think you could change the PR to a `Draft` state so that it can't be approved and merged. From when I did that for checking flaky tests in CI, you can still run GitHub workflows on a draft PR.

khanhntd:
> Did you confirm that removing the macos cache path did not impact the build and test time?

Yes I have tested it and showed it in the description. For your convenience, here are the tests
* Without caching this path runner/Library/Caches/go-build:
https://github.com/aws/amazon-cloudwatch-agent/runs/4990690788?check_suite_focus=true
https://github.com/aws/amazon-cloudwatch-agent/runs/4990745483?check_suite_focus=true

* Compared when cached this path runner/Library/Caches/go-build:
https://github.com/aws/amazon-cloudwatch-agent/runs/4990443776?check_suite_focus=true

khanhntd:
> BTW if you are running verifications on changes to the GitHub workflow, I think you could change the PR to a `Draft` state so that it can't be approved and merged. From when I did that for checking flaky tests in CI, you can still run GitHub workflows on a draft PR.

Nice tips though. Forgot that we could release a draft PR instead of filing one.

jhnlsn:
Overall, I don't think you can skip building based on a cache hit.  the only thing we are caching is the output of the testing and the dependencies.  If code changes, it doesn't look like a new build/test will be triggered.

khanhntd:
> Overall, I don't think you can skip building based on a cache hit. the only thing we are caching is the output of the testing and the dependencies. If code changes, it doesn't look like a new build/test will be triggered.

It only works when retrying the workflow instead of making a new commits. That's why we use ```github.run_id``` that would be unique between commits but will be the same when retrying the workflow.

jhnlsn:
Ok, i see that now, which brings up another issue.  This means that you're only caching for the run, removing the original value of cache, which was to cache dependent packages and successful test runs so they didn't have to be downloaded again and the test didn't have to run again of no changes were made.

jhnlsn:
Closing this out out in favor of breaking out each unit test workflow into its own flow

Issue #170, opened 2021-02-09
Graceful shutdown


kao73:
How should I configure the CloudWatch Agent to make sure all collected data pushed away when I stop the agent?
I tried many options, but wasn't able to make it work as expected.

We use metrics to collect statistics in our application. Thus, we expect the metrics to be precise enough.
We have several ASG clusters with installed CloudWatch Agent with StatsD plugin on each EC2. There are several modules sending metrics to the agent. We found some metrics lost when ASG scales down. During investigation and some manual tests we found the agent doesn't publish collected data even if the amazon-cloudwatch-agent service stopped gracefully.

Can it be just incorrect configuration or WAD feature?

My configuration:
`{
  "agent": {
    "metrics_collection_interval": 60,
    "omit_hostname": true,
    "logfile": "/var/log/amazon-cloudwatch-agent.log"
  },
  "metrics": {
    "namespace": "MyNameSpace",
    "append_dimensions": {
      "AutoScalingGroupName": "${aws:AutoScalingGroupName}"
    },
    "metrics_collected": {
      "statsd": {
        "metrics_collection_interval": 60,
        "metrics_aggregation_interval": 0
      },
      "mem": {
        "metrics_collection_interval": 60,
        "metrics_aggregation_interval": 0,
        "measurement": [
          "used_percent"
        ]
      }
    }
  }
}`

For manual test we used:
* start the agent service
* send 1000 data points for a single metric by netcat
* wait about 20 seconds
* stop the agent service

Thanks.

pingleig:
I think if you send a SIGTERM to the agent, it should flush the data in buffer. btw: We are not following telegraf closely, which support flushing without shutdown using SIGUSER1 https://github.com/influxdata/telegraf/pull/7366

Shigerello:
Fluentd supports a variety of flushing and shutting-down options using signals as well.

https://docs.fluentd.org/deployment/signals

> SIGUSR1
> Forces the buffered messages to be flushed and reopens Fluentd's log. Fluentd will try to flush the current buffer (both memory and file) immediately, and keep flushing at flush_interval.

mihaileu:
I have same issue

[agent] Hang on, flushing any cached metrics before shutdown

doesn't flush the last aggregated stats. Is there a workaround or a fix for this ?

PR #350, opened 2022-02-01
Make cloudwatchlogs's pusher wait for the final flush to complete before returning

brandondahler:
# Description of the issue
During graceful shutdown, the cloudwatchlogs output plugin signals its goroutine to do a final flush; however, it doesn't actually wait for this goroutine to finish.  

This results in buffered metric entries being silently dropped if the agent's exit wins the race against the goroutine's flush/api call.

## Reproduction Steps
This was originally identified when testing against the EMF log metrics collector:

1. Import required configuration: 
   ```
   cat <<EOF > test.json
   {
     "agent": {
       "run_as_user": "cwagent",
       "debug": true
     },
     "logs": {
       "metrics_collected": {
         "emf": { }
       }
     }
   }
   EOF
   
   sudo amazon-cloudwatch-agent-ctl -a fetch-config -c file:test.json -s
   ```
2. Submit EMF-formatted log entry over UDP to the agent:
   ```
   echo '{"_aws":{"Timestamp":1574109732004,"LogGroupName":"Foo","CloudWatchMetrics":[{"Namespace":"MyApp","Dimensions":[["Operation"]],"Metrics":[{"Name":"ProcessingLatency","Unit":"Milliseconds"}]}]},"Operation":"Aggregator","ProcessingLatency":100}' > /dev/udp/0.0.0.0/25888; \
   sleep 1; \
   sudo amazon-cloudwatch-agent-ctl -a stop
   ```

### Expected Behavior
Log stream entry under Foo is created with the provided contents.

### Actual Behavior
No entry is created.

# Description of changes
Forces the Stop method to wait for the start goroutine via a WaitGroup that closes when the start method completes gracefully.  Previously the Stop method would close the stop channel, but wouldn't actually wait for the buffered events to be published.

# License

# Tests
* Removed the extra sleep from the associated test that previously was required because it didn't synchronously wait before.
* Tested with reproduction steps above and noted success with updated code.


brandondahler:
Potentially related to #170, though that issue seems to be using the cloudwatch output as opposed to the cloudwatchlogs output.

plugins/outputs/cloudwatchlogs/pusher.go
```diff
@@ -55,6 +55,7 @@ type pusher struct {
 
 	initNonBlockingChOnce sync.Once
 	startNonBlockCh       chan struct{}
+	wg                    sync.WaitGroup
```
SaxyPandaBear, line 58: Can we follow the structure that already exists in the `cloudwatch` output plugin where this attribute is a pointer to a WaitGroup?
```suggestion
	wg                    *sync.WaitGroup
```

SaxyPandaBear (commented):
The way this is written currently, there would be a separate `sync.WaitGroup` for each pusher, where we would have N pushers for N log files, IIRC. So that means N wait groups also.

When the agent gets shutdown, there isn't a scenario where we would really need to wait separately for each wait group. I think we can follow the structure that already exists in the `cloudwatch` plugin where we pass in a pointer to a WaitGroup when constructing the pusher struct, and then call `wg.Done()` in the `Stop()` function of the pusher, and we `wg.Wait()` one level up in the `cloudwatchlogs.go`.

We could store a single WaitGroup at the top level, and then just call `Wait()` on the group when we `Close()` the cloudwatchlogs struct. Could also restructure this change so that we don't `Add()` in the constructor function, and instead we add in the `start()` function of the pusher. I'm mostly indifferent if we `defer` the `Done()` call where you have it now, versus moving to the `Stop()` function of the pusher. 

--- 
Now, that being said, you referenced #170, and it does mildly concern me that there is a bug filed for the `cloudwatch` output plugin and I am advising to follow that same structure. It would be nice if we could add a clear test for validating that the graceful shutdown of the cloudwatchlogs struct does not drop data silently (if there isn't already one).

SaxyPandaBear:
Yup. I noticed this too @brandondahler and addressed it in the interim with https://github.com/aws/amazon-cloudwatch-agent/pull/352. Can you merge master into your PR branch?

SaxyPandaBear (approved):
Confirmed on the latest version of the agent that the repro steps fail (the Log group "Foo" is not created). 

Then confirmed that using the code from this PR (built locally then scp onto my EC2 for testing) that the log group gets created. 

LGTM

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -61,17 +61,23 @@ type CloudWatchLogs struct {
...
+	cwDests         map[Target]*cwDest
 }
 
 func (c *CloudWatchLogs) Connect() error {
 	return nil
 }
 
 func (c *CloudWatchLogs) Close() error {
+	close(c.pusherStopChan)
+	c.pusherWaitGroup.Wait()
```
adam-mateen, line 75: Is there a scenario where users would prefer the CWA to stop before sending pending data?
I am wondering about when CloudWatch APIs are timing out.
There is no timeout in this `Close()` function.
So we must rely on on each Pusher to have a reasonable timeout.

Right now I thinking potentially delaying the shutdown, even if it is for a 60 second HTTP timeout, would be preferred by _almost all_ users.

adam-mateen: It looks to me like `func (p *pusher) send` does not deterministically return in less than 60 seconds because each new  pusher is created like this:
```
pusher := NewPusher(t, client, c.ForceFlushInterval.Duration, maxRetryTimeout, c.Log)
```
where
```
maxRetryTimeout    = 14*24*time.Hour + 10*time.Minute
```
I recommend changing the `send()` function to check the status of:
```
stop                <-chan struct{}
```
at the end of the function before looping and retrying.
So I recommend adding an explicit timeout on

brandondahler: > Is there a scenario where users would prefer the CWA to stop before sending pending data?

From a user perspective, I don't personally see any reason to want CWA to drop pending data.  If the user expects/requires CWA to shutdown within a specific timeframe, I would expect them to do so by using the appropriate signals -- namely sending SIGKILL if they actually want the agent to stop immediately..

> Right now I thinking potentially delaying the shutdown, even if it is for a 60 second HTTP timeout, would be preferred by almost all users.

From a practical perspective, most service managers work similar to systemd.  By default systemd will wait 90 seconds after the process is sent the SIGTERM signal, if it still hasn't quit it'll then send a SIGKILL..

brandondahler: > I recommend changing the send() function to check the status of:
> ```
> stop                <-chan struct{}
> ```
>
> at the end of the function before looping and retrying.

Addressed this by changing the `time.Sleep(wait)` to a select for either the stop channel or `time.After(wait)`.  This effectively causes all retries to be immediately stopped, which should guarantee that the WaitGroup will return almost immediately.

I added a test to verify this behavior as well.

brandondahler:
Addressed review and rebased on latest master branch.

adam-mateen (approved):
approved

SaxyPandaBear (approved):

Issue #353, opened 2022-02-02
Breakout unit testing per OS


jhnlsn:
We currently have 2 workflow files.  1 for windows and 1 for Mac and Linux.

The issue with the mac/linux build/test workflow file is that if the run fails in 1 OS, we are required to rerun everything all over again for both OS, which is both costly and time consuming.

Breaking out into multiple workflow files should help solve this issue.

PR #354, opened 2022-02-02
Breakout unit testing per OS. Issue: #353

GuillaumeBchd:
# Description of the issue
The "build.yml" workflow does both linux and macos build and test. This could lead to the failure of both of them if one of them fails and then require to rerun everything all over again for both OS, which is both costly and time consuming.
Link to the issue: https://github.com/aws/amazon-cloudwatch-agent/issues/353

# Description of changes
Split the "build.yml" workflow into "build-test-linux.yml" and "build-test-macos.yml".

# License

# Tests
Pushed the same code in another fork and have run both of these workflow successfully .

### DIFF
--- a/.github/workflows/build-test-linux.yml
+++ b/.github/workflows/build-test-linux.yml
@@ -30,9 +30,6 @@ jobs:
         go-version: ~1.15.15
       id: go
 
-    - name: Configure git with longpath enabled (for windows)
-      run: git config --global core.longpaths true
-
     - name: Check out code
       uses: actions/checkout@v2
       with:
--- a/.github/workflows/build-test-linux.yml
+++ b/.github/workflows/build-test-linux.yml
@@ -19,8 +19,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  build-linux:
-    name: Build-linux
+  linux-unittest:
     runs-on: ubuntu-latest
     steps:  
 
--- a/.github/workflows/build-test-linux.yml
+++ b/.github/workflows/build-test-linux.yml
@@ -43,7 +43,6 @@ jobs:
         path: |
               ~/.cache/go-build
               ~/go/pkg/mod
-              /home/runner/.cache/go-build
         key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
 
     - name: Test
--- a/.github/workflows/build-test-linux.yml
+++ b/.github/workflows/build-test-linux.yml
@@ -1,9 +1,10 @@
-name: build-and-test-linux
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT
 
+name: build-and-test-linux
 on:
   push:
     branches: [master, main]
-
     paths-ignore:
       - '**.md'
       - '.github/**'
--- a/.github/workflows/build-test-linux.yml
+++ b/.github/workflows/build-test-linux.yml
@@ -28,7 +28,6 @@ jobs:
       uses: actions/setup-go@v2
       with:
         go-version: ~1.15.15
-      id: go
 
     - name: Check out code
       uses: actions/checkout@v2
--- a/.github/workflows/build-test-linux.yml
+++ b/.github/workflows/build-test-linux.yml
@@ -1,0 +1,68 @@
+name: build-and-test-linux
+
+on:
+  push:
+    branches:
+      - master
+
+    paths-ignore:
+      - '**.md'
+      - '.github/**'
+      - '!.github/workflows/build-*'
+      
+
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref_name }}
+  cancel-in-progress: true
+
+jobs:
+  build-linux:
+    name: Build-linux
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+    steps:  
+
+    - name: Set up Go 1.x
+      uses: actions/setup-go@v2
+      with:
+        go-version: ~1.15.15
+      id: go
+
+    - name: Configure git with longpath enabled (for windows)
+      run: git config --global core.longpaths true
+
+    - name: Check out code
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+        submodules: 'true'
+
+    - name: Debug go.mod
+      run: cat go.mod
+
+    - name: Cache build output
+      uses: actions/cache@v2
+      with:
+        path: |
+              ~/.cache/go-build
+              ~/go/pkg/mod
+              /home/runner/.cache/go-build
+        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+        restore-keys: |
+          ${{ runner.os }}-go-
+
+    - name: Test
+      run: make test
+
+    - name: Upload coverage to Codecov
+      uses: codecov/codecov-action@v2
+      with:
+        verbose: true
+
+    - name: Build
+      run: make build
--- a/.github/workflows/build-test-linux.yml
+++ b/.github/workflows/build-test-linux.yml
@@ -35,9 +35,6 @@ jobs:
         fetch-depth: 0
         submodules: 'true'
 
-    - name: Debug go.mod
-      run: cat go.mod
-
     - name: Cache build output
       uses: actions/cache@v2
       with:
--- a/.github/workflows/build-test-linux.yml
+++ b/.github/workflows/build-test-linux.yml
@@ -47,8 +47,6 @@ jobs:
               ~/go/pkg/mod
               /home/runner/.cache/go-build
         key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
-        restore-keys: |
-          ${{ runner.os }}-go-
 
     - name: Test
       run: make test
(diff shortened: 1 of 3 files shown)


.github/workflows/build-test-macos.yml
```diff
@@ -1,4 +1,4 @@
-name: build-and-test
+name: build-and-test-macos
```
SaxyPandaBear, line 1: We only need one workflow that uploads test coverage. Can you remove the upload step from this workflow?

GuillaumeBchd: This should be ok now

khanhntd:
Hi @GuillaumeBchd , thanks for opening up a PR for this. Still did not expect this though after opening my PR https://github.com/aws/amazon-cloudwatch-agent/pull/355. However, there are some changes but you can look at my changes and will answer all of the changes though

.github/workflows/build-test-macos.yml
```diff
@@ -0,0 +1,63 @@
...
+      fail-fast: false
+    steps:  
+
+    - name: Set up Go 1.x
+      uses: actions/setup-go@v2
+      with:
+        go-version: ~1.15.15
+      id: go
+
+    - name: Configure git with longpath enabled (for windows)
```
khanhntd, line 36: Please discard this since it is not used for either linux or macOS.

.github/workflows/build-test-macos.yml
```diff
@@ -0,0 +1,63 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref_name }}
+  cancel-in-progress: true
+
+jobs:
+  build-macos:
+    name: Build-macos
+    runs-on: macos-latest
+    strategy:
```
khanhntd, line 26: Please discard this since we are not using matrix for github.

.github/workflows/build-test-macos.yml
```diff
@@ -0,0 +1,63 @@
+name: build-and-test-macos
+
+on:
+  push:
+    branches:
```
khanhntd, line 5: Please changed to ```branches: [master, main]``` since we intend to rename the branches in the future.

.github/workflows/build-test-linux.yml
```diff
@@ -61,7 +48,10 @@ jobs:
     - name: Cache build output
       uses: actions/cache@v2
       with:
-        path: ${{ matrix.cachepath }}
+        path: |
+              ~/.cache/go-build
+              ~/go/pkg/mod
+              /home/runner/.cache/go-build
         key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
         restore-keys: |
```
khanhntd, line 56: We don't need the restore keys since it would cache the old dependencies package.

.github/workflows/build-test-macos.yml
```diff
@@ -0,0 +1,63 @@
...
+
+    - name: Cache build output
+      uses: actions/cache@v2
+      with:
+        path: |
+              ~/Library/Caches/go-build
+              ~/go/pkg/mod
+              /Users/runner/Library/Caches/go-build
+        key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+        restore-keys: |
```
khanhntd, line 56: We don't need the restore keys since it would cache the old dependencies package.

.github/workflows/build-test-macos.yml
```diff
@@ -0,0 +1,63 @@
...
+          ${{ runner.os }}-go-
+
+    - name: Test
+      run: make test
+
```
khanhntd, line 61: Please  add the 
```
      - name: Upload coverage to Codecov
        uses: codecov/codecov-action@v2
        with:
          verbose: true
```

.github/workflows/build-test-macos.yml
```diff
@@ -0,0 +1,63 @@
...
+    runs-on: macos-latest
+    strategy:
+      fail-fast: false
+    steps:  
+
+    - name: Set up Go 1.x
+      uses: actions/setup-go@v2
+      with:
+        go-version: ~1.15.15
+      id: go
```
khanhntd, line 34: We don't use ```id``` so better to discard it.

khanhntd:
Please also changes to ```make test``` after using ```choco install make``` and ```make build`` for the window pipeline https://github.com/aws/amazon-cloudwatch-agent/blob/master/.github/workflows/build-test-windows.yml#L45

.github/workflows/build-test-macos.yml
SaxyPandaBear: Why do we want to have all of the runners publish test coverage

khanhntd: Assuming we retrying a single workflow and due to facts that our test are flakiness, so sometimes we cannot cover all of our tests. Therefore, uploading a test coverage for single workflow is to address this problem.

.github/workflows/build-test-linux.yml
```diff
@@ -61,7 +48,10 @@ jobs:
     - name: Cache build output
       uses: actions/cache@v2
       with:
-        path: ${{ matrix.cachepath }}
+        path: |
+              ~/.cache/go-build
+              ~/go/pkg/mod
+              /home/runner/.cache/go-build
```
khanhntd, line 54: We can delete the home runner cache based on the tests I have run: 
* Test with a PR and found unexpected results without caching this path runner/Library/Caches/go-build:
https://github.com/aws/amazon-cloudwatch-agent/runs/4990690788?check_suite_focus=true
https://github.com/aws/amazon-cloudwatch-agent/runs/4990745483?check_suite_focus=true

* Compared when cached this path runner/Library/Caches/go-build:
https://github.com/aws/amazon-cloudwatch-agent/runs/4990443776?check_suite_focus=true

khanhntd: This can be the same when cache build packages for window and mac

.github/workflows/build-test-linux.yml
```diff
@@ -20,24 +20,11 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  build:
-    name: Build
-    runs-on: ${{ matrix.os }}
+  build-linux:
```
khanhntd, line 23: Please change this to ```linux-unittest``` for consistency with [build-and-test-window](https://github.com/aws/amazon-cloudwatch-agent/blob/master/.github/workflows/build-test-windows.yml#L22)

.github/workflows/build-test-linux.yml
```diff
@@ -20,24 +20,11 @@ concurrency:
...
+    name: Build-linux
+    runs-on: ubuntu-latest
     strategy:
```
khanhntd, line 26: Please discard this since we are not using matrix.

.github/workflows/build-test-linux.yml
```diff
@@ -0,0 +1,57 @@
...
+      with:
+        go-version: ~1.15.15
+
+    - name: Check out code
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+        submodules: 'true'
+
+    - name: Debug go.mod
```
khanhntd, line 37: We don't need this since we don't use ```restores key``` anymore.

khanhntd:
Please delete ```build.yaml``` too if we implement these changes.

.github/workflows/build-test-macos.yml
```diff
@@ -75,4 +54,4 @@ jobs:
         verbose: true
```
khanhntd, line 55: Please also add 
```
      - name: Upload coverage to Codecov
        uses: codecov/codecov-action@v2
        with:
          verbose: true
```

.github/workflows/build-test-linux.yml
```diff
@@ -0,0 +1,57 @@
+name: build-and-test-linux
```
khanhntd, line 1: Would be good to add license for these yaml file though
```
# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
# SPDX-License-Identifier: MIT
```

.github/workflows/build-test-macos.yml
jhnlsn: From what I can remember, we do need coverage uploads for each OS because some tests don't run on every runner

GuillaumeBchd:
> Please delete `build.yaml` too if we implement these changes.

It should already be deleted

khanhntd:
> > Please delete `build.yaml` too if we implement these changes.
> 
> It should already be deleted

You are right. My mistake though.

.github/workflows/build-test-macos.yml
SaxyPandaBear: Is there no conflict/overriding that happens when we have multiple coverages being published for the same PR? I'll have to look into that again.

khanhntd (approved):
LGTM.

.github/workflows/build-test-macos.yml
SaxyPandaBear: Uploading coverage from all of the OS workflows does not address test flakiness. That being said, it does make sense that we need to run it on all our OS because windows tests would only get run on the windows workflow, and Codecov does [automatic report merging](https://docs.codecov.com/docs/merging-reports), so there's no conflict there. I was under the impression that different code coverage reports would overwrite previous ones on the same commit.

.github/workflows/build-test-macos.yml
SaxyPandaBear: Read up on Codecov and there's no issue: https://docs.codecov.com/docs/merging-reports We should upload coverage from all of the workflows

SaxyPandaBear (approved):

PR #357, opened 2022-02-03
Switch the toToml unit test from using string to Toml struct comparison

ZhenyuTan-amz:
…ct comparison for more tolerance to toml format changes

# Description of the issue
Current unit test using string comparison is fragile, tests fails too easy, when the order of an array is different or even extra empty spaces and lines while the toml config is actually having same meaning.

# Description of changes
Using burnt sushi package to parse the toml into config struct and compare struct instead of hard code string values.

# Tests
Added the tests to test the accuracy of the comparison.


translator/totomlconfig/toTomlConfig_test.go
```diff
@@ -61,127 +55,150 @@ func TestLogMetricAndLog(t *testing.T) {
...
+	actualOutput := ToTomlConfig(input)
+	checkIfIdenticalToml(t, "./tomlConfigs/agentToml.conf", actualOutput)
+}
+
+func checkTomlTranslation(t *testing.T, jsonPath string, desiredTomlPath string, os string) {
+	agent.Global_Config = *new(agent.Agent)
+	translator.SetTargetPlatform(os)
+	var input interface{}
+	err := json.Unmarshal([]byte(ReadFromFile(jsonPath)), &input)
+	require.NoError(t, err)
```
SaxyPandaBear, line 196: Nitpick: we use both `assert.NoError()` and `require.NoError()`

translator/totomlconfig/tomlConfigs/tomlConfig.go
```diff
@@ -0,0 +1,338 @@
...
+		Cpu []cpuConfig
+		Disk []diskConfig
+		DiskIo []diskioConfig
+		Eththool []ethtoolConfig
+		K8sapiserver []k8sApiServerConfig
+		Logfile []logFileConfig
+		Mem []memConfig
+		Net []netConfig
+		NetStat []netStatConfig
+		NvidiaSmi []nvidiaSmi 						`toml:"nvidia_smi"`
```
SaxyPandaBear, line 42: Not sure if it's just my browser but the spacing looks off on some of these.

translator/totomlconfig/tomlConfigs/tomlConfig.go
```diff
@@ -0,0 +1,338 @@
+package tomlConfigs
+
+type (
+	TomlConfig struct {
```
SaxyPandaBear, line 4: What led to this struct? This feels weird to me because this is just a Go struct that mirrors the `schema.json`. Is there a way we can have one or the other, but not both? I understand that this is intended for the unit tests, but it strikes me as redundant.

ZhenyuTan-amz: Good question, it is required by the BurntSushi toml parser, it's the object representing the whole config object, like Java you can simple parse a config into a POJO, this is the Go equivalent of that POJO.

ZhenyuTan-amz: And we recently discover that schema.json is not a necessary thing to maintain, unfortunately this struct is, for the parser to work as intended.

SaxyPandaBear (approved):

translator/totomlconfig/toTomlConfig_test.go
```diff
@@ -202,3 +219,21 @@ func resetContext() {
 
 	os.Setenv("ProgramData", "c:\\ProgramData")
 }
+
```
aateeqi, line 222: Could we add a comment about how the two tomls are being evaluated into the go object? This may also be useful when someone needs to add or extend the toToml.go file.

aateeqi (approved):

PR #361, opened 2022-02-16
Increase Max Dimensions To 30

sethAmazon:
# Description of the issue
Want to have a greater max number of dimensions

# Description of changes
Increase the hardcoded value from 10 to 30.

# Tests
make test

Backend fails as expected for too many metrics. backend will need to support more metrics. 
Screen Shot 2022-02-16 at 3 51 44 PM


plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -530,11 +530,11 @@ func (c *CloudWatch) BuildMetricDatum(point telegraf.Metric) []*cloudwatch.Metri
 }
 
 // Make a list of Dimensions by using a Point's tags. CloudWatch supports up to
-// 10 dimensions per metric so we only keep up to the first 10 alphabetically.
+// 30 dimensions per metric so we only keep up to the first 30 alphabetically.
 // This always includes the "host" tag if it exists.
 func BuildDimensions(mTags map[string]string) []*cloudwatch.Dimension {
 
-	const MaxDimensions = 10
+	const MaxDimensions = 30
```
SaxyPandaBear, line 537: If we're already making changes, instead of just modifying this value can you pull this out to a constant here? That way, we can just refer to it in the test rather than redefine it.

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -555,6 +555,7 @@ func BuildDimensions(mTags map[string]string) []*cloudwatch.Dimension {
 
 	for _, k := range keys {
 		if len(dimensions) >= MaxDimensions {
+			log.Printf("D! max MaxDimensions %v is less than than number of dimensions %v thus only taking the max number", MaxDimensions, len(dimensions))
```
SaxyPandaBear, line 558: The wording in this log message seems off. If I have 100 dimensions, it would read:
```
D! max MaxDimensions 30 is greater than number of dimensions 100 thus only taking the max number
```

SaxyPandaBear: Ah you changed it while I was reviewing. Fair enough

plugins/outputs/cloudwatch/aggregator_test.go
```diff
@@ -109,12 +109,12 @@ func TestAggregator_MultipleAggregationPeriods(t *testing.T) {
 	aggregator.AddMetric(m)
 
 	assertNoMetricsInChan(t, metricChan)
-	assertMetricContent(t, metricChan, aggregationInterval * 2, m, expectedFieldContent{"value", 3, 1, 3, 6, "",
+	assertMetricContent(t, metricChan, aggregationInterval * 3, m, expectedFieldContent{"value", 3, 1, 3, 6, "",
```
SaxyPandaBear, line 112: I'm mildly suspicious of this because I saw in https://github.com/aws/amazon-cloudwatch-agent/pull/334/commits/da00fff5c83eb5aa660a654b2236dca7de1561eb that it did not fix the test. I've been rerunning the MacOS workflow on this PR when I remember to and it passes consistently so not sure why it failed more often than succeeded with a larger timeout from the PR that originally refactored this. Part of me thinks it could be more related to the additional sleep time on line 143 where there wasn't a sleep before. 

I would rather figure out the root issue than play around with sleep timers on unit tests, but don't want this to become a cause of significant confusion for people submitting PRs. 

For my own sanity, can you change the 1 second sleep you added back to 0 seconds and see if it still consistently passes in GitHub CI? I am curious if that test has cascading issues that affect these other tests. I haven't dug around the source code this is testing too much.

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -530,11 +531,9 @@ func (c *CloudWatch) BuildMetricDatum(point telegraf.Metric) []*cloudwatch.Metri
 }
 
 // Make a list of Dimensions by using a Point's tags. CloudWatch supports up to
-// 10 dimensions per metric so we only keep up to the first 10 alphabetically.
+// 30 dimensions per metric so we only keep up to the first 30 alphabetically.
```
adam-mateen, line 534: Is there a GitHub issue that this is resolving?
It would be nice to link to more details on why this change from 10 to 30 is being done.
Did CloudWatch backend announce they can support 30 dimensions per metric?
If so, any idea why they chose 30 instead of 40, 50, etc.?

sethAmazon:
@adam-mateen we can't link internal issues but I can create an issue on gh and copy.

sethAmazon:
As for the MaxDimensions being public I just copied from the function but yeah it should be package level.

adam-mateen (approved):
Approved.
We should wait for the CloudWatch backend to support 30 dimensions before releasing a new CWA with this change though.

SaxyPandaBear:
I don't think there's any need to approve this until the backend changes are ready since we have to wait for them anyway. I would just leave the PR open until then

SaxyPandaBear:
Untagging from v51 since this is unlikely to be supported on the backend by our next release

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -40,6 +40,7 @@ const (
 	highResolutionTagKey           = "aws:StorageResolution"
 	defaultRetryCount              = 5 // this is the retry count, the total attempts would be retry count + 1 at most.
 	backoffRetryBase               = 200
+	MaxDimensions                  = 30
```
sethAmazon, line 43: Need to be public for the integration test

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -40,6 +40,7 @@ const (
 	highResolutionTagKey           = "aws:StorageResolution"
 	defaultRetryCount              = 5 // this is the retry count, the total attempts would be retry count + 1 at most.
 	backoffRetryBase               = 200
+    MaxDimensions                  = 30
```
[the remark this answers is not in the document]
khanhntd: That happens to me sometimes too. To ensure it, `make fmt` is your choice though.

go.mod
```diff
@@ -2,9 +2,9 @@ module github.com/aws/amazon-cloudwatch-agent
 
 go 1.13
 
-replace github.com/influxdata/telegraf => github.com/aws/telegraf v0.10.2-0.20201015165757-4470de2d306b
+replace github.com/influxdata/telegraf => github.com/aws/telegraf v0.10.2-0.20220509174810-d4f1bd11c120
```
sethAmazon, line 5: Updating the aws sdk causes golang.org/x/net to update this we need to update telegraf.

SaxyPandaBear: Do we need to update everything in this PR? should we pin to a specific version of the AWS SDK until the telegraf upgrade is complete?

go.mod
```diff
@@ -41,7 +41,7 @@ replace golang.zx2c4.com/wireguard v0.0.20200121 => golang.zx2c4.com/wireguard v
 require (
 	github.com/BurntSushi/toml v0.3.1
 	github.com/Jeffail/gabs v1.4.0
-	github.com/aws/aws-sdk-go v1.30.15
+	github.com/aws/aws-sdk-go v1.44.9
```
sethAmazon, line 44: This was not required but Gary Taylor said to do it.

go.mod
khanhntd: Agree. Upgrade with `aws-sdk` is exceptional since it does not directly involve with our main packages and only involves in cleaning `integration-test` resources.

integration/test/metrics_number_dimension/metrics_number_dimension_test.go
```diff
@@ -56,17 +54,26 @@ func TestNumberMetricDimension(t *testing.T) {
...
 
 	for _, parameter := range parameters {
 		//before test run
 		log.Printf("resource file location %s find target %t input number dimension %d metric name %s",
 			parameter.resourcePath, parameter.findTarget, parameter.numberDimensionsInCW, parameter.metricName)
 
-		target := fmt.Sprintf(targetString, maxDimension, parameter.numberDimensionsInCW)
+		target := fmt.Sprintf(targetString, cwPlugin.MaxDimensions, parameter.numberDimensionsInCW)
 
 		t.Run(fmt.Sprintf("resource file location %s find target %t", parameter.resourcePath, parameter.findTarget), func(t *testing.T) {
```
khanhntd, line 78: Instead of using `resource file location %s find target %t`, would be better to initialize a `name` into struct so it would clarify what each test would be instead of find the corresponding json.

khanhntd (commented):
Can you help me clarify why we would need to create a list of metrics after we ensure that [no metrics are being found](https://github.com/aws/amazon-cloudwatch-agent/pull/361/files#diff-e4de5a78363c2467f52e8738fc0b8cf0ca044e3666faacaf34e52c63b16ba4ffR105-R115)

go.mod
SaxyPandaBear: If this causes a ripple effect of forcing an upgrade in telegraf and is not explicitly required for the change in max dimensions, I wouldn't do this upgrade now.

.github/workflows/integrationTest.yml
```diff
@@ -610,7 +610,7 @@ jobs:
             fi
       
       - name: Terraform destroy
-        if: ${{ cancelled() && steps.ecs-fargate-integration-test.outputs.cache-hit != 'true'
+        if: ${{ cancelled() && steps.ecs-fargate-integration-test.outputs.cache-hit != 'true' }}
```
SaxyPandaBear, line 613: I was going to put in a PR for this today after I saw that my fork failed because of it. Would rather get that fix in sooner than including it in this PR.

go.mod
```diff
@@ -2,9 +2,9 @@ module github.com/aws/amazon-cloudwatch-agent
 
 go 1.13
 
-replace github.com/influxdata/telegraf => github.com/aws/telegraf v0.10.2-0.20201015165757-4470de2d306b
+replace github.com/influxdata/telegraf => github.com/aws/telegraf v0.10.2-0.20220509174810-d4f1bd11c120
 
-replace github.com/shirou/gopsutil => github.com/aws/telegraf/patches/gopsutil v0.0.0-20201015165757-4470de2d306b // Temporary fix, pending PR https://github.com/shirou/gopsutil/pull/957
+replace github.com/shirou/gopsutil => github.com/aws/telegraf/patches/gopsutil v0.0.0-20220509174810-d4f1bd11c120 // Temporary fix, pending PR https://github.com/shirou/gopsutil/pull/957
```
SaxyPandaBear, line 7: I'm really hesitant about making these changes. Let's leave these versions be where we can, and only upgrade things that are necessary to support the max dimension change (which I don't think requires any version bump)

integration/test/metrics_number_dimension/metrics_number_dimension_test.go
```diff
@@ -56,17 +54,26 @@ func TestNumberMetricDimension(t *testing.T) {
...
+		{
+			resourcePath:         "resources/30_dimension",
+			findTarget:           false,
+			numberDimensionsInCW: 30,
+			metricName:           "mem_used_percent",
+		},
+		{
+			resourcePath:         "resources/35_dimension",
+			findTarget:           true,
+			numberDimensionsInCW: 30,
```
SaxyPandaBear, line 66: Should this be 35?

go.mod
khanhntd: @SaxyPandaBear  Right. Then what we could do is create a separate branch from `upgrade-telegraf` and ensure it works from there too to avoid any unnecessary regression. For now, we will hold on merging until this work on `upgrade-telegraf`.

integration/test/metrics_number_dimension/metrics_number_dimension_test.go
khanhntd: @SaxyPandaBear It should be 30 since CWAgent would drop metric's dimension if the dimensions > maxDimension based on my understanding.

.github/workflows/integrationTest.yml
khanhntd: Andrew has created a separated PR for it https://github.com/aws/amazon-cloudwatch-agent/pull/461.

integration/test/metrics_number_dimension/metrics_number_dimension_test.go
sethAmazon: Yeah that is how it works for statsd. For append dimension it should fail on the config check. My bad on that.

sethAmazon: I rewrote the test for this.

.github/workflows/integrationTest.yml
sethAmazon: It will auto resolve then if this fix is already in.

integration/test/metrics_number_dimension/metrics_number_dimension_test.go
```diff
@@ -52,36 +47,45 @@ func TestNumberMetricDimension(t *testing.T) {
...
-
-		t.Run(fmt.Sprintf("resource file location %s find target %t", parameter.resourcePath, parameter.findTarget), func(t *testing.T) {
+		t.Run(fmt.Sprintf("resource file location %s find target %t", parameter.resourcePath, parameter.failToStart), func(t *testing.T) {
 			test.CopyFile(parameter.resourcePath+configJSON, configOutputPath)
-			test.StartAgent(configOutputPath)
+			err := test.StartAgent(configOutputPath, false)
+
+			// for append dimension we auto fail over 30 for custom metrics (statsd we collect remove dimension and continue)
+			if parameter.failToStart && err == nil {
+				t.Fatalf("Agent should not have started for append %v dimension", parameter.resourcePath)
```
SaxyPandaBear, line 79: Should this be 
```suggestion
				t.Fatalf("Agent should not have started for append %v dimension", parameter.numberDimensionsInCW)
```
Maybe I just don't understand the test table params anymore

integration/test/metrics_number_dimension/metrics_number_dimension_test.go
```diff
@@ -52,36 +47,45 @@ func TestNumberMetricDimension(t *testing.T) {
...
 
-		target := fmt.Sprintf(targetString, maxDimension, parameter.numberDimensionsInCW)
-
-		t.Run(fmt.Sprintf("resource file location %s find target %t", parameter.resourcePath, parameter.findTarget), func(t *testing.T) {
+		t.Run(fmt.Sprintf("resource file location %s find target %t", parameter.resourcePath, parameter.failToStart), func(t *testing.T) {
 			test.CopyFile(parameter.resourcePath+configJSON, configOutputPath)
-			test.StartAgent(configOutputPath)
+			err := test.StartAgent(configOutputPath, false)
+
+			// for append dimension we auto fail over 30 for custom metrics (statsd we collect remove dimension and continue)
```
sethAmazon, line 77: Go output starts at the time of failure so the failure message gets chopped off. Thus have to use if there is an error and just assume reason.

integration/test/metrics_number_dimension/metrics_number_dimension_test.go
sethAmazon: I will change this after the integration tests are done running so they will not have commit hash issues on pull.

SaxyPandaBear (approved):
lgtm

Issue #360, opened 2022-02-10
Tests in aggregator_test are flaky


SaxyPandaBear:
This was attempted to be addressed in #334, and is tidied up, but the test is still flaky

SaxyPandaBear:
An example test failure in CI can be found [here](https://github.com/aws/amazon-cloudwatch-agent/actions/runs/1821455052/attempts/4). This should be investigated soon so there are less blips in GitHub actions when running PRs and merging to the main branch.

SaxyPandaBear:
Note that increasing the timeouts as part of #366  has alleviated this a little bit, but the tests are stilly flaky

SaxyPandaBear:
Closed by #619

PR #366, opened 2022-02-18
Increase Aggregator Test Timeout

sethAmazon:
# Description of the issue
Failing on macos github runner

# Description of changes
Increase timeout to find metrics by 1 second

# Tests
pr test


aateeqi (approved):

SaxyPandaBear (approved):
I still have my hesitancies with this, but it does seem to alleviate some pains specific to MacOS testing. I don't think this fully addresses the issue described in #360, but I think it can help short term.

PR #368, opened 2022-02-21
Upload Integration Test Binary To S3 Instead Of Building On EC2 And A…

sethAmazon:
# Description of the issue
Adding oracle and rhel ami to integration tests 

# Description of changes
Adds oracle and rhel ami

# Tests
Ran integration tests on my fork https://github.com/sethAmazon/amazon-cloudwatch-agent/actions/runs/1877875828


integration/test/ca_bundle/ca_bundle_test.go
```diff
@@ -42,8 +42,6 @@ func TestBundle(t *testing.T) {
 		{dataInput: "resources/integration/ssl/with/original/bundle", findTarget: true},
 		//Do not look for ca bundle should not connect thus target string found
 		{dataInput: "resources/integration/ssl/without/bundle", findTarget: true},
-		//Do not look for ca bundle should not connect thus target string found
-		{dataInput: "resources/integration/ssl/without/bundle", findTarget: true},
```
sethAmazon, line 46: This is a repeated tests not sure why I merged this.

.github/workflows/integrationTest.yml
```diff
@@ -28,31 +82,81 @@ jobs:
...
+          aws-access-key-id: ${{ secrets.TERRAFORM_AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.TERRAFORM_AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-west-2
+
+      - name: Cache if success
+        id: ec2-linux-integration-test
+        uses: actions/cache@v2
+        with:
+          path: |
+            RELEASE_NOTES
```
sethAmazon, line 142: Just need something for the path and the version is a generated file.

integration/terraform/ec2/linux/centOS8-setup.sh
```diff
@@ -0,0 +1,22 @@
+# You can use this script to install all required deps for integration testing to a centos8 based linux (rhel8 ol8)
```
jhnlsn, line 1: Where does this setup script get called from?

sethAmazon: No where. These are the commends I used to add the required deps to an ami before building the image.

sethAmazon: I take an ami lets say rhel8  -> create an ec2 instance -> add these deps -> build a new ami that I use in the test

.github/workflows/integrationTest.yml
```diff
@@ -15,7 +15,61 @@ concurrency:
...
+      - name: Cache binaries
+        id: cached_binaries
+        uses: actions/cache@v2
+        with:
+          key: "cached_binaries_${{ github.sha }}"
+          path: build
+
+      - name: Build Binaries
+        if: steps.cached_binaries.outputs.cache-hit != 'true'
+        run: make build package-rpm package-deb
```
SaxyPandaBear, line 64: Do we plan on adding Windows/MacOS integration tests to this?

.github/workflows/integrationTest.yml
```diff
@@ -61,11 +165,10 @@ jobs:
           -var="ami=${{ matrix.arrays.ami }}"
           -var="ca_cert_path=${{ matrix.arrays.caCertPath }}"
           -var="arc=${{ matrix.arrays.arc }}"
+          -var="binary_name=${{ matrix.arrays.binaryName }}"
 
       - name: Terraform destroy
-        if: always()
+        if: ${{ always() && steps.ec2-linux-integration-test.outputs.cache-hit != 'true' }}
```
SaxyPandaBear, line 171: Do we need the `always()` here if we are including the cache hit check?

integration/localstack/docker-compose.yml
```diff
@@ -3,7 +3,9 @@ version: "3.8"
 services:
   localstack:
     container_name: "${LOCALSTACK_DOCKER_NAME-localstack_main}"
-    image: localstack/localstack
+    # @TODO use latest when this is fixed https://github.com/localstack/localstack/issues/5502
+    # Use 0.12.20 since this is last version that worked for now
+    image: localstack/localstack:0.12.20
```
jhnlsn, line 8: With this test going from a handful of instances to a wide assortment of instances, is there opportunity for localstack to run in a central location instead of on every instance where the test is running?  Running in a mode of many integration tests to one localstack?

sethAmazon: It is a bit complicated setting this up I think. I can create a ticket for that and take care of it in a new pr.

.github/workflows/integrationTest.yml
sethAmazon: So far I have not created any windows or mac integration tests. That can be added as needed.

.github/workflows/integrationTest.yml
sethAmazon: Yes since if we do not include it and the test fails this will not run. Always forces a run even if a step before fails. The cache hit make sure this does not run on cache hit.

SaxyPandaBear (approved):
LGTM

PR #372, opened 2022-02-22
Address flaky tests in aggregator

### LINKED ISSUE #360: Tests in aggregator_test are flaky

SaxyPandaBear:
# Description of the issue
Closes #360 

# Description of changes
Noticed that some of the tests just used `close(shutdownChan)` at the end of the test without waiting for the WaitGroup to complete as part of the aggregator shutdown. Not 100% positive that this fixes the flakiness but it's a pretty big smoking gun. 

# License

# Tests
Reverted sleep increases from previous PR and ran locally, going to verify on GitHub CI as well.


plugins/outputs/cloudwatch/aggregator_test.go
```diff
@@ -21,118 +21,110 @@ var metricName = "metric1"
...
-	aggregator.AddMetric(m)
+	agg.AddMetric(m)
 
 	fields = map[string]interface{}{"value": 3}
 	m, _ = metric.New(metricName, tags, fields, timestamp)
-	aggregator.AddMetric(m)
+	agg.AddMetric(m)
 
-	tags = map[string]string{"d1key": "d1value", "d2key": "d2value", aggregationIntervalTagKey: (2 * aggregationInterval).String()}
+	tags = map[string]string{"d1key": "d1value", "d2key": "d2value", aggregationIntervalTagKey: (3 * aggregationInterval).String()}
```
SaxyPandaBear, line 97: My theory is that the flakiness stems from the aggregation interval. Originally, this was set to 2x the aggregation interval, and the test would wait 2x aggregation interval for a metric. Because of the timing, this sometimes missed the mark. A PR later attempted to update the aggregation interval itself, which did not alleviate the flakiness. 

Then a later PR than that increased the wait time for the tests, which alleviated the flakiness. Based on this, I believe that so long as the wait time > aggregation interval, this will significantly increase the effectiveness of the unit test.

SaxyPandaBear: https://github.com/aws/amazon-cloudwatch-agent/runs/5294867347?check_suite_focus=true disproves my theory. Will keep looking

SaxyPandaBear:
Revisit this later

PR #373, opened 2022-02-22
Add Verify Linux Ctl Integration Test

sethAmazon:
# Description of the issue
Sanity test agent

# Description of changes
Add integration test sanity

# Tests
Ran code on ec2 instance

### DIFF
--- a/integration/test/sanity/resources/verifyLinuxCtlScript.sh
+++ b/integration/test/sanity/resources/verifyLinuxCtlScript.sh
@@ -1,0 +1,128 @@
+#!/bin/sh
+
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT
+
+assertStatus() {
+  keyToCheck="${1:-}"
+  expectedVal="${2:-}"
+
+  grepKey='unknown'
+  case "${keyToCheck}" in
+    cwa_running_status)
+        grepKey="\"status\""
+        ;;
+    cwa_config_status)
+        grepKey="\"configstatus\""
+        ;;
+    cwoc_running_status)
+        grepKey="\"cwoc_status\""
+        ;;
+    cwoc_config_status)
+        grepKey="\"cwoc_configstatus\""
+        ;;
+    *)  echo "Invalid Key To Check: ${keyToCheck}" >&2
+        exit 1
+        ;;
+  esac
+
+  result=$(/usr/bin/amazon-cloudwatch-agent-ctl -a status | grep "${grepKey}" | awk -F: '{print $2}' | sed 's/ "//; s/",//')
+
+  if [ "${result}" = "${expectedVal}" ]; then
+      echo "In step ${step}, ${keyToCheck} is expected"
+  else
+      echo "In step ${step}, ${keyToCheck} is NOT expected. (actual="${result}"; expected="${expectedVal}")"
+      exit 1
+  fi
+}
+
+# init
+step=0
+/usr/bin/amazon-cloudwatch-agent-ctl -a remove-config -c all -o all
+/usr/bin/amazon-cloudwatch-agent-ctl -a stop
+
+step=1
+/usr/bin/amazon-cloudwatch-agent-ctl -a status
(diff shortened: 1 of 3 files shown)


integration/test/sanity/sanity_test.go
```diff
@@ -0,0 +1,16 @@
...
+// +build integration
+
+package sanity
+
+import (
+	"github.com/aws/amazon-cloudwatch-agent/integration/test"
+	"testing"
+)
+
+func TestAgentStatus(t *testing.T) {
```
khanhntd, line 14: Where it would trigger this file though? Just curiosity

integration/test/sanity/resources/verifyLinuxCtlScript.sh
```diff
@@ -0,0 +1,125 @@
+#!/bin/sh
```
khanhntd, line 1: Maybe add a license in here would be great though.

integration/test/sanity/sanity_test.go
```diff
@@ -0,0 +1,16 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
khanhntd, line 1: Nice on moving integration test out to Github.

SaxyPandaBear:
Just trying to understand the flow for the integration tests.

We use terraform to spin up the EC2 instance. Then we run `make integration-test` on it to have it execute our integration tests from within the host? One thing I think makes me a little concerned about this test suite in particular is that it does some destructive things on the host by removing configs, stopping and restarting the agent, etc. Would that potentially put the test host in a bad state wherein subsequent integration tests wouldn't work?

integration/test/sanity/sanity_test.go
sethAmazon: https://github.com/aws/amazon-cloudwatch-agent/blob/master/integration/terraform/ec2/linux/main.tf#L31

sethAmazon:
It is up to us to make sure that tests keep state clean. This can be an issue in any solution to integration tests. The good part of this system is that it brings up a new ec2 and destroys at the end so we don't have an ec2 instance that can be in a bad state from 1 ci run to another.

sethAmazon:
My recommendation is to have the test running put host is required state before running.

SaxyPandaBear:
Maybe I'm misunderstanding. When we run `make integration-test`, does that not run all of the integration tests on the same host sequentially? I wasn't referring to bringing up new instances between CI executions. I meant putting the host in a bad state during test execution for the next test in the same CI action.

sethAmazon:
They run 1 at time.

aateeqi (approved):

integration/test/sanity/resources/verifyLinuxCtlScript.sh
```diff
@@ -0,0 +1,128 @@
...
+
+step=11
+/usr/bin/amazon-cloudwatch-agent-ctl -a fetch-config -o default -s
+assertStatus "cwa_running_status" "running"
+assertStatus "cwoc_running_status" "running"
+assertStatus "cwa_config_status" "not configured"
+assertStatus "cwoc_config_status" "configured"
+
+step=12
+/usr/bin/amazon-cloudwatch-agent-ctl -a stop
```
SaxyPandaBear, line 124: What I'm trying to answer is this: 

**Are integration tests run on the same host in the same environment, i.e.: not isolated tests in separate containers?**

What this script will do (if it succeeds) is stop the agent in step 12. If this happens, will this cause other integration tests to fail because they assume CWA is running?

sethAmazon: No because the other integration tests starts the agent. Though it does seem that the other test runs first

SaxyPandaBear: Okay cool. Just wanted to make sure.

As far as your last note, I don't think that the test order is something we can rely on, which is why I wanted to make sure that running this test wouldn't affect the other integration tests that run.

SaxyPandaBear (approved):
LGTM

PR #376, opened 2022-02-24
Clean Integration Test AMI That Are Older Than 60 Days

sethAmazon:
# Description of the issue
AMI are being created by our pipelines but not cleaned. 

# Description of changes
clean out ami older than 60 days 

# Tests
Tested on my fork


integration/clean/clean_ami.go
```diff
@@ -0,0 +1,85 @@
...
+
+func main() {
+	err := cleanAMI()
+	if err != nil {
+		log.Fatalf("errors cleaning %v", err)
+	}
+}
+
+func cleanAMI() []error {
+	log.Printf("Begin to clean EC2 Instances")
```
SaxyPandaBear, line 27: Don't need to use Printf. Can just use Print.

integration/clean/clean_ami.go
```diff
@@ -0,0 +1,85 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+// +build clean
+
+package main
+
+import (
+	"context"
+	"github.com/aws/aws-sdk-go-v2/aws"
```
SaxyPandaBear, line 10: Any particular reason we're taking a new dependency on v2? At least, it looks new based on the go mod changes

integration/clean/clean_ami.go
```diff
@@ -0,0 +1,85 @@
...
+
+	describeImagesInput := ec2.DescribeImagesInput{Filters: []types.Filter{nameFilter}}
+	describeImagesOutput, err := ec2client.DescribeImages(cxt, &describeImagesInput)
+	if err != nil {
+		return []error{err}
+	}
+
+	var errors []error
+	for _, image := range describeImagesOutput.Images {
+		creationDate, err := smithyTime.ParseDateTime(*image.CreationDate)
```
SaxyPandaBear, line 64: Why are we using an external library for parsing the datetime here? What's special about the format?

integration/clean/clean_ami.go
```diff
@@ -0,0 +1,85 @@
...
+		log.Printf("image name %v image id %v experation date %v creation date parsed %v image creation date raw %v",
```
SaxyPandaBear, line 65: How would this log work if the date parsing failed?

integration/clean/clean_ami.go
```diff
@@ -0,0 +1,85 @@
...
+		"cloudwatch-agent-integration-test-rhel8-4*",
+		"cloudwatch-agent-integration-test-ol8-1*",
+		"cloudwatch-agent-integration-test-ol8-2*",
+		"cloudwatch-agent-integration-test-ol8-3*",
+		"cloudwatch-agent-integration-test-ol8-4*",
+	}}
+
+	//get instances to delete
+	daysToKeep := 60
+	keepDuration := -1 * time.Hour * 24 * time.Duration(daysToKeep)
```
SaxyPandaBear, line 53: Can make these consts/vars at the top level

integration/clean/clean_ami.go
sethAmazon: Just to make it future proof. I spoke with @jhnlsn and he said we do not have plans to upgrade to v2 int he near future. The aws sdk still gets security updates.

integration/clean/clean_ami.go
sethAmazon: This is because of the format of the *image.CreationDate. smithyTime is what I called smithyTime "github.com/aws/smithy-go/time". Which contains the conversion from aws string format to time struct. I think this looks cleaner than having to hardcode the format and this module is owned by aws so we should be good here.

SaxyPandaBear: I just saw that this isn't necessarily external, but still confused about this. What requires the smithy-time library that isn't solved in the time stdlib?

integration/clean/clean_ami.go
sethAmazon: My bad err check should go before this line.

integration/clean/clean_ami.go
sethAmazon: Yes

integration/clean/clean_ami.go
SaxyPandaBear: Noted. Yeah that's my bad for not seeing `/aws/` in the path of the import. I'm fine with this.

integration/clean/clean_ami.go
sethAmazon: Yes

SaxyPandaBear (approved):

jhnlsn (changes requested):
Can you include a README in the clean folder that describes what is going on.  Can it include a summary of how AMIs are created

sethAmazon:
I added to the readme. Please review again.

integration/terraform/ec2/linux/README.md
```diff
@@ -10,3 +10,15 @@ Instance assumptions
...
+
+**How are ami built?**
+
+AMI builder pipeline builds the ami
+
+The pipeline installs required packages and updates ami software
+
+This process generates a new ami we can then use for testing
+
+Required packages https://github.com/aws/amazon-cloudwatch-agent/blob/master/integration/terraform/ec2/linux/README.md
```
SaxyPandaBear, line 23: Kind of confusing to have this link to itself. IIRC you can use link anchors in Markdown, but don't recall how.

integration/clean/README.md
```diff
@@ -0,0 +1,3 @@
+**What does the cleaner do?**
+
+Cleaner cleans out old ami (ami older than 60 days)
```
SaxyPandaBear, line 3: Not sure what @jhnlsn was looking for, but for me, I would like to see more of an explanation here as to what the cleaner actually does (I know that it cleans out old AMIs), and why we need to have this.

integration/terraform/ec2/linux/README.md
```diff
@@ -10,3 +10,15 @@ Instance assumptions
 6. make
 7. aws-cli
 8. CloudWatchAgentServerRole is attached
+9. crontab
```
SaxyPandaBear, line 13: What do we use crontab for? Just curious.

jhnlsn (approved):

integration/clean/clean_ami.go
```diff
@@ -0,0 +1,86 @@
...
+		"cloudwatch-agent-integration-test-al2*",
+		"cloudwatch-agent-integration-test-rhel8-base*",
+		"cloudwatch-agent-integration-test-rhel8-1*",
+		"cloudwatch-agent-integration-test-rhel8-2*",
+		"cloudwatch-agent-integration-test-rhel8-3*",
+		"cloudwatch-agent-integration-test-rhel8-4*",
+		"cloudwatch-agent-integration-test-ol8-1*",
+		"cloudwatch-agent-integration-test-ol8-2*",
+		"cloudwatch-agent-integration-test-ol8-3*",
+		"cloudwatch-agent-integration-test-ol8-4*",
```
aateeqi, line 52: Since all the ami's share the `cloudwatch-agent-integration-test`, would it make since we save this into a const string and use it to make it less wordy and easier to extend the list?

aateeqi:
LGTM :shipit:

aateeqi (approved):

PR #379, opened 2022-02-25
Fixing windows event messages truncated on Windows server 2022

ZhenyuTan-amz:
# Description of the issue
 windows event messages truncated in half on Windows server 2022

# Description of changes
The returned value of buffer used from syscall to wevtapi.dll changed from bytes used to characters count, which is half of what is used to return. Which misleading the UTF16ToUTF8Bytes slice the buffer in half.

# License

# Tests
Add unit test with new return length.


plugins/inputs/windows_event_log/wineventlog/wineventlog.go
```diff
@@ -327,6 +327,7 @@ func (w *windowsEventLog) getRecord(evtHandle EvtHandle) (*windowsEventLogRecord
 	if err != nil {
 		return nil, fmt.Errorf("UTF16ToUTF8Bytes() err %v", err)
 	}
+
```
sethAmazon, line 330: Can the formatting changes come in a diff pr?

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -424,7 +424,6 @@ func (c *CloudWatch) BuildMetricDatum(point telegraf.Metric) []*cloudwatch.Metri
 	//https://www.ardanlabs.com/blog/2013/08/understanding-slices-in-go-programming.html
 	var datums []*cloudwatch.MetricDatum
 	for k, v := range point.Fields() {
-		log.Printf("D! #Fields of the current point: k - %s v - %f", k, v)
```
sethAmazon, line 427: Why was the log removed?

translator/totomlconfig/sampleConfig/complete_linux_config.conf
```diff
@@ -153,6 +153,8 @@
     role_arn = "metrics_role_arn_value_test"
     rollup_dimensions = [["ImageId"], ["InstanceId", "InstanceType"], ["d1"], []]
     tagexclude = ["host", "metricPath"]
+    [outputs.cloudwatch.drop_original_metrics]
```
sethAmazon, line 156: Why is linux being changed?

sethAmazon:
Is there a way we can just test this function against windows 2022 and windows 2019? I know we can run test based on os and arc but not sure about os version.

translator/totomlconfig/sampleConfig/complete_linux_config.conf
SaxyPandaBear: Changes propagated from https://github.com/aws/amazon-cloudwatch-agent/pull/357

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
SaxyPandaBear: It's only a handful of formatting changes and they're all purely cosmetic. I think they're fine to keep in here.

SaxyPandaBear:
Can you add `windows-latest` as a target OS for the Windows GitHub actions? As is, the CI workflows don't really confirm anything for this change because I set it to strictly use Windows 2019.

plugins/outputs/cloudwatch/cloudwatch.go
ZhenyuTan-amz: I added this log in previous PR for debugging purpose(more insights about what fields each data point contains), but end up finding it will become really noisy when rolling up dimensions in debug mode. Because the log happen every time the rolling happens.

translator/totomlconfig/sampleConfig/complete_linux_config.conf
ZhenyuTan-amz: Removed.

.github/workflows/build-test-windows.yml
```diff
@@ -51,3 +51,36 @@ jobs:
 
       - name: Run build
         run: make build
+
+  windows-lateest-unittest:
+    runs-on: windows-2019
```
SaxyPandaBear, line 56: This should point to windows-latest

.github/workflows/build-test-windows.yml
```diff
@@ -51,3 +51,36 @@ jobs:
 
       - name: Run build
         run: make build
+
+  windows-lateest-unittest:
```
SaxyPandaBear, line 55: Any reason these are two separate jobs instead of reusing the same job but defining a matrix for the runner OS?

SaxyPandaBear: Also there's a typo in the job name

.github/workflows/build-test-windows.yml
```diff
@@ -19,7 +19,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  windows-unittest:
+  windows-2019-unittest:
```
SaxyPandaBear, line 22: Not sure where I stand on it yet, but I at least wanted to bring up the discussion on how we want to handle this going forward. Pre-2022 Windows Server, this bug exists and we have no guarantee that they'll patch it in existing OS versions. So long as we support them, we must ensure that we don't have a regression, but I think ideally we would want to phase this out to just handle the fixed implementation that Windows Server 2022 has now.

plugins/inputs/windows_event_log/wineventlog/utils.go
```diff
@@ -78,10 +79,21 @@ func CreateQuery(path string, levels []string) (*uint16, error) {
...
+	if IsTheEndOfContent(in, length) {
+		i = int(length)
+		if i%2 != 0 {
+			i--
+		}
+	} else {
+		log.Printf("W! Buffer used: %d is returning as double byte character count, doubling it for decoding", length)
+		i = int(length) * 2
+	}
+	if i > cap(in) {
```
SaxyPandaBear, line 95: I think technically we can merge this with the above else block. Haven't dealt with math functions in Go, but would be something like
```go
else {
  i = math.Min(cap(in), int(length)*2)
}
```

plugins/inputs/windows_event_log/wineventlog/utils.go
```diff
@@ -99,6 +111,28 @@ func UTF16ToUTF8Bytes(in []byte, length uint32) ([]byte, error) {
...
 
+func IsTheEndOfContent(in []byte, length uint32) bool {
+	// scan next 100 characters, if any of them is none '0', return false
+	i := int(length)
+
+	if i%2 != 0 {
+		i -= 1
+	}
+	max := len(in)
+	if i+100 < max {
```
SaxyPandaBear, line 122: Is there any specific reason we try to scan 100 ahead? Or is this just arbitrary?

plugins/inputs/windows_event_log/wineventlog/utils.go
```diff
@@ -99,6 +111,28 @@ func UTF16ToUTF8Bytes(in []byte, length uint32) ([]byte, error) {
 	return decoded, err
 }
 
+func IsTheEndOfContent(in []byte, length uint32) bool {
```
SaxyPandaBear, line 114: I know that this function is implicitly tested via the outer function and the new unit test. 

That being said, is there a specific reason this function is made public (IIRC capitalizing it makes it accessible through the package)? If we're exposing this I think it warrants adding unit tests for this specific function. Otherwise, I think we should make it private scoped

plugins/inputs/windows_event_log/wineventlog/utils.go
```diff
@@ -99,6 +111,28 @@ func UTF16ToUTF8Bytes(in []byte, length uint32) ([]byte, error) {
...
+	if i%2 != 0 {
+		i -= 1
+	}
+	max := len(in)
+	if i+100 < max {
+		max = i+100
+	}
+
+	for ; i < max - 2; i += 2 {
+		v1 := uint16(in[i+2]) | uint16(in[i+1])<<8
```
SaxyPandaBear, line 127: I will have to reread this again to understand it, but would be nice to have a comment that explains why we do the bitwise OR here

plugins/inputs/windows_event_log/wineventlog/utils.go
ZhenyuTan-amz: Golang min only works with float64, which is annoying to convert back and forth.

plugins/inputs/windows_event_log/wineventlog/utils.go
ZhenyuTan-amz: It is combining two bytes(8 bits) into uint16.

plugins/inputs/windows_event_log/wineventlog/utils.go
ZhenyuTan-amz: Good call, leaving it private make more sense.

plugins/inputs/windows_event_log/wineventlog/utils.go
ZhenyuTan-amz: Kind of arbitrary to avoid false negative caused by empty spaces, lines, bad luck with continuous '0's

plugins/inputs/windows_event_log/wineventlog/utils.go
SaxyPandaBear: Okay that's fine by me. No point making it more confusing.

.github/workflows/build-test-windows.yml
```diff
@@ -20,7 +20,10 @@ concurrency:
 
 jobs:
   windows-unittest:
-    runs-on: windows-2019
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ windows-2019, windows-latest]
```
adam-mateen, line 26: 👍

plugins/inputs/windows_event_log/wineventlog/utils.go
```diff
@@ -78,10 +79,21 @@ func CreateQuery(path string, levels []string) (*uint16, error) {
...
+	// which is half of the actual buffer used by byte(what older Windows OS returns), checking if the length
+	//land on the end of used buffer, if no, double it.
+	var i int
+	if isTheEndOfContent(in, length) {
+		i = int(length)
+		if i%2 != 0 {
+			i--
+		}
+	} else {
+		log.Printf("W! Buffer used: %d is returning as double byte character count, doubling it for decoding", length)
```
adam-mateen, line 92: Should this really be warning?
If every single Windows Event on WinServer2022 will have this issue then it will be very noisy.

SaxyPandaBear: I would agree. I'm not familiar with this code but I would be concerned about the frequency we emit this warn level log. I think that this should be removed entirely (after the tests/issue is confirmed to be fixed with these changes), or at the bare minimum moved to debug-level logging before merging this in.

.github/workflows/build-test-windows.yml
```diff
@@ -20,7 +20,10 @@ concurrency:
 
 jobs:
   windows-unittest:
-    runs-on: windows-2019
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ windows-2019 ]
```
SaxyPandaBear, line 26: If we're on the same page, this is going to include `windows-latest` in a separate PR that addresses a different Windows 2022, bug, right?

plugins/inputs/windows_event_log/wineventlog/utils.go
ZhenyuTan-amz: Yeah, thinking about that, to avoid the noise, maybe a debug is more suitable.

.github/workflows/build-test-windows.yml
ZhenyuTan-amz: Correct.

SaxyPandaBear (dismissed):
I see the windows-latest test fails, but as I understand, that bug is going to be addressed separately. The windows-2019 job was cancelled (likely because of the windows-latest failure), but the unit tests did succeed, so looks good to me

SaxyPandaBear (approved):

plugins/inputs/windows_event_log/wineventlog/utils.go
adam-mateen: You can skip the conversion and bit shift and just do:
```
for ; i < max - 2; i += 2 {
    if in[i+1] != 0 || in[1+2] != 0 {
        return false
    }
```
Or just
```
for ; i < max; i++ {
    if in[1] != 0 {
        return false
    }
}
```

PR #384, opened 2022-02-28
Have LocalStack On 1 EC2 Instance

sethAmazon:
# Description of the issue
Local stack runs on all hosts 

# Description of changes
Local stack will now only run on 1 host

# Tests
Ran action on my fork


.github/workflows/integrationTest.yml
```diff
@@ -67,9 +67,45 @@ jobs:
...
+        id: localstack
+        run: >
+          echo run terraform and execute test code &&
+          terraform apply --auto-approve
+          -var="ssh_key=${PRIVATE_KEY}"
+          -var="github_repo=${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}.git"
+          -var="github_sha=${GITHUB_SHA}" &&
+          LOCAL_STACK_HOST_NAME=$(terraform output -raw public_dns) &&
+          echo $LOCAL_STACK_HOST_NAME &&
+          echo "::set-output name=local_stack_host_name::$LOCAL_STACK_HOST_NAME" &&
```
SaxyPandaBear, line 104: Might just not understand how the output part of GitHub actions works. Does doing the `::set-output` tell the GitHub runner to emit this as an output? It looks weird to me cause it's in an `echo` call, so just confirming

integration/localstack/docker-compose.yml
```diff
@@ -11,7 +11,7 @@ services:
       - "127.0.0.1:53:53"
       - "127.0.0.1:53:53/udp"
       - "127.0.0.1:443:443"
-      - "127.0.0.1:4566:4566"
+      - "0.0.0.0:4566:4566"
```
SaxyPandaBear, line 14: Not specific to this change, last I saw when using LocalStack, I thought that they were consolidating to a single port for their services endpoint. Do we need `4566` AND `4571`, or is 4571 for another service?

integration/localstack/ls_tmp/snakeoil.conf
```diff
@@ -14,4 +14,5 @@ CN = localhost
 subjectAltName = @alt_names
 [alt_names]
 DNS.1 = localhost
-DNS.2 = localhost.localstack.cloud
\ No newline at end of file
+DNS.2 = localhost.localstack.cloud
+DNS.3 = *.us-west-2.compute.amazonaws.com
```
SaxyPandaBear, line 18: Do we need to specify `us-west-2` for the DNS? Part of me thinks we could/should be able to point this to any other commercial region if necessary.

integration/terraform/ec2/localstack/README.md
```diff
@@ -0,0 +1,19 @@
...
+3. docker-compose
+4. git
+5. openssl
+6. aws-cli
+7. CloudWatchAgentServerRole is attached
+8. crontab
+
+Tag the ec2 instance with LocalStackIntegrationTestInstance
+
+There is no guarantee of destroy if the pipeline state is cancelled. Since create happens before tests job and cancel happens after test job.
```
SaxyPandaBear, line 17: Can you clarify on this? Are you saying that if the pipeline is cancelled, we will have orphaned EC2 instances in the account that won't be cleaned up? If so, I think we should add to the cleaner (not part of this PR) to also clean up hosts used by these integration tests that are stale.

.github/workflows/integrationTest.yml
sethAmazon: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions yes look at the docs

integration/localstack/docker-compose.yml
sethAmazon: I originally copied this from the docker-compose sample in the localstack gh https://github.com/localstack/localstack. I'm not sure what 4571 is used for.

integration/terraform/ec2/localstack/README.md
sethAmazon: I re read the gh docs and it seems this my understanding was incorrect. Will delete this.

integration/localstack/docker-compose.yml
SaxyPandaBear: Finally found the old version of the [README](https://github.com/localstack/localstack/blob/2df3eca88ed0215f8d99ed05fd2d7ddb35636d34/README.md).

> Note: Starting with version 0.11.0, all APIs are exposed via a single edge service, which is accessible on http://localhost:4566/ by default

I was using LocalStack for a project back in 2020, shortly after they started migrating stuff to their edge port endpoint, so I recalled seeing 4571 (I think this is where their /health endpoint was hosted) when I was working with it, but moved everything to 4566 later.

integration/localstack/ls_tmp/snakeoil.conf
sethAmazon: It seems you can’t wildcard sub domains for ssl
So *.domain.com works but
*.*.domain will not work
https://cheapsslsecurity.com/blog/how-to-secure-multi-level-subdomains/#:~:text=What%20is%20a%20Wildcard%20Certificate,any%20website%20with%20multiple%20subdomains.

integration/localstack/docker-compose.yml
sethAmazon: I can remove

integration/terraform/ec2/localstack/README.md
```diff
@@ -0,0 +1,17 @@
...
+3. docker-compose
+4. git
+5. openssl
+6. aws-cli
+7. CloudWatchAgentServerRole is attached
+8. crontab
+
+Tag the ec2 instance with LocalStackIntegrationTestInstance
+
+Output will be the public dns to be used for https connection and the proper ssl pem files (uploaded to s3)
```
adam-mateen, line 17: How public is public?
Any security concern over _something else_ using our LocalStack?

integration/terraform/ec2/localstack/variables.tf
```diff
@@ -0,0 +1,44 @@
...
+}
+
+variable "iam_instance_profile" {
+  type = string
+  default = "CloudWatchAgentServerRole"
+}
+
+variable "vpc_security_group_ids" {
+  type = list(string)
+  default = ["sg-013585129c1f92bf0"]
```
adam-mateen, line 18: Is there anything special or interesting about this security group?
allows incoming connections on port 443? 80?

integration/terraform/ec2/localstack/README.md
sethAmazon: These are short lived instances. I do not think this will be a problem. @jhnlsn thoughts.

integration/terraform/ec2/localstack/variables.tf
sethAmazon: This is the security group that I have been using for testing. Mandatory open is 22 (ssh -ip must be open to all address as the connect comes from runner-), 4566 (secure and unsecure -can come within the same security group-) on in. Out, not sure since the default is allow all (I assume 80, 443, and 4556 out are the only required).

adam-mateen (approved):

Issue #330, opened 2021-12-15
ECS Service Discovery: Default Network Mode


griffinmyers:
When evaluating tasks for ECS Service Discovery, this repo uses the network mode defined on a task's task definition to resolve which [IP](https://github.com/aws/amazon-cloudwatch-agent/blob/1df147021064e850a90101b124a4377940112b4e/internal/ecsservicediscovery/decoratedtask.go#L70-L92) or [port](https://github.com/aws/amazon-cloudwatch-agent/blob/1df147021064e850a90101b124a4377940112b4e/internal/ecsservicediscovery/decoratedtask.go#L94-L123) should be used as the scrape target. 

Unfortunately, these bits of logic aren't sensitive to the default network mode that kicks in when not specified in a task definition: 

> **`networkMode`**
> 
> Type: string
> 
> Required: no
> 
> The Docker networking mode to use for the containers in the task. For Amazon ECS tasks hosted on Amazon EC2 Linux instances, the valid values are none, bridge, awsvpc, and host. **If no network mode is specified, the default network mode is bridge**. For Amazon ECS tasks hosted on Amazon EC2 Windows instances, the valid values are default, and awsvpc. **If no network mode is specified, the default network mode is used.**
> 
> https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html

Are folks open to updating this behavior to handle the default cases, and if so are their opinions about exactly how this logic should be handled? I'm happy to draft up a PR, but wanted to get some sense of directional alignment here first. 

Thanks!

javabrett:
Thanks @griffinmyers .

Otherwise stated, `TaskDefinition.NetworkMode == nil` is valid and should select the platform-default network mode, but at:

https://github.com/aws/amazon-cloudwatch-agent/blob/c3894b985db74b51a8dc07f087e1a2da698dff2b/internal/ecsservicediscovery/decoratedtask.go#L68-L73

... fails to resolve the private IP address and the task if silently dropped and filtered-out.  On Linux, the logic should be per `bridge` and operate as if that mode was explicitly-defined.

griffinmyers:
@javabrett exactly. I think we'll additionally want to apply this logic when determining which port to scrape:

https://github.com/aws/amazon-cloudwatch-agent/blob/c3894b985db74b51a8dc07f087e1a2da698dff2b/internal/ecsservicediscovery/decoratedtask.go#L97-L100

adam-mateen:
[...]

PR #385, opened 2022-02-28
ECS Service Discovery: Fixup implicit network mode

griffinmyers:
# Description of the issue

I originally submitted https://github.com/aws/amazon-cloudwatch-agent/issues/330 with the goal of correcting the service discovery behavior of a task when its definition did not specify a network mode.

Very graciously, @javabrett submitted a fix for this via https://github.com/aws/amazon-cloudwatch-agent/pull/335. Upon testing this out today, I found out we're not _quite_ there.

Crucially, when network mode is undefined in a task definition, it will parse as `nil`, not the empty string. Correcting this in our unit tests reveals the behavior isn't working as desired, since the `getPrivateIp` function isn't properly handling the `nil` case. I believe this addresses the issue.

# Description of changes

1. Updated unit tests to more accurately reflect the value of a task's network mode when unset in ECS and unmarshaled here (`nil`). This yields a broken test. 
2. Fixed the test by handling the `nil` case more appropriately in `decoratedtask/getPrivateIP()`. 

# License

# Tests

I've corrected a unit test.

### DIFF
--- a/internal/ecsservicediscovery/decoratedtask.go
+++ b/internal/ecsservicediscovery/decoratedtask.go
@@ -72,12 +72,13 @@ func addExporterLabels(labels map[string]string, labelKey string, labelValue *st
 // Get the private ip of the decorated task.
 // Return "" when fail to get the private ip
 func (t *DecoratedTask) getPrivateIp() string {
-	if t.TaskDefinition.NetworkMode == nil || *t.TaskDefinition.NetworkMode == ecs.NetworkModeNone {
+	networkMode := aws.StringValue(t.TaskDefinition.NetworkMode)
+	if networkMode == ecs.NetworkModeNone {
 		return ""
 	}
 
 	// AWSVPC: Get Private IP from tasks->attachments (ElasticNetworkInterface -> privateIPv4Address)
-	if *t.TaskDefinition.NetworkMode == ecs.NetworkModeAwsvpc {
+	if networkMode == ecs.NetworkModeAwsvpc {
 		for _, v := range t.Task.Attachments {
 			if aws.StringValue(v.Type) == "ElasticNetworkInterface" {
 				for _, d := range v.Details {
--- a/internal/ecsservicediscovery/decoratedtask_test.go
+++ b/internal/ecsservicediscovery/decoratedtask_test.go
@@ -175,7 +175,6 @@ func Test_ExportDockerLabelBasedTarget_Fargate_AWSVPC(t *testing.T) {
 	assert.Equal(t, "/metrics", target.Labels["__metrics_path__"])
 	assert.Equal(t, "/metrics", target.Labels["ECS_PROMETHEUS_METRICS_PATH"])
 
-
 }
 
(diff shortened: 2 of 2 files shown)


internal/ecsservicediscovery/decoratedtask.go
```diff
@@ -72,12 +72,13 @@ func addExporterLabels(labels map[string]string, labelKey string, labelValue *st
 // Get the private ip of the decorated task.
 // Return "" when fail to get the private ip
 func (t *DecoratedTask) getPrivateIp() string {
-	if t.TaskDefinition.NetworkMode == nil || *t.TaskDefinition.NetworkMode == ecs.NetworkModeNone {
```
griffinmyers, line 75: When `t.TaskDefinition.NetworkMode == nil`, we _want_ the logic to fall through to the bottom of the function, where

```
if t.EC2Info != nil {
    return t.EC2Info.PrivateIP
}
```

should return the correct IP for bridge mode.

javabrett (approved):

internal/ecsservicediscovery/decoratedtask_test.go
```diff
@@ -403,18 +402,20 @@ func buildWorkloadEC2BridgeDynamicPort(dockerLabel bool, taskDef bool, serviceNa
...
 
 func Test_ExportMixedSDTarget_EC2_Bridge_DynamicPort(t *testing.T) {
-	testExportMixedSDTarget_EC2_Bridge_DynamicPort(t, ecs.NetworkModeBridge, 2)
+	networkMode := ecs.NetworkModeBridge
+	testExportMixedSDTarget_EC2_Bridge_DynamicPort(t, &networkMode, 2)
 }
 
 func Test_ExportMixedSDTarget_EC2_Bridge_DynamicPort_With_Implicit_NetworkMode(t *testing.T) {
-	testExportMixedSDTarget_EC2_Bridge_DynamicPort(t, "", 2)
+	testExportMixedSDTarget_EC2_Bridge_DynamicPort(t, nil, 2)
```
adam-mateen, line 410: https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html
If users could set networkmode to "" in their ECS task definition, then it would be good to keep the test for "", don't you think?

griffinmyers: Good instinct! I'm happy to add the test back (and, I guess ensure it behaves same as not specifying the network mode), but don't believe ECS will allow you to set `""` as the value. 

I ran a quick test (setting `networkMode: ""`) and received this error:

```
ClientException:  is not a valid network mode. Must be one of [bridge,host,none,awsvpc]
```

What do you think?

adam-mateen: Ah, okay.
Thanks for trying it anyways.
Approved.

adam-mateen (approved):

SaxyPandaBear (approved):

griffinmyers:
Thanks @javabrett, @adam-mateen, and @SaxyPandaBear!

PR #402, opened 2022-03-11
cut scan length to 20 bytes

ZhenyuTan-amz:
# Description of the issue
Performance degrade when high log throughput 

# Description of changes
reduce the number of bytes scanning when determine the buffer used returned by Microsoft windows event log api should be doubled.

# License

# Tests
Manually test new build on soaking host and observe the usage of CPU and mem back to good stat.


SaxyPandaBear (approved):
LGTM since this is an interim fix

PR #403, opened 2022-03-12
Redo winevt log length detection

ZhenyuTan-amz:
# Description of the issue
During windows event log decoding, do not necessarily need to scan buffer to determine the buffer size used for every windows event log message.

# Description of changes
Only scan once and remember the result

# License

# Tests
Manually verified on test host, add unit test.


plugins/inputs/windows_event_log/wineventlog/utils.go
```diff
@@ -112,8 +105,31 @@ func UTF16ToUTF8Bytes(in []byte, length uint32) ([]byte, error) {
 	return decoded, err
 }
 
+func UTF16ToUTF8BytesForWindowsEventBuffer(in []byte, length uint32) ([]byte, error) {
+	// Since Windows server 2022, the returned value of used buffer represents for double bytes char count,
+	// which is half of the actual buffer used by byte(what older Windows OS returns), checking if the length
+	//land on the end of used buffer, if no, double it.
+	if NumbersOfBytesPerCharacter == UnknownBytesPerCharacter {
+		if isTheEndOfContent(in, length) {
+			log.Printf("I! Buffer used: %d is returning as single byte character count", length)
```
SaxyPandaBear, line 114: not sure if this is still WIP but noticed that these are info logs. I think this is fine because it should only be logged once.. I think.

plugins/inputs/windows_event_log/wineventlog/utils.go
```diff
@@ -112,8 +105,31 @@ func UTF16ToUTF8Bytes(in []byte, length uint32) ([]byte, error) {
 	return decoded, err
 }
 
+func UTF16ToUTF8BytesForWindowsEventBuffer(in []byte, length uint32) ([]byte, error) {
+	// Since Windows server 2022, the returned value of used buffer represents for double bytes char count,
+	// which is half of the actual buffer used by byte(what older Windows OS returns), checking if the length
+	//land on the end of used buffer, if no, double it.
+	if NumbersOfBytesPerCharacter == UnknownBytesPerCharacter {
```
SaxyPandaBear, line 112: I haven't looked at the code much but my concern here is concurrency - though I think maybe it's not a super big issue since it would be on the same machine so the outcome would (theoretically) be the same. Is there any potential for a race condition when trying to evaluate `NumberOfBytesPerCharacter` when calling this function? 

When I think about it, I don't think there's really much harm, but figured I'd bring it up just in case.

plugins/inputs/windows_event_log/wineventlog/utils.go
```diff
@@ -112,8 +105,31 @@ func UTF16ToUTF8Bytes(in []byte, length uint32) ([]byte, error) {
...
+	if i > cap(in) {
+		i = cap(in)
+	}
+
+	return UTF16ToUTF8Bytes(in, uint32(i))
+}
+
 func isTheEndOfContent(in []byte, length uint32) bool {
-	// scan next 20 bytes, if any of them is none '0', return false
+	// scan next 100 bytes, if any of them is none '0', return false
```
SaxyPandaBear, line 132: Honestly I think rather than constantly updating this value in the comment, we could just refer to the variable name. Changes here are noisy and don't really add a lot of value - also it can get out of sync pretty easily

plugins/inputs/windows_event_log/wineventlog/utils_test.go
```diff
@@ -15,51 +15,54 @@ import (
 
 func TestPayload_Value(t *testing.T) {
 	// one null test
+	NumbersOfBytesPerCharacter = 0
```
SaxyPandaBear, line 18: I'm fine with this because I understand that we need to reset the state for the tests, but just to make that more obvious why we're setting this, can we pull this out into a helper function like
```go
func resetState() {
    NumberOfBytesPerCharacter = 0
}
```
That makes it very obvious why we set this, and then if/when more tests get added here it's easier to remember to also do this in new tests.

plugins/inputs/windows_event_log/wineventlog/utils.go
```diff
@@ -80,21 +83,11 @@ func CreateQuery(path string, levels []string) (*uint16, error) {
 }
 
 func UTF16ToUTF8Bytes(in []byte, length uint32) ([]byte, error) {
```
SaxyPandaBear, line 85: Does this function get called externally anymore after your changes? Should this be made private scoped?

plugins/inputs/windows_event_log/wineventlog/utils.go
```diff
@@ -112,8 +105,31 @@ func UTF16ToUTF8Bytes(in []byte, length uint32) ([]byte, error) {
...
+	// Since Windows server 2022, the returned value of used buffer represents for double bytes char count,
+	// which is half of the actual buffer used by byte(what older Windows OS returns), checking if the length
+	//land on the end of used buffer, if no, double it.
+	if NumbersOfBytesPerCharacter == UnknownBytesPerCharacter {
+		if isTheEndOfContent(in, length) {
+			log.Printf("I! Buffer used: %d is returning as single byte character count", length)
+			NumbersOfBytesPerCharacter = 1
+		} else {
+			log.Printf("I! Buffer used: %d is returning as double byte character count, doubling it to get the whole buffer content.", length)
+			NumbersOfBytesPerCharacter = 2
```
SaxyPandaBear, line 118: Thinking about this more now, if we double the number of characters that get read in, does this mean that compared to Windows 2019, CWA running on Windows 2022 would potentially use twice as much memory when consuming Windows event logs?

ZhenyuTan-amz: No, they consumed same amount of space, the API behavior change is Windows 2022 decides to return n for a message with n characters while Windows 2019 returns 2n, because they are 2-bytes characters.

PR #406, opened 2022-03-20
Add wait group for avoiding checking file exists before removing file

khanhntd:
# Description of the issue
The `TestLogsFileAutoRemoval` has been in a race condition of removing the first file by using `auto_removal`; however, in some case, the first file has not been removed by the time the [check first file exists](https://github.com/aws/amazon-cloudwatch-agent/blob/master/plugins/inputs/logfile/logfile_test.go#L441-L442) occurs. Therefore, adding `WaitGroup` to wait for the second file to be completely created, then it would delete the first file and the [check first file exists](https://github.com/aws/amazon-cloudwatch-agent/blob/master/plugins/inputs/logfile/logfile_test.go#L441-L442) will be able to pass. 

# Description of changes
Add [WaitGroup](https://www.golangprograms.com/goroutines.html) for `TestLogsFileAutoRemoval`.

# License

# Tests
By using `make test` over 10 times, I have confidence that the credibility of this change has worked.

### DIFF
--- a/plugins/inputs/logfile/logfile_test.go
+++ b/plugins/inputs/logfile/logfile_test.go
@@ -408,8 +408,6 @@ func TestLogsFileAutoRemoval(t *testing.T) {
 	go func() {
 		defer wg.Done()
 
-		time.Sleep(1 * time.Second)
-
 		// create a new file matching configured pattern
 		tmpfile2, err = createTempFile("", filePrefix)
 		require.NoError(t, err)
--- a/plugins/inputs/logfile/logfile_test.go
+++ b/plugins/inputs/logfile/logfile_test.go
@@ -360,6 +360,8 @@ func TestLogsFileRemove(t *testing.T) {
 
 //When another file is created for the same file config and the file config has auto_removal as true, the old files will stop at EOF and removed afterwards
 func TestLogsFileAutoRemoval(t *testing.T) {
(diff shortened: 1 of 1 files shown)


plugins/inputs/logfile/logfile_test.go
```diff
@@ -400,7 +402,12 @@ func TestLogsFileAutoRemoval(t *testing.T) {
 			os.Remove(tmpfile2.Name())
 		}
 	}()
+
+	wg.Add(1)
+	
 	go func() {
+		defer wg.Done()
```
SaxyPandaBear, line 409: If we have this, do we still need the sleep?

khanhntd: Not anymore. I have tested without the sleep.

SaxyPandaBear: Can you remove the sleep on line [411](https://github.com/aws/amazon-cloudwatch-agent/pull/406/files#diff-c757df81e1ad75b84c4059eef343eb2b855921ec0206e402b73f3e030c592b48R411) then?

khanhntd: Deleted. Have not noticed that. Only notice the sleep I intend to update though. Thanks for reminding.

SaxyPandaBear (approved):

Issue #399, opened 2022-03-09
"Drop Prometheus metrics with unsupported types" due to cwagent built-in scrape_duration_seconds


ecerulm:
After configuring prometheus I get ` Drop Prometheus metrics with unsupported types. Only Gauge, Counter and Summary are supported.` and I'm forced to run with `debug: true` to see the offending metrics, 

It turns that the metrics that are lacking a type are `scrape_duration_seconds` , `scrape_samples_scraped`, `scape_samples_post_metric_relabeling`, `scrape_series_added`. **Those metrics must be autogenerated by amazon-cloudwatch-agent itself**, because those are not present on the prometheus target. 

```
2022-03-09T13:48:57Z D! metric name changed from "" to "scrape_duration_seconds" during relabel
2022-03-09T13:48:57Z D! metric name changed from "" to "scrape_samples_scraped" during relabel
2022-03-09T13:48:57Z D! metric name changed from "" to "scrape_samples_post_metric_relabeling" during relabel
2022-03-09T13:48:57Z D! metric name changed from "" to "scrape_series_added" during relabel
2022-03-09T13:48:57Z I! Drop Prometheus metrics with unsupported types. Only Gauge, Counter and Summary are supported.
2022-03-09T13:48:57Z I! Please enable CWAgent debug mode to view the first 1000 dropped metrics
2022-03-09T13:48:57Z D! [1/1000] Unsupported Prometheus metric: up with type:
2022-03-09T13:48:57Z D! [2/1000] Unsupported Prometheus metric: scrape_duration_seconds with type:
2022-03-09T13:48:57Z D! [3/1000] Unsupported Prometheus metric: scrape_samples_scraped with type:
2022-03-09T13:48:57Z D! [4/1000] Unsupported Prometheus metric: scrape_samples_post_metric_relabeling with type:
2022-03-09T13:48:57Z D! [5/1000] Unsupported Prometheus metric: scrape_series_added with type:
```

There are a couple of problems with the info message
* It's distracting, it forces me to go an restart the agent in debug mode to see if the dropped metric is an "actual" real metric that I care about
* It doesn't say how many metrics were dropped. If the message said `Drop 4 Prometheus metrics` then I would be able to safely ignore it as long as the number is always 4. 
* There is no way that I know of to get rid of that message. **Can I do something in the configuration to assign a type to those metrics?**

It would be great if those metrics are autoassigned the right type so I don't see the message or that instructions are provided to fix the type.

[...]

PR #407, opened 2022-03-21
Add not-dropping-unsupported-metrics for Prometheus plugins

khanhntd:
# Description of the issue
Currently, we only support three kinds of Prometheus metrics: [Gauge, Counter, Sum](https://github.com/aws/amazon-cloudwatch-agent/blob/master/plugins/inputs/prometheus_scraper/metrics_filter.go#L24-L28). However, for an unsupported metrics  such as [Histogram, Info, StateSet](https://github.com/prometheus/prometheus/blob/main/model/textparse/interface.go#L94-L103). This would drop the metric as the CWAgent's default behavior. Moreover, the customer needs to set `debug` mode to true in order to see what metrics have been dropped. However, this would cause an annoyance for their CWAgent's experience since between need to set debug to true to see unimportant metrics and setting the unimportant as unknown metrics, pushing those and not experience the dropping issues., they could choose the latter. Therefore, creating an option for customer to **set unsupported metrics as Unknown** and ignore the dropping messages.

# Description of changes
Adding flags `drop_unsupported_metric` and if flags `drop_unsupported_metric` are set to false, setting Unsupported Metrics as `textparse.MetricTypeUnknown`

# Issues to keep track
https://github.com/aws/amazon-cloudwatch-agent/issues/399

# License

# Tests
Scrapping unsupported metrics from an application and test.

### DIFF
--- a/plugins/inputs/prometheus_scraper/metrics_filter.go
+++ b/plugins/inputs/prometheus_scraper/metrics_filter.go
@@ -41,13 +41,15 @@ func (mf *MetricsFilter) Filter(pmb PrometheusMetricBatch,dropUnsupportedMetrics
 							mf.droppedMetrics = make(map[string]string)
 						}
 					}
+
 				}
+				
+				continue
 			} else {
 				pm.metricType = textparse.MetricTypeUnknown
 			}
-		} else {
-			result = append(result, pm)
 		}
+		result = append(result, pm)
 	}
 	return
 }
--- a/plugins/inputs/prometheus_scraper/metrics_filter.go
+++ b/plugins/inputs/prometheus_scraper/metrics_filter.go
@@ -26,12 +26,12 @@ func (mf *MetricsFilter) Filter(pmb PrometheusMetricBatch,dropUnsupportedMetrics
 				if mf.droppedMetrics == nil {
 					mf.droppedMetrics = make(map[string]string, mf.maxDropMetricsLogged)
 					log.Println("I! Drop Prometheus metrics with unsupported types. Only Gauge, Counter and Summary are supported.")
-					log.Println("I! Please enable CWAgent debug mode to view the first %d dropped metrics", mf.maxDropMetricsLogged)
+					log.Printf("I! Please enable CWAgent debug mode to view the first %d dropped metrics \n", mf.maxDropMetricsLogged)
 				}
 
 				if !mf.hitMaxLimit && (len(mf.droppedMetrics) < mf.maxDropMetricsLogged) {
 					if _, ok := mf.droppedMetrics[pm.metricName]; !ok {
-						log.Println("D! [%d/%d] Unsupported Prometheus metric: %s with type: %s",
+						log.Printf("D! [%d/%d] Unsupported Prometheus metric: %s with type: %s",
 							len(mf.droppedMetrics)+1,
 							mf.maxDropMetricsLogged, pm.metricName,
 							pm.metricType)
@@ -48,6 +48,7 @@ func (mf *MetricsFilter) Filter(pmb PrometheusMetricBatch,dropUnsupportedMetrics
 				pm.metricType = textparse.MetricTypeUnknown
 			}
 		}
+		
 		result = append(result, pm)
 	}
 	return
--- a/plugins/inputs/prometheus_scraper/metrics_filter.go
+++ b/plugins/inputs/prometheus_scraper/metrics_filter.go
@@ -10,7 +10,6 @@ import (
 
 const (
 	MaxDropMetricsLogged = 1000
-	dropUnsupportedMetrics = false
 )
 
 type MetricsFilter struct {
@@ -20,7 +19,7 @@ type MetricsFilter struct {
 }
 
 // Filter out and Log the unsupported metric types
-func (mf *MetricsFilter) Filter(pmb PrometheusMetricBatch) (result PrometheusMetricBatch) {
+func (mf *MetricsFilter) Filter(pmb PrometheusMetricBatch,dropUnsupportedMetrics bool) (result PrometheusMetricBatch) {
 	for _, pm := range pmb {
 		if !pm.isGauge() && !pm.isCounter() && !pm.isSummary() {
 			if dropUnsupportedMetrics {
--- a/plugins/inputs/prometheus_scraper/metrics_filter.go
+++ b/plugins/inputs/prometheus_scraper/metrics_filter.go
@@ -26,7 +26,7 @@ func (mf *MetricsFilter) Filter(pmb PrometheusMetricBatch,dropUnsupportedMetrics
 				if mf.droppedMetrics == nil {
 					mf.droppedMetrics = make(map[string]string, mf.maxDropMetricsLogged)
 					log.Println("I! Drop Prometheus metrics with unsupported types. Only Gauge, Counter and Summary are supported.")
-					log.Printf("I! Please enable CWAgent debug mode to view the first %d dropped metrics \n", mf.maxDropMetricsLogged)
+					log.Printf("I! Please enable CWAgent debug mode to view the first %d dropped metrics or set drop_unsupported_metric as false to set these metrics as Unknown type. \n", mf.maxDropMetricsLogged)
 				}
 
 				if !mf.hitMaxLimit && (len(mf.droppedMetrics) < mf.maxDropMetricsLogged) {
--- a/plugins/inputs/prometheus_scraper/metrics_filter.go
+++ b/plugins/inputs/prometheus_scraper/metrics_filter.go
@@ -5,10 +5,12 @@ package prometheus_scraper
 
 import (
 	"log"
+	"github.com/prometheus/prometheus/pkg/textparse"
 )
 
 const (
 	MaxDropMetricsLogged = 1000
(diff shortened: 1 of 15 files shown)


plugins/inputs/prometheus_scraper/metrics_filter.go
```diff
@@ -18,31 +19,37 @@ type MetricsFilter struct {
 }
 
 // Filter out and Log the unsupported metric types
-func (mf *MetricsFilter) Filter(pmb PrometheusMetricBatch) (result PrometheusMetricBatch) {
+func (mf *MetricsFilter) Filter(pmb PrometheusMetricBatch,dropUnsupportedMetrics bool) (result PrometheusMetricBatch) {
```
SaxyPandaBear, line 22: looks like this might need `go fmt`

plugins/inputs/prometheus_scraper/metrics_filter_test.go
```diff
@@ -30,54 +30,119 @@ func buildTestData(pass int, drop int) (result PrometheusMetricBatch) {
...
 }
 
 func TestMetricsFilterFilter_ExceedMaxDroppedMetrics(t *testing.T) {
+	var dropUnsupportedMetric bool
+	//Filter Metric with DroppingUnsupportedMetric equals to true
+	metricFilterWithDroppingUnsupportedMetric := &MetricsFilter{maxDropMetricsLogged: 3}
+
+	dropUnsupportedMetric = true
+	metricBatchWithDroppingUnsupportedMetric := buildTestData(2, 12)
+	metricBatchWithDroppingUnsupportedMetric = metricFilterWithDroppingUnsupportedMetric.Filter(metricBatchWithDroppingUnsupportedMetric,dropUnsupportedMetric)
```
SaxyPandaBear, line 126: This bit of code looks like it is duplicated.

Two things:
1. What is the purpose of the last two lines? We build the test data and then always filter out unsupported metrics? 
2. Can all of this logic be put inside of the `buildTestData()` helper function? 
3.

plugins/inputs/prometheus_scraper/metrics_handler.go
```diff
@@ -47,7 +48,7 @@ func (mh *metricsHandler) handle(pmb PrometheusMetricBatch) {
 	pmb = mh.mtHandler.Handle(pmb)
 
 	// Filter out Histogram and untyped Metrics and adding logging
-	pmb = mh.filter.Filter(pmb)
+	pmb = mh.filter.Filter(pmb,mh.dropUnsupportedMetric)
```
SaxyPandaBear, line 51: `go fmt`

plugins/inputs/prometheus_scraper/metrics_type_handler.go
```diff
@@ -55,6 +55,10 @@ func (pm *PrometheusMetric) isSummary() bool {
 	return pm.metricType == textparse.MetricTypeSummary
 }
 
+func (pm *PrometheusMetric) isUnknown() bool {
```
SaxyPandaBear, line 58: I see that this function is called [here](https://github.com/aws/amazon-cloudwatch-agent/pull/407/files#diff-7afaf53868de6a1aec096eab2439312727c4ca612d2f54eb5ebed758c1feac8aR47) which happens to be in the same package so it is accessible. Should this function be made public?

translator/totomlconfig/sampleConfig/prometheus_config_windows.json
```diff
@@ -6,6 +6,7 @@
     "metrics_collected": {
       "prometheus": {
         "cluster_name": "TestCluster",
+        "drop_unsupported_metric": false,
```
SaxyPandaBear, line 9: Is there any new test config with this value set to `true`? 

Furthermore, does the value in the JSON have to exist - is this a non-breaking change in the schema config?

translator/translate/logs/metrics_collected/prometheus/ruleDropUnsupportMetric.go
```diff
@@ -0,0 +1,30 @@
...
+
+const (
+	SectionKeyDropUnsupportedMetric = "drop_unsupported_metric"
+	DefaultDropUnsupportedMetric = true
+)
+
+type DropUnsupportedMetric struct {
+}
+
+func (d *DropUnsupportedMetric) ApplyRule(input interface{}) (returnKey string, returnVal interface{}) {
```
SaxyPandaBear, line 15: Can you add a unit test for this?

khanhntd:
It will be assigned as [Unknown type](https://github.com/aws/amazon-cloudwatch-agent/pull/407/files#diff-8164b096d957cf02b91a37300e6c77d8b5586eef76a0b1c24a279cad7f6f557bR47-R49) and submit [this unknown metric type without any calculating](https://github.com/aws/amazon-cloudwatch-agent/pull/407/files#diff-7afaf53868de6a1aec096eab2439312727c4ca612d2f54eb5ebed758c1feac8aR47-R49)

translator/totomlconfig/sampleConfig/prometheus_config_windows.json
khanhntd: No this param does not need to be existed since the default value for [this param has been set to true](https://github.com/aws/amazon-cloudwatch-agent/pull/407/files#diff-8cf20937efbe2691d764eb3fdee2185782480ef35e9c4f566f1180d89324698fR9), which is the usual behavior for CWAgent when filtering Prometheus's metrics.

plugins/inputs/prometheus_scraper/metrics_type_handler.go
khanhntd: Are there any plugins besides `Prometheus` using these types? If not, I don't see any reason to make them public though.

plugins/inputs/prometheus_scraper/metrics_handler.go
khanhntd: No worries. I will let's everyone review them before making a final change using `make fmt` though.

translator/translate/logs/metrics_collected/prometheus/ruleDropUnsupportMetric.go
khanhntd: The schema test is a unit test for [this rule](https://github.com/aws/amazon-cloudwatch-agent/pull/407/files#diff-cc5c0c08c341aac1f4d061da8ae056f92d2d6c01e4e4286d693e95b8068eef72R21). Meaning after translating the CWAgent's configuration file to toml file with the predefined rules, it would validate between the expected toml file and the actual one.

plugins/inputs/prometheus_scraper/metrics_filter_test.go
khanhntd: 1. Technically yes, we build two types of metrics: counter which supposed to be pass and histogram, which supposed to be dropped.
2. Great idea though. WIll try to do that.

plugins/inputs/prometheus_scraper/metrics_filter.go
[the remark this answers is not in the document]
khanhntd: I have changed the necessary changes within the package. So no regression has been detected so far. Also confirming this by using `make build`. Only thing left to do is test on my desktop.

khanhntd:
> > and submit [this unknown metric type without any calculating](https://github.com/aws/amazon-cloudwatch-agent/pull/407/files#diff-7afaf53868de6a1aec096eab2439312727c4ca612d2f54eb5ebed758c1feac8aR47-R49)
> 
> So what happens on the customer's side? Are the metrics useful in any way?

For your information, the metrics got submitted with the type: unknown metric is a similar behavior to [AWS Otel Collector](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/prometheusreceiver/internal/otlp_metricsbuilder.go#L61-L81) and [Telegraf](https://github.com/influxdata/telegraf/blob/master/plugins/inputs/prometheus/parser.go#L74-L77). Moreover, there are two benefits regarding applying this feature:
* Ignoring the debugged corresponding messages, meaning not showing the dropping of metrics with an unsupported type
* Keeping the useful metrics, which could be in these types: Histogram, Info. 
* (Hidden benefit) Customer will give us opinion on how we can deal with these unknown type such as the Histogram one.

For the second question, the metrics will be useful for them. The only thing we have changed regarding these metrics so far is their type. Moreover, it make more sense for me to not dropping these metrics and let the customer experience the `calculated metrics` and give their opinion on how we calculate these metrics afterwards. For example, `the Histogram` or `the Info` metric type. What I am considering right now is between these two solutions:
* Do we honor the metrics type and do not set the metric type as unknown for unsupported metrics? (more regression for unsupported metric types but benefit for the customer in grouping these metrics type)
* Do we not honor the metrics type and set the metric type as unknown for unsupported metrics? (opposite)
For what happened on customer side, I am currently building and testing it out before giving any  updates

ebarault:
hi @khanhntd @SaxyPandaBear 
any chance this MR lands?
currently all my spark metrics are dropped with errors like these:
`E! metricsHandler NO metaData for metrics_test_driver_LiveListenerBus_queue_streams_listenerProcessingTime_MeanRate`

ebarault:
hi @khanhntd , why closing this MR ?

PR #410, opened 2022-03-23
Do Not Run montudor/action-zip@v1 When Previous Build Pass

sethAmazon:
# Description of the issue
Fails this job when previous test passes. 

# Description of changes
do not run this job on previous pass. 

# Tests
None


SaxyPandaBear (approved):
Sounds like this is why the step was failing when rerunning. LGTM

Issue #404, opened 2022-03-14
fetch-config should error if the input file is not valid JSON


ecerulm:
Currently the following command:

```
sudo /opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent-ctl -a fetch-config -m ec2 -s -c file:/opt/aws/amazon-cloudwatch-agent/var/cwagent-config.json
```
will overwrite `/opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.toml`  even if the input JSON file is **not valid JSON**. 

It will print out **No json config files found, use the default one**

```
2022/03/14 10:10:41 unable to scan config dir /opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.d with error: unable to parse json, error: invalid character '}' after array element
No json config files found, use the default one
```

But the exit value is still 0 (indicating no error). My proposal is that if the input file is invalid **it should error (non zero exit code) and refuse to overwrite the TOML file.**. Otherwise it's very easy to miss that the configuration was incorrectly updated.

Full output: 

``` 
****** processing amazon-cloudwatch-agent ******
/opt/aws/amazon-cloudwatch-agent/bin/config-downloader --output-dir /opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.d --download-source file:/opt/aws/amazon-cloudwatch-agent/var/cwagent-config.json --mode ec2 --config /opt/aws/amazon-cloudwatch-agent/etc/common-config.toml --multi-config default
Successfully fetched the config and saved in /opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.d/file_cwagent-config.json.tmp
Start configuration validation...
/opt/aws/amazon-cloudwatch-agent/bin/config-translator --input /opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.json --input-dir /opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.d --output /opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.toml --mode ec2 --config /opt/aws/amazon-cloudwatch-agent/etc/common-config.toml --multi-config default
2022/03/14 10:07:14 Reading json config file path: /opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.d/file_cwagent-config.json.tmp ...
[...]

PR #412, opened 2022-03-24
Add strict validation flag for validating strictly on passing multi_config value as default or appending

khanhntd:
# Description of the issue
Currently, if the `multi-config` set as `default` or `append` and if the agent was not able to [merge the configuration](https://github.com/aws/amazon-cloudwatch-agent/blob/master/translator/cmdutil/translatorutil.go#L206) (happens in the agent translator), [the agent will use the default json configuration ](https://github.com/aws/amazon-cloudwatch-agent/blob/master/translator/jsonconfig/mergeJsonConfig.go#L31-L33) depends on the OS to replace the old agent's running json configuration. This will cause 2 tremendous errors:
* Stop the running agent and replace the running agent with the default json configuration file if flag -s is turning on.
* Misleading the customer in not showing the corresponding error when the translation from json to toml is not succeed.

[ecerulm](https://github.com/ecerulm) and [SaxyPandaBear](https://github.com/SaxyPandaBear) have offered two interesting solutions:
* Add a flag for strict validation.
* Add a corresponding error when translation failed.

The latter would cause translator to exit on every fail translation; however, for the former, it would maintain the CWAgent behavior: if translator failed, the CWAgent will use the default json configuration; however, if the flag `strict_validation` turns on, it would exit after translator would fail .Therefore, decide to have this as a final solution. `Add a flag for strict validation`

# Issues to keep track
https://github.com/aws/amazon-cloudwatch-agent/issues/404

# Description of changes
Add a `strict_validation` flag for `config-translator` and also for `amazon-cloudwatch-agent-ctl`

# License

# Tests
Use `make release` to build binaries for `amazon-cloudwatch-agent-ctl` and `config-translator`, test with invalid configuration to validate the changes. The below image is the expect result:
image


cmd/config-translator/translator.go
```diff
@@ -32,6 +32,8 @@ func initFlags() {
 	var inputMode = flag.String("mode", "ec2", "Please provide the mode, i.e. ec2, onPrem")
 	var inputConfig = flag.String("config", "", "Please provide the common-config file")
 	var multiConfig = flag.String("multi-config", "remove", "valid values: default, append, remove")
+	var strictValidation = flag.String("strict", "false", "Strict validation for agent json confil file. Valid values: true, false")
```
SaxyPandaBear, line 35: We should use `flag.Bool` https://pkg.go.dev/flag#Bool so we don't accept arbitrary string values for this config

khanhntd: True. Current fixing behind the scene with `make release`.

khanhntd: On second thought, after using `flag.Bool`, there is certain behavior for it: we need to pass as `strict-validation=false` (only work with =); otherwise, it would always return true if we use flag strict-validation without equal sign. Do we want that to happen? Personally I would say no [since most of our documents start cwagent without equal side](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/install-CloudWatch-Agent-commandline-fleet.html). Therefore, would need to add some validations in the `config-translator` for this flag.

cmd/config-translator/translator.go
```diff
@@ -32,6 +32,8 @@ func initFlags() {
 	var inputMode = flag.String("mode", "ec2", "Please provide the mode, i.e. ec2, onPrem")
 	var inputConfig = flag.String("config", "", "Please provide the common-config file")
 	var multiConfig = flag.String("multi-config", "remove", "valid values: default, append, remove")
+	var strictValidation = flag.Bool("strict-validation", false, "Return exit code 99 if validating agent's json config fails. Valid values: true, false")
```
SaxyPandaBear, line 35: One thing I thought about that I think we need to test just so we know how this `bool` flag behaves.

We would want to confirm how to pass in the flag on different OSes, namely *NIX (Linux, Unix) versus Windows.

go.mod
```diff
@@ -42,23 +42,23 @@ require (
 	github.com/BurntSushi/toml v0.3.1
 	github.com/Jeffail/gabs v1.4.0
 	github.com/aws/aws-sdk-go v1.30.15
-	github.com/aws/aws-sdk-go-v2 v1.13.0 // indirect
-	github.com/aws/aws-sdk-go-v2/config v1.13.1 // indirect
-	github.com/aws/aws-sdk-go-v2/service/ec2 v1.29.0 // indirect
-	github.com/aws/smithy-go v1.10.0 // indirect
+	github.com/aws/aws-sdk-go-v2 v1.13.0
```
SaxyPandaBear, line 45: Why is this getting changed? Didn't @sethAmazon already fix this by running `go mod tidy`? Can you please run that on your workspace too?

packaging/darwin/amazon-cloudwatch-agent-ctl
```diff
@@ -65,6 +65,8 @@ UsageString="
             file:<file-path>:                       file path on the host
             all:                                    all existing configs. Only apply to remove-config action.
 
+        -v: Return exit code 99 if validating agent's json config fails.
```
SaxyPandaBear, line 68: This is probably a nitpick, but I disagree very strongly about the flag name we are choosing for this. Typically, `-v` is reserved for "verbose" or "version". I don't think that we should use this flag name for strict validation. Is there any way we can use a longer name than a single character? I would like this to just be `-strict` if possible.

packaging/dependencies/amazon-cloudwatch-agent-ctl
```diff
@@ -79,6 +81,8 @@ UsageString="
         -s: optionally restart after configuring the agent configuration
             this parameter is used for 'fetch-config', 'append-config', 'remove-config' action only.
 
+        -v: Return exit code 99 if validating agent's json config fails.
```
SaxyPandaBear, line 84: Same comment about naming

packaging/windows/amazon-cloudwatch-agent-ctl.ps1
```diff
@@ -32,6 +34,7 @@ $UsageString = @"
                 [-c default|all|ssm:<parameter-store-name>|file:<file-path>]
                 [-o default|all|ssm:<parameter-store-name>|file:<file-path>]
                 [-s]
+                [-v]
```
SaxyPandaBear, line 37: At what point does the `StrictValidation` get mapped to this flag?

translator/jsonconfig/mergeJsonConfig.go
```diff
@@ -25,11 +25,12 @@ func MergeJsonConfigMaps(jsonConfigMapMap map[string]map[string]interface{}, def
 				return util.GetJsonMapFromJsonBytes([]byte(config.DefaultECSJsonConfig()))
 			}
 		}
-		if multiConfig == "remove" {
-			fmt.Println("No json config files found, please provide config, exit now")
+		
+		if strictValidation || multiConfig == "remove" {
```
SaxyPandaBear, line 29: Personally, I think we should use a different exit code, and thus a different conditional branch for when the JSON fails validation - a failed JSON configuration translation is not the same as missing a JSON file in my opinion.

go.mod
khanhntd: Maybe I have not merged the changes from @sethAmazon yet. Moreover, I used `go mod tidy` to make this changes so properly does no different from @sethAmazon 's changes. However, making sure by pulling the latest changes.

cmd/config-translator/translator.go
khanhntd: Yes. I have tested on different platforms. Only needs to pass with as `--strict-validation`, for example: `./config-translator --config ./agent-config.json --strict-validation` without passing any values(if passing, should be with --strict-validation=true)

translator/jsonconfig/mergeJsonConfig.go
khanhntd: Cause the current code is exit after detecting there is no json configuration was founding (since the json is invalid); therefore, exiting with `ERR_CODE_NOJSONFILE` . However, I got stuck between [exit after founding there is no json configuration](https://github.com/aws/amazon-cloudwatch-agent/pull/412/files#diff-e64720159e8e883ef2291f9df929db8abbc5901003168fc2568090ad638addfdR31) because of invalid json and [during the invalidation](https://github.com/aws/amazon-cloudwatch-agent/pull/412/files#diff-eb5d801c3c6e31c5d7ef534897e7e095a1f6344597db7d95adce51d962ec1038R186-R200) because of how we handle error. However, fixing it after some thoughts again.

packaging/darwin/amazon-cloudwatch-agent-ctl
khanhntd: Agree on this part. -v stands for version or verbose. However, if using `long command`, current `getopts` would only suitable for single character. Therefore, utilizing `https://github.com/UrsaDK/getopts_long` to provide addition long arguments for current flags and `strict` flag

translator/jsonconfig/mergeJsonConfig.go
khanhntd: After thinking some thoughts again, the ```fmt.Println("No json config files found, please provide config, exit now")``` makes me thinking of **we only output this error when the json is missing**. However, that's not true since if that json configuration was not found, it would error before coming to this condition (image 1). Therefore, the only possible errors that come to this condition would be `invalid json configuration`. So changed the error from `ERR_CODE_NOJSONFILE` to `ERR_CODE_INVALIDJSONFILE` if that makes sense.
image

packaging/darwin/amazon-cloudwatch-agent-ctl
SaxyPandaBear: I'm open to other suggestions though. I would err towards just using a different character rather than pulling in another library just to allow using a long name `-strict`. @ecerulm suggested this so would love feedback from them on what they envision it could/should be. 

I think one thing we could do is try to find examples of other established tools that have some form of fail-fast or strict validation mechanism to see what the consensus is for the naming of the flag used for it.

packaging/darwin/amazon-cloudwatch-agent-ctl
```diff
@@ -252,23 +255,87 @@ cwa_config() {
      fi
 }
 
+getopts_long() {
```
SaxyPandaBear, line 258: If this is what is required in order to support using `-strict` as a flag, I don't think it's worthwhile. See previous comment about looking at alternative single char flags to use other than `-v`

translator/jsonconfig/mergeJsonConfig.go
SaxyPandaBear: When we exit, can we not distinguish between there being an error parsing a JSON file versus no JSON file being supplied? At the end of the day, I think this change is fine because technically having no JSON file is "invalid".

packaging/dependencies/amazon-cloudwatch-agent-ctl
```diff
@@ -50,16 +51,17 @@ UsageString="
...
             start:                                  start the agent process.
             status:                                 get the status of the agent process.
+            validate:                               validate the agent's configuration
             fetch-config:                           apply config for agent, followed by -c or -o or both. Target config can be based on location (ssm parameter store name, file name), or 'default'.
             append-config:                          append json config with the existing json configs if any, followed by -c. Target config can be based on the location (ssm parameter store name, file name), or 'default'.
             remove-config:                          remove config for agent, followed by -c or -o or both. Target config can be based on the location (ssm parameter store name, file name), or 'all'.
             set-log-level:                          sets the log level, followed by -l to provide the level in all caps.
 
-        -m: mode
+        -m/--mode: mode
```
SaxyPandaBear, line 64: Doing this as a separate review so I only block on this.

---
I am hesitant on including a change that allows us to use long-name flags for existing functionality. I'm mostly concerned of scope creep, because the original ask was to add a new flag for just forcing strict validation. I'm going back on my suggestion/request earlier on using `-strict` long form, now that I see what is required to make that work, but I want others to weigh in on this.

Do we benefit enough from including this change as part of the PR, or should should we (1) change the strict validation flag back to just a single char flag and (2) back out this specific change and put it in a new issue to address separately?

SaxyPandaBear (changes requested):

packaging/darwin/amazon-cloudwatch-agent-ctl
khanhntd: That's also another thing I am stucking when inventing another character for `--strict`, thought of change `s` from restart to strict but it's a default behavior for CWAgent entire time (kinda one way door a little bit).

packaging/darwin/amazon-cloudwatch-agent-ctl
khanhntd: That's not only the reason why I support using long arguments though. Another reason for this is when using `--help`, it would duplicate the `UsageString` with the error invalid option. Moreover, using long arguments also indicates their meaning. So in the future, we would need to support long arguments as well.

translator/jsonconfig/mergeJsonConfig.go
khanhntd: Agree on later comments. The way we handling error right now is: checking if file exist from [config download] (https://github.com/aws/amazon-cloudwatch-agent/blob/master/cmd/config-downloader/downloader.go#L195-L217) -> error parsing a JSON file from [config translator]. Therefore, yes we can handle an error code for parsing JSON file and no JSON file being supplied. However, this still does not change anything much though. `config-translator` is only for translator, not fetching so the `fmt.Println("No json config files found, please provide config, exit now")` does not make sense to me.

packaging/dependencies/amazon-cloudwatch-agent-ctl
khanhntd: I would agree on add a single char flag for `strict` and add a long form for it though. You can see it in my last reply https://github.com/aws/amazon-cloudwatch-agent/pull/412#discussion_r835402859

packaging/darwin/amazon-cloudwatch-agent-ctl
SaxyPandaBear: I understand why we would want to include it. I guess the question is, do we want to include that in this specific PR or address it separately? That's what I want some outside opinions on.

packaging/darwin/amazon-cloudwatch-agent-ctl
SaxyPandaBear: Yeah, I wouldn't change the meaning of another flag. Like I said, I think we should look at other software for potential ideas for reasonable flags that we can use here for the naming of this feature flag. Personally, I thought maybe `-f` and change the meaning to `--failfast` or something like that rather than just thinking of some variation of `strict-validation`. I think it's worth taking the time to look at other examples for what makes the most sense here before committing.

translator/jsonconfig/mergeJsonConfig.go
ZhenyuTan-amz: Does this makes output become "invalid json configuration" when customer doesn't provide a json config at all? Still think we should distinguish missing file vs invalid file.

packaging/darwin/amazon-cloudwatch-agent-ctl
```diff
@@ -162,8 +164,9 @@ cwa_pid() {
 cwa_config() {
      config_location="${1:-}"
      restart="${2:-}"
-     mode="${3:-}"
-     multi_config="${4:-}"
+     strict_validation="${3:-}"
```
sethAmazon, line 167: Should strict be the 5th instead?

translator/jsonconfig/mergeJsonConfig.go
sethAmazon: I would say no json file is invalid json file, but I do agree it would be better to separate.

khanhntd: If that's the case, change how we handle errors a little bit. There would be lots of scenario but able to write the final code for these scenarios.

packaging/darwin/amazon-cloudwatch-agent-ctl
khanhntd: I am fine with either of these since my habit of grouping boolean things together.

translator/jsonconfig/mergeJsonConfig.go
```diff
@@ -16,20 +17,29 @@ import (
...
 				return util.GetJsonMapFromJsonBytes([]byte(config.DefaultECSJsonConfig()))
 			}
 		}
-		if multiConfig == "remove" {
-			fmt.Println("No json config files found, please provide config, exit now")
-			os.Exit(config.ERR_CODE_NOJSONFILE)
+
+		if errors.Is(inputJsonDirPathErr, os.ErrNotExist) {
+			//When there is no input from flag --input, --input-dir and cannot find agent's config through containerized environment
+			log.Printf("No agent's json config was found from containerized environment or file path.")
```
SaxyPandaBear, line 32: I'm not sure we _need_ to specify about containerized environments. What is the reasoning behind including that in the log?

translator/jsonconfig/mergeJsonConfig.go
```diff
@@ -16,20 +17,29 @@ import (
 	"github.com/aws/amazon-cloudwatch-agent/translator/util"
 )
 
-func MergeJsonConfigMaps(jsonConfigMapMap map[string]map[string]interface{}, defaultJsonConfigMap map[string]interface{}, multiConfig string) (map[string]interface{}, error) {
+func MergeJsonConfigMaps(jsonConfigMapMap map[string]map[string]interface{}, defaultJsonConfigMap map[string]interface{}, multiConfig string, strictValidation bool, inputJsonDirPathErr error ) (map[string]interface{}, error) {
```
SaxyPandaBear, line 20: Personally, I don't like the idea of passing in an `error` as an input to a function.

translator/jsonconfig/mergeJsonConfig.go
```diff
@@ -16,20 +17,29 @@ import (
...
-			fmt.Println("No json config files found, please provide config, exit now")
-			os.Exit(config.ERR_CODE_NOJSONFILE)
+
+		if errors.Is(inputJsonDirPathErr, os.ErrNotExist) {
+			//When there is no input from flag --input, --input-dir and cannot find agent's config through containerized environment
+			log.Printf("No agent's json config was found from containerized environment or file path.")
+			if strictValidation{
+				os.Exit(config.ERR_CODE_NOJSONFILE)
+			}
+			log.Printf("Use the default agent's json config.")
```
SaxyPandaBear, line 36: This feels like a bad code smell to me. I could understand returning from a function early, but having a log/other actions afterwards if the condition isn't met, but I don't feel like that same logic applies to this scenario where we abruptly terminate the process with an exit call. 

In this code, we check if the error is a file/dir doesn't exist, and then if there is strict validation, we terminate. In the `else if`, if there is strict validation or we want to remove configs, then we log exit. Otherwise, we fail through. 

I haven't figured out how to rework this, but I will think about this more.

translator/jsonconfig/mergeJsonConfig.go
```diff
@@ -16,20 +17,29 @@ import (
...
+		} else if strictValidation || multiConfig == "remove" {
+			log.Printf("Invalid agent's json config files, please provide an appropriate config.")
+			os.Exit(config.ERR_CODE_INVALIDJSONFILE)
 		} else {
```
SaxyPandaBear, line 40: I don't think we need this `else` here. Both of the conditions before it cause the process to end abruptly. If you took out the final else and kept the log statements, I feel like it would be just as valid.

translator/jsonconfig/mergeJsonConfig.go
```diff
@@ -16,20 +17,29 @@ import (
 	"github.com/aws/amazon-cloudwatch-agent/translator/util"
 )
 
-func MergeJsonConfigMaps(jsonConfigMapMap map[string]map[string]interface{}, defaultJsonConfigMap map[string]interface{}, multiConfig string) (map[string]interface{}, error) {
+func MergeJsonConfigMaps(jsonConfigMapMap map[string]map[string]interface{}, defaultJsonConfigMap map[string]interface{}, multiConfig string, strictValidation bool, inputJsonDirPathErr error ) (map[string]interface{}, error) {
 	if jsonConfigMapMap == nil || len(jsonConfigMapMap) == 0 {
 		if os.Getenv(config.USE_DEFAULT_CONFIG) == config.USE_DEFAULT_CONFIG_TRUE {
```
SaxyPandaBear, line 22: Should we honor this logic for a non ECS environment? I don't know the context behind this, but it seems like it could be relevant.

translator/cmdutil/translatorutil.go
```diff
@@ -132,61 +132,69 @@ func GenerateMergedJsonConfigMap(ctx *context.Context) (map[string]interface{},
...
-			if info.IsDir() {
-				return nil
-			}
-
-			if filepath.Ext(path) == context.TmpFileSuffix {
-				// .tmp files
-				if ctx.MultiConfig() == "default" || ctx.MultiConfig() == "append" {
-					jsonConfigMap, err := getJsonConfigMap(path, ctx.Os())
+				if info.Mode()&os.ModeSymlink != 0 {
+					log.Printf("Find symbolic link %s \n", path)
```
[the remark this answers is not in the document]
khanhntd: For this error log message, I have not changed anything. It shows the change because there was a space tab for it though.

translator/jsonconfig/mergeJsonConfig.go
khanhntd: Either way it's the same meaning. Therefore, it is possible to honor this logic for a non ECS and ECS environment. Besides, I only want to specific about log messages as possible.

translator/jsonconfig/mergeJsonConfig.go
khanhntd: Agree. Passing a boolean instead of `error`. as an input to a function.

translator/jsonconfig/mergeJsonConfig.go
khanhntd: I would want to keep the ```log.Printf("Use the default agent's json config.")```. However, change code to node use the else after os.Exit 
```
		if isNotExistError {
			//When there is no input from flag --input, --input-dir and cannot find agent's config through containerized environment
			log.Printf("No agent's json config was found.")
			if strictValidation{
				os.Exit(config.ERR_CODE_NOJSONFILE)
			}
		} else {
			log.Printf("Invalid agent's json config files")
			if strictValidation || multiConfig == "remove" {
				os.Exit(config.ERR_CODE_INVALIDJSONFILE)
			}
		}
		
		log.Printf("Use the default agent's json config.")
		return defaultJsonConfigMap, nil
```

translator/jsonconfig/mergeJsonConfig.go
khanhntd: Personally, I would agree on logs/action afterwards actions afterwards would not be a best practice. Cause the whole reason I writing this one was to simplifying the code and manage all error states in `MergeJsonConfigMaps`. However, if that's not what we desired, then keeping an original code would be fine by me. Cause the question I only keep question is: how was we able to differentiate between `Invalid Config` and `NoConfig` through the function `MergeJsonConfigMaps`. If its only serves as a helper one then only `logging Invalid`  would be a better idea for me.

translator/jsonconfig/mergeJsonConfig.go
SaxyPandaBear: We aren't including any information about what the error is. Would we be able to distinguish between the file not existing on a host versus not being accessible in a container? If not then I think changing this log just makes it more confusing.

packaging/darwin/amazon-cloudwatch-agent-ctl
```diff
@@ -36,7 +36,7 @@ readonly ALL_CONFIG='all'
 UsageString="
 
 
-        usage: amazon-cloudwatch-agent-ctl -a stop|start|status|fetch-config|append-config|remove-config [-m ec2|onPremise|auto] [-c default|ssm:<parameter-store-name>|file:<file-path>] [-s]
+        usage: amazon-cloudwatch-agent-ctl -a stop|start|status|fetch-config|append-config|remove-config [-m ec2|onPremise|auto] [-c default|ssm:<parameter-store-name>|file:<file-path>] [-s] [-v]
```
SaxyPandaBear, line 39: I still think that `-v` is not the flag we should use here because of how it can be interpreted. Did you take a look at comparable functionality in other executables for ideas? 

I took a look myself briefly and couldn't find anything that has a similar requirement to what you're trying to add. 

---
Honestly, I think it's worthwhile to take a step back and think about why we want this to be an optional flag that we can apply on any of the actions from the ctl script, and why we wouldn't just want to add a new action `validate`.

It would be a little different than what is asked for in the original GitHub issue, but I think semantically, adding a `validate` action might make more sense - I would like input from others like @ecerulm and @ZhenyuTan-amz on this.

So we could do something like:
```bash
/usr/bin/amazon-cloudwatch-agent-ctl -a validate -c default -o default
```

And we could just have the action run the translator and parrot the exit code from that executable. If we want to make a shift to have everything customer-facing centralized in the `ctl` script, then I think that adding a new action would make more sense than adding a new flag here.

SaxyPandaBear (commented):
I think that we should discuss next week about what we really want to accomplish out of this feature, and how this aligns or conflicts with the path we want to lead with the agent and how people use it.

cmd/config-translator/translator.go
```diff
@@ -32,6 +32,8 @@ func initFlags() {
 	var inputMode = flag.String("mode", "ec2", "Please provide the mode, i.e. ec2, onPrem")
 	var inputConfig = flag.String("config", "", "Please provide the common-config file")
 	var multiConfig = flag.String("multi-config", "remove", "valid values: default, append, remove")
+	var strictValidation = flag.Bool("strict-validation", false, "Return exit code 98 if validating agent's json config fails and 99 if no config json was found.")
```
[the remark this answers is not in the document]
khanhntd: For the default behavior, the translator will have a non-exit zero code when **--multi-config is default or append** even if the translator is having an invalid config or having no json file since it would use a default JSON config for that particular OS. So the cx have get used to it for a long time; therefore, I honest don't want to change this behavior for them though.

SaxyPandaBear:
After discussing today, I think the simplest course of action with the least amount of splash damage for customers is to just update the log that gets emitted when the translator fails to parse the JSON:
```
2022/03/14 10:10:41 unable to scan config dir /opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.d with error: unable to parse json, error: invalid character '}' after array element
No json config files found, use the default one
```
We could update the above to just include our standard telegraf `E!` prefix, even though it doesn't actually use telegraf, just to make it more evident that there is an error in the log. One thing that does bother me which I think we could/should update is the timestamp format between the different executables, but that's a problem for another day.

ecerulm:
> After discussing today, I think the simplest course of action with the least amount of splash damage for customers is to just update the log that gets emitted when the translator fails to parse the JSON

@SaxyPandaBear,  should this PR get closed then?

PR #413, opened 2022-03-25
Update go.mod dependencies 

khanhntd:
# Description of the issue
Update go.mod dependencies with does not have breaking changes.

# License

# Tests
Using `make test` and `integration test` to test all the packages changes.


SaxyPandaBear:
I don't think that this PR is necessary. There are already two newer PRs with more value that both update the go mod/sum

PR #416, opened 2022-03-29
update error handling to account for ResourceAlreadyExists gracefully

### LINKED ISSUE #415: Don't emit error log when log group already created
Given the following assumptions about your environment:
1. Running CloudWatch agent on multiple instances
2. CloudWatch agent is expected to create new log group and log streams
3. Log group is shared across instances

When the agent starts up independently on each host, the agent will attempt to submit logs. Then, when it sees that it needs to create the log group and log streams, it will do so.

However, because the different instances of the CloudWatch agent don't know about each other in the fleet, there can be a race condition where one agent creates the log group before the others attempt to, which results in subsequent calls to create the log group to fail. This gets emitted as an error log like `E! [outputs.cloudwatchlogs] creating group fail due to : ResourceAlreadyExistsException: The specified log group already exists`.

This creates unnecessary noise because the agent successfully created the log group from one of the hosts. 

The simplest solution would be to log that the agent successfully created a resource, however this doesn't solve the problem where we publish an error log when failing to create a log group that already exists. What we should do is make the error handling more intelligent such that when it encounters an error that the log group or log stream already exist, this is not a failure state (also don't log this as an error).

SaxyPandaBear:
# Description of the issue
Closes #415 

Currently, when we try to create a log stream and log group (typically when the agent first starts up and sends its first log), there is a potential race condition that can occur when two different instances of the agent attempt to create the same log group. In `pusher.go`, we reuse the same `error` variable, so in the scenario where:
1. Agent A tries to push to CWL and log group doesn't exist
2. Agent B tries to push to CWL and log group doesn't exist
3. Agent B creates log group
4. Agent A fails to create log group because it already exists

The agent log will be confusing, because there will be an instance where the log says the log group does not exist yet, and simultaneously there will be an error that says the log group already exists.

# Description of changes
1. Moved all of the log statements in the `createLogGroupAndStream()` function to debug level, so we don't parrot the same error information twice at error level
2. Added a check before returning, such that if the error we are about to return is due to `ResourceAlreadyExistsException`, we return nil instead, indicating no issue
3. Added debug log when attempting to update the log retention policy
4. Captures the error when failing to put the log retention policy when the log group doesn't exist yet, because that shouldn't be considered a failure state (error level log).
5. Added debug logs to show that the log group or stream was created successfully when no err was returned, or that the retention policy was updated successfully

# License

# Tests
Wasn't able to 100% reproduce the issue, but I think the logging updates are sufficient.
1. Set up a CloudFormation template that spawns a bunch of independent EC2 instances
2. Each EC2 instance downloads the RPM and config, and starts the agent
6. Each agent attempts to publish to the same log groups but different log streams

The relevant agent logs from the instance that started publishing first. Note that the other agent logs don't enter the race condition posited in the original issue, but I don't want to spend a lot of time trying to reproduce that exact scenario since this change is more overarching than that one situation.
```
2022-03-29T20:59:54Z D! [outputs.cloudwatchlogs] Log group MyTestLogGroup11111a not created yet: ResourceNotFoundException: The specified log group does not exist.
2022-03-29T20:59:54Z D! [outputs.cloudwatchlogs] Log group MyTestLogGroup22222a not created yet: ResourceNotFoundException: The specified log group does not exist.
2022-03-29T20:59:59Z D! [outputs.cloudwatchlogs] creating stream fail due to : ResourceNotFoundException: The specified log group does not exist.
2022-03-29T20:59:59Z D! [outputs.cloudwatchlogs] creating stream fail due to : ResourceNotFoundException: The specified log group does not exist.
2022-03-29T20:59:59Z D! [outputs.cloudwatchlogs] successfully created log group MyTestLogGroup11111a. Retrying log stream MyTestLogStream11111a
2022-03-29T20:59:59Z D! [outputs.cloudwatchlogs] successfully created log group MyTestLogGroup22222a. Retrying log stream i-0d30e8222e219283d-agent-log22222a
2022-03-29T20:59:59Z D! [outputs.cloudwatchlogs] successfully created log stream MyTestLogStream11111a
2022-03-29T20:59:59Z D! [outputs.cloudwatchlogs] successfully created log stream i-0d30e8222e219283d-agent-log22222a
2022-03-29T20:59:59Z D! [outputs.cloudwatchlogs] successfully updated log retention policy for log group MyTestLogGroup11111a
2022-03-29T20:59:59Z D! [outputs.cloudwatchlogs] successfully updated log retention policy for log group MyTestLogGroup22222a
```
The config file used:
```
{
    "agent": {
        "debug": true,
        "run_as_user": "root"
    },
    "logs": {
        "logs_collected": {
            "files": {
                "collect_list": [
                    {
                        "file_path": "/tmp/logfile.txt",
                        "log_group_name": "MyTestLogGroup11111a",
                        "log_stream_name": "MyTestLogStream11111a",
                        "retention_in_days": 1
                    },
                    {
                        "log_group_name": "MyTestLogGroup22222a",
                        "log_stream_name": "{instance_id}-agent-log22222a",
                        "retention_in_days": 1,
                        "filters": [
                            {
                                "type": "include",
                                "expression": "D!"
                            },
                            {
                                "type": "include",
                                "expression": "cloudwatchlogs"
                            },
                            {
                                "type": "exclude",
                                "expression": "Buffer fullness"
                            },
                            {
                                "type": "exclude",
                                "expression": "Retried"
                            },
                            {
                                "type": "exclude",
                                "expression": "Pusher published"
                            }
                        ]
                    }
                ]
            }
        }
    }
}
```

### DIFF
--- a/plugins/outputs/cloudwatchlogs/pusher.go
+++ b/plugins/outputs/cloudwatchlogs/pusher.go
@@ -344,9 +344,14 @@ func (p *pusher) createLogGroupAndStream() error {
 					LogStreamName: &p.Stream,
 				})
 			} else {
-				p.Log.Errorf("creating group fail due to : %v \n", err)
+				p.Log.Debugf("creating group fail due to : %v \n", err)
 			}
 		}
+	}
+
+	if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == cloudwatchlogs.ErrCodeResourceAlreadyExistsException {
+		p.Log.Debugf("Resource was already created. %v\n", err)
+		return nil // if the log group or log stream already exist, this is not worth returning an error for
 	}
 
 	return err
--- a/plugins/outputs/cloudwatchlogs/pusher.go
+++ b/plugins/outputs/cloudwatchlogs/pusher.go
@@ -330,27 +330,31 @@ func (p *pusher) createLogGroupAndStream() error {
 		LogStreamName: &p.Stream,
 	})
 
-	if err != nil {
-		p.Log.Debugf("creating stream fail due to : %v", err)
-		if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == cloudwatchlogs.ErrCodeResourceNotFoundException {
-			_, err = p.Service.CreateLogGroup(&cloudwatchlogs.CreateLogGroupInput{
-				LogGroupName: &p.Group,
-			})
+	if err == nil {
+		p.Log.Debugf("successfully created log stream %v", p.Stream)
+		return nil
+	}
 
-			// create stream again if group created successfully.
+	p.Log.Debugf("creating stream fail due to : %v", err)
+	if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == cloudwatchlogs.ErrCodeResourceNotFoundException {
+		_, err = p.Service.CreateLogGroup(&cloudwatchlogs.CreateLogGroupInput{
+			LogGroupName: &p.Group,
+		})
+
+		// attempt to create stream again if group created successfully.
+		if err == nil {
+			p.Log.Debugf("successfully created log group %v. Retrying log stream %v", p.Group, p.Stream)
+			_, err = p.Service.CreateLogStream(&cloudwatchlogs.CreateLogStreamInput{
+				LogGroupName:  &p.Group,
+				LogStreamName: &p.Stream,
+			})
 			if err == nil {
-				p.Log.Debugf("successfully created log group %v. Retrying log stream %v", p.Group, p.Stream)
-				_, err = p.Service.CreateLogStream(&cloudwatchlogs.CreateLogStreamInput{
-					LogGroupName:  &p.Group,
-					LogStreamName: &p.Stream,
-				})
-				if err == nil {
-					p.Log.Debugf("successfully created log stream %v", p.Stream)
-				}
-			} else {
-				p.Log.Debugf("creating group fail due to : %v", err)
+				p.Log.Debugf("successfully created log stream %v", p.Stream)
 			}
+		} else {
+			p.Log.Debugf("creating group fail due to : %v", err)
 		}
+
 	}
 
 	if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == cloudwatchlogs.ErrCodeResourceAlreadyExistsException {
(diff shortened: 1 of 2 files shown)


plugins/outputs/cloudwatchlogs/pusher.go
```diff
@@ -330,23 +330,36 @@ func (p *pusher) createLogGroupAndStream() error {
...
-				p.Log.Errorf("creating group fail due to : %v \n", err)
+				p.Log.Debugf("successfully created log stream %v", p.Stream)
 			}
+		} else {
+			p.Log.Debugf("creating group fail due to : %v", err)
 		}
+
+	}
+
+	if awsErr, ok := err.(awserr.Error); ok && awsErr.Code() == cloudwatchlogs.ErrCodeResourceAlreadyExistsException {
```
khanhntd, line 360: IMO, please put handling error `ErrCodeResourceAlreadyExistsException` before [`ErrCodeResourceNotFoundException`](https://github.com/aws/amazon-cloudwatch-agent/pull/416/files#diff-9a19b49be59e15932ccf98dbea45ca58f52d90639bbfef76e2cbc93dc646cb88R339)

SaxyPandaBear: Doing that is a bigger change. The existing code reuses the `err` variable, and attempts the following:
1. Try to create log stream
2. If it fails, check if it's because the log group doesn't exist yet (ResourceNotFound)
3. Then, try to create the log group
4. If it succeeds, then try to create the log stream again
I didn't change this flow.

Trying to check for `ResourceAlreadyExists` first would change it to:
1. Try to create log stream
2. If it fails because it already exist, return nil
3. Else, if it fails because the log group doesn't exist yet
4. Then try to create the log group
5. If it fails because it already exists, OR the log group was created successfully
6. Then try to create the log stream again

5 in the second scenario is why I didn't update the order to begin with, opting to keep the existing flow so when it fails through to this check, we can catch the error gracefully at the very end.

khanhntd: I see. I have not notice that you reuse the `err` variable for the second create `Log Stream`, which is to avoid the 5 condition. Agree with this changes.

plugins/outputs/cloudwatchlogs/pusher_test.go
```diff
@@ -790,7 +790,7 @@ func TestResendWouldStopAfterExhaustedRetries(t *testing.T) {
 	var logbuf bytes.Buffer
 	log.SetOutput(io.MultiWriter(&logbuf, os.Stdout))
 
-	stop, p := testPreparation(1, &s, 10*time.Millisecond, time.Second)
+	stop, p := testPreparation(-1, &s, 10*time.Millisecond, time.Second)
```
khanhntd, line 793: A question: why these `retentionInDays` were change from 1 to -1 (infinite) ?

aateeqi (approved):

PR #418, opened 2022-03-29
Move conflicting retention check to translator

aateeqi:
# Description of the issue

During customer call, the agent stopped without logging any errors to agent log when conflicting retentions are set and agent panics. 

# Description of changes

During config translation, checks for conflicting retention setting for the same log group

# License


plugins/inputs/logfile/logfile.go
```diff
@@ -412,7 +412,9 @@ func (t *LogFile) checkForDuplicateRetentionSettings() {
 		if configMap[logGroup] != 0 {
 			// different retentions has been set for the same log group, panic and stop the agent
 			if configMap[logGroup] != fileConfig.RetentionInDays {
-				panic(fmt.Sprintf("error: The Log Group has differing retentionInDays values defined in two or more places. Log Group Name: %v", fileConfig.LogGroupName))
+				msg := fmt.Sprintf("error: The Log Group %v has differing retentionInDays values defined in two or more places.", fileConfig.LogGroupName)
+				t.Log.Errorf(msg)
```
SaxyPandaBear, line 416: Nitpick 
```suggestion
				t.Log.Error(msg)
```

SaxyPandaBear:
Have you done any manual validation that this logs the error message we want in the agent log before panicking?

khanhntd:
Please add the corresponding config in the sample config: some of invalid config such as your case (set different log group's retention) and a valid one. Another thing would be having this testing as a schema validation phase since I would prefer fail fast over detecting in agent's log or journalctl's log to find this particular problem.

SaxyPandaBear:
@khanhntd While I agree that we should fail fast, I don't think we should remove the check from the agent code. **It should be duplicated**. I feel like I remember having a conversation about this in a previous PR but I looked through them and don't see it so it might be in Slack. 

The problem with removing the validation in the main agent code is that a customer can use a TOML file they create themselves, without passing a JSON through the translator. Because of this, we can get into an undefined/undesirable state where the customer's configuration does not match what we allow.

plugins/inputs/logfile/logfile.go
```diff
@@ -149,8 +149,6 @@ func (t *LogFile) FindLogSrc() []logs.LogSrc {
 
 	t.cleanUpStoppedTailerSrc()
 
-	// If a log group has differing retentionInDays values defined in multiple places, stop the agent
-	t.checkForDuplicateRetentionSettings()
```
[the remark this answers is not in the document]
aateeqi: @SaxyPandaBear Since the agent was not intended to be used by directly supplying the toml file, I don't think its a case we should necessarily support. Restarting the agent through an external source seems like something that would be more common and I do agree, without the agent code validation we wouldn't be protected from that. 

Currently when the agent gets restarted from an external source, wouldn't we be in danger from other invalid inputs?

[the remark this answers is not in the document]
SaxyPandaBear: > Currently when the agent gets restarted from an external source, wouldn't we be in danger from other invalid inputs as well?

Yes, and that's why knowing this, I don't want to add something that contributes to this issue.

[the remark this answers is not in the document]
adam-mateen: I would like to avoid duplicate code.
The typical and only supported usage for CloudWatch Agent is to give a JSON configuration and then run the translator to convert to TOML.
I do not think we should worry about customers modifying TOML directly.

Checking configuration validity during translation seems preferable to waiting until the Agent program begins.
It happens earlier and the error output can be returned directly to stdout/stderr instead of users having to check log files.

translator/translate/logs/logs_collected/files/collect_list/collect_list_test.go
```diff
@@ -611,3 +611,115 @@ func TestLogFilters(t *testing.T) {
 	}}
 	assert.Equal(t, expectVal, val)
 }
+
+func TestRetention(t *testing.T) {
```
adam-mateen, line 615: I recommend making the name more specific:
`func TestRetentionDifferentLogGroups`

translator/translate/logs/logs_collected/files/collect_list/collect_list_test.go
```diff
@@ -611,3 +611,115 @@ func TestLogFilters(t *testing.T) {
...
+			{
+				"file_path":"path1",
+       			"log_group_name":"test1",
+				"retention_in_days":5
+			}
+		]
+	}`), &input)
+	if e != nil {
+		assert.Fail(t, e.Error())
+	}
```
adam-mateen, line 708: Could change:
```
	if e != nil {
		assert.Fail(t, e.Error())
	}
```
to:
```
assert.Equal(e, WhateverTheExpectedErrorIs)
```
It would be more specific.

translator/translate/logs/logs_collected/windows_events/collect_list/collectlist_test.go
```diff
@@ -67,3 +68,143 @@ func TestApplyRule(t *testing.T) {
...
+
+	var actual interface{}
+
+	error := json.Unmarshal([]byte(rawJsonString), &input)
+	if error == nil {
+		_, actual = c.ApplyRule(input)
+		assert.Equal(t, 0, len(translator.ErrorMessages))
+		assert.Equal(t, expected, actual)
+	} else {
+		panic(error)
```
adam-mateen, line 147: Why use `panic()` in test code?
Why not just fail an assert?

translator/translate/logs/util/filter_retention.go
```diff
@@ -0,0 +1,40 @@
...
+			if retention, ok := logConfigMap["retention_in_days"].(int); ok {
+				if logGroup, ok := logConfigMap["log_group_name"].(string); ok {
+					// if retention is 0, -1 or less it's either invalid or default
+					logGroup = strings.ToLower(logGroup)
+					if retention < 1 {
+						continue
+					}
+					// if the configMap[logGroup] exists, retention has been set for the same logGroup somewhere
+					if configMap[logGroup] != 0 {
+						// different retentions has been set for the same log group, panic and stop the agent
```
adam-mateen, line 23: Update comment - Does not panic or stop the agent, right?

translator/translate/logs/util/filter_retention.go
```diff
@@ -0,0 +1,40 @@
...
+	for _, logConfig := range logConfigs {
+		if logConfigMap, ok := logConfig.(map[string]interface{}); ok {
+			// skip if retention is not set
+			if retention, ok := logConfigMap["retention_in_days"].(int); ok {
+				if logGroup, ok := logConfigMap["log_group_name"].(string); ok {
+					// if retention is 0, -1 or less it's either invalid or default
+					logGroup = strings.ToLower(logGroup)
+					if retention < 1 {
+						continue
+					}
```
adam-mateen, line 20: You could do:
```
if retention, ok := logConfigMap["retention_in_days"].(int); ok {
    if retention < 1 {
        continue
    }
    if logGroup, ok := logConfigMap["log_group_name"].(string); ok {
.
.
.
```
Move the retention value check immediately after getting the retention value.

translator/translate/logs/logs_collected/windows_events/collect_list/collectlist_test.go
SaxyPandaBear: There's a lot of test code like this. I was going to go through and clean up the test code so it doesn't panic. Personally I wouldn't block on this because I'm going to be cleaning up afterwards.

plugins/inputs/logfile/logfile.go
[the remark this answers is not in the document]
SaxyPandaBear: I'm on board if we're all in agreement that if a customer messes up their TOML file, that's completely on them. Is this documented in some way?

translator/translate/logs/logs_collected/windows_events/collect_list/collectlist_test.go
aateeqi: Yeah it was copy paste pasta but i'll change it in the new test cases and we can clean up the the previous code later like Andrew said.

plugins/inputs/logfile/logfile.go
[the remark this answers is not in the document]
aateeqi: I don't think we explicitly document that we convert Json to Toml (we do print this in the logs) and I think it makes sense. This use case is for customers who have understood how the CWagent flow is working internally but we shouldn't have to add extra support for use cases that we didn't design and test the agent for.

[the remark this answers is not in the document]
SaxyPandaBear: > output can be returned directly to stdout/stderr instead of users having to check log files

The output still shows up in log files, assuming that people use the agent through the ctl script. But either way, good enough for me.

translator/translate/logs/util/filter_retention.go
```diff
@@ -0,0 +1,40 @@
+package util
```
SaxyPandaBear, line 1: This is a nitpick, but I think that the name of the file is misleading. I think it would make sense to use a more directed verb, like `translator/translate/logs/util/validate_retention.go`

translator/translate/logs/util/filter_retention.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"fmt"
+	"github.com/aws/amazon-cloudwatch-agent/translator"
+	"strings"
+)
+
+func CheckForConflictingRetentionSettings(logConfigs []interface{}, currPath string) []interface{} {
```
SaxyPandaBear, line 9: Also a nitpick, but this name is very verbose. While on the one hand, I'd rather be verbose than ambiguous, I think that the name is really long and we could make the function name more succinct.

```suggestion
func ValidateLogRetentionSettings(logConfigs []interface{}, currPath string) []interface{} {
```

translator/translate/logs/util/filter_retention.go
```diff
@@ -0,0 +1,40 @@
...
+	configMap := make(map[string]int)
+	for _, logConfig := range logConfigs {
+		if logConfigMap, ok := logConfig.(map[string]interface{}); ok {
+			// skip if retention is not set
+			if retention, ok := logConfigMap["retention_in_days"].(int); ok {
```
SaxyPandaBear, line 14: Because I'm paranoid, where do the `retention_in_days` and `log_group_name` keys come from? If they come from a centralized, public set of constants in another file, I would prefer to reference those. If they're spread out, I still think we should at least pull these out into private scoped constants in this file, like:
```go
const (
    logRetentionKey = "retention_in_days"
    logGroupKey = "log_group_name"
)
```

adam-mateen (dismissed):

translator/translate/logs/util/filter_retention.go
```diff
@@ -0,0 +1,40 @@
...
+						// Retention for a log group has been configured in multiple places. Unset it so that the retention api is only called once
+						logConfigMap["retention_in_days"] = -1
+					} else {
+						configMap[logGroup] = retention
+					}
+				}
+			}
+		}
+	}
+	return logConfigs
```
SaxyPandaBear, line 38: 🤔 Does this function return the `logConfigs` parameter unmodified? I see above there is a line of code that sets the retention to the default for a given log config, but not sure if that propagates back up, outside of this function. I won't lose sleep over this, though.

SaxyPandaBear (commented):
Not sure what's wrong with GitHub for me, but this didn't publish

SaxyPandaBear (approved):
At the end of the day, it looks fine to merge since what I'm asking for is mostly cosmetic. If you want to merge, go for it. If you want to update based on my comments, I'll keep an eye out for a new revision and will happily approve that

translator/translate/logs/util/filter_retention.go
aateeqi: Yeah this is confirmed through running the unit tests

translator/translate/logs/util/validate_retention.go
```diff
@@ -0,0 +1,44 @@
...
+					// if the configMap[logGroup] exists, retention config for the log group was already included earlier
+					if configMap[logGroup] != 0 {
+						// different retentions are attempted to be configured for the same log group, add error message to fail translation
+						if configMap[logGroup] != retention {
+							translator.AddErrorMessages(
+								currPath,
+								fmt.Sprintf("Different retention_in_days values can't be set for the same log group: %v", logGroup))
+						}
+						// Retention for a log group has been configured in multiple places. Unset it so that the retention api is only called once
+						logConfigMap["retention_in_days"] = -1
```
SaxyPandaBear, line 35: Missed this reference to the retention key

SaxyPandaBear (approved):

translator/translate/logs/util/validate_retention.go
aateeqi: 🤦

PR #421, opened 2022-03-30
Add writing to CWAgent's log before a panic

khanhntd:
# Description of the issue
Currently, panic() does not write log message to CWAgent's log. Therefore, add a `log.Printf` before a panic based on these criteria
* As long as it does not involve in test since I believe it would be better to use exit for [short term termination](https://stackoverflow.com/a/28473339)
* Has a `sprintf` or `printf` before a panic.
* Does not involve with `internal` package since most encountered errors has been at surface levels instead of internal package and avoid having unnecessary logs.  

# License

# Tests
Use `make test` to test the changes. After making some test with Linux, Mac and Windows, they are having the same behavior:
* Without using recover panic
image

* With using recover for panic
image


cmd/config-downloader/downloader.go
```diff
@@ -125,13 +125,15 @@ func main() {
 	if inputConfig != "" {
 		f, err := os.Open(inputConfig)
 		if err != nil {
-			fmt.Printf("E! Failed to open Common Config: %v\n", err)
-			panic(err)
+			errorMessage := fmt.Sprintf("E! Failed to open Common Config: %v\n", err)
+			log.Printf(errorMessage)
```
SaxyPandaBear, line 129: So, two things.
1. I was reading up on the differences between `fmt.Print()` and `log.Print()` and from what I saw online, by default, the `log` package writes to stderr whereas `fmt` writes to stdout. **This would result in a change in behavior**. 
2. If we determine it's okay to accept the above change in behavior, this could be simplified to 
```suggestion
			log.Panicf("E! Failed to open Common Config: %v\n", err)
```

cmd/config-downloader/downloader.go
```diff
@@ -154,11 +157,12 @@ func main() {
 	if region == "" && downloadLocation != locationDefault {
 		fmt.Println("Unable to determine aws-region.")
 		if mode == config.ModeEC2 {
-			errorMessage = fmt.Sprintf("Please check if you can access the metadata service. For exampe, on linux, run 'wget -q -O - http://169.254.169.254/latest/meta-data/instance-id && echo' ")
+			errorMessage = fmt.Sprintf("E! Please check if you can access the metadata service. For example, on linux, run 'wget -q -O - http://169.254.169.254/latest/meta-data/instance-id && echo' ")
```
SaxyPandaBear, line 160: I don't think that the downloader executable points to the telegraf logger in any way, so I don't think that the `E!` notation helps much outside of letting us know semantically that this is an "error". Just calling that out. It's not a breaking change so I don't mind keeping it.

cmd/config-translator/translator_test.go
```diff
@@ -19,14 +19,14 @@ import (
 func checkIfSchemaValidateAsExpected(t *testing.T, jsonInputPath string, shouldSuccess bool, expectedErrorMap map[string]int) {
 	actualErrorMap := make(map[string]int)
 
-	jsonInputMap, err := util.GetJsonMapFromFile(jsonInputPath)
-	if err != nil {
-		t.Fatalf("Failed to get json map from %v with error: %v", jsonInputPath, err)
+	jsonInputMap, e := util.GetJsonMapFromFile(jsonInputPath)
+	if e != nil {
+		t.Fatalf("Failed to get json map from %v with error: %v", jsonInputPath, e)
```
SaxyPandaBear, line 24: This can be simplified to:
```suggestion
	assert.NoError(t, e)
```

There is a lot of code like this and I wanted to clean up our test code myself, so I'm fine with leaving this as is in your PR because I'll just update it in mine.

internal/containerinsightscommon/nodeCapacity.go
```diff
@@ -18,7 +18,7 @@ type NodeCapacity struct {
 
 func NewNodeCapacity() *NodeCapacity {
 	if _, err := os.Lstat("/rootfs/proc"); os.IsNotExist(err) {
-		panic("/rootfs/proc doesn't exists")
+		panic("E! /rootfs/proc doesn't exists")
```
SaxyPandaBear, line 21: How is this error getting logged? This just adds the telegraf log error prefix but I don't think that this alone writes to a log about the error.

translator/config/schema_test.go
```diff
@@ -12,9 +12,9 @@ import (
 )
 
 func TestGetJsonSchema(t *testing.T) {
-	jsonFile, err := ioutil.ReadFile("./schema.json")
-	if err != nil {
-		panic(err)
+	jsonFile, e := ioutil.ReadFile("./schema.json")
+	if e != nil {
+		panic(e)
```
SaxyPandaBear, line 17: This shouldn't panic, but I'm going to clean up our test code so you don't have to change it.
```suggestion
	assert.NoError(t, e)
```

translator/translate/metrics/append_dimensions/append_dimensions_test.go
```diff
@@ -11,19 +11,19 @@ import (
 )
 
 func TestAppendDimensions(t *testing.T) {
-	e := new(appendDimensions)
+	a := new(appendDimensions)
 	//Check whether override default config
 	var input interface{}
-	err := json.Unmarshal([]byte(`{
+	e := json.Unmarshal([]byte(`{
```
SaxyPandaBear, line 17: This feels like an unnecessary change. Why did you change `err` to `e`? This caused you to change the variable name of the dimension.

cmd/config-translator/translator_test.go
```diff
@@ -19,14 +19,14 @@ import (
 func checkIfSchemaValidateAsExpected(t *testing.T, jsonInputPath string, shouldSuccess bool, expectedErrorMap map[string]int) {
 	actualErrorMap := make(map[string]int)
 
-	jsonInputMap, err := util.GetJsonMapFromFile(jsonInputPath)
-	if err != nil {
-		t.Fatalf("Failed to get json map from %v with error: %v", jsonInputPath, err)
+	jsonInputMap, e := util.GetJsonMapFromFile(jsonInputPath)
```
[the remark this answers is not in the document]
khanhntd: It's for differentiate between error in usual function and in testing function. However, most of others companies using golang would use `err` in their function instead. That's why I need third party opinion a little bit.

translator/config/schema_test.go
khanhntd: Agree. Thought of doing in another PR but prefer left it as it is.

translator/translate/metrics/append_dimensions/append_dimensions_test.go
khanhntd: It's for differentiate between error in usual function and in testing function. However, most of others companies using golang would use err in their function instead. That's why I need third party opinion a little bit.

cmd/config-downloader/downloader.go
khanhntd: Agree on this. Just my habit of grouping things to error.

plugins/inputs/logfile/tail/watch/polling.go
```diff
@@ -39,7 +40,9 @@ func (fw *PollingFileWatcher) BlockUntilExists(t *tomb.Tomb) error {
 			return tomb.ErrDying
 		}
 	}
-	panic("unreachable")
+	errorMessage := fmt.Sprintf("E! Unreachable to file %s with PollingFileWatcher",fw.Filename)
```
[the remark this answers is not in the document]
khanhntd: Agree. Thought deleting panics would do some breaking changes so keeping their.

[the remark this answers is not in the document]
khanhntd: Agree. Thought deleting panics would do some breaking changes so keeping their.

internal/containerinsightscommon/nodeCapacity.go
khanhntd: That is one of the criteria I mentioned: `most encountered errors has been at surface levels instead of internal package and avoid having unnecessary logs.`. However, if it is needed, I would add logs for this as well.

plugins/inputs/logfile/tail/tail.go
```diff
@@ -593,7 +596,7 @@ func (tail *Tail) exitOnDeletion() {
 // with the last chunk of variable size.
 func partitionString(s string, chunkSize int) []string {
 	if chunkSize <= 0 {
-		panic("invalid chunkSize")
+		panic("Invalid chunkSize")
```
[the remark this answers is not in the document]
khanhntd: It does not write log before panicking. Same criteria that I mentioned before: `most encountered errors has been at surface levels instead of internal package and avoid having unnecessary logs`. However, if it is needed, I would add writing to logs for this as well

plugins/inputs/logfile/tail/tail.go
```diff
@@ -465,7 +465,10 @@ func (tail *Tail) waitForChanges() error {
 	case <-tail.Dying():
 		return ErrStop
 	}
-	panic("unreachable")
+	errorMessage := fmt.Sprintf("E! Unreachable to file %s",tail.Filename)
```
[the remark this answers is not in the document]
khanhntd: Agree. Thought deleting panics would do some breaking changes so keeping their.

cmd/config-translator/translator_test.go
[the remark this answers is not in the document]
SaxyPandaBear: I don't understand what the concern of having it referred to as `err` is. This is in a test so there's no reason anyone other than us would use this.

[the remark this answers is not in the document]
khanhntd: I'm good with either so it should not be hard to change it back.

cfg/aws/refreshable_shared_credentials_provider.go
```diff
@@ -22,7 +22,7 @@ type Refreshable_shared_credentials_provider struct {
 func (p *Refreshable_shared_credentials_provider) Retrieve() (credentials.Value, error) {
 
 	p.SetExpiration(time.Now().Add(p.ExpiryWindow), 0)
-	creds, e := p.sharedCredentialsProvider.Retrieve()
+	creds, err := p.sharedCredentialsProvider.Retrieve()
```
SaxyPandaBear, line 25: I'm okay with this, but if we're going to be updating a lot of variable names just because we want to change the name of the variables, it would help if that was in a separate PR. This PR already has a lot of changes because of all of the panic logging we are adding. Including cosmetic changes to unrelated source code for variable names makes this harder to review

cmd/config-downloader/downloader.go
```diff
@@ -125,26 +125,24 @@ func main() {
 	if inputConfig != "" {
 		f, err := os.Open(inputConfig)
 		if err != nil {
-			fmt.Printf("E! Failed to open Common Config: %v\n", err)
-			panic(err)
+			log.Panicf("E! Failed to open Common Config: %v", err)
```
SaxyPandaBear, line 128: Did you confirm on all major OS types that these logs are visible in the agent log even though they are going to stderr instead of stdout?

cmd/config-downloader/downloader.go
```diff
@@ -187,7 +185,8 @@ func main() {
 
 	locationArray := strings.SplitN(downloadLocation, locationSeparator, 2)
 	if locationArray == nil || len(locationArray) < 2 && downloadLocation != locationDefault {
-		panic(fmt.Sprintf("downloadLocation %s is malformated.\n", downloadLocation))
+		errorMessage := fmt.Sprintf("E! downloadLocation %s is malformated.", downloadLocation)
+		log.Panicf(errorMessage)
```
SaxyPandaBear, line 189: If we're using `Panicf`, we don't need to use `fmt.Sprintf` to create the error message beforehand.

cmd/config-translator/translator.go
```diff
@@ -90,16 +89,16 @@ func main() {
...
 	if !ctx.RunInContainer() {
 		// run as user only applies to non container situation.
-		current, e := user.Current()
-		if e == nil && current.Name == "root" {
+		current, err := user.Current()
+		if err == nil && current.Name == "root" {
 			runAsUser, err := cmdutil.DetectRunAsUser(mergedJsonConfigMap)
 			if err != nil {
-				panic("E! Failed to detectRunAsUser\n")
+				log.Panicf("E! Failed to detectRunAsUser")
```
SaxyPandaBear, line 101: If we aren't passing in any other parameters for the format string, this could just be:
```suggestion
				log.Panic("E! Failed to detectRunAsUser")
```

plugins/inputs/logfile/tail/tail.go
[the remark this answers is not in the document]
SaxyPandaBear: Why do we not want to write any log for this if it panics? Are you saying that telegraf is catching this before terminating?

plugins/inputs/windows_event_log/wineventlog/utils_test.go
```diff
@@ -46,15 +46,15 @@ func TestDecodingWithDifferentBufferUsedReturned(t *testing.T) {
...
 	data, _ := hex.DecodeString(originalHexStr)
-	bufferUsed := len(data)/2
+	bufferUsed := len(data) / 2
```
SaxyPandaBear, line 49: Why is go fmt inconsistent when doing operations? I noticed that when formatting, it removes spaces around `+` but here, it added spaces around `/`

translator/cmdutil/translatorutil.go
```diff
@@ -101,16 +101,16 @@ func RunSchemaValidation(inputJsonMap map[string]interface{}) (*gojsonschema.Res
 func checkSchema(inputJsonMap map[string]interface{}) {
 	result, err := RunSchemaValidation(inputJsonMap)
 	if err != nil {
-		panic(err.Error())
+		log.Panicf("E! Failed to run schema validation because of %v", err)
 	}
 	if result.Valid() {
-		fmt.Println("Valid Json input schema.")
+		log.Print("I! Valid Json input schema.")
```
SaxyPandaBear, line 107: Why did this get changed to a log? This introduces more inconsistencies. We use `fmt.Print` for a bunch of things to write to stdout for normal success paths, but this changes this specific function to still log to stderr

translator/config/os.go
```diff
@@ -22,11 +22,14 @@ func ToValidOs(os string) string {
...
 	formattedOs := strings.ToLower(os)
 	for _, val := range supportedOs {
 		if formattedOs == val {
 			return formattedOs
 		}
 	}
-	panic(fmt.Sprintf("%v is not a supported osType type", os))
+
+	log.Panicf("E! %v is not a supported osType type", os)
+	return ""
```
SaxyPandaBear, line 34: This is unreachable. Frankly, I dislike this because the logic here is fragile. I think that this function should not panic at all. It should either:
1. return an empty string and an error if the input is not a valid OS
2. just return an empty string and code that uses it needs to know to check for the empty string when using the return value from this function.

If that's too much to add to this PR, I would revert the changes to this function and address it separately.

translator/totomlconfig/toTomlConfig.go
```diff
@@ -58,9 +58,9 @@ func ToTomlConfig(c interface{}) string {
 	_, val := r.ApplyRule(c)
 	buf := bytes.Buffer{}
 	enc := toml.NewEncoder(&buf)
-	e := enc.Encode(val)
-	if e != nil {
-		panic(e)
+	err := enc.Encode(val)
+	if err != nil {
+		panic(err)
```
SaxyPandaBear, line 63: Why don't we log this err?

translator/translate/logs/logs_collected/windows_events/collect_list/collectlist_test.go
```diff
@@ -59,11 +59,11 @@ func TestApplyRule(t *testing.T) {
 
 	var actual interface{}
 
-	error := json.Unmarshal([]byte(rawJsonString), &input)
-	if error == nil {
+	e := json.Unmarshal([]byte(rawJsonString), &input)
```
SaxyPandaBear, line 62: This is not consistent with all of the other name changing. Why is this not `err`?

translator/translate/metrics/append_dimensions/append_dimensions_test.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 func TestAppendDimensions(t *testing.T) {
-	e := new(appendDimensions)
+	a := new(appendDimensions)
```
SaxyPandaBear, line 14: The changes to this file can be reverted.

translator/translate/metrics/drop_origin/drop_origin.go
```diff
@@ -36,7 +36,7 @@ func (do *dropOrigin) ApplyRule(input interface{}) (returnKey string, returnVal
 						if reflect.TypeOf(value).String() == "string" {
 							droppingDimensions = append(droppingDimensions, value.(string))
 						} else {
-							log.Panic("Fail to translate the JSON config, invalid format of dropping dimensions.")
+							log.Panicf("Fail to translate the JSON config, invalid format of dropping dimensions.")
```
SaxyPandaBear, line 39: Why does this need `Panicf`?

translator/translate/metrics/metrics_collect/gpu/nvidiaSmi_test.go
```diff
@@ -43,7 +43,7 @@ func TestNoFieldConfig(t *testing.T) {
 func TestFullConfig(t *testing.T) {
 	n := new(NvidiaSmi)
 	var input interface{}
-	err := json.Unmarshal([]byte(`{"nvidia_gpu":{"measurement": [
+	e := json.Unmarshal([]byte(`{"nvidia_gpu":{"measurement": [
```
SaxyPandaBear, line 46: Why is this changed to `e`? It's not consistent.

translator/translate/metrics/rollup_dimensions/rollupDimensions_test.go
```diff
@@ -13,13 +13,13 @@ import (
 )
 
 func TestRollupDimensions(t *testing.T) {
-	e := new(rollupDimensions)
+	r := new(rollupDimensions)
```
SaxyPandaBear, line 16: This change can be reverted

SaxyPandaBear:
Also it would be great to share any manual testing results with the built agent executables with this new change in behavior to use the `log` package instead of `fmt` for writing out our logging. I am a little concerned about what will happen with where the logs go. I think a requirement before approving is to take one of the code paths that leads to a `log.Panic` call and reproduce it on Darwin, Linux and Windows in order to confirm that the log shows up in the same place as where it used to before your changes

cmd/config-translator/translator.go
khanhntd: Have not notice that. I used `log.Panic` for some but not noticing this until now. Nice catch

translator/translate/logs/logs_collected/windows_events/collect_list/collectlist_test.go
khanhntd: Fixed it in the newest commit. Nice catch on these though.

translator/translate/metrics/append_dimensions/append_dimensions_test.go
khanhntd: Also fixed it in the newest commit. Another nice catch on these though.

translator/config/os.go
khanhntd: Agree on fixing it in a separate PR and fixed it in the newest commit.

translator/translate/agent/ruleLogFile.go
```diff
@@ -33,7 +33,8 @@ func getDefaultValue() string {
 	case config.OS_TYPE_WINDOWS:
 		return util.GetWindowsProgramDataPath() + "\\Amazon\\AmazonCloudWatchAgent\\Logs\\amazon-cloudwatch-agent.log"
 	default:
-		panic(fmt.Sprintf("Unsupported platform %v for logRule", targetPlatform))
+		log.Panicf("Unsupported platform %v for logRule", targetPlatform)
+		return ""
```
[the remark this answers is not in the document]
khanhntd: Agree on the aforementioned solution and will fix it in a separate PR and fixed it in the newest commit.

cmd/config-downloader/downloader.go
khanhntd: Fixed it in the newest commit. Nice catch on these though.

translator/totomlconfig/toTomlConfig.go
khanhntd: Cause I thought [this log](https://github.com/aws/amazon-cloudwatch-agent/blob/master/translator/cmdutil/translatorutil.go#L45-L47) would be enough for us to track down  the issue. However, it could be from `BurntSushi` itself so reconsider to add the error.

translator/cmdutil/translatorutil.go
khanhntd: Mostly changing this would signal us that CWAgent translated successfully and moving on to other executables in the log. I am considering changing most of the fmt to `log.Print` since some could help us in tracking down the issues though.

cmd/config-downloader/downloader.go
SaxyPandaBear: Just want to confirm this before approving

SaxyPandaBear (approved):

PR #434, opened 2022-04-08
Deleting ssm parameter store after 30 days

khanhntd:
# Description of the issue
Instead of letting the integration test do the cleaning, utilizing the [aws-go-sdk-v2](https://github.com/aws/aws-sdk-go-v2) to clean the SSM parameter store to avoid throttling limit for AWS account.

# Description of changes
Add cleaning 

# License

# Tests
Using `make integration-cleaner` to confirm the result


integration/clean/clean_aws_resource.go
```diff
@@ -0,0 +1,145 @@
...
+
+const (
+	daysToKeep   = 30
+	keepDuration = -1 * time.Hour * 24 * time.Duration(daysToKeep)
+)
+
+var expirationDate = time.Now().UTC().Add(keepDuration)
+
+func main() {
+	log.Println("Begin to clean EC2 AMI")
```
SaxyPandaBear, line 31: I think that if we wanted to print when we're starting an action, you could just put it at the beginning of the `cleanAMI()` function itself. Otherwise, as we add more things to clean up here, we are forced to add more prints to this main function.

integration/clean/clean_aws_resource.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+	ec2Type "github.com/aws/aws-sdk-go-v2/service/ec2/types"
+	"github.com/aws/aws-sdk-go-v2/service/ssm"
+	ssmType "github.com/aws/aws-sdk-go-v2/service/ssm/types"
+	smithyTime "github.com/aws/smithy-go/time"
+)
+
+const (
+	daysToKeep   = 30
```
SaxyPandaBear, line 24: What is the reason for going from 60 to 30? I'm mostly fine with it. I just wanted to understand why the change.

integration/clean/clean_aws_resource.go
```diff
@@ -0,0 +1,145 @@
...
+
+func main() {
+	log.Println("Begin to clean EC2 AMI")
+	cleanAMI()
+
+	log.Println("Begin to clean SSM Parameter Store")
+	cleanSSMParameterStore()
+
+	log.Println("Finished cleaning resources.")
+}
```
SaxyPandaBear, line 38: On the one hand, I see value in cleanly exiting successfully always. On the other hand, what this does is it swallows issues that we would never notice because the job always succeeds. Without a notification in place, this doesn't matter either way. 

Personally, my opinion is that we should retain the existing behavior where we failed the job when there is an error, and we can revisit that when we work towards implementing some notification mechanism so we know that there is an issue with the job and someone needs to fix it.

integration/clean/clean_aws_resource.go
```diff
@@ -0,0 +1,145 @@
...
+	ssmType "github.com/aws/aws-sdk-go-v2/service/ssm/types"
+	smithyTime "github.com/aws/smithy-go/time"
+)
+
+const (
+	daysToKeep   = 30
+	keepDuration = -1 * time.Hour * 24 * time.Duration(daysToKeep)
+)
+
+var expirationDate = time.Now().UTC().Add(keepDuration)
```
SaxyPandaBear, line 28: Rather than use a global here, can you instantiate the expiration date inside of the `main` function, and then pass it in as an argument to the functions it needs to go to?

integration/clean/clean_aws_resource.go
```diff
@@ -0,0 +1,145 @@
...
+				continue
+			}
+
+			log.Printf("Trying to delete Parameter Store with name %s and creation date %v", *parameter.Name, *parameter.LastModifiedDate)
+
+			deleteParameterInput := ssm.DeleteParameterInput{Name: parameter.Name}
+
+			if _, err := ssmClient.DeleteParameter(ctx, &deleteParameterInput); err != nil {
+				log.Printf("Failed to delete Parameter Store with name %s because of %v", *parameter.Name, err)
+				return
```
SaxyPandaBear, line 86: My two issues with this are that:
1. Along the lines of what I said above, until we have a way to consistently alert us that there is an issue with this GitHub action, we should be capturing errors and failing the job.
2. This behavior is inconsistent even with the changes made in the AMI cleaning function. In that one, you kept the errors array, but then at the end check the length of the array to log something. 
Please update this to return an error array like it was originally

integration/clean/clean_aws_resource.go
```diff
@@ -0,0 +1,145 @@
...
+			_, err := ec2client.DeregisterImage(ctx, &deregisterImageInput)
+			if err != nil {
+				errors = append(errors, err)
+			}
+		}
+	}
+
+	if len(errors) != 0 {
+		log.Printf("Deleted some of AMIs failed because of %v", err)
+		return
```
SaxyPandaBear, line 143: This return statement doesn't add value. Whether the errors array is empty or not, we still return after this anyway.

integration/clean/clean_aws_resource.go
SaxyPandaBear: The reason I didn't ask for this in the original PR was because it looked like we were only going to be cleaning AMIs. Since you're extending the functionality of the script, I figured now would be a good time to parameterize rather than having a global value that gets instantiated when the main gets invoked.

integration/clean/clean_aws_resource.go
khanhntd: I'm good with either way since it serves the same purpose to me.

integration/clean/clean_aws_resource.go
khanhntd: Every time we do a release, it would rather spin up the new AMI, relevant resources,... However, these release could mostly takes up to 2 - 5 weeks. Therefore, reduce to delete resources more frequently.

integration/clean/clean_aws_resource.go
khanhntd: True though. I intend to separate the go files and parameterize if we intend to extends the function more, same as `aws-otel-collector` does. However, I'm good with doing now

integration/clean/clean_aws_resource.go
khanhntd: I prefer the former since it would be good for cleaning all the resources without having to hold back. However, if we prefer to fix ASAP when the errors occur, I'm good with that too.

integration/clean/clean_aws_resource.go
khanhntd: Fixed the inconsistency between cleaning resources's behavior. Nice catch on this

integration/clean/clean_aws_resource.go
khanhntd: I added the `log.Printf` after this `len(errors)` and changes back to `log.Fatalf` when the cleaning encouters errors.

integration/clean/clean_aws_resource.go
SaxyPandaBear: Can you elaborate? What I'm saying is that regardless of errors or not, currently (even with this change), we are slow to react to issues with this GitHub action. By making it succeed even when there are errors with deleting resources, it makes it harder for us to tell that something is wrong.

integration/clean/clean_aws_resource.go
```diff
@@ -0,0 +1,158 @@
...
+		if describeParametersOutput.NextToken == nil {
+			break
+		}
+
+		nextToken = describeParametersOutput.NextToken
+	}
+
+	if len(errors) != 0 {
+		log.Fatalf("Deleted some of Parameter Store failed because of %v", err)
+		return
```
SaxyPandaBear, line 100: There's no way to reach this `return` statement if you are calling `log.Fatalf()` above it.

integration/clean/clean_aws_resource.go
```diff
@@ -0,0 +1,158 @@
...
+
+func cleanSSMParameterStore(expirationDate time.Time) {
+	log.Println("Begin to clean SSM Parameter Store")
+
+	ctx := context.Background()
+	defaultConfig, err := config.LoadDefaultConfig(ctx)
+
+	if err != nil {
+		log.Printf("Load default config failed because of %v", err)
+		return
```
SaxyPandaBear, line 47: I'm confused. Further down, we keep track of a list of errors. If there are any errors, we fail the execution. Why does this not fail the execution as well?

khanhntd: Forgot to change these lines. However, fixed it in the newest commit with the current failing behavior.

integration/clean/clean_aws_resource.go
khanhntd: To be more specific, failing a job early would  make us tackle handle errors frequently. However, the resources for release's integration test are not created frequently, only in a short amount of time and the release happens every 3 months. On the other hand, if we shallow the errors, they would shrimp down the successes `deleted resources` and during the next month, we only need to encounter the errors though. So the question was: do we want to tackle the errors only once a month(infrequently) or tacking the errors immediately? Giving the circumstances of release one a month and shrimping down the resources to delete one a month and tackle only that time, I would prefer the *shallow errors* way.  However, if we prefer to tackle immediately, I am good by that too.

integration/clean/clean_aws_resource.go
```diff
@@ -0,0 +1,154 @@
...
+	ctx := context.Background()
+	defaultConfig, err := config.LoadDefaultConfig(ctx)
+
+	if err != nil {
+		log.Fatalf("Load default config failed because of %v", err)
+	}
+
+	ssmClient := ssm.NewFromConfig(defaultConfig)
+
+	//Allow to load all th since the default respond is paginated auto scaling groups.
```
ZhenyuTan-amz, line 51: Typo?

khanhntd: Yes. My fault for this typo. Fixed it with the newest commit

integration/clean/clean_aws_resource.go
```diff
@@ -0,0 +1,154 @@
...
+		
+		deregisterImageInput := ec2.DeregisterImageInput{ImageId: image.ImageId}
+		
+		if _, err := ec2client.DeregisterImage(ctx, &deregisterImageInput); err != nil {
+			errors = append(errors, err)
+		}
+	}
+
+	if len(errors) != 0 {
+		log.Fatalf("Deleted some of AMIs failed because of %v", err)
```
SaxyPandaBear, line 150: Same as above, and I don't think I caught this in the original review of that PR. `err` here is unhelpful because it refers to creating the context. Also if we get here, err will always be nil.

integration/clean/clean_aws_resource.go
```diff
@@ -0,0 +1,154 @@
...
+
+		if describeParametersOutput.NextToken == nil {
+			break
+		}
+
+		nextToken = describeParametersOutput.NextToken
+	}
+
+	if len(errors) != 0 {
+		log.Fatalf("Deleted some of Parameter Store failed because of %v", err)
```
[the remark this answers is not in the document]
khanhntd: Make sense. However, I have formatted the cleaner package and changed the behavior to `err` for all the cleaner package.

integration/clean/clean_aws_resource.go
khanhntd: Make sense. However, I have formatted the cleaner package and changed the behavior to err for all the cleaner package.

sethAmazon:
How does the action run this?

khanhntd:
> How does the action run this?

Forgot the Github Action. Added the new github workflow @sethAmazon

integration/clean/dedicated_host/clean.go
```diff
@@ -1,54 +1,45 @@
...
-	}
-}
+const (
+	Type     = "dedicated_host"
+	tagName  = "tag:Name"
+	tagValue = "IntegrationTestMacDedicatedHost"
+)
 
-func cleanDedicatedHost() error {
+func Clean(ctx context.Context, expirationDate time.Time) error {
```
SaxyPandaBear, line 24: A point of code cleanliness, I think, would be to define an interface for resource cleaners and have each of these different cleaners implement the interface, though that might be overkill.

```go
type Cleaner interface {
    Clean(ctx context.Context, expirationDate time.Time) error
    Type() string
}
```

integration/clean/main.go
```diff
@@ -0,0 +1,76 @@
...
+	"strings"
+	"time"
+
+	"github.com/aws/amazon-cloudwatch-agent/integration/clean/ami"
+	"github.com/aws/amazon-cloudwatch-agent/integration/clean/dedicated_host"
+	"github.com/aws/amazon-cloudwatch-agent/integration/clean/ssm"
+)
+
+const (
+	defaultDaysToKeep = 5
```
SaxyPandaBear, line 20: This default value does not line up with the default value in the GitHub action. I'm indifferent as to which one we use, and ultimately the action value will overwrite this every time, but I'd rather have them match.

integration/clean/main.go
```diff
@@ -0,0 +1,76 @@
...
+		case dedicated_host.Type:
+			if err := dedicated_host.Clean(ctx, expirationDate); err != nil {
+				log.Printf("%v", err)
+			}
+		case ssm.Type:
+			if err := ssm.Clean(ctx, expirationDate); err != nil {
+				log.Printf("%v", err)
+			}
+		default:
+			log.Printf("Skipping invalid cleaner '%s'. Please see -h for options.", cleaner)
```
SaxyPandaBear, line 71: Do we expect anyone to run this main function themselves outside of the GitHub action? This is a nitpick, but I don't think referencing `-h` adds much value unless we expect people to build and run this code locally.

integration/clean/dedicated_host/clean.go
SaxyPandaBear: Although it would look elegant, I don't think this is worthwhile. I'm still leaving it here though.

integration/clean/main.go
```diff
@@ -0,0 +1,76 @@
...
+	if strings.Contains(cleanersToRun, cleanerTypeAll) {
+		cleanersToRun = cleanerOptions
+	}
+
+	// The date used to determine if a resource can be cleaned.
+	// Anything older than the date will be removed.
+	expirationDate := time.Now().UTC().Add(keepDuration)
+	log.Printf("Expiration date set to %v", expirationDate)
+
+	for _, cleaner := range strings.Split(cleanersToRun, delimiter) {
```
SaxyPandaBear, line 56: I will have to think about this more. 

I am not sure I understand the value of consolidating all of this code into one main that executes the different cleaners. Also, you noted in the PR description `to avoid throttling limit for AWS account`, but I can't gleam from this PR how this addresses throttling limits.

integration/clean/main.go
khanhntd: I would especially when we are developing integration test which created local resources on their own account (Seth has done this before)

.github/workflows/clean-aws-resources.yml
```diff
@@ -0,0 +1,46 @@
...
+
+env:
+  DAYS_TO_KEEP: 30
+
+jobs:
+  clean-resources:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      max-parallel: 3
```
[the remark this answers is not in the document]
khanhntd: Below 3 is fine since there would be a [limit to concurrent job](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration#usage-limits) across the whole workflow instead of a single workflow

integration/clean/main.go
khanhntd: Initially I thought this would address the **throttling limit for the internal integration test**; however, changing the purpose to reduce the billing and also avoid reaching the quota limit for each corresponding service.

integration/clean/main.go
khanhntd: Agree. It would better to consistence with the Github Workflow. Changed it in the newest commit.

.github/workflows/clean-aws-resources.yml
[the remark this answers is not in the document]
SaxyPandaBear: > its better to prioritize the integration test

Can you explain how limiting concurrency on job executions for this helps to prioritize integration tests? I'm trying to understand what we benefit from setting a max parallelism on this job.

integration/clean/main.go
SaxyPandaBear: Can you update the README for this at `integration/clean/README.md` to reflect the changes you are introducing to it?

integration/clean/dedicated_host/clean.go
khanhntd: I thought of it but agree on its overkill a little bit and we don't maintain this much.

integration/clean/main.go
khanhntd: Yes nice catch. Have changed the READ ME and how to use the code with the variables.

.github/workflows/clean-aws-resources.yml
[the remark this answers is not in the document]
khanhntd: For example: if we have a **Free Github Plan** which only has 20 concurrent jobs, and assuming we will release CWAgent by using the GitHub workflow, there will be 2 scenarios:
*  If cleaner runs before CWAgent release, the `max-parallel`, which is 3, allows the cleaner to run 3 concurrent job. However, it would reduce the CWAgent release 's integration test to 17.
* If CWAgent release runs before cleaner, the cleaner would consume the remaining concurrent jobs which are allowed in the entire workflow, which could be 1, 2, or 3. 
Hope that make sense.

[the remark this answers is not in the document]
adam-mateen: If you remove the limit, what is the worst case?

Do we really need to be concerned about hitting the concurrent job limit?

Even if the limit is reached, won't the next job just wait for the next available "runner"?
(So worst case is just a delay)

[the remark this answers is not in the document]
khanhntd: Worst case for this cleaning you are right though.

integration/clean/README.md
```diff
@@ -7,12 +7,22 @@ The cleaner first searches for ami names (these are the ami created by the pipel
 
 Then checks to see if the creation date is greater than 60 days. (The aws sdk v2 gives creation date as a pointer to string. To convert to golang time we use the aws smithy go time. This allows us to compare to 60 days in past time)
 
-If the ami is older than 60 days old then we delete the ami
+If the ami is older than 30 days old as default (can be configured) then we delete the ami
```
SaxyPandaBear, line 10: If we're going to indicate that this is configurable, we should say how it is configured.

integration/clean/README.md
```diff
@@ -7,12 +7,22 @@ The cleaner first searches for ami names (these are the ami created by the pipel
...
 The cleaner first searches for dedicated host tag Name:IntegrationTestMacDedicatedHost
 
-Then checks to see if the creation date is greater than 26 hours and host status is available
+Then checks to see if the creation date is greater than 30 days as default (can be configured) and host status is available
 
-Delete is true
\ No newline at end of file
+Delete is true
+
+# How to use the script?
```
SaxyPandaBear, line 20: Should provide an example here of running it, not just the abstract parameters below.

integration/clean/README.md
```diff
@@ -7,12 +7,22 @@ The cleaner first searches for ami names (these are the ami created by the pipel
...
 ###Cleans dedicated hosts for mac
 
 The cleaner first searches for dedicated host tag Name:IntegrationTestMacDedicatedHost
 
-Then checks to see if the creation date is greater than 26 hours and host status is available
+Then checks to see if the creation date is greater than 30 days as default (can be configured) and host status is available
 
-Delete is true
\ No newline at end of file
+Delete is true
```
SaxyPandaBear, line 18: What does this mean?

integration/clean/README.md
khanhntd: Good point. Add three examples for deleting AWS resources.

integration/clean/README.md
khanhntd: Good point to me again. Have changed the document how to configure this parameter

.github/workflows/clean-aws-resources.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.TERRAFORM_AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.TERRAFORM_AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{ matrix.region }}
+
+      - name: Update days to keep
+        if: github.event_name == 'workflow_dispatch'
+        run: echo "DAYS_TO_KEEP=${{ github.event.inputs.daystokeep }}" >> $GITHUB_ENV
```
SaxyPandaBear, line 40: Are you sure this works? I'm not sure that piping to the `$GITHUB_ENV` would do anything. Is there any documentation that points to doing this

integration/clean/README.md
```diff
@@ -1,18 +1,42 @@
...
-
-The cleaner first searches for dedicated host tag Name:IntegrationTestMacDedicatedHost
-
-Then checks to see if the creation date is greater than 26 hours and host status is available
-
-Delete is true
\ No newline at end of file
+## 1. What does the cleaner do?
+The cleaner will clean the AWS resources which are created from Integration Test.
+* CloudWatch Agent Integration Test
```
SaxyPandaBear, line 3: This line is unnecessary

integration/clean/README.md
```diff
@@ -1,18 +1,42 @@
...
+
+### 1.1 Cleans AMI 
+The cleaner first searches for AMI' name that begins with `cloudwatch-agent-integration-test`
```
SaxyPandaBear, line 6:
```suggestion
The cleaner first searches for AMI name that begins with `cloudwatch-agent-integration-test`
```
extra `'` char

integration/clean/README.md
```diff
@@ -1,18 +1,42 @@
...
+
+### 1.3 Cleans SSM Parameter Store
+The cleaner first searches for SSM Parameter Stores' name that begins with `AmazonCloudWatch`
+Then checks to see if the creation date is greater than 30 days as default. However, you can 
+change the 30 days by using variable `keep` as [the below example](README.md#2-how-to-use-the-script).
+
+## 2. How to use the script?
+By running the below command **with or without replacing these variables**:
+* **keep:** days to keep your resources (i.e 5 which is deleting resources after existing for 5 days)
+* **clean:** resources need to be cleaned (only support dedicated_host, ssm, ami)
```
SaxyPandaBear, line 23: those are the only clean scripts we have so why are we calling that it is only supported for these?

integration/clean/main.go
```diff
@@ -0,0 +1,76 @@
...
+				log.Printf("%v", err)
+			}
+		case dedicated_host.Type:
+			if err := dedicated_host.Clean(ctx, expirationDate); err != nil {
+				log.Printf("%v", err)
+			}
+		case ssm.Type:
+			if err := ssm.Clean(ctx, expirationDate); err != nil {
+				log.Printf("%v", err)
+			}
```
SaxyPandaBear, line 69: What bothers me about this is that it's added work to hook up a new cleaner here. For example, if we introduce k8s or ECS integration testing, we will probably want to add some form of cleaning up either the docker images, the clusters, or something - not sure what exactly yet. 

When we eventually add those integration tests, I wouldn't be surprised if we end up needing to add yet another switch statement here. It is fine in this scenario but makes it a little more cumbersome to add new things. 

Look at telegraf as an example. If someone contributes a new plugin to telegraf, the maintainers of telegraf don't have to also include a code change in the main code to accept that new plugin, as far as I know. If the point here is to consolidate and make it easier to use all of these different cleaner functions, we should treat them like individual cleaner "plugins" that we can use interchangeably. Adding the `Type` field to each cleaner should have been the first clue that this is a necessary change. We should not need to add a new switch statement here if we were to add a new cleaner.

integration/clean/README.md
khanhntd: Nice catch on wording. I should have change to `only have` instead of `only support`; support is more towards customer instead of developers.

integration/clean/README.md
khanhntd: Nice catch. I have fixed it in the newest commit.

integration/clean/README.md
khanhntd: Nice catch. I have fixed it in the newest commit.

.github/workflows/clean-aws-resources.yml
khanhntd: Yes I am sure this would work. Here is a document regarding how to [set an environment variable using bash script](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable). Moreover,

.github/workflows/clean-aws-resources.yml
```diff
@@ -0,0 +1,45 @@
...
+  DAYS_TO_KEEP: 30
+
+jobs:
+  clean-resources:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        resource: [ami, dedicated_host, ssm]
+        region: [ us-east-1, us-east-2, us-west-2, eu-west-1 ]
```
SaxyPandaBear, line 26: Do our integration tests on GitHub create any resources in a region outside of us-west-2? What is this for?

integration/clean/main.go
khanhntd: That's a good point of `telegraf`. Maybe I'm more biased regarding how `open-telemetry` framework. Even though the `open-telemetry` can treat these as individual cleaner "plugins"; however, it still provides the options for down-stream repo to consolidate their options instead of `telegraf` which only specify their options only to the specific functions. However, consider the small cleaner functions, I personally more favor to `open-telemetry` framework than towards `telegraf`. But It's good to have a new pair of eyes on this.

.github/workflows/clean-aws-resources.yml
khanhntd: Ack. I initially thought the our open-source integration-test uses the same account as the internal integration test. Therefore, cleaning in those regions. However, I have changed to only us-west-2 in those commits.

SaxyPandaBear (changes requested):
Changing this conflicts with the existing GitHub action in `internal-pipeline-dedicated-host-cleaner.yml`.

khanhntd:
> 

IMO, its won't conflict with the internal-pipeline cleaning dedicated host since both goals achieve the same thing, one would be more than 30 days one would be more than 1 days.

khanhntd:
Added `clean-internal-resources` job to `clean-aws-resources` workflow which only deletes `dedicated_host` in a daily manner; with this, we will able to avoid conflict `internal-pipeline-dedicated-host-cleaner.yml`

SaxyPandaBear (commented):
This has gone through a lot of edits, and I don't think that the title of the PR nor the description accurately describes the intended purpose of this PR. I think that we should discuss what we want to achieve with this more, as a team. I don't think that this needs a design doc or anything, but every revision feels like a bandaid on the last one.

Personally, I think that we should close this PR, discuss what we want to accomplish, make a GitHub issue detailing what we want to accomplish and then execute on that plan.

khanhntd:
Agree it has been through a lot of changes; however, the purpose still having the same effects; cleaning SSM Parameter with refactored code. So in a way, I could open a new PR for this (similar code) but agree we should discuss with everyone. Since after the Internal Integration Test  has passed; therefore, I would say this PR only serves as a sideway project and can be treated as low priority.

khanhntd:
Closing this because open another PR that reflects all the comments/ changes and will making a final decision on that PR https://github.com/aws/amazon-cloudwatch-agent/pull/476

PR #443, opened 2022-04-15
Add set of requirements for PR and add go linter tools for checking security, format, simplify...

khanhntd:
# Description of the issue
By using [`golangci-lint`](https://golangci-lint.run/usage/install/) with appropriate tools such as `gosec, golint, govet, simplify`, we can confirm that every code is up-to-standards and able to identify security errors, before merging it. One thing to note for is that this `golangci-lint` is only for development level instead of pipelines level (i.e when we deploy a hot-fix but simplify does not pass)

# License

# Tests
Run `golangci-lint run` with my branch.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -10,6 +10,11 @@ By submitting this pull request, I confirm that you can use, modify, copy, and r
 # Tests
 _Describe what tests you have done._
 
+#Requirements
+_Before commit the code, please do the following steps._
+1. Run `make fmt` and `make fmt-sh`
+2. Run [golangci-lint run](https://golangci-lint.run/usage/install/)
+
 
 
 
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -13,7 +13,7 @@ _Describe what tests you have done._
 #Requirements
 _Before commit the code, please do the following steps._
 1. Run `make fmt` and `make fmt-sh`
-2. Run [golangci-lint run](https://golangci-lint.run/usage/install/)
+2. Run `make linter`
 
 
 
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ GOIMPORTS_OPT?= -w -local $(CW_AGENT_IMPORT_PATH)
 
 GOIMPORTS = $(TOOLS_BIN_DIR)/goimports
 SHFMT = $(TOOLS_BIN_DIR)/shfmt
-
+LINTER = $(TOOLS_BIN_DIR)/golangci-lint
 release: clean test build package-rpm package-deb package-win package-darwin
 
 nightly-release: release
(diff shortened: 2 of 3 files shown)


.github/pull_request_template.md
```diff
@@ -10,6 +10,11 @@ By submitting this pull request, I confirm that you can use, modify, copy, and r
 # Tests
 _Describe what tests you have done._
 
+#Requirements
+_Before commit the code, please do the following steps._
+1. Run `make fmt` and `make fmt-sh`
+2. Run [golangci-lint run](https://golangci-lint.run/usage/install/)
```
SaxyPandaBear, line 16: I don't think that we should force people to need to run this external lint. Why would we need to do this if we are already requiring `make fmt`? The checks should be consistent.

SaxyPandaBear:
If we are going to add https://golangci-lint.run/usage/install/, then we should get rid of or update our existing `make fmt` to use that. I don't want to have multiple linters.

.github/pull_request_template.md
khanhntd: `make fmt` only formats code, while `golangci-lint` would do other thing such as security check, simplifying it as well. One thing we could do to make consistency is remove `goimport` and `gofmt` as default linter in `golangci-lint`

khanhntd:
> If we are going to add https://golangci-lint.run/usage/install/, then we should get rid of or update our existing `make fmt` to use that. I don't want to have multiple linters.

Somehow, I would prefer having a separate linters for focusing on specifically jobs though such as using `gofmt` and `goimport` only for formatting `golang` while `golinter` do everything else besides `gofmt` and `goimport`. A sample when running `golangci-lint`(lots of thing to fix but not focusing on what files has been changed with `goimport` or `gofmt`.
image

.github/pull_request_template.md
ZhenyuTan-amz: Can we discuss the necessity and benefits of this? I would much more prefer only need to run build/make which have gofmt embedded and not worry about those two extra steps added here. I think there're lint and security check we can add in Github for PRs, so you won't need to do it locally.

khanhntd: Some are not having with `github-action` such as `simply`, `govet`, `gosec`, `misspell` based on my finding so far. However, we can inject the `golangci-lint` to `Makefile` as a binary (same as GoImport if that make sense)

khanhntd: Adding installing binaries in the Makefile for `golangci-lint`, so we could use `make linter` to use the golangci-linter without any specification.

Makefile
```diff
@@ -139,6 +142,9 @@ fmt: install-tools
 fmt-sh: install-tools
 	${SHFMT} -w -d -i 5 .
 
+linter: install-tools
```
aateeqi, line 145: can we change `linter` to `lint` everywhere since it should describe the job not the tool

Makefile
```diff
@@ -139,6 +142,9 @@ fmt: install-tools
...
+	${LINTER} run ./...
```
aateeqi, line 146: why are we not using `golint` as described here: https://sparkbox.com/foundry/go_vet_gofmt_golint_to_code_check_in_Go

khanhntd: This includes `go-lint` but it [disabled by default](https://golangci-lint.run/usage/linters/#disabled-by-default)

Makefile
khanhntd: Make sense. Fixed it in the newest commit.

sethAmazon (approved):

PR #445, opened 2022-04-15
Include metric name in warning message when value is negative.

### LINKED ISSUE #288: The error/warning message "Value cannot be negative: -1" should point out the affected metric
I can see the following  entries in my `/var/log/amazon/amazon-cloudwatch-agent/amazon-cloudwatch-agent.log`

```
2021-10-06T13:18:00Z W! Value cannot be negative: -1
2021-10-06T13:18:00Z W! Value cannot be negative: -1
2021-10-06T13:18:00Z W! Value cannot be negative: -1
2021-10-06T13:18:00Z W! Value cannot be negative: -1
2021-10-06T13:18:00Z W! Value cannot be negative: -1
```

The messages should point out where the value -1 is coming from.  In my case I assume is one of the metrics that collectd is sending to the 25826 port. But there are too many to inspect, I think cloudwatch agent should print what is the source of that value.

adam-mateen:
# Description of the issue
Resolves #288 

# Description of changes
Change the `Distribution` interface's `AddEntry()` and `AddEntryWithUnit()` to return an error.
Then check that return value and conditionally log an error message including the metric name and value.

# License

# Tests
Updated unit tests to check return value.
Verified all unit tests pass.


SaxyPandaBear:
Overall looks fine to me. One thing that I think I want to see, just because I can't visualize it, is how often these could get logged. I'm a little concerned about bloating the log file with warning logs.

adam-mateen:
> Overall looks fine to me. One thing that I think I want to see, just because I can't visualize it, is how often these could get logged. I'm a little concerned about bloating the log file with warning logs.

It would not log any more than before.

Current code has:
```
log.Printf("W! Value cannot be negative: %v", value)
```
Which I removed since it did not log the metric name.
Unfortunately the function which logged that did not have access to the metric name, so an error must be returned so that the calling function can then print both metric name and value.

The frequency of this warning message depends on the types of metrics being collected and the likelihood that a negative value is collected.

SaxyPandaBear (dismissed):
LGTM. forgot that there was an existing warn log so there's no way for this to really increase the log file size that badly.

plugins/outputs/cloudwatch/cloudwatch_test.go
```diff
@@ -96,7 +96,8 @@ func TestBuildMetricDatums(t *testing.T) {
 	}
 
 	invalidDistribution := distribution.NewDistribution()
-	invalidDistribution.AddEntry(-1, 1)
+	err := invalidDistribution.AddEntry(-1, 1)
+	assert.Error(err)
```
khanhntd, line 100: Instead of only `assert.Error`, using `assert.EqualErrorf(t, err, expectedErrorMsg, "Error should be: %v, got: %v", expectedErrorMsg, err)` would be a better idea to confirm if the metrics are actually have that kind of error: . In this case, the negative value.

khanhntd:
Another thing to consider for which Andrew has pointed out already: the frequency and the time we log this error. Do we consider warning the customer every time we receive this negative value even if it is separate time? What I consider is adding to the `buffer` (same as how we treat unsupported type for Prometheus's metrics) and once it reaches 1000 metrics, it would show the errors and  `a final drop message instead` with a solution to that. Based on customer's view, what I really want is the final message and the solution to help overcoming that error.

plugins/outputs/cloudwatch/cloudwatch_test.go
adam-mateen: Good idea, will do.

adam-mateen:
> Another thing to consider for which Andrew has pointed out already: the frequency and the time we log this error. Do we consider warning the customer every time we receive this negative value even if it is separate time? What I consider is adding to the `buffer` (same as how we treat unsupported type for Prometheus's metrics) and once it reaches 1000 metrics, it would show the errors and `a final drop message instead` with a solution to that. Based on customer's view, what I really want is the final message and the solution to help overcoming that error.

In general, "noisy" logs can be solved by respecting the definitions of log levels.
if "negative value" metrics were a normal occurrence, or something that the user can safely ignore, then I would drop it down to an INFO level.
I kept it at WARN because user will have unexpected behavior when CWA collects a negative value.
And it is recommended the user debugs why negative values are collected.

There may be other scenarios a typically unexpected log message suddenly starts getting spammed in the log. In this case it would be nice to have a generic library allowing us to log a message with throttling, e.g.
```
maxMessagesPerMinute = 100
mylib.Printf_with_throttle(maxMessagesPerMinute, "my format string %s, %v", foo, 123)
```

khanhntd:
> > Another thing to consider for which Andrew has pointed out already: the frequency and the time we log this error. Do we consider warning the customer every time we receive this negative value even if it is separate time? What I consider is adding to the `buffer` (same as how we treat unsupported type for Prometheus's metrics) and once it reaches 1000 metrics, it would show the errors and `a final drop message instead` with a solution to that. Based on customer's view, what I really want is the final message and the solution to help overcoming that error.
> 
> In general, "noisy" logs can be solved by respecting the definitions of log levels. If "negative value" metrics were a normal occurrence, or something that the user can safely ignore, then I would drop it down to an INFO level. I kept it at WARN because user will have unexpected behavior when CWA collects a negative value. And it is recommended _most_ users debugswhy negative values are collected.
> 
> There may be other scenarios a typically unexpected log message suddenly starts getting spammed in the log. In this case it would be nice to have a generic library allowing us to log a message with throttling, e.g.
> 
> ```
> maxMessagesPerMinute := 100
> mylib.PrintWithThrottle(maxMessagesPerMinute, "my format string %s, %v", foo, 123)
> ```

True for respecting the definitions of log levels. Would be nice to have that but well, only buffer as a last resolution I guess.
Another underlying error we have been neglecting so far is the corresponding action when the customer faces an error. As for best practice I have read a long time ago, errors should  [have been accompanied by solution](https://uxplanet.org/how-to-write-good-error-messages-858e4551cd4). Even though we do not understand why the customer application emits negative value; however, we could suggest this has been solved by these ways ..... instead of letting the customer finding the customer on their owns which can sometime not a good source of truth.

SaxyPandaBear:
> In this case it would be nice to have a generic library allowing us to log a message with throttling

Agreed, but I don't think it's worth including in this PR.

Still LGTM

SaxyPandaBear (approved):

PR #448, opened 2022-04-19
Clean Dedicated Hosts From AWS Account For Mac Tests

sethAmazon:
# Description of the issue
Need a way to clean dedicated hosts 

# Description of changes
Clean host that have been created 26 hr ago

# Tests
Ran the cleaner.

### DIFF
--- a/.github/workflows/clean_ami.yml
+++ b/.github/workflows/clean_ami.yml
@@ -1,7 +1,7 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: MIT
 
-name: Monthly Resources Cleaner
+name: AMI Monthly Resources Cleaner
 
 on:
   schedule:
@@ -9,10 +9,11 @@ on:
   workflow_dispatch:
 
 jobs:
-  clean-resources:
+  clean-ami:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - uses: actions/setup-go@v3
 
       - name: Configure AWS Credentials
         uses: aws-actions/configure-aws-credentials@v1
@@ -21,5 +22,5 @@ jobs:
           aws-secret-access-key: ${{ secrets.TERRAFORM_AWS_SECRET_ACCESS_KEY }}
           aws-region: us-west-2
 
-      - name: Clean old aws resources
-        run: make integration-cleaner
+      - name: Clean old ami
+        run: go run ./integration/clean/clean_ami/clean_ami.go --tags=clean
--- a/.github/workflows/clean_dedicated_host.yml
+++ b/.github/workflows/clean_dedicated_host.yml
@@ -1,0 +1,26 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT
+
+name: Dedicated Host Daily Resources Cleaner
+
+on:
+  schedule:
+    - cron: "0 0 * * *" # Run Every Day At Midnight
+  workflow_dispatch:
+
+jobs:
+  clean-dedicated-hosts:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v3
+
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.TERRAFORM_AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.TERRAFORM_AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-west-2
+
+      - name: Clean old dedicated host
+        run: go run ./integration/clean/clean_dedicated_host/clean_dedicated_host.go --tags=clean
--- a/integration/clean/clean_dedicated_host/clean_dedicated_host.go
+++ b/integration/clean/clean_dedicated_host/clean_dedicated_host.go
@@ -1,0 +1,80 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+//go:build clean
+// +build clean
+
+package main
+
+import (
+	"context"
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+	"github.com/aws/aws-sdk-go-v2/service/ec2/types"
+	"log"
+	"time"
+)
+
+// Can't release a host if it was being used within the last 24 hr add 2 hr as a buffer
+const keepDurationDedicatedHost = -1 * time.Hour * 26
+const tagName = "tag:Name"
+const tagValue = "IntegrationTestMacDedicatedHost"
+
+func main() {
+	err := cleanDedicatedHost()
+	if err != nil {
+		log.Fatalf("errors cleaning %v", err)
+	}
+}
+
+func cleanDedicatedHost() error {
+	log.Print("Begin to clean EC2 Dedicated Host")
+
+	expirationDateDedicatedHost := time.Now().UTC().Add(keepDurationDedicatedHost)
+	cxt := context.Background()
+	defaultConfig, err := config.LoadDefaultConfig(cxt)
+	if err != nil {
+		return err
+	}
+	ec2client := ec2.NewFromConfig(defaultConfig)
+
+	dedicatedHosts, err := getDedicatedHost(cxt, ec2client)
+	if err != nil {
+		return err
+	}
+
+	dedicatedHostIds := make([]string, 0)
+	for _, dedicatedHost := range dedicatedHosts {
+		log.Printf("dedicated host id %v experation date %v dedicated host creation date raw %v",
+			*dedicatedHost.HostId, expirationDateDedicatedHost, *dedicatedHost.AllocationTime)
+		if expirationDateDedicatedHost.After(*dedicatedHost.AllocationTime) && dedicatedHost.State == types.AllocationStateAvailable {
+			log.Printf("Try to delete dedicated host %s tags %v launch-date %s", *dedicatedHost.HostId, dedicatedHost.Tags, *dedicatedHost.AllocationTime)
+			dedicatedHostIds = append(dedicatedHostIds, *dedicatedHost.HostId)
+		}
+	}
+
+	if len(dedicatedHostIds) == 0 {
+		log.Printf("No dedicated hosts to release")
+		return nil
+	}
+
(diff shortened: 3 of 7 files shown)


integration/clean/clean_ami.go
```diff
@@ -8,27 +8,19 @@ package main
...
-		log.Fatalf("errors cleaning %v", err)
-	}
-}
+var expirationDateAMI = time.Now().UTC().Add(keepDurationAMI)
 
 const daysToKeep = 60
-const keepDuration = -1 * time.Hour * 24 * time.Duration(daysToKeep)
-
-var expirationDate = time.Now().UTC().Add(keepDuration)
+const keepDurationAMI = -1 * time.Hour * 24 * time.Duration(daysToKeep)
```
SaxyPandaBear, line 23: Does this name need to be changed? It's lowercase so AFAIK it shouldn't conflict with the duration used in the other file.

integration/clean/clean_ami.go
```diff
@@ -8,27 +8,19 @@ package main
...
+	"time"
 )
 
-func main() {
-	err := cleanAMI()
-	if err != nil {
-		log.Fatalf("errors cleaning %v", err)
-	}
-}
+var expirationDateAMI = time.Now().UTC().Add(keepDurationAMI)
```
SaxyPandaBear, line 20: I'm not a fan of defining the expiration date at the top like this, depending on `time.Now()`. Can you add a `time.Time` parameter to the `cleanAMI` function, and then do this computation inside the function?

integration/clean/clean_dedicated_host.go
```diff
@@ -0,0 +1,70 @@
...
+	"log"
+	"time"
+)
+
+// Can't release a host if it was being used within the last 24 hr add 2 hr as a buffer
+var expirationDateDedicatedHost = time.Now().UTC().Add(keepDurationDedicatedHost)
+
+const keepDurationDedicatedHost = -1 * time.Hour * 26
+
+func cleanDedicatedHost() []error {
```
SaxyPandaBear, line 24: After looking through the function, there's no point at which we would ever return an errors array that has > 1 element in it. Can you change this to 
```suggestion
func cleanDedicatedHost() error {
```
and just return a single err

integration/clean/clean_dedicated_host.go
```diff
@@ -0,0 +1,70 @@
...
+	releaseDedicatedHost := ec2.ReleaseHostsInput{HostIds: dedicatedHostIds}
+	_, err = ec2client.ReleaseHosts(cxt, &releaseDedicatedHost)
+	if err != nil {
+		return []error{err}
+	}
+
+	return nil
+}
+
+func getDedicatedHost(cxt context.Context, ec2client *ec2.Client, nameFilter types.Filter) ([]types.Host, []error) {
```
SaxyPandaBear, line 63: Typically I'm okay with defining small functions like this, but it feels inconsistent. [Lines 54-58](https://github.com/aws/amazon-cloudwatch-agent/pull/448/files#diff-3abfb2bd5ed887a356d3af5724ab99d1980fdd4eff3bd3fe8bc82324dafec894R54-R58) do basically the same thing this function does with a different EC2 API, but it's in the main function. In this case, I don't have a preference as to whether we functionalize everything or just dump it all in the main function, as long as it's mostly consistent.

integration/clean/clean_dedicated_host.go
```diff
@@ -0,0 +1,70 @@
...
+	cxt := context.Background()
+	defaultConfig, err := config.LoadDefaultConfig(cxt)
+	if err != nil {
+		return []error{err}
+	}
+	ec2client := ec2.NewFromConfig(defaultConfig)
+
+	// Get list of dedicated host
+	nameFilter := types.Filter{Name: aws.String("tag:Name"), Values: []string{
+		"IntegrationTestMacDedicatedHost",
```
SaxyPandaBear, line 36: Missed this somehow - I think that this name should be put at the top as a const.

integration/clean/clean.go
```diff
@@ -0,0 +1,23 @@
...
+package main
+
+import (
+	"log"
+)
+
+func main() {
+	err := cleanAMI()
+	if err != nil {
+		log.Fatalf("errors cleaning ami err %v", err)
```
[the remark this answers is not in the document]
sethAmazon: I moved to 2 separate actions

.github/workflows/clean_ami.yml
```diff
@@ -21,5 +22,5 @@ jobs:
           aws-secret-access-key: ${{ secrets.TERRAFORM_AWS_SECRET_ACCESS_KEY }}
           aws-region: us-west-2
 
-      - name: Clean old aws resources
-        run: make integration-cleaner
\ No newline at end of file
+      - name: Clean old ami
+        run: go run ./integration/clean/clean_ami/clean_ami.go --tags=clean
```
SaxyPandaBear, line 26: Part of me thinks that we can rework the make command to allow us to pass in specific tags, and then we could have something like an `ami` tag, and a `dedicatedhost` tag or something and then keep it all wrapped inside the Makefile. Not sure. That was my gut reaction but I don't know how clean that would be.

.github/workflows/clean_ami.yml
```diff
@@ -1,18 +1,19 @@
 # Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 # SPDX-License-Identifier: MIT
 
-name: Monthly Resources Cleaner
+name: Clean AMI Monthly Resources Cleaner
```
SaxyPandaBear, line 4: Nitpick: I think the name is redundant. Personally, I think we can drop the "Clean" at the beginning of the name

.github/workflows/clean_dedicated_host.yml
```diff
@@ -0,0 +1,38 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT
+
+name: Clean Dedicated Host Daily Resources Cleaner
```
SaxyPandaBear, line 4: Nitpick: I think the name is redundant. Personally, I think we can drop the "Clean" at the beginning of the name

.github/workflows/clean_dedicated_host.yml
```diff
@@ -0,0 +1,38 @@
...
+          aws-access-key-id: ${{ secrets.TERRAFORM_AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.TERRAFORM_AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-west-2
+
+      - name: Clean old dedicated host
+        run: go run ./integration/clean/clean_dedicated_host/clean_dedicated_host.go --tags=clean
+
+      # @TODO delete this when we remove the internal cw agent integration pipeline
+      # In the mean time just set this secret to your account creds
+      - name: Configure AWS Credentials
```
SaxyPandaBear, line 30: IDK how granular we can get with configurations, but is there a way that we could specify that this specific step should only run in **this** repo, and not on people's forks? I know we can configure triggers to only work on specific repos, but not sure about individual steps.

integration/clean/clean_dedicated_host/clean_dedicated_host.go
```diff
@@ -0,0 +1,83 @@
...
+		return nil
+	}
+
+	releaseDedicatedHost := ec2.ReleaseHostsInput{HostIds: dedicatedHostIds}
+	_, err = ec2client.ReleaseHosts(cxt, &releaseDedicatedHost)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
SaxyPandaBear, line 68: Didn't notice this the first time. We could just simplify this to `return err` because if there is an error, we would return a non-nil error, but if there is no error, returning `err` would just result in returning nil, which does the same thing as this code block.

integration/clean/clean_dedicated_host/clean_dedicated_host.go
```diff
@@ -0,0 +1,83 @@
...
+		return err
+	}
+	ec2client := ec2.NewFromConfig(defaultConfig)
+
+	dedicatedHosts, err := getDedicatedHost(cxt, ec2client)
+	if err != nil {
+		return err
+	}
+
+	dedicatedHostIds := make([]string, 0)
```
SaxyPandaBear, line 47: Nit: I think it's weird to allocate a capacity of zero here. The maximum capacity we could have is `len(dedicatedHosts)`, so we could just always allocate an array of that size.

.github/workflows/clean_dedicated_host.yml
```diff
@@ -0,0 +1,38 @@
...
+  clean-dedicated-hosts:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v3
+
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.TERRAFORM_AWS_ACCESS_KEY_ID }}
```
SaxyPandaBear, line 21: I didn't realize/notice that GH actions outside of the integration tests use these secrets. I think it warrants consolidating a README that just outlines the different secrets needed to run things on your fork. Not a blocker for this PR, and I'm more than willing to follow up with a PR for that after this gets merged.

.github/workflows/clean_dedicated_host.yml
SaxyPandaBear: If not, is there a way that we could add a conditional here, like "if INTERNAL_AWS_ACCESS_KEY_ID is nonempty, run this step"? 

Basically, I just don't want to set up another credential on my fork if I don't have to.

.github/workflows/clean_ami.yml
sethAmazon: I don't think it worth wrapping the makefile since I moved the cleaners into separate packages.

integration/clean/clean_dedicated_host/clean_dedicated_host.go
sethAmazon: If you do that you can't check if the list is empty. len(dedicatedHostIds) == 0  will always return false.

.github/workflows/clean_dedicated_host.yml
sethAmazon: The workflow is not valid. .github/workflows/clean_dedicated_host.yml (Line: 30, Col: 13): Unrecognized named-value: 'secrets'. Located at position 1 within expression: secrets.INTERNAL_AWS_ACCESS_KEY_ID .github/workflows/clean_dedicated_host.yml (Line: 38, Col: 13): Unrecognized named-value: 'secrets'. Located at position 1 within expression: secrets.INTERNAL_AWS_ACCESS_KEY_ID
 I'm just going to move this out to another workflow for now

integration/clean/clean_dedicated_host/clean_dedicated_host.go
SaxyPandaBear: I _thought_ that len and cap would return different values for `make([]string, 10)`, but I was wrong. Ignore this

PR #449, opened 2022-04-19
Print Out Dedicated Host State In Clean Action

sethAmazon:
# Description of the issue
View dedicated host state in logs 

# Description of changes
Print dedicated host state in logs

### DIFF
--- a/integration/clean/clean_dedicated_host/clean_dedicated_host.go
+++ b/integration/clean/clean_dedicated_host/clean_dedicated_host.go
@@ -46,8 +46,8 @@ func cleanDedicatedHost() error {
 
 	dedicatedHostIds := make([]string, 0)
 	for _, dedicatedHost := range dedicatedHosts {
-		log.Printf("dedicated host id %v experation date %v dedicated host creation date raw %v",
-			*dedicatedHost.HostId, expirationDateDedicatedHost, *dedicatedHost.AllocationTime)
+		log.Printf("dedicated host id %v experation date %v dedicated host creation date raw %v host state %v",
+			*dedicatedHost.HostId, expirationDateDedicatedHost, *dedicatedHost.AllocationTime, dedicatedHost.State)
 		if expirationDateDedicatedHost.After(*dedicatedHost.AllocationTime) && dedicatedHost.State == types.AllocationStateAvailable {
 			log.Printf("Try to delete dedicated host %s tags %v launch-date %s", *dedicatedHost.HostId, dedicatedHost.Tags, *dedicatedHost.AllocationTime)
 			dedicatedHostIds = append(dedicatedHostIds, *dedicatedHost.HostId)


integration/clean/clean_dedicated_host/clean_dedicated_host.go
```diff
@@ -46,8 +46,8 @@ func cleanDedicatedHost() error {
 
 	dedicatedHostIds := make([]string, 0)
 	for _, dedicatedHost := range dedicatedHosts {
-		log.Printf("dedicated host id %v experation date %v dedicated host creation date raw %v",
-			*dedicatedHost.HostId, expirationDateDedicatedHost, *dedicatedHost.AllocationTime)
+		log.Printf("dedicated host id %v experation date %v dedicated host creation date raw %v host state %v",
+			*dedicatedHost.HostId, expirationDateDedicatedHost, *dedicatedHost.AllocationTime, dedicatedHost.State)
```
SaxyPandaBear, line 50: I didn't look at the actual struct, but should this resolve a pointer, like `*dedicatedHost.State`? Asking cause the other references to the host struct do that. No point in adding this to the log if it just prints out a memory address

sethAmazon: type Host struct {

	// The time that the Dedicated Host was allocated.
	AllocationTime *time.Time

	// Indicates whether the Dedicated Host supports multiple instance types of the
	// same instance family. If the value is on, the Dedicated Host supports multiple
	// instance types in the instance family. If the value is off, the Dedicated Host
	// supports a single instance type only.
	AllowsMultipleInstanceTypes AllowsMultipleInstanceTypes

	// Whether auto-placement is on or off.
	AutoPlacement AutoPlacement

	// The Availability Zone of the Dedicated Host.
	AvailabilityZone *string

	// The ID of the Availability Zone in which the Dedicated Host is allocated.
	AvailabilityZoneId *string

	// Information about the instances running on the Dedicated Host.
	AvailableCapacity *AvailableCapacity

	// Unique, case-sensitive identifier that you provide to ensure the idempotency of
	// the request. For more information, see Ensuring Idempotency
	// (https://docs.aws.amazon.com/AWSEC2/latest/APIReference/Run_Instance_Idempotency.html).
	ClientToken *string

	// The ID of the Dedicated Host.
	HostId *string

	// The hardware specifications of the Dedicated Host.
	HostProperties *HostProperties

	// Indicates whether host recovery is enabled or disabled for the Dedicated Host.
	HostRecovery HostRecovery

	// The reservation ID of the Dedicated Host. This returns a null response if the
	// Dedicated Host doesn't have an associated reservation.
	HostReservationId *string

	// The IDs and instance type that are currently running on the Dedicated Host.
	Instances []HostInstance

	// Indicates whether the Dedicated Host is in a host resource group. If
	// memberOfServiceLinkedResourceGroup is true, the host is in a host resource
	// group; otherwise, it is not.
	MemberOfServiceLinkedResourceGroup *bool

	// The ID of the Amazon Web Services account that owns the Dedicated Host.
	OwnerId *string

	// The time that the Dedicated Host was released.
	ReleaseTime *time.Time

	// The Dedicated Host's state.
	State AllocationState

	// Any tags assigned to the Dedicated Host.
	Tags []Tag

	noSmithyDocumentSerde
}

for the struct this is not a pointer

SaxyPandaBear (approved):

PR #451, opened 2022-04-21
Add Integration Test For Metric Number Of Dimension

sethAmazon:
# Description of the issue
Add number of dimension integration test

# Description of changes
Add number of dimension integration test

# Tests
https://github.com/sethAmazon/amazon-cloudwatch-agent/actions/runs/2198728626


integration/terraform/ec2/linux/main.tf
```diff
@@ -24,7 +24,8 @@ resource "aws_instance" "integration-test" {
       "export AWS_REGION=${var.region}",
       "echo run tests with the tag integration, one at a time, and verbose",
       "cd ~/amazon-cloudwatch-agent",
-      "go test ./integration/test/... -p 1 -v --tags=${var.tag}"
+      "echo run sanity test && go test ./integration/test/sanity -p 1 -v --tags=integration",
```
SaxyPandaBear, line 27: This is interesting. Question - is there any reason we wouldn't want to break out the sanity test as a separate workflow from the integration tests? 🤔

integration/test/metrics_number_dimension/metrics_number_dimension_test.go
```diff
@@ -0,0 +1,133 @@
...
+const configOutputPath = "/opt/aws/amazon-cloudwatch-agent/bin/config.json"
+const configJSON = "/config.json"
+const namespace = "MetricNumberDimensionTest"
+const instanceId = "InstanceId"
+const append = "append"
+
+//Let the agent run for 2 minutes. This will give agent enough time to call server
+const agentRuntime = 2 * time.Minute
+
+var targetString = "max MaxDimensions %v is less than than number of dimensions %v thus only taking the max number"
```
SaxyPandaBear, line 31: This feels like bad code smell to me. It took me a bit to realize that this global gets modified in the tests. I think that we should not modify this inside the test. We should create a new string off of this template, and then pass _that_ value into the `outputLogContainsTarget` function.

integration/test/metrics_number_dimension/metrics_number_dimension_test.go
```diff
@@ -0,0 +1,133 @@
...
+//This test uses a pem file created for the local stack endpoint to be able to connect via ssl
+func TestBundle(t *testing.T) {
+
+	parameters := []input{
+		{dataInput: "resources/10_dimension",
+			findTarget:           false,
+			inputNumberDimension: 10,
+			metricName:           "mem_used_percent"},
+		// @TODO add when https://github.com/aws/amazon-cloudwatch-agent/pull/361 is merged
+		// {dataInput: "resources/30_dimension", findTarget: false, inputNumberDimension: 30, metricName: "mem_used_percent"},
```
SaxyPandaBear, line 55: If we were to turn on this test today, would this have `findTarget = true` to pass the test? If that's the case, I think that what we can do is
1. uncomment these test cases so we can merge this fully
2. modify the other PR to account for this by flipping the bool here, so we can illustrate that our integration tests do what we want them to do 😄

integration/test/metrics_number_dimension/metrics_number_dimension_test.go
```diff
@@ -0,0 +1,133 @@
...
+		// {dataInput: "resources/35_dimension", findTarget: true, inputNumberDimension: 35, metricName: "mem_used_percent"},
+	}
+
+	for _, parameter := range parameters {
+		//before test run
+		log.Printf("resource file location %s find target %t input number dimension %d metric name %s",
+			parameter.dataInput, parameter.findTarget, parameter.inputNumberDimension, parameter.metricName)
+
+		targetString = fmt.Sprintf(targetString, 30, parameter.inputNumberDimension)
```
SaxyPandaBear, line 64: Wouldn't this turn into:
```
max MaxDimensions 30 is less than than number of dimensions 10 thus only taking the max number
```
Is that correct? Also feels like bad code smell to hard code 30 inside the test. I think it should be moved out to a constant, unless this is supposed to change with the test cases? But I don't think it should if I recall correctly.

integration/test/metrics_number_dimension/metrics_number_dimension_test.go
```diff
@@ -0,0 +1,133 @@
...
+			// test for target string
+			output := test.ReadAgentOutput(agentRuntime)
+			containsTarget := outputLogContainsTarget(output)
+			if (parameter.findTarget && !containsTarget) || (!parameter.findTarget && containsTarget) {
+				t.Errorf("Find target is %t contains target is %t", parameter.findTarget, containsTarget)
+			}
+
+			// test for cloud watch metrics
+			cxt := context.Background()
+			dimensionFilter := buildDimensionFilterList(parameter.inputNumberDimension - 1)
```
SaxyPandaBear, line 82: Can you explain why we pass in `- 1` here? I see that we add 1 back to it inside the `buildDimensionFilterList`, so this feels confusing to me.

integration/test/metrics_number_dimension/metrics_number_dimension_test.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+type metric struct {
+	name  string
+	value string
+}
+
+//Must run this test with parallel 1 since this will fail if more than one test is running at the same time
+//This test uses a pem file created for the local stack endpoint to be able to connect via ssl
+func TestBundle(t *testing.T) {
```
SaxyPandaBear, line 47: I think this is copy pasted from the other test. I think the test name should be changed (and some of the wording in the comments since this isn't verifying SSL certs)

integration/test/metrics_number_dimension/metrics_number_dimension_test.go
```diff
@@ -0,0 +1,133 @@
...
+				Dimensions: dimensionFilter,
+			}
+			data, err := client.ListMetrics(cxt, &listMetricsInput)
+			if err != nil {
+				t.Errorf("Error getting metric data %v", err)
+			}
+			if len(data.Metrics) == 0 {
+				metrics := make([]metric, len(dimensionFilter))
+				for i, filter := range dimensionFilter {
+					var value string
```
SaxyPandaBear, line 96: The dimension filters get created in the below `buildDimensionFiltersList` function. When would the value be nil?

integration/test/metrics_number_dimension/metrics_number_dimension_test.go
```diff
@@ -0,0 +1,133 @@
...
+			listMetricsInput := cloudwatch.ListMetricsInput{
+				MetricName: aws.String(parameter.metricName),
+				Namespace:  aws.String(namespace),
+				Dimensions: dimensionFilter,
+			}
+			data, err := client.ListMetrics(cxt, &listMetricsInput)
+			if err != nil {
+				t.Errorf("Error getting metric data %v", err)
+			}
+			if len(data.Metrics) == 0 {
```
SaxyPandaBear, line 93: Is there any way to get a consistent/semi-consistent estimate of the expected number of metrics? Basically this is the same as saying when we ListMetrics, we expect at least one metric to come back, which doesn't tell us much - hypothetical situation, but assuming that we had a bug in the metric publishing when we try to append more than 30 dimensions that causes the agent to die after submitting the metrics once, this would not help us catch that (maybe it shouldn't?). 

I'd prefer this to be fleshed out more as a generalized validation function, but maybe I'm just biased because that's how I did the log validation.

integration/test/metrics_number_dimension/metrics_number_dimension_test.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/aws/aws-sdk-go-v2/service/cloudwatch/types"
+	"log"
+	"strings"
+	"testing"
+	"time"
+)
+
+const configOutputPath = "/opt/aws/amazon-cloudwatch-agent/bin/config.json"
+const configJSON = "/config.json"
+const namespace = "MetricNumberDimensionTest"
```
SaxyPandaBear, line 24: Does reusing the namespace here mess up the test? 

I know when I was doing the integration tests for CWL, I ran into a problem verifying the data that got sent because I used a shared log group/stream, and since the test gets run across different OSes in parallel, that messed up the expected number of logs.

integration/test/metrics_number_dimension/metrics_number_dimension_test.go
```diff
@@ -0,0 +1,133 @@
...
+	name  string
+	value string
+}
+
+//Must run this test with parallel 1 since this will fail if more than one test is running at the same time
+//This test uses a pem file created for the local stack endpoint to be able to connect via ssl
+func TestBundle(t *testing.T) {
+
+	parameters := []input{
+		{dataInput: "resources/10_dimension",
```
SaxyPandaBear, line 50: I don't think that `dataInput` is a very clear name for what this represents - might want to revisit the cert tests too, but not a blocker. The value here is a directory/path to a specific config file, right?

integration/test/agent_util.go
```diff
@@ -93,3 +97,30 @@ func ReplaceLocalStackHostName(pathIn string) {
 		log.Fatal(fmt.Sprint(err) + string(out))
 	}
 }
+
+func GetInstanceId() string {
+	// this uses the {instance_id} placeholder in the agent configuration,
+	// so we need to determine the host's instance ID for validation
```
SaxyPandaBear, line 103: I think the context in the CWL integration tests gets lost by moving this comment over. Can you move this comment block back to where it was? I am fine with pulling the code here as a function in this util file, just don't think the comment here is 100% relevant generally - it's referencing the `{instance_id}` placeholder in log configuration.

integration/terraform/ec2/linux/main.tf
sethAmazon: Sanity is here to test if things are working on the box. Will tell us our agent is working. I like sanity for every test case. Just makes sense to me. We can talk about this.

integration/test/agent_util.go
sethAmazon: It makes sense in both context since we need the instance id for validation for both tests. I will still move.

integration/test/metrics_number_dimension/metrics_number_dimension_test.go
sethAmazon: No this is fine. I have no issue.

integration/test/metrics_number_dimension/metrics_number_dimension_test.go
sethAmazon: It will not find true since the other pr is not merged. The log target is added in that pr.

integration/test/metrics_number_dimension/metrics_number_dimension_test.go
sethAmazon: Problem we don't have this hardcoded until the other pr is merged. Since the log line and the 30 was added in that pr.

integration/test/metrics_number_dimension/metrics_number_dimension_test.go
sethAmazon: We would catch this by the finding the debug line in the log.

sethAmazon: Also agent does not fail on more than 30 metrics it just strips dimension to 30 then continues

integration/terraform/ec2/linux/main.tf
SaxyPandaBear: When we break out all of the integration tests by directory, does that mean we would use different hosts to run like the CA bundle tests versus the dimensions tests versus the CWL tests? If that's the case, then yeah this makes sense to me. I think when I was reading it last night, I was still in the frame of reference where the tests would be executed on the same hosts, so we only need to run sanity once, not per directory.

integration/test/agent_util.go
SaxyPandaBear: Right. We _use_ the instance ID get we fetch from IMDS in both tests, but I think the comment in question is misleading because it specifically points out the log configuration, which is not relevant to using the `{aws:InstanceId}` aggregation dimension, but that's just me being super nitpicky about it

integration/test/metrics_number_dimension/metrics_number_dimension_test.go
sethAmazon: I intended to merge now and just comment out the other tests when the other is merged. I can also remove the hard coded value when that happens.

integration/test/metrics_number_dimension/resources/30_dimension/config.json
```diff
@@ -0,0 +1,53 @@
+{
```
SaxyPandaBear, line 1: This is not a blocker, but it makes more sense in my head to not include the 30 and 35 dimensions configs at all in this PR, and then just introduce these configs when merging the other one.

SaxyPandaBear (approved):
LGTM

PR #452, opened 2022-04-21
Fix auto_removal feature so log lines are NOT dropped from old file

### LINKED ISSUE #381: auto_removal will drop log lines from old file when new file discovered
If a user has a log config with:
```
 "file_path": "/tmp/foo.log*",
 "auto_removal": true
```
And they write to a file `/tmp/foo.log.2020-01-02` very rapidly, it is possible CWAgent could fall behind tailing that file.
And when the user starts writing to `/tmp/foo.log.2020-01-03` it is possible CWAgent has not finished sending all those lines from the old file.

In this case it will intentionally drop lines.
But the user doesn't want to drop lines 😢 

In `FindLogSrc()`:
```
if _, ok := dests[filename]; ok {
	continue
} else if fileconfig.AutoRemoval { // This logic means auto_removal does not work with publish_multi_logs
	for _, dst := range dests {
		dst.tailer.StopAtEOF() // Stop all other tailers in favor of the newly found file
	}
}
```

`StopAtEOF()` has a comment making it seem like the intention was read until the end...
```
// StopAtEOF stops tailing as soon as the end of the file is reached.
func (tail *Tail) StopAtEOF() error {
	tail.Kill(errStopAtEOF)
	return tail.Wait()
}
```
However `Tail.sendLine()` stops and drops lines if `Tail.Dying()` is set for any reason, including `errStopAtEOF`:
```
// sendLine sends the line(s) to Lines channel, splitting longer lines
// if necessary. Return false if rate limit is reached.
func (tail *Tail) sendLine(line string, offset int64) bool {
	now := time.Now()
	lines := []string{line}

	// Split longer lines
	if tail.MaxLineSize > 0 && len(line) > tail.MaxLineSize {
		lines = partitionString(line, tail.MaxLineSize)
	}

	for i, line := range lines {
		// This select is to avoid blockage on the tail.Lines chan
		select {
		case tail.Lines <- &Line{line, now, nil, offset}:
		case <-tail.Dying():
			tail.dropCnt += len(lines) - i
			return true
		}
	}
```

Please consider changing CWAgent to wait for tailer to actually read and report all lines from the old file before auto removing it.

adam-mateen:
# Description of the issue
Fix #381 

# Description of changes
Fix Tail.sendLine() to NOT drop log lines when the tailer is dying because of errStopAtEOF.

# License

# Tests
Updated existing unit tests.
Ran all tests and verified they pass.

```
go test ./plugins/inputs/logfile -run TestLogsFileAutoRemoval -count=100 -v

=== RUN   TestLogsFileAutoRemoval
    logfile_test.go:376: Created temp file, /var/folders/7c/rx6r0v6n44j4pwnc_77gt46w0000gs/T/file_auto_removal2801393868
2022/05/16 08:41:34 The state file  for /var/folders/7c/rx6r0v6n44j4pwnc_77gt46w0000gs/T/file_auto_removal2801393868 does not exist: stat : no such file or directory
    logfile_test.go:397: Fill temp file with sufficient lines to be read.
2022/05/16 08:41:34 W! [inputs.tail] Stopping tail as file no longer exists: /var/folders/7c/rx6r0v6n44j4pwnc_77gt46w0000gs/T/file_auto_removal668799011
    logfile_test.go:430: Verify every line written to the temp file is received.
    logfile_test.go:376: Created temp file, /var/folders/7c/rx6r0v6n44j4pwnc_77gt46w0000gs/T/file_auto_removal2014352509
    logfile_test.go:444: Verify child completed.
2022/05/16 08:41:34 The state file  for /var/folders/7c/rx6r0v6n44j4pwnc_77gt46w0000gs/T/file_auto_removal2014352509 does not exist: stat : no such file or directory
    logfile_test.go:397: Fill temp file with sufficient lines to be read.
    logfile_test.go:430: Verify every line written to the temp file is received.
    logfile_test.go:447: Completed before timeout (as expected)
    logfile_test.go:451: Verify 1st temp file was auto deleted.
2022/05/16 08:41:35 W! [logfile] Failed to auto remove file /var/folders/7c/rx6r0v6n44j4pwnc_77gt46w0000gs/T/file_auto_removal2014352509: remove /var/folders/7c/rx6r0v6n44j4pwnc_77gt46w0000gs/T/file_auto_removal2014352509: no such file or directory
    logfile_test.go:466: Verify 1st tmp file created and discovered.
    logfile_test.go:469: Completed before timeout (as expected)
--- PASS: TestLogsFileAutoRemoval (1.40s)
```

```
go test ./plugins/inputs/logfile/tail -v

=== RUN   TestNotTailedCompeletlyLogging
--- PASS: TestNotTailedCompeletlyLogging (1.60s)
=== RUN   TestStopAtEOF
timeout waiting for StopAtEOF() (as expected)
StopAtEOF() completed (as expected)
--- PASS: TestStopAtEOF (1.00s)
PASS
ok  	github.com/aws/amazon-cloudwatch-agent/plugins/inputs/logfile/tail	2.744s
```

### DIFF
--- a/plugins/inputs/logfile/logfile.go
+++ b/plugins/inputs/logfile/logfile.go
@@ -167,9 +167,11 @@ func (t *LogFile) FindLogSrc() []logs.LogSrc {
 
 			if _, ok := dests[filename]; ok {
 				continue
-			} else if fileconfig.AutoRemoval { // This logic means auto_removal does not work with publish_multi_logs
+			} else if fileconfig.AutoRemoval {
+				// This logic means auto_removal does not work with publish_multi_logs
 				for _, dst := range dests {
-					dst.tailer.StopAtEOF() // Stop all other tailers in favor of the newly found file
+					// Stop all other tailers in favor of the newly found file
+					dst.tailer.StopAtEOF()
 				}
 			}
 
--- a/plugins/inputs/logfile/logfile_test.go
+++ b/plugins/inputs/logfile/logfile_test.go
@@ -417,7 +417,9 @@ func createWriteRead(t *testing.T, prefix string, logFile *LogFile, done chan bo
 	logSrc, evts := getLogSrc(t, logFile)
 	defer (*logSrc).Stop()
 	defer close(evts)
-	const numLines int = 1000
+	// Choose a large enough number of lines so that even high-spec hosts will not
+	// complete receiving logEvents before the 2nd createWriteRead() goroutine begins.
+	const numLines int = 100000
 	const msg string = "this is the best log line ever written to a file"
 	writeLines(t, file, numLines, msg)
 	file.Close()
--- a/plugins/inputs/logfile/logfile_test.go
+++ b/plugins/inputs/logfile/logfile_test.go
@@ -358,100 +358,128 @@ func TestLogsFileRemove(t *testing.T) {
 	tt.Stop()
 }
 
-//When another file is created for the same file config and the file config has auto_removal as true, the old files will stop at EOF and removed afterwards
-func TestLogsFileAutoRemoval(t *testing.T) {
-	var wg sync.WaitGroup
+// getLogSrc returns a LogSrc from the given LogFile, and the channel for output.
+// Verifies 1 and only 1 LogSrc is discovered.
+func getLogSrc(t *testing.T, logFile *LogFile) (*logs.LogSrc, chan logs.LogEvent) {
+	logSources := logFile.FindLogSrc()
+	require.Equal(t, 1, len(logSources))
+	logSource := logSources[0]
+	evts := make(chan logs.LogEvent)
+	logSource.SetOutput(func(e logs.LogEvent) {
+		if e != nil {
+			evts <- e
+		}
+	})
+	return &logSource, evts
+}
 
+func writeSomeLines(t *testing.T, file *os.File, numLines int, msg string) {
+	for i := 0; i < numLines; i++ {
+		//time.Sleep(1 * time.Millisecond)
+		//fmt.Printf("write i %v\n", i)
+		_, err := file.WriteString(msg + "\n")
+		require.NoError(t, err)
+	}
+}
+
+// TestLogsFileAutoRemoval verifies when a new file matching the configured
+// FilePath is discovered, the old file will be automatically deleted after
+// being read to the end-of-file.
+func TestLogsFileAutoRemoval(t *testing.T) {
+	// Override global in tailersrc.go.
 	multilineWaitPeriod = 10 * time.Millisecond
-	logEntryString := "anything"
+
 	filePrefix := "file_auto_removal"
 	tmpfile1, err := createTempFile("", filePrefix)
+	fmt.Printf("Created 1st temp file, %s\n", tmpfile1.Name())
 	require.NoError(t, err)
 
-	_, err = tmpfile1.WriteString(logEntryString + "\n")
-	require.NoError(t, err)
-	tmpfile1.Sync()
-	tmpfile1.Close()
-
-	tt := NewLogFile()
-	tt.Log = TestLogger{t}
-	tt.FileConfig = []FileConfig{{
+	logFile := NewLogFile()
+	defer logFile.Stop()
+	logFile.Log = TestLogger{t}
+	logFile.FileConfig = []FileConfig{{
 		FilePath:      filepath.Join(filepath.Dir(tmpfile1.Name()), filePrefix+"*"),
 		FromBeginning: true,
 		AutoRemoval:   true,
 	}}
-	tt.FileConfig[0].init()
-	tt.started = true
+	logFile.FileConfig[0].init()
+	logFile.started = true
 
-	lsrcs := tt.FindLogSrc()
-	if len(lsrcs) != 1 {
-		t.Fatalf("%v log src was returned when 1 should be available", len(lsrcs))
-	}
+	logSource, evts := getLogSrc(t, logFile)
+	defer (*logSource).Stop()
 
-	lsrc := lsrcs[0]
-	evts := make(chan logs.LogEvent)
-	lsrc.SetOutput(func(e logs.LogEvent) {
-		if e != nil {
-			evts <- e
-		}
-	})
+	fmt.Println("Fill temp file with sufficient lines to be read.")
+	numLogLinesToWrite := 10000
+	logEntryString := "this is the best log line ever written to a file"
+	writeSomeLines(t, tmpfile1, numLogLinesToWrite, logEntryString)
+	tmpfile1.Close()
 
 	var tmpfile2 *os.File
+	var logSource2 *logs.LogSrc
+	var	evts2 chan logs.LogEvent
 	defer func() {
+		if logSource2 != nil {
+			(*logSource2).Stop()
+		}
 		if tmpfile2 != nil {
+			tmpfile2.Close()
 			os.Remove(tmpfile2.Name())
 		}
 	}()
 
-	wg.Add(1)
+	fmt.Println("Verify every line written to the first temp file is received.")
+	// Do this in a goroutine in case there is a bug and it hangs.
+	readerDone := make(chan bool)
+	// Need to make sure tmpfile2 is created before accessing evts2
+	fileCreatorDone := make(chan bool)
 
 	go func() {
-		defer wg.Done()
-
-		// create a new file matching configured pattern
-		tmpfile2, err = createTempFile("", filePrefix)
-		require.NoError(t, err)
-
-		_, err = tmpfile2.WriteString(logEntryString + "\n")
-		require.NoError(t, err)
-
+		defer close(readerDone)
+		for i := 0; i < numLogLinesToWrite; i++ {
+			logEvent := <- evts
+			require.Equal(t, logEntryString, logEvent.Message())
+			if i != numLogLinesToWrite / 2 {
+				continue
+			}
+			// Halfway through receiving events create a new temp file.
+			// Still expect to recv all events from the first file.
+			// Need to do this in a goroutine since FindLogSrc()
+			// will block until tailer reaches EOF on first temp file.
+			go func() {
+				defer close(fileCreatorDone)
+				tmpfile2, err = createTempFile("", filePrefix)
+				fmt.Printf("Created 2nd temp file, %s\n", tmpfile2.Name())
+				require.NoError(t, err)
+				_, err = tmpfile2.WriteString(logEntryString + "\n")
+				require.NoError(t, err)
+				logSource2, evts2 = getLogSrc(t, logFile)
+			}()
+		}
 	}()
 
-	e := <-evts
-	if e.Message() != logEntryString {
-		t.Errorf("Wrong log found from first file: \n%v\nExpecting:\n%v\n", e.Message(), logEntryString)
-	}
-	defer lsrc.Stop()
-
-	for {
-		lsrcs = tt.FindLogSrc()
-		if len(lsrcs) > 0 {
-			break
-		}
-		time.Sleep(1 * time.Second)
+	fmt.Println("Verify reader completed.")
+	select {
+	case <-readerDone:
+		fmt.Println("Completed before timeout (as expected)")
+	case <-time.After(time.Second * 10):
+		t.Fatalf("timeout waiting for reader")
 	}
 
-	lsrc = lsrcs[0]
-	lsrc.SetOutput(func(e logs.LogEvent) {
-		if e != nil {
-			evts <- e
-		}
-	})
-
-	e = <-evts
-	if e.Message() != logEntryString {
-		t.Errorf("Wrong log found from 2nd file: \n% x\nExpecting:\n% x\n", e.Message(), logEntryString)
+	fmt.Println("Verify 2nd tmp file created and discovered.")
+	select {
+	case <-fileCreatorDone:
+		fmt.Println("Completed before timeout (as expected)")
+	case <-time.After(time.Second * 10):
+		t.Fatalf("timeout waiting for 2nd temp file.")
 	}
 
-	//Use Wait Group to avoid race condition between opening tmpfile2 to delete tmpfile1 with auto_removal and opening tmpfile1
-	//to check it exist
-	wg.Wait()
+	fmt.Println("Verify message in 2nd temp file.")
+	e2 := <- evts2
+	assert.Equal(t, logEntryString, e2.Message())
 
+	fmt.Println("Verify 1st temp file was auto deleted.")
 	_, err = os.Open(tmpfile1.Name())
 	assert.True(t, os.IsNotExist(err))
-
-	lsrc.Stop()
-	tt.Stop()
 }
 
 func TestLogsTimestampAsMultilineStarter(t *testing.T) {
--- a/plugins/inputs/logfile/logfile_test.go
+++ b/plugins/inputs/logfile/logfile_test.go
@@ -358,6 +358,26 @@ func TestLogsFileRemove(t *testing.T) {
 	tt.Stop()
 }
 
+func setupLogFileForTest(t *testing.T, file *os.File, prefix string) *LogFile {
+	logFile := NewLogFile()
+	logFile.Log = TestLogger{t}
+	logFile.FileConfig = []FileConfig{{
+		FilePath:      filepath.Join(filepath.Dir(file.Name()), prefix+"*"),
+		FromBeginning: true,
+		AutoRemoval:   true,
+	}}
+	logFile.FileConfig[0].init()
+	logFile.started = true
+	return logFile
+}
+
+func makeTempFile(t *testing.T, prefix string) *os.File {
+	file, err := createTempFile("", prefix)
+	t.Logf("Created temp file, %s\n", file.Name())
+	require.NoError(t, err)
+	return file
+}
+
 // getLogSrc returns a LogSrc from the given LogFile, and the channel for output.
 // Verifies 1 and only 1 LogSrc is discovered.
 func getLogSrc(t *testing.T, logFile *LogFile) (*logs.LogSrc, chan logs.LogEvent) {
@@ -373,12 +393,62 @@ func getLogSrc(t *testing.T, logFile *LogFile) (*logs.LogSrc, chan logs.LogEvent
 	return &logSource, evts
 }
 
-func writeSomeLines(t *testing.T, file *os.File, numLines int, msg string) {
+func writeLines(t *testing.T, file *os.File, numLines int, msg string) {
+	t.Log("Fill temp file with sufficient lines to be read.")
 	for i := 0; i < numLines; i++ {
-		//time.Sleep(1 * time.Millisecond)
-		//fmt.Printf("write i %v\n", i)
 		_, err := file.WriteString(msg + "\n")
 		require.NoError(t, err)
+	}
+}
+
+// createWriteRead creates a temp file, writes to it, then verifies events
+// are received. If isParent is true, then spawn a 2nd goroutine for createWriteRead.
+// Close the given channel when complete to let caller know it was successful.
+func createWriteRead(t *testing.T, prefix string, logFile *LogFile, done chan bool, isParent bool) {
+	// Let caller know when the goroutine is done.
+	defer close(done)
+	// done2 is only passed to child if this is the parent.
+	done2 := make(chan bool)
+	file := makeTempFile(t, prefix)
+	if isParent {
+		logFile = setupLogFileForTest(t, file, prefix)
+		defer logFile.Stop()
+	}
+	logSrc, evts := getLogSrc(t, logFile)
+	defer (*logSrc).Stop()
+	defer close(evts)
+	const numLines int = 1000
+	const msg string = "this is the best log line ever written to a file"
+	writeLines(t, file, numLines, msg)
+	file.Close()
+	if !isParent {
+		// Child creates 2nd temp file which is NOT auto removed.
+		defer os.Remove(file.Name())
+	}
+	t.Log("Verify every line written to the temp file is received.")
+	for i := 0; i < numLines; i++ {
+		logEvent := <- evts
+		require.Equal(t, msg, logEvent.Message())
+		if i != numLines / 2 {
+			continue
+		}
+		// Halfway through start another goroutine to create another temp file.
+		if isParent {
+			go createWriteRead(t, prefix, logFile, done2, false)
+		}
+	}
+	// Only wait for child if it was spawned
+	if isParent {
+		t.Log("Verify child completed.")
+		select {
+		case <-done2:
+			t.Log("Completed before timeout (as expected)")
+		case <-time.After(time.Second * 5):
+			require.Fail(t, "timeout waiting for child")
+		}
+		t.Log("Verify 1st temp file was auto deleted.")
+		_, err := os.Open(file.Name())
+		assert.True(t, os.IsNotExist(err))
 	}
 }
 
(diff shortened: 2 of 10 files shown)


plugins/inputs/logfile/tailersrc.go
```diff
@@ -76,6 +76,8 @@ type tailerSrc struct {
 	startTailerOnce sync.Once
 	cleanUpFns      []func()
 }
+// Verify tailerSrc implements LogSrc
+var _ logs.LogSrc = (*tailerSrc)(nil)
```
adam-mateen, line 80: These are optional changes.
I noticed tailerSrc did not actually implement all methods of the interface because some methods used pointer-receiver, others used value-receiver.

https://go.dev/tour/methods/8
says:
>In general, all methods on a given type should have either value or pointer receivers, but not a mixture of both.

plugins/inputs/logfile/tail/tail_test.go
```diff
@@ -71,22 +76,46 @@ func TestNotTailedCompeletlyLogging(t *testing.T) {
...
+	go func() {
+		tail.StopAtEOF()
+		close(done)
+	}()
+
+	// Verify the goroutine is blocked indefinitely.
+	select {
+	case <-done:
+		t.Fatalf("StopAtEOF() completed unexpectedly")
+	case <-time.After(time.Second * 1):
```
adam-mateen, line 96: Waiting 1 second is arbitrary.
It could be increased, but I am trying not to increase unit test execution time _too much_.

plugins/inputs/logfile/logfile_test.go
```diff
@@ -358,100 +358,128 @@ func TestLogsFileRemove(t *testing.T) {
 	tt.Stop()
 }
 
-//When another file is created for the same file config and the file config has auto_removal as true, the old files will stop at EOF and removed afterwards
-func TestLogsFileAutoRemoval(t *testing.T) {
```
adam-mateen, line 362: This old test seems to create a temp file and write 1 line to it.
Then spawn a goroutine to create a 2nd temp file.
However, the test read the 1 line from the  1st temp file before the goroutine runs and create the 2nd file, so it was not an effective test.

plugins/inputs/logfile/logfile_test.go
```diff
@@ -358,100 +358,128 @@ func TestLogsFileRemove(t *testing.T) {
...
+		//fmt.Printf("write i %v\n", i)
+		_, err := file.WriteString(msg + "\n")
+		require.NoError(t, err)
+	}
+}
+
+// TestLogsFileAutoRemoval verifies when a new file matching the configured
+// FilePath is discovered, the old file will be automatically deleted after
+// being read to the end-of-file.
+func TestLogsFileAutoRemoval(t *testing.T) {
```
adam-mateen, line 388: I realize this test case is 96 lines long and has 2 goroutines (not counting deferred cleanup).
So if you see a way to simplify or shorten it please let me know.

I created the 2 helper functions `getLogSrc()` and `writeSomeLines()` to shorten `TestLogsFileAutoRemoval()`.

plugins/inputs/logfile/logfile_test.go
```diff
@@ -358,100 +358,128 @@ func TestLogsFileRemove(t *testing.T) {
...
+	// Override global in tailersrc.go.
 	multilineWaitPeriod = 10 * time.Millisecond
-	logEntryString := "anything"
+
 	filePrefix := "file_auto_removal"
 	tmpfile1, err := createTempFile("", filePrefix)
+	fmt.Printf("Created 1st temp file, %s\n", tmpfile1.Name())
```
adam-mateen, line 394: I left in a lot of `fmt.Print` calls in the test because they definitely help debug when running `go test  ./plugins/inputs/logfile -v`.
They also make the order of operations between goroutines clear.
I am fine removing them if someone strongly opposed it.

plugins/inputs/logfile/tail/tail.go
```diff
@@ -510,8 +510,13 @@ func (tail *Tail) sendLine(line string, offset int64) bool {
 		select {
 		case tail.Lines <- &Line{line, now, nil, offset}:
 		case <-tail.Dying():
-			tail.dropCnt += len(lines) - i
-			return true
+			if tail.Err() == errStopAtEOF {
+				// Try sending, even if it blocks.
+				tail.Lines <- &Line{line, now, nil, offset}
```
SaxyPandaBear, line 515: 🤔  maybe this is just such an extreme case it wouldn't happen, but don't we want to return something here? Part of me thinks that it doesn't matter since this only gets executed when the tail routine is `Dying()`.

plugins/inputs/logfile/tail/tail_test.go
```diff
@@ -8,8 +8,12 @@ import (
 	"strings"
 	"testing"
 	"time"
+
+	"gotest.tools/assert"
```
SaxyPandaBear, line 12: This looks like a different assert package than what we normally import. Not sure if this was intentional. Didn't see an update to the go.mod or go.sum so I'm not too bothered but wanted to call it out.

plugins/inputs/logfile/tail/tail_test.go
```diff
@@ -8,8 +8,12 @@ import (
...
 )
 
+var linesWrittenToFile int = 10
```
SaxyPandaBear, line 15: Unless I skimmed this file incorrectly, this looks like it could be a const.

plugins/inputs/logfile/tail/tail_test.go
SaxyPandaBear: Just to confirm, would the channel ever get closed if we waited long enough? 🤔  I'm fine with this, just wanted to see.

plugins/inputs/logfile/tail/tail_test.go
```diff
@@ -71,22 +76,46 @@ func TestNotTailedCompeletlyLogging(t *testing.T) {
...
+	case <-done:
+		t.Fatalf("StopAtEOF() completed unexpectedly")
+	case <-time.After(time.Second * 1):
+		fmt.Println("timeout waiting for StopAtEOF() (as expected)")
+	}
+
+	assert.Equal(t, errStopAtEOF, tail.Err())
+
+	// Read to EOF
+	for i := 0; i < linesWrittenToFile - 3; i++ {
```
SaxyPandaBear, line 103: I haven't connected the dots on where the `- 3` comes from for this. Can you elaborate how this becomes the end of file?

plugins/inputs/logfile/tailersrc.go
SaxyPandaBear, line 80: For this, what happens if we don't implement LogSrc? Does this fail the build? I think that's fair. You could make a unit test that creates a tailerSrc and then cast it to LogSrc, and verify it casted ok:
```go
_, ok := someTailer.(logs.LogSrc)
assert.True(t, ok)
```
or something like that. Putting it in here looks a little weird to me, but not a blocker.

plugins/inputs/logfile/tail/tail.go
```diff
@@ -390,7 +391,7 @@ func (tail *Tail) tailFileSync() {
 						if errReadLine == nil {
 							tail.sendLine(line, tail.curOffset)
 						} else {
-							break
+							return
```
SaxyPandaBear, line 394: 🤔  because of the return [a couple lines down](https://github.com/aws/amazon-cloudwatch-agent/pull/452/files#diff-b4cbda30ad21153305e4658dca8a5844580e52b12544b5147237a3315d185284R400), this doesn't change any behavior, right?

SaxyPandaBear (commented):
Haven't gone through `logfile_test.go` yet but have to step away and don't want my comments to disappear

plugins/inputs/logfile/tail/tail.go
adam-mateen: There is only 1 for loop in the `tailFileSync()` method.
And there is no code after the end of the for loop.
So a `break` is effectively the same as `return` here.
It caught my eye and confused me why `break` was used on this line, when `return` is used _everywhere else_ within `tailFileSync()`.

So this change is just for consistency and to reduce my confusion.

plugins/inputs/logfile/tail/tail.go
adam-mateen: This is inside a for loop, although it typically loops just once, unless the line is huge.
Regardless, no I don't think we need to return.
The return type of this method is `bool` and there is a `return true` at the bottom of the method.
If the `Tail` is dying because of `errStopAtEOF` we want to act as if it is NOT dying at all.

plugins/inputs/logfile/tail/tail_test.go
adam-mateen: Vscode auto import.
I will fix it.
Good catch.

plugins/inputs/logfile/tail/tail_test.go
adam-mateen: I forgot const existed...
Will fix.

plugins/inputs/logfile/tail/tail_test.go
adam-mateen: Channel should not close until `StopAtEOF()` returns, and that should not happen until all lines are read from the file.
`setup()` writes 10 lines, and this test case has only read 3 so far.

plugins/inputs/logfile/tail/tail_test.go
adam-mateen: The test already called a function `readThreelines()`.

plugins/inputs/logfile/tailersrc.go
adam-mateen: If `tailerSrc` did not fully implement `logs.LogSrc` then this would be a compile time error.
I think compile time error is preferable to a seperate test.

To be honest, I was iffy about adding this because it seems counterproductive to Golang not requiring a class to explicitly list out the interfaces it implements.

In this case it actually helped me realize that the interface was not being met, so it definitely provided value.

plugins/inputs/logfile/logfile_test.go
SaxyPandaBear: Rather than use `fmt.Print()`, I'd switch to using the test logger: `t.Log()` I think?

plugins/inputs/logfile/tailersrc.go
SaxyPandaBear: Makes sense to me. I'm fine keeping it like this and thanks for the link for context.

plugins/inputs/logfile/logfile_test.go
```diff
@@ -358,100 +358,128 @@ func TestLogsFileRemove(t *testing.T) {
...
+		if e != nil {
+			evts <- e
+		}
+	})
+	return &logSource, evts
+}
 
+func writeSomeLines(t *testing.T, file *os.File, numLines int, msg string) {
+	for i := 0; i < numLines; i++ {
+		//time.Sleep(1 * time.Millisecond)
```
SaxyPandaBear, line 378: looks like you could remove the commented out code here.

plugins/inputs/logfile/logfile_test.go
```diff
@@ -358,100 +358,128 @@ func TestLogsFileRemove(t *testing.T) {
...
+	logFile.Log = TestLogger{t}
+	logFile.FileConfig = []FileConfig{{
 		FilePath:      filepath.Join(filepath.Dir(tmpfile1.Name()), filePrefix+"*"),
 		FromBeginning: true,
 		AutoRemoval:   true,
 	}}
-	tt.FileConfig[0].init()
-	tt.started = true
+	logFile.FileConfig[0].init()
+	logFile.started = true
```
SaxyPandaBear, line 406: not super important, but looks like this set up (minus the defer `Stop()`) _could_ be extracted into a helper, but you'd only save like 5 lines of code in this test.

plugins/inputs/logfile/logfile_test.go
```diff
@@ -358,100 +358,128 @@ func TestLogsFileRemove(t *testing.T) {
...
+	var	evts2 chan logs.LogEvent
 	defer func() {
+		if logSource2 != nil {
+			(*logSource2).Stop()
+		}
 		if tmpfile2 != nil {
+			tmpfile2.Close()
 			os.Remove(tmpfile2.Name())
 		}
 	}()
```
SaxyPandaBear, line 428: _Could_ extract this out to a real function rather than in-lining it, but again only a couple lines saved. That being said, I don't think I see a similar cleanup for tmpfile1.

plugins/inputs/logfile/logfile_test.go
```diff
@@ -358,100 +358,128 @@ func TestLogsFileRemove(t *testing.T) {
...
-		lsrcs = tt.FindLogSrc()
-		if len(lsrcs) > 0 {
-			break
-		}
-		time.Sleep(1 * time.Second)
+	fmt.Println("Verify reader completed.")
+	select {
+	case <-readerDone:
+		fmt.Println("Completed before timeout (as expected)")
+	case <-time.After(time.Second * 10):
```
SaxyPandaBear, line 464: Is this 10 seconds meaningful? Ultimately this is for catching a hanging test, so it's not a big deal to leave it like this if this is an arbitrary wait time. I just wanted to see if there was a reason why it was so long.

plugins/inputs/logfile/logfile_test.go
```diff
@@ -358,100 +358,128 @@ func TestLogsFileRemove(t *testing.T) {
...
+				defer close(fileCreatorDone)
+				tmpfile2, err = createTempFile("", filePrefix)
+				fmt.Printf("Created 2nd temp file, %s\n", tmpfile2.Name())
+				require.NoError(t, err)
+				_, err = tmpfile2.WriteString(logEntryString + "\n")
+				require.NoError(t, err)
+				logSource2, evts2 = getLogSrc(t, logFile)
+			}()
+		}
 	}()
```
SaxyPandaBear, line 458: I'll look at this again in the morning, but this _looks_ like it could be pulled out/parameterized as a separate validation function - it would mess up some of the channel reads down below, but I don't think it would be terrible. At the end of the day, I'm not going to lose sleep if we have a unit test with 100 lines of code in it - skimming over this test, it is almost like an integration test in what it validates.

plugins/inputs/logfile/tail/tail.go
```diff
@@ -510,8 +510,13 @@ func (tail *Tail) sendLine(line string, offset int64) bool {
 		select {
 		case tail.Lines <- &Line{line, now, nil, offset}:
 		case <-tail.Dying():
-			tail.dropCnt += len(lines) - i
-			return true
+			if tail.Err() == errStopAtEOF {
+				// Try sending, even if it blocks.
+				tail.Lines <- &Line{line, now, nil, offset}
+			} else {
```
khanhntd, line 516: Curious question (don't need to address in this PR): if there would be an initial stop request but some of the customers want to read until the EOF, would we want to stop the request or sending until it reaches EOF and sends those lines?

plugins/inputs/logfile/logfile_test.go
adam-mateen: Good call.
Will fix.

plugins/inputs/logfile/logfile_test.go
adam-mateen: No, if the test passes it doesn't matter how long the timeout is.
It just needs to be large enough that any host running this unit test (e.g. 1 core, 500 MHz) can write/read/recv_events.
I will lower it to 5.

plugins/inputs/logfile/logfile_test.go
adam-mateen: I did some refactoring.
Split out code into some new functions.
`TestLogsFileAutoRemoval()` is now 14 lines long.
The longest helper function is 47 lines.

SaxyPandaBear (commented):
Forgot to hit submit.

plugins/inputs/logfile/tail/tail_test.go
```diff
@@ -71,22 +76,46 @@ func TestNotTailedCompeletlyLogging(t *testing.T) {
...
+	done := make(chan bool)
+	go func() {
+		tail.StopAtEOF()
+		close(done)
+	}()
+
+	// Verify the goroutine is blocked indefinitely.
+	select {
+	case <-done:
+		t.Fatalf("StopAtEOF() completed unexpectedly")
```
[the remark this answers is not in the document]
adam-mateen: Agreed.
Will fix.

plugins/inputs/logfile/tail/tail_test.go
```diff
@@ -56,6 +60,7 @@ func (l *testLogger) Info(args ...interface{}) {
 
 func TestNotTailedCompeletlyLogging(t *testing.T) {
 	tmpfile, tail, tlog := setup(t)
+	tmpfile.Close()
```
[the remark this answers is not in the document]
adam-mateen: Now it makes sense why it wasn't already here.
Will remove.
Thanks.

plugins/inputs/logfile/tail/tail.go
adam-mateen: `sendLine()` will keep getting called until `Tail` reaches EOF.
Which is the behavior _I think_ all users want and expect.

plugins/inputs/logfile/logfile_test.go
```diff
@@ -358,100 +358,116 @@ func TestLogsFileRemove(t *testing.T) {
...
-			break
+// createWriteRead creates a temp file, writes to it, then verifies events
+// are received. If isParent is true, then spawn a 2nd goroutine for createWriteRead.
+// Close the given channel when complete to let caller know it was successful.
+func createWriteRead(t *testing.T, prefix string, logFile *LogFile, done chan bool, isParent bool) {
+	// Let caller know when the goroutine is done.
+	defer close(done)
+	// done2 is only passed to child if this is the parent.
+	done2 := make(chan bool)
+	file := makeTempFile(t, prefix)
```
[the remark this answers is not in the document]
adam-mateen: Not really.
If the program (test) never reached the `file.Close()` line then it must have crashed.
And if it crashed, then all file handles are released/closed anyway.
At least in `C` you don't need to bother calling `free()` on memory or closing files when the program is exiting.
https://stackoverflow.com/questions/15246833/is-it-a-good-practice-to-close-file-descriptors-on-exit

plugins/inputs/logfile/tail/tail_test.go
```diff
@@ -71,22 +75,46 @@ func TestNotTailedCompeletlyLogging(t *testing.T) {
...
+
+	// Read to EOF
+	for i := 0; i < linesWrittenToFile - 3; i++ {
+		<-tail.Lines
+	}
+
+	// Verify StopAtEOF() has completed.
+	select {
+	case <-done:
+		fmt.Println("StopAtEOF() completed (as expected)")
```
SaxyPandaBear, line 109: One more place where it should use `t.Log`

adam-mateen: Oh wow, sorry 😅 
I will do a thorough code search this time...

SaxyPandaBear:
```
--- FAIL: TestLogsFileAutoRemoval (0.06s)
    logfile_test.go:376: Created temp file, /tmp/file_auto_removal239475258
    logfile_test.go:397: Fill temp file with sufficient lines to be read.
    logfile_test.go:428: Verify every line written to the temp file is received.
    logfile_test.go:376: Created temp file, /tmp/file_auto_removal696020049
    logfile_test.go:385: 
        	Error Trace:	logfile_test.go:385
        	            				logfile_test.go:417
        	            				asm_amd[64](https://github.com/aws/amazon-cloudwatch-agent/runs/6159312728?check_suite_focus=true#step:7:64).s:1374
        	Error:      	Not equal: 
        	            	expected: 1
        	            	actual  : 0
        	Test:       	TestLogsFileAutoRemoval
    logfile_test.go:442: Verify child completed.
    logfile_test.go:445: Completed before timeout (as expected)
    logfile_test.go:449: Verify 1st temp file was auto deleted.
    logfile_test.go:451: 
        	Error Trace:	logfile_test.go:451
        	            				logfile_test.go:463
        	Error:      	Should be true
        	Test:       	TestLogsFileAutoRemoval
    logfile_test.go:464: Verify 1st tmp file created and discovered.
    logfile_test.go:467: Completed before timeout (as expected)
FAIL
coverage: [66](https://github.com/aws/amazon-cloudwatch-agent/runs/6159312728?check_suite_focus=true#step:7:66).9% of statements
```
Don't recall seeing this test fail the last time. Just noting this here in case this is a source of flakiness

SaxyPandaBear:
@adam-mateen 
```
Error: plugins\inputs\windows_event_log\windows_event_log_test.go:10:2: imported and not used: "fmt"
```
looks like this needs to be cleaned up for the Windows tests. Otherwise LGTM

SaxyPandaBear:
@adam-mateen can you try running the unit tests locally on a **Linux** machine? It seems to me like there is some flakiness introduced with the testing changes here for some reason for the linux unit tests

adam-mateen:
> @adam-mateen can you try running the unit tests locally on a **Linux** machine? It seems to me like there is some flakiness introduced with the testing changes here for some reason for the linux unit tests

will do

SaxyPandaBear (commented):
A little concerned about test flakiness on Linux, but just giving verbal confirmation that the PR looks good to me. Because of my concern, won't be formally approving without hearing back about what happens when you run the tests on Linux to see if it reproduces flakiness. Just based on the GitHub runner, it seemed to work every other execution, which isn't very good. I know we already have some tests that are flaky on windows and macos, but I just want to avoid adding something else that we have to constantly retry

SaxyPandaBear (approved):
lgtm

khanhntd (approved):
LGTM.

plugins/inputs/logfile/logfile_test.go
```diff
@@ -417,7 +417,9 @@ func createWriteRead(t *testing.T, prefix string, logFile *LogFile, done chan bo
 	logSrc, evts := getLogSrc(t, logFile)
 	defer (*logSrc).Stop()
 	defer close(evts)
-	const numLines int = 1000
+	// Choose a large enough number of lines so that even high-spec hosts will not
+	// complete receiving logEvents before the 2nd createWriteRead() goroutine begins.
+	const numLines int = 100000
```
adam-mateen, line 422: I ran this test 50 times on my macbook and on a high spec EC2 instance (m5.2xlarge) running AL2:
```
go test ./plugins/inputs/logfile -run TestLogsFileAutoRemoval -count=50 -v
```

Example output from my macbook:
```
=== RUN   TestLogsFileAutoRemoval
    logfile_test.go:376: Created temp file, /var/folders/7c/rx6r0v6n44j4pwnc_77gt46w0000gs/T/file_auto_removal2578471407
2022/05/16 08:39:48 The state file  for /var/folders/7c/rx6r0v6n44j4pwnc_77gt46w0000gs/T/file_auto_removal2578471407 does not exist: stat : no such file or directory
    logfile_test.go:397: Fill temp file with sufficient lines to be read.
2022/05/16 08:39:48 W! [inputs.tail] Stopping tail as file no longer exists: /var/folders/7c/rx6r0v6n44j4pwnc_77gt46w0000gs/T/file_auto_removal3828958604
    logfile_test.go:430: Verify every line written to the temp file is received.
    logfile_test.go:376: Created temp file, /var/folders/7c/rx6r0v6n44j4pwnc_77gt46w0000gs/T/file_auto_removal1902835504
    logfile_test.go:444: Verify child completed.
2022/05/16 08:39:49 The state file  for /var/folders/7c/rx6r0v6n44j4pwnc_77gt46w0000gs/T/file_auto_removal1902835504 does not exist: stat : no such file or directory
    logfile_test.go:397: Fill temp file with sufficient lines to be read.
    logfile_test.go:430: Verify every line written to the temp file is received.
    logfile_test.go:447: Completed before timeout (as expected)
    logfile_test.go:451: Verify 1st temp file was auto deleted.
2022/05/16 08:39:49 W! [logfile] Failed to auto remove file /var/folders/7c/rx6r0v6n44j4pwnc_77gt46w0000gs/T/file_auto_removal1902835504: remove /var/folders/7c/rx6r0v6n44j4pwnc_77gt46w0000gs/T/file_auto_removal1902835504: no such file or directory
    logfile_test.go:466: Verify 1st tmp file created and discovered.
    logfile_test.go:469: Completed before timeout (as expected)
--- PASS: TestLogsFileAutoRemoval (1.33s)
```

And on my AL2 EC2 instance:
```
=== RUN   TestLogsFileAutoRemoval
    logfile_test.go:376: Created temp file, /tmp/file_auto_removal890554952
2022/05/16 13:40:35 The state file  for /tmp/file_auto_removal890554952 does not exist: stat : no such file or directory
    logfile_test.go:397: Fill temp file with sufficient lines to be read.
    logfile_test.go:428: Verify every line written to the temp file is received.
2022/05/16 13:40:35 W! [inputs.tail] Stopping tail as file no longer exists: /tmp/file_auto_removal235327309
    logfile_test.go:376: Created temp file, /tmp/file_auto_removal076276999
2022/05/16 13:40:35 The state file  for /tmp/file_auto_removal076276999 does not exist: stat : no such file or directory
    logfile_test.go:442: Verify child completed.
    logfile_test.go:397: Fill temp file with sufficient lines to be read.
    logfile_test.go:428: Verify every line written to the temp file is received.
    logfile_test.go:445: Completed before timeout (as expected)
    logfile_test.go:449: Verify 1st temp file was auto deleted.
2022/05/16 13:40:36 W! [logfile] Failed to auto remove file /tmp/file_auto_removal076276999: remove /tmp/file_auto_removal076276999: no such file or directory
    logfile_test.go:464: Verify 1st tmp file created and discovered.
    logfile_test.go:467: Completed before timeout (as expected)
--- PASS: TestLogsFileAutoRemoval (1.03s)
```

khanhntd (approved):
There is a flakey test in linux 
```
--- FAIL: TestOffsetDoneCallBack (10.01s)
    tailersrc_test.go:312: Not enough logs have been processed, only 30 are processed
FAIL
```
However, everything besides that looks good to me

plugins/inputs/logfile/logfile.go
```diff
@@ -167,9 +167,11 @@ func (t *LogFile) FindLogSrc() []logs.LogSrc {
 
 			if _, ok := dests[filename]; ok {
 				continue
-			} else if fileconfig.AutoRemoval { // This logic means auto_removal does not work with publish_multi_logs
+			} else if fileconfig.AutoRemoval {
+				// This logic means auto_removal does not work with publish_multi_logs
 				for _, dst := range dests {
-					dst.tailer.StopAtEOF() // Stop all other tailers in favor of the newly found file
+					// Stop all other tailers in favor of the newly found file
+					dst.tailer.StopAtEOF()
```
adam-mateen, line 174: this no longer blocks!

PR #453, opened 2022-04-22
Add scrapping ECS_CONTAINER_METADATA_URI_V4 for ECS 

### LINKED ISSUE #118: Fargate instances incorrectly detected as OnPrem

khanhntd:
Fixed #118 

# Description of the issue
During 2020 Oct,  for running Task Definition with Container Agent > `v1.39.0`,  the CWAgent needs to scrap the `ECS_CONTAINER_METADATA_URI_V4` in order for it to get the required information such as TaskArn, ClusterName, otherwise, it would fall down to use provider credential file. Would need to create an integration test on each corresponding metadata.

# Description of changes
Adding `ECS_CONTAINER_METADATA_URI_V4` able to detect its on ECS EC2 and able to scrap metadata

# License

# Tests
Create a Fargate Instance and let's the CWAgent run without scrapping any metrics. For more information on how I able to run: https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/ContainerInsights-Prometheus-install-ECS.html. For the result,
image

### DIFF
--- a/amazon-cloudwatch-container-insights/cloudwatch-agent-dockerfile/source/Dockerfile
+++ b/amazon-cloudwatch-container-insights/cloudwatch-agent-dockerfile/source/Dockerfile
@@ -9,8 +9,6 @@ ARG TARGETARCH
 RUN mkdir -p /go/src/github.com/aws/amazon-cloudwatch-agent/
 WORKDIR /go/src/github.com/aws/amazon-cloudwatch-agent/
 
-ENV GOPROXY=direct
-
 ARG GO111MODULE="on"
 ENV GO111MODULE=${GO111MODULE}
 
--- a/amazon-cloudwatch-container-insights/cloudwatch-agent-dockerfile/source/Dockerfile
+++ b/amazon-cloudwatch-container-insights/cloudwatch-agent-dockerfile/source/Dockerfile
@@ -6,6 +6,8 @@ FROM $GO_IMAGE as builder
 # NOTE: This arg will be populated by docker buildx
 # https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope
 ARG TARGETARCH
+    ARG GOPROXY="direct"
+ENV GOPROXY=${GOPROXY}
 RUN mkdir -p /go/src/github.com/aws/amazon-cloudwatch-agent/
 WORKDIR /go/src/github.com/aws/amazon-cloudwatch-agent/
 
--- a/amazon-cloudwatch-container-insights/cloudwatch-agent-dockerfile/source/Dockerfile
+++ b/amazon-cloudwatch-container-insights/cloudwatch-agent-dockerfile/source/Dockerfile
@@ -9,6 +9,8 @@ ARG TARGETARCH
 RUN mkdir -p /go/src/github.com/aws/amazon-cloudwatch-agent/
 WORKDIR /go/src/github.com/aws/amazon-cloudwatch-agent/
 
+ENV GOPROXY=direct
+
 ARG GO111MODULE="on"
 ENV GO111MODULE=${GO111MODULE}
 
--- a/amazon-cloudwatch-container-insights/cloudwatch-agent-dockerfile/source/Dockerfile
+++ b/amazon-cloudwatch-container-insights/cloudwatch-agent-dockerfile/source/Dockerfile
@@ -6,8 +6,6 @@ FROM $GO_IMAGE as builder
 # NOTE: This arg will be populated by docker buildx
 # https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope
 ARG TARGETARCH
-    ARG GOPROXY="direct"
-ENV GOPROXY=${GOPROXY}
 RUN mkdir -p /go/src/github.com/aws/amazon-cloudwatch-agent/
 WORKDIR /go/src/github.com/aws/amazon-cloudwatch-agent/
 
--- a/translator/util/ecsutil/ecsutil.go
+++ b/translator/util/ecsutil/ecsutil.go
@@ -70,9 +70,14 @@ func (e *ecsUtil) IsECS() bool {
 func (e *ecsUtil) getECSMetadata() (em *ecsMetadataResponse, err error) {
 	// Based on endpoint to get ECS metadata, for more information on the respond, https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint.html
 	if v4MetadataEndpoint, ok := os.LookupEnv(v4MetadataEndpointEnv); ok {
-		em, err = e.getMetadataResponse(v4MetadataEndpoint + "/task")
+		if v4FinalMetadataEndpoint, ok := v4MetadataEndpoint.(string); ok {
+			em, err = e.getMetadataResponse(v4FinalMetadataEndpoint + "/task")
+		}
 	} else if v3MetadataEndpoint, ok := os.LookupEnv(v3MetadataEndpointEnv); ok {
-		em, err = e.getMetadataResponse(v3MetadataEndpoint + "/task")
+		if v3FinalMetadataEndpoint, ok := v3MetadataEndpoint.(string); ok {
+			em, err = e.getMetadataResponse(v3FinalMetadataEndpoint + "/task")
+		}
+		
 	} else {
 		em, err = e.getMetadataResponse(v2MetadataEndpoint)
 	}
--- a/translator/util/ecsutil/ecsutil.go
+++ b/translator/util/ecsutil/ecsutil.go
@@ -32,11 +32,9 @@ type ecsUtil struct {
 	httpClient *httpclient.HttpClient
 }
 
-var ecsUtilInstance *ecsUtil
-
 var ecsUtilOnce sync.Once
 
-func GetECSUtilSingleton() *ecsUtil {
+func GetECSUtilSingleton() (ecsUtilInstance *ecsUtil) {
 	ecsUtilOnce.Do(func() {
 		ecsUtilInstance = initECSUtilSingleton()
 	})
--- a/translator/util/ecsutil/ecsutil.go
+++ b/translator/util/ecsutil/ecsutil.go
@@ -32,9 +32,11 @@ type ecsUtil struct {
 	httpClient *httpclient.HttpClient
 }
 
+var ecsUtilInstance *ecsUtil
+
 var ecsUtilOnce sync.Once
 
-func GetECSUtilSingleton() (ecsUtilInstance *ecsUtil) {
+func GetECSUtilSingleton() *ecsUtil {
 	ecsUtilOnce.Do(func() {
 		ecsUtilInstance = initECSUtilSingleton()
 	})
--- a/translator/util/ecsutil/ecsutil.go
+++ b/translator/util/ecsutil/ecsutil.go
@@ -107,14 +107,3 @@ func (e *ecsUtil) parseRegion(em *ecsMetadataResponse) {
 	}
 	e.Region = splitedContent[3]
 }
-
-// There is only one format for ClusterArn (https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_Cluster.html)
-// arn:aws:ecs:region:aws_account_id:cluster/cluster-name
-func (e *ecsUtil) parseClusterName(em *ecsMetadataResponse) {
-	splitedContent := strings.Split(em.Cluster, "/")
-	// When splitting the ClusterName with /, the last is always the cluster name
-	if len(splitedContent) == 0 {
-		log.Printf("E! Invalid cluster arn: %s", em.Cluster)
-	}
-	e.Cluster = splitedContent[len(splitedContent)-1]
-}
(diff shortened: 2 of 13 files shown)


translator/util/ecsutil/ecsutil.go
```diff
@@ -65,26 +68,29 @@ func (e *ecsUtil) IsECS() bool {
...
 	if err != nil {
 		return
 	}
 
 	err = json.Unmarshal(resp, em)
 	if err != nil {
-		log.Printf("E! unable to parse resp from ecsmetadata endpoint, error: %v", err)
-		log.Printf("D! resp content is %s", string(resp))
+		log.Printf("E! Unable to parse response from ecsmetadata endpoint, error: %v", err)
+		log.Printf("D! Respond content is %s", string(resp))
```
SaxyPandaBear, line 93: Nitpick: I think this should be `Response content`, not `Respond content`. Or just drop it entirely:
```suggestion
		log.Printf("D! Content is %s", string(resp))
```

translator/util/sdkutil.go
```diff
@@ -25,18 +25,12 @@ const (
...
 var DetectRegion func(mode string, credsConfig map[string]string) string = detectRegion
 var DetectCredentialsPath func() string = detectCredentialsPath
-var runInAws = os.Getenv(config.RUN_IN_AWS)
 
 func DetectAgentMode(configuredMode string) string {
 	if configuredMode != "auto" {
 		return configuredMode
 	}
-
-	if runInAws == config.RUN_IN_AWS_TRUE {
```
SaxyPandaBear, line 35: Can you explain why you are removing this? From what I recall, this was a pretty impactful bug for running ECS where the agent couldn't determine that it's running on an EC2 but should have.

translator/util/sdkutil.go
```diff
@@ -93,15 +87,13 @@ func detectRegion(mode string, credsConfig map[string]string) (region string) {
...
-	// try to get region from ecs metadata
-	if region == "" && mode == config.ModeEC2 {
-		fmt.Println("I! detect region from ecs")
-		region = defaultECSRegion()
+		fmt.Println("I! Detect region from EC2/ECS")
+		if region = defaultEC2Region(); region == "" {
+			// Try to get region from ecs metadata
+			region = defaultECSRegion()
+		}
 	}
```
SaxyPandaBear, line 95: This _looks_ equivalent to me, but is there a way to unit test this?

translator/util/ecsutil/ecsutil.go
```diff
@@ -65,26 +68,29 @@ func (e *ecsUtil) IsECS() bool {
 }
 
 func (e *ecsUtil) getECSMetadata() (em *ecsMetadataResponse, err error) {
-	// choose available endpoint
-	if v3MetadataEndpoint, ok := os.LookupEnv(v3MetadataEndpointEnv); !ok {
-		em, err = e.getMetadataResponse(v2MetadataEndpoint)
-	} else {
+	// Based on endpoint to get ECS metadata, for more information on the respond, https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint.html
+	if v4MetadataEndpoint, ok := os.LookupEnv(v4MetadataEndpointEnv); ok {
+		em, err = e.getMetadataResponse(v4MetadataEndpoint + "/task")
```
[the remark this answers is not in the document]
SaxyPandaBear: I don't think I understand your comment about panicking here. I also don't see existing code that references shared credentials files or anything. Can you please elaborate on what this comment means?

SaxyPandaBear (commented):
Can you elaborate on what the behavior was before this change? I think I can infer based on the screenshot you provided for testing, but would rather hear it explained and confirmed in the PR than guess.

translator/util/sdkutil.go
khanhntd: Yes I believe the issue was the agent was not able to scrap `metadataendpoint` on `awsvpc network` which happens on ECS using Task Definition > v1.39 because the issue happened around when the task definition new version got release. However, after adding the endpoint`ECS_CONTAINER_METADATA_URI_V4`, it would solve the issue. For upcoming release, I will talk with @sethAmazon to make an integration test based on different task definition and will help us in recognize the changes instead of relying on the flag. However, I could keeping the flag until the `integration test` got done to avoid the impact if that make sense.

translator/util/sdkutil.go
khanhntd: For ECS, an integration test is needed since I have not known to replicate the `Metadataendpont`. For a EC2, it would be the same since it also depends on [EC2Metadata endpoint](https://github.com/aws/amazon-cloudwatch-agent/blob/master/tool/util/util.go#L136-L155)

translator/util/ecsutil/ecsutil.go
```diff
@@ -65,26 +68,29 @@ func (e *ecsUtil) IsECS() bool {
 }
 
 func (e *ecsUtil) getECSMetadata() (em *ecsMetadataResponse, err error) {
-	// choose available endpoint
-	if v3MetadataEndpoint, ok := os.LookupEnv(v3MetadataEndpointEnv); !ok {
-		em, err = e.getMetadataResponse(v2MetadataEndpoint)
-	} else {
+	// Based on endpoint to get ECS metadata, for more information on the respond, https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-metadata-endpoint.html
```
sethAmazon, line 71: Do we still need to support v2 and v3? If endpoint is there but get metadata fails should we fall back?

sethAmazon: IDC either way.

translator/util/sdkutil.go
sethAmazon: Yeah I would keep this until a integration tests is built for this.

translator/util/ecsutil/ecsutil.go
SaxyPandaBear: I think we still need to support v2 and v3 so long as people can create ECS clusters that still rely on them. I think we'd have to investigate further where the drop off points are for v2 and v3 before we could take out support for them.

khanhntd:
Based on my understanding, the flow for getting the credentials based on flow would be:
* **Step 1:** CWAgent will determine [onPrem or EC2](https://github.com/aws/amazon-cloudwatch-agent/blob/master/translator/util/sdkutil.go#L30-L51). For EC2, the criteria would be on metadata endpoint, whether it would be ECS or EC2.
* **Step 2:** After determine the mode, the credentials will be based on:
* Based on `onPrem`, the credentials would take the [default profile](https://github.com/aws/amazon-cloudwatch-agent/blob/master/translator/util/sdkutil.go#L160-L162) and [create a  credentials session](https://github.com/aws/amazon-cloudwatch-agent/blob/master/translator/util/sdkutil.go#L72-L79)
* Based on `EC2`, it would follow   the following procedures
```
  ## Credentials are loaded in the following order
  ## 1) Assumed credentials via STS if role_arn is specified
  ## 2) explicit credentials from 'access_key' and 'secret_key'
  ## 3) shared profile from 'profile'
  ## 4) environment variables
  ## 5) shared credentials file
  ## 6) EC2 Instance Profile
```
* **Step 3:** Submit metrics/log etc
**Note**: On step 2, `onPrem`  credentials are considered before following credential procedure. I have not figured out why this is the behavior.

translator/util/sdkutil.go
khanhntd: On second though, take back my word a little bit. It seems there are a workaround in [build an AWSVPC Fargate](https://github.com/aws/amazon-cloudwatch-agent/blob/master/internal/ecsservicediscovery/decoratedtask_test.go#L50-L136). Would need to take a look at it.

translator/util/sdkutil.go
khanhntd: On second though, should not delete this since EKS Fargate does not know how to detect `onPrem` or not so would need to keep back this flag.

translator/util/ecsutil/ecsutil.go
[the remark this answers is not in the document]
khanhntd: Here is a full explanation based on my understanding https://github.com/aws/amazon-cloudwatch-agent/pull/453#issuecomment-1107099265

SaxyPandaBear:
So after this change, people using ECS on Fargate don't need the `RUN_IN_AWS` flag set? But EKS users do?

khanhntd:
@SaxyPandaBear Yes. That's what I have tested so far. To avoid the impact for EKS users,  should we build a separate CWAgent image for Fargate (we don't have a unit test for this but it does not impact much)  until we could detect the mode `onEC2`  on Fargate? Difference between `onEC2` and `onPrem`  is how the CWAgent configure its credential, for more information

translator/util/ecsutil/ecsutil.go
```diff
@@ -97,7 +103,18 @@ func (e *ecsUtil) parseRegion(em *ecsMetadataResponse) {
...
 }
+
+// There is only one format for ClusterArn (https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_Cluster.html)
+// arn:aws:ecs:region:aws_account_id:cluster/cluster-name
+func (e *ecsUtil) parseClusterName(em *ecsMetadataResponse) {
+	splitedContent := strings.Split(em.Cluster, "/")
+	// When splitting the ClusterName with /, the last is always the cluster name
+	if len(splitedContent) == 0 {
+		log.Printf("E! Invalid cluster arn: %s", em.Cluster)
+	}
```
SaxyPandaBear, line 118: If the ARN is invalid, should we still try to assign the cluster ARN to the last element?

translator/util/sdkutil.go
```diff
@@ -36,14 +35,15 @@ func DetectAgentMode(configuredMode string) string {
...
-
+	
 	if defaultEC2Region() != "" {
 		fmt.Println("I! Detected the instance is EC2")
 		return config.ModeEC2
 	}
 
 	if defaultECSRegion() != "" {
 		fmt.Println("I! Detected the instance is ECS")
+		fmt.Printf("I! Cluster: %s, region: %s \n",defaultECSRegion(),defaultClusterName())
```
SaxyPandaBear, line 46: The values don't map to the right things. The format string is Cluster, Region, but the parameters passed in are in the order of Region, Cluster

translator/util/ecsutil/ecsutil.go
SaxyPandaBear: Never mind. I see it in the old code. I would feel better if it was ported over more accurately to the existing code in `get_ecs_cluster_name`

translator/util/ecsutil/ecsutil.go
```diff
@@ -48,13 +50,14 @@ func initECSUtilSingleton() (newInstance *ecsUtil) {
...
 	ecsMetadataResponse, err := newInstance.getECSMetadata()
+
 	if err != nil {
 		log.Printf("I! access ECS task metadata fail with response %v, assuming I'm not running in ECS.\n", err)
 		return
 	}
 
 	newInstance.parseRegion(ecsMetadataResponse)
-	newInstance.Cluster = ecsMetadataResponse.Cluster
+	newInstance.parseClusterName(ecsMetadataResponse)
```
SaxyPandaBear, line 60: Personally, I prefer explicitly setting the value like it did originally rather than expecting that the `parseClusterName` sets the value for the `Cluster` field, but I'm fine with this.

amazon-cloudwatch-container-insights/cloudwatch-agent-dockerfile/source/Dockerfile
```diff
@@ -9,6 +9,8 @@ ARG TARGETARCH
 RUN mkdir -p /go/src/github.com/aws/amazon-cloudwatch-agent/
 WORKDIR /go/src/github.com/aws/amazon-cloudwatch-agent/
 
+ENV GOPROXY=direct
```
SaxyPandaBear, line 12: Can you confirm why this is needed?

translator/util/sdkutil.go
khanhntd: Sorry forgot to delete it. Mostly it for my testing on other PR.

amazon-cloudwatch-container-insights/cloudwatch-agent-dockerfile/source/Dockerfile
khanhntd: Forgot to delete it again when building docker image though. Deleted it in the newest commit.

translator/util/ecsutil/ecsutil.go
khanhntd: I wrote this code based on behavior of [parsing Region from taskARN](https://github.com/aws/amazon-cloudwatch-agent/pull/453/files#diff-8b21e9e828737c79625db57c62d2f87fabcd2e60434cf9b1c4573e4c15da5324R102-R109). However, I have not had more information regarding this to change behavior. However, for one short notice is that even though the `clusterName` is invalid based on what I have skimmed, it would not affect the whole flow so I am still considering whether we need to fatal if clusterName is invalid.

translator/util/ecsutil/ecsutil.go
khanhntd: I would prefer the same since its does not directly use the Cluster Name but the `Cluster ARN` in this case. Therefore, implicitly setting the value would be fine as long as its clarify the meaning though.

SaxyPandaBear (approved):
LGTM

khanhntd:
> LGTM

Forgot about the test and update sys dependency. Fixed it in the newest commit.

SaxyPandaBear (approved):
lgtm

PR #455, opened 2022-04-26
Change behavior for handling file offsets for rotated files

### LINKED ISSUE #447: Missing Logs With Python Rotating File Handler And Minimal Log Messages
There is potential for logs to go missing and to have partial logging messages when using a rotating file handler within an application that produces very few logs. 

The behavior seems to be when a log file is rotated (ex: `logfile.log` is renamed to `logfile.log.2022-04-18`) and a new message is written to `logfile.log`, if the byte length of the single new message is >= the byte length of the entire old log file, then the new message will be entirely missing or partially missing.

We are using the latest version of the CloudWatch Agent, no code modifications have been made, and there is nothing unusual about our environment.

We are mitigating this issue currently just by decreasing the rate at which log files get rotated.  

Minimal reproducible example:

CloudWatch agent `config.json`:
```json
{
    "agent": {
      "run_as_user": "root",
      "credentials": {
        "role_arn": "arn:aws:iam::------------:role/EC2AdminRole"
      }
    },
    "logs": {
      "logs_collected": {
        "files": {
          "collect_list": [
            {
              "file_path": "/home/ec2-user/amazon-cloudwatch-agent/logdest/logs.log*",
              "log_group_name": "custom-agent-logs",
              "log_stream_name": "{instance_id}"
            }
          ]
        }
      }
    }
  }
  
```
`log_generator.py`:
```python
import glob
import os
import logging
import time
import json

# get root logger
logger = logging.getLogger()
logger.setLevel(logging.INFO)

# rotate our log file every 3 minutes
handler = TimedRotatingFileHandler("logdest/logs.log", when="M", interval=3)
logger.addHandler(handler)

# log a message
logging.info(json.dumps({"Metric": "12345"*10}))
# sleep for 4 minutes so that file will rotate upon next log message
time.sleep(60*4)
# log another message (this one will not appear since byte length of message == byte length of old log file)
logging.info(json.dumps({"Metric": "09876"*10}))
# sleep for another 4 minutes so that file will rotate upon next log message
time.sleep(60*4)
# this message will be partially written
logging.info({"Metric": "1234567890"*10})
```

CloudWatch Logs:
*Notice that the second log message is entirely missing, and that the third message is missing its first 65 bytes.*
| Timestamp | Message | Bytes |
| - | - | - |
2022-04-18T11:04:49.539-04:00|{"Metric": "12345123451234512345123451234512345123451234512345"}|64
2022-04-18T11:13:21.539-04:00|45678901234567890123456789012345678901234567890'}|49

Log Files on EC2:
`logs.log` (114 bytes):
```
{'Metric': '1234567890123456789012345678901234567890123456789012345678901234567890123456789012345678901234567890'}
```
`logs.log.2022-04-18_15-09` (64 bytes):
```
{"Metric": "09876098760987609876098760987609876098760987609876"}
```
`logs.log.2022-04-18_15-05` (64 bytes):
```
{"Metric": "12345123451234512345123451234512345123451234512345"}
```

SaxyPandaBear:
# Description of the issue
Closes #447 

# Description of changes
* Updates the default behavior for our tail plugin configuration to automatically attempt to reopen the file after it is deleted. This change in behavior causes the agent to read the file from the beginning rather than use the offset, so we can't miss log lines
* Updated unit tests that validated the previous behavior, and refactored the Recreate test since it was very long, breaking it up into smaller functions
* Added an integration test based on the repro case provided in #447 to validate 
* I couldn't figure out how to reproduce the issue natively in the integration test code itself, natively in Go, so I opted to copy the repro case and execute that instead. This does what I wanted it to, however it introduces a dependency on Python in the AMIs - I'm working through updating the AMIs currently so don't have a successful run of the integration tests currently but there shouldn't be significant code changes so still publishing the PR now.
* Updated integration test README to reflect missing gcc requirement, and new Python requirement.
* Updated tail `reopen` function to retry on permission failures to account for Windows being slow to recreate files - this is noticeable in unit tests on Windows, but was not noticeable in manual/soak testing on Windows test hosts for this.

# License

# Tests
I manually reproduced the issue on a Linux (AL2) and Windows EC2, and also built the agent locally and verified that the code change fixes the issue. 

After this, I set up pairs of hosts - one host with the latest release of the agent (v1.247350) and one with my custom build.

For each of these pairs of hosts, I performed a simple test:
1. On AL2, consume log lines at 2TPS, rotating regularly 
2. On Windows 2019 Server, consume log lines at 2TPS, rotating regularly
3. On AL2, have the agent consume log lines and configure the agent to automatically remove old files, once. This is to verify that there is no obvious mem/CPU leak when the agent auto removes files with this change.

After I validated all of this and set up the monitoring dashboard for these hosts, I wrote an integration test to reproduce the issue, in order to verify that my code change resolves the issue via an automated integration test.

## logging code to validate auto removal
```python
import logging

logger = logging.getLogger('test')
logger.setLevel(logging.INFO)

# write 100 logs to the log file
fh = logging.FileHandler('foo.log.1')
logger.addHandler(fh)

for i in range(100):
    logger.info('This is the first batch: %d', i)
logger.removeHandler(fh)

fh = logging.FileHandler('foo.log.2')
logger.addHandler(fh)

for i in range(100):
    logger.info('This is the second batch: %d', i)
logger.removeHandler(fh)

fh = logging.FileHandler('foo.log.3')
logger.addHandler(fh)

for i in range(100):
    logger.info('This is the third batch: %d', i)
```
For this, I had the agent monitor `/home/ec2-user/foo.log*` so it would pick up the latest log file each time I stopped writing to the old one, and validated that the agent deleted the files `foo.log.1` and `foo.log.2`.

## Dashboard for metrics
Screen Shot 2022-04-26 at 8 53 06 AM
Screen Shot 2022-04-26 at 8 53 27 AM
Screen Shot 2022-04-26 at 8 54 27 AM

## Integration testing
Added an integration test to accurately reproduce the scenario provided in #447. I tweaked the original script to rotate more quickly, since there is a strict 15 minute timeout currently for the integration tests. The source Python script is located in `integration/test/cloudwatchlogs/resources/write_and_rotate_logs.py` and is invoked directly in the integration test, hence why Python is a required dependency in the integration test AMIs now, with this change. 

### Output of integration test with `ReOpen` change reverted (set back to false)
```
aws_instance.integration-test (remote-exec):         2022-04-25T22:41:17Z I! [logagent] piping log from i-04bc8d8d08febf163/i-04bc8d8d08febf163Rotated(/tmp/rotate_me.log) to cloudwatchlogs with retention -1
aws_instance.integration-test (remote-exec):         2022-04-25T22:41:27Z W! [outputs.cloudwatchlogs] Retried 0 time, going to sleep 151.497393ms before retrying.
aws_instance.integration-test (remote-exec):         2022-04-25T22:41:32Z W! [inputs.tail] Stopping tail as file no longer exists: /tmp/rotate_me.log
aws_instance.integration-test (remote-exec):         2022-04-25T22:41:32Z I! [logagent] Log src has stopped for i-04bc8d8d08febf163/i-04bc8d8d08febf163Rotated(/tmp/rotate_me.log)
aws_instance.integration-test (remote-exec):         2022-04-25T22:41:33Z I! [inputs.logfile] Reading from offset 65 in /tmp/rotate_me.log
aws_instance.integration-test (remote-exec):         2022-04-25T22:41:33Z I! [logagent] piping log from i-04bc8d8d08febf163/i-04bc8d8d08febf163Rotated(/tmp/rotate_me.log) to cloudwatchlogs with retention -1
aws_instance.integration-test (remote-exec):         2022-04-25T22:41:47Z W! [inputs.tail] Stopping tail as file no longer exists: /tmp/rotate_me.log
aws_instance.integration-test (remote-exec):         2022-04-25T22:41:47Z I! [logagent] Log src has stopped for i-04bc8d8d08febf163/i-04bc8d8d08febf163Rotated(/tmp/rotate_me.log)
aws_instance.integration-test (remote-exec):         2022-04-25T22:41:48Z I! [inputs.logfile] Reading from offset 65 in /tmp/rotate_me.log
aws_instance.integration-test (remote-exec):         2022-04-25T22:41:48Z I! [logagent] piping log from i-04bc8d8d08febf163/i-04bc8d8d08febf163Rotated(/tmp/rotate_me.log) to cloudwatchlogs with retention -1
aws_instance.integration-test (remote-exec):         2022-04-25T22:42:02Z I! Profiler is stopped during shutdown
aws_instance.integration-test (remote-exec):         2022-04-25T22:42:02Z I! [agent] Hang on, flushing any cached metrics before shutdown

aws_instance.integration-test (remote-exec): 2022/04/25 22:42:17 Checking i-04bc8d8d08febf163/i-04bc8d8d08febf163Rotated since 2022-04-25T22:41:01Z for 3 expected logs
aws_instance.integration-test (remote-exec): 2022/04/25 22:42:17 Found log: {"Metric": "12345123451234512345123451234512345123451234512345"}
aws_instance.integration-test (remote-exec): 2022/04/25 22:42:17 Found log: 45678901234567890123456789012345678901234567890'}
aws_instance.integration-test (remote-exec): 2022/04/25 22:42:17 Done paginating log events for i-04bc8d8d08febf163/i-04bc8d8d08febf163Rotated and found 2 logs
aws_instance.integration-test (remote-exec):     cwl_util.go:128:
aws_instance.integration-test (remote-exec):         	Error Trace:	cwl_util.go:128
aws_instance.integration-test (remote-exec):         	            				publish_logs_test.go:144
aws_instance.integration-test (remote-exec):         	Error:      	"[{"Metric": "12345123451234512345123451234512345123451234512345"} 45678901234567890123456789012345678901234567890'}]" should have 3 item(s), but has 2
aws_instance.integration-test (remote-exec):         	Test:       	TestRotatingLogsDoesNotSkipLines
aws_instance.integration-test (remote-exec):     cwl_util.go:130:
aws_instance.integration-test (remote-exec):         	Error Trace:	cwl_util.go:130
aws_instance.integration-test (remote-exec):         	            				publish_logs_test.go:144
aws_instance.integration-test (remote-exec):         	Error:      	Not equal:
aws_instance.integration-test (remote-exec):         	            	expected: "{\"Metric\": \"09876098760987609876098760987609876098760987609876\"}"
aws_instance.integration-test (remote-exec):         	            	actual  : "45678901234567890123456789012345678901234567890'}"

aws_instance.integration-test (remote-exec):         	            	Diff:
aws_instance.integration-test (remote-exec):         	            	--- Expected
aws_instance.integration-test (remote-exec):         	            	+++ Actual
aws_instance.integration-test (remote-exec):         	            	@@ -1 +1 @@
aws_instance.integration-test (remote-exec):         	            	-{"Metric": "09876098760987609876098760987609876098760987609876"}
aws_instance.integration-test (remote-exec):         	            	+45678901234567890123456789012345678901234567890'}
aws_instance.integration-test (remote-exec):         	Test:       	TestRotatingLogsDoesNotSkipLines
aws_instance.integration-test (remote-exec): --- FAIL: TestRotatingLogsDoesNotSkipLines (75.67s)
```
Note that I included some `tail` output from the agent log file to see what happened, and it shows the same output as the repro case. Also, you can see that the two logs (instead of expected 3) queried from CloudWatch Logs match the expected outcome of the repro case, where the first log line is fully visible, the second log line is completely missing, and the third log line is published truncated. 

### Output of integration test with change
https://github.com/SaxyPandaBear/amazon-cloudwatch-agent/actions/runs/2229185212
From AL2 test from this integration test run:
```
aws_instance.integration-test (remote-exec): === RUN   TestRotatingLogsDoesNotSkipLines
aws_instance.integration-test (remote-exec): 2022/04/26 20:34:08 Found instance id i-0ece89a3e737ca584
aws_instance.integration-test (remote-exec): 2022/04/26 20:34:08 Copy File resources/config_log_rotated.json to /opt/aws/amazon-cloudwatch-agent/bin/config.json
aws_instance.integration-test (remote-exec): 2022/04/26 20:34:08 File resources/config_log_rotated.json abs path /home/ec2-user/amazon-cloudwatch-agent/integration/test/cloudwatchlogs/resources/config_log_rotated.json
aws_instance.integration-test (remote-exec): 2022/04/26 20:34:08 File : resources/config_log_rotated.json copied to : /opt/aws/amazon-cloudwatch-agent/bin/config.json
aws_instance.integration-test (remote-exec): 2022/04/26 20:34:08 Agent has started
aws_instance.integration-test: Still creating... [2m50s elapsed]
aws_instance.integration-test (remote-exec):     publish_logs_test.go:99: Writing logs and rotating
aws_instance.integration-test: Still creating... [3m0s elapsed]
aws_instance.integration-test: Still creating... [3m10s elapsed]
aws_instance.integration-test: Still creating... [3m20s elapsed]
aws_instance.integration-test: Still creating... [3m30s elapsed]
aws_instance.integration-test: Still creating... [3m40s elapsed]
aws_instance.integration-test (remote-exec): 2022/04/26 20:35:08 Agent is stopped
aws_instance.integration-test (remote-exec): 2022/04/26 20:35:08 Checking i-0ece89a3e737ca584/i-0ece89a3e737ca584Rotated since 2022-04-26T20:34:08Z for 3 expected logs
aws_instance.integration-test (remote-exec): 2022/04/26 20:35:08 Done paginating log events for i-0ece89a3e737ca584/i-0ece89a3e737ca584Rotated and found 3 logs
aws_instance.integration-test (remote-exec): --- PASS: TestRotatingLogsDoesNotSkipLines (60.69s)
aws_instance.integration-test (remote-exec): PASS
aws_instance.integration-test (remote-exec): ok  	github.com/aws/amazon-cloudwatch-agent/integration/test/cloudwatchlogs	122.873s
```

#Requirements
1. Run `make linter`


integration/test/cloudwatchlogs/publish_logs_test.go
```diff
@@ -93,6 +93,53 @@ func TestWriteLogsToCloudWatch(t *testing.T) {
...
+	// TODO: this only works for EC2 based testing
+	client := imds.NewFromConfig(c)
+	metadata, err := client.GetInstanceIdentityDocument(ctx, &imds.GetInstanceIdentityDocumentInput{})
+	if err != nil {
+		t.Fatalf("Error occurred while retrieving EC2 instance ID: %v", err)
+	}
+	instanceId := metadata.InstanceID
+	log.Printf("Found instance id %s", instanceId)
+	logGroup := instanceId
+	logStream := instanceId + "Rotated"
```
SaxyPandaBear, line 118: The reason I opted to use a different log stream name here is because as I was trying out the integration test, sometimes it would pick up log events from the other CWL integration test. Rather than introducing a sleep timer here to account for that, I just broke this out into it's own log group/stream.

integration/test/cwl_util.go
```diff
@@ -75,6 +74,62 @@ func ValidateLogs(t *testing.T, logGroup, logStream string, numExpectedLogs int,
 	assert.Equal(t, numExpectedLogs, numLogsFound)
 }
 
+// ValidateLogsInOrder takes a log group, log stream, a list of specific log lines and a timestamp.
+// It should query the given log stream for log events, and then confirm that the log lines that are
+// returned match the expected log lines.
+func ValidateLogsInOrder(t *testing.T, logGroup, logStream string, logLines []string, since time.Time) {
```
SaxyPandaBear, line 80: The code here looks very similar to the original `ValidateLogs` function, and I did start to consolidate them, but decided against it because they validate different things, and breaking that out to generalize this pattern seemed like it would cause more confusion than anything else. Namely, this function validates that the log events we get from GetLogEvents matches exactly, in order, the provided `logLines` input, whereas the original validation function only checks the number of log events. Another difference, though not as big, is that this uses the `StartFromHead` API parameter to sort the events chronologically when retrieving them from CWL.

Basically, I didn't want to create a generalized pattern for doing "validation" inside a validation function, like:
```go
func ValidateLogs(t *testing.T, ..., fn func(events []OutputLogEvents) bool) {
  // get log events since timestamp
  evts := paginagateAndDoStuff()

  // perform generalized validation
  if !fn(evts) {
     t.Fail()
  }
}
```
Though cleaner than duplicating this code, I didn't think it provided much value. If anyone feels strongly that I should still consolidate, I can do that. I'm just explaining why I stopped myself from doing it before publishing the PR

integration/test/cloudwatchlogs/resources/write_and_rotate_logs.py
```diff
@@ -0,0 +1,23 @@
+import json
```
sethAmazon, line 1: Can this be written in go? One of the goals of moving away from our current tests to keep single language tests. Like before we had java test but a golang code base now we want golang tests with the golang code base.

integration/test/cloudwatchlogs/publish_logs_test.go
```diff
@@ -93,6 +93,53 @@ func TestWriteLogsToCloudWatch(t *testing.T) {
...
+	// so we need to determine the host's instance ID for validation
+	ctx := context.Background()
+	c, err := config.LoadDefaultConfig(ctx)
+	if err != nil {
+		// fail fast so we don't continue the test
+		t.Fatalf("Error occurred while creating SDK config: %v", err)
+	}
+
+	// TODO: this only works for EC2 based testing
+	client := imds.NewFromConfig(c)
```
sethAmazon, line 110: I think getting the instance id already has a method in the util.

SaxyPandaBear: Ah good point. I forgot if I merged main back into my branch or not. I'll fix this

integration/test/cloudwatchlogs/publish_logs_test.go
sethAmazon: These tests should be running 1 at a time. What command did you use to start this test or did you run in the integration tests pipeline. If you ran in the pipeline and the tests were not running 1 at a time that is a bug.

integration/test/cloudwatchlogs/publish_logs_test.go
```diff
@@ -93,6 +93,53 @@ func TestWriteLogsToCloudWatch(t *testing.T) {
 	}
 }
 
+// Validate https://github.com/aws/amazon-cloudwatch-agent/issues/447
+func TestRotatingLogsDoesNotSkipLines(t *testing.T) {
```
sethAmazon, line 97: Should we split this out into a new dir so it can run at the same time or is it fine being apart of the same test?

integration/test/cwl_util.go
sethAmazon: I think this is fine.

integration/test/cloudwatchlogs/publish_logs_test.go
SaxyPandaBear: Personally I think it's fine being in here, but I'll defer to you if you think it should be separated into it's own dir and executed concurrently. I feel like this is pretty small in the grand scheme of our testing strategy.

integration/test/cloudwatchlogs/publish_logs_test.go
SaxyPandaBear: I can dig up one of the old pipeline test runs, but basically what happened was the call to GetLogEvents would pick up the very last log line from the previous test.

integration/test/cloudwatchlogs/publish_logs_test.go
sethAmazon: IDC you wrote the code, your choice.

integration/test/cloudwatchlogs/publish_logs_test.go
SaxyPandaBear:
```
aws_instance.integration-test (remote-exec):         	Error Trace:	cwl_util.go:127
aws_instance.integration-test (remote-exec):         	            				publish_logs_test.go:137
aws_instance.integration-test (remote-exec):         	Error:      	"[   Apr 25 18:21:43.997 - [bar] #99 This is a log line. 12345678901234567890123456789012345678901234567890]" should have 3 item(s), but has 5
```
Ignore the actual len assertion, I messed that up. But this test from https://github.com/SaxyPandaBear/amazon-cloudwatch-agent/runs/6162948889?check_suite_focus=true has `Apr 25 18:21:43.997 - [bar] #99 This is a log line.`, which is from the above test. So like I said, rather than introducing more sleeping (and potentially needing to bump up the test timeout), I just opted to direct these logs to a different stream.

integration/test/cloudwatchlogs/publish_logs_test.go
SaxyPandaBear: I'll leave it here then cause it's easier on me

integration/test/cloudwatchlogs/publish_logs_test.go
SaxyPandaBear: Fixed this

sethAmazon (approved):

SaxyPandaBear:
```
aws_instance.integration-test (remote-exec): === RUN   TestRotatingLogsDoesNotSkipLines
aws_instance.integration-test (remote-exec): 2022/04/26 14:30:49 Found instance id i-0153f97e734d72bd4
aws_instance.integration-test (remote-exec): 2022/04/26 14:30:49 Copy File resources/config_log_rotated.json to /opt/aws/amazon-cloudwatch-agent/bin/config.json
aws_instance.integration-test (remote-exec): 2022/04/26 14:30:49 File resources/config_log_rotated.json abs path /home/ec2-user/amazon-cloudwatch-agent/integration/test/cloudwatchlogs/resources/config_log_rotated.json
aws_instance.integration-test (remote-exec): 2022/04/26 14:30:49 File : resources/config_log_rotated.json copied to : /opt/aws/amazon-cloudwatch-agent/bin/config.json
aws_instance.integration-test (remote-exec): 2022/04/26 14:30:49 Agent has started
aws_instance.integration-test: Still creating... [3m0s elapsed]
aws_instance.integration-test (remote-exec):     publish_logs_test.go:99: Writing logs and rotating
aws_instance.integration-test: Still creating... [3m10s elapsed]
aws_instance.integration-test: Still creating... [3m20s elapsed]
aws_instance.integration-test: Still creating... [3m30s elapsed]
aws_instance.integration-test: Still creating... [3m40s elapsed]
aws_instance.integration-test: Still creating... [3m50s elapsed]
aws_instance.integration-test (remote-exec): 2022/04/26 14:31:50 Agent is stopped
aws_instance.integration-test (remote-exec): 2022/04/26 14:31:50 Checking i-0153f97e734d72bd4/i-0153f97e734d72bd4Rotated since 2022-04-26T14:30:49Z for 3 expected logs
aws_instance.integration-test (remote-exec): 2022/04/26 14:31:50 Done paginating log events for i-0153f97e734d72bd4/i-0153f97e734d72bd4Rotated and found 3 logs
aws_instance.integration-test (remote-exec): --- PASS: TestRotatingLogsDoesNotSkipLines (61.20s)
aws_instance.integration-test (remote-exec): PASS
aws_instance.integration-test (remote-exec): ok  	github.com/aws/amazon-cloudwatch-agent/integration/test/cloudwatchlogs	124.025s
```
Successfully ran the integ test

SaxyPandaBear:
```
--- FAIL: TestLogsFileRemove (0.50s)
    logfile_test.go:357: tailerSrc should have reopened after deletion
FAIL
```
Interesting.

integration/test/cwl_util.go
```diff
@@ -75,6 +75,65 @@ func ValidateLogs(t *testing.T, logGroup, logStream string, numExpectedLogs int,
...
+			break
+		}
+
+		nextToken = output.NextForwardToken
+	}
+
+	// Validate that each of the logs are found, in order and in full.
+	assert.Len(t, foundLogs, len(logLines))
+	for i := 0; i < len(logLines); i++ {
+		expected := strings.ReplaceAll(logLines[i], "'", "\"")
```
khanhntd, line 131: Personally, `assert.Equal(t, expected, actual)` would confirm `assert.Len(t, foundLogs, len(logLines))` but I guess its fine.

plugins/inputs/logfile/logfile_test.go
```diff
@@ -714,99 +716,74 @@ func TestLogsFileWithInvalidOffset(t *testing.T) {
...
+	// the file should have rotated, and we should get a new log line
+	validateLogMessageConsumed(t, evts, line2)
 
-	e = <-evts
-	if e.Message() != expectedContent {
-		t.Errorf("Wrong log found after file replacement: \n% x\nExpecting:\n% x\n", e.Message(), expectedContent)
+	go recreateLogFileWithNewMessage(t, files, tmpfile, line3)
+	select {
+	case tmpfile = <-files:
+	case <-time.After(5 * time.Second):
```
khanhntd, line 783: This could be use with 10 * delay

khanhntd:
Not sure why `ReOpen` false would be the expected behavior for CWAgent if the offset would be the issue? If now `ReOpen` should have expected to be true, should be set this to always true instead of keeping this flag?

integration/test/cwl_util.go
SaxyPandaBear: The original behavior could truncate part of the log that gets published. We want to validate that the entire log message was sent to CWL.

integration/test/cloudwatchlogs/resources/write_and_rotate_logs.py
SaxyPandaBear: Appreciate the context. I think to get this through the door, I'll hold off on trying to reinvent the wheel in Go, but might revisit this when I free up.

SaxyPandaBear:
```
--- FAIL: TestOffsetDoneCallBack (10.01s)
    tailersrc_test.go:312: Not enough logs have been processed, only 30 are processed
FAIL
```
noting this for myself as I've seen this fail before but not sure if I made the test flakier

plugins/inputs/logfile/tail/tail.go
```diff
@@ -182,11 +187,25 @@ func (tail *Tail) closeFile() {
...
+				// Retry to allow for some leniency for when the file on the host becomes available.
+				tail.Logger.Debugf("Access denied on %s. Retried %d times so far", tail.Filename, numTries)
+				if numTries < fileOpenMaxRetries {
+					tail.Logger.Debugf("Sleeping for %v ms and retrying", waitDuration.Milliseconds())
+					time.Sleep(waitDuration)
+				}
+				numTries += 1
+				waitDuration *= 2
+				continue
+			}
```
adam-mateen, line 208: 100ms + 200ms + 400ms + 800ms + 1600ms = 3100ms.
I think that is fine since it is only for an edge case on the Windows OS.
Otherwise I'd suggest adding checks in any long running functions/methods to `tail.Dying()` or `tail.Err()` and return early.

plugins/inputs/logfile/tail/tail.go
```diff
@@ -182,11 +187,25 @@ func (tail *Tail) closeFile() {
...
 			if os.IsNotExist(err) {
 				tail.Logger.Debugf("Waiting for %s to appear...", tail.Filename)
 				if err := tail.watcher.BlockUntilExists(&tail.Tomb); err != nil {
```
adam-mateen, line 211: So does this mean the `tail` never dies now?
If the user is monitoring `/tmp/foo*.log` and continuously creates files like:
```
foo1.log
foo2.log
foo3.log
...
```
Even if the old files get deleted will the `tail` live on, stuck in `BlockUntilExists()`?

SaxyPandaBear: I just noticed this myself and am fixing this

SaxyPandaBear: Let me back up

> Even if the old files get deleted will the tail live on, stuck in BlockUntilExists()

I was concerned about this myself. That was partly why I had the last set of metrics I provided in the PR description.
Screen Shot 2022-04-27 at 10 02 21 AM

Let me pull debug logs from both of those hosts

---
What I was referring to "fixing" just now is that I messed up the retry on file permission errors.

adam-mateen: I would expect without `auto_removal` that this function would wait forever for the file to get recreated.

SaxyPandaBear: That's what I expect too. The question is the impact. Will repurpose some hosts I used for this to see what happens when I constantly rotate to a new file.

SaxyPandaBear: What I'm doing now for validating is I'm using this script to generate 10,000 log files:
```
import logging
import time

logger = logging.getLogger('test')
logger.setLevel(logging.INFO)

def write_batch(logger, batch):
    for i in range(100):
        logger.info("This is the %dth batch: %d", batch, i)
        time.sleep(0.001)

for batch in range(1, 10001):
    file_name = 'foo.log.'+str(batch)
    fh = logging.FileHandler(file_name)
    logger.addHandler(fh)
    write_batch(logger, batch)
    logger.removeHandler(fh)
    time.sleep(0.1)
```
I set auto_removal to false on both test hosts.

After all of the log files get written, CPU/memory spike up on both hosts, which makes sense because the agent is now tailing 10k log files, though the log files aren't being updated after the script finishes.

Now, I'm going through and deleting some of the log files, like `rm foo.log.1*` to delete 1000 of the log files, and monitoring the CPU/mem usage as I remove log files to monitor. Basically, I want to see how drastically they diverge in held CPU and memory when I delete log files that it tries to constantly reopen. 

Will post my findings here as I get more data.

SaxyPandaBear: Screen Shot 2022-04-27 at 2 53 07 PM
I'm not very happy with my findings. Only sharing the process CPU usage because the other metrics align pretty closely despite the change in behavior. This is something I would want to drill down on before merging this change. 

Something worth noting - as I was looking at existing unit test behavior, I saw that there is a timer for cleaning up a tail process if the file is closed. This is controlled by the `exitOnDeletion()` function. 

This is currently, with my change here, not executed. 
```go
	if !config.ReOpen {
		go t.exitOnDeletion()
	}
```
I'm thinking that based on the existing durations for this monitoring, which maxes out at 5 minutes, I think that it is safe to remove the `if` check here and just always run this goroutine. I'll make that change and subsequent unit test changes, and then redo this test.

SaxyPandaBear: Forgot to update on this: what happens is the `exitOnDeletion()` never detects that the file is deleted, so this does nothing to resolve the issue. At this point, I feel like this was wasted effort and I need to pivot. I have test hosts and monitoring set up so it should be quick to get stuff set up to monitor for leaks in whatever other solution I come up with, but unfortunately for @akozd , I don't think this is a viable solution without serious degradation.

SaxyPandaBear:
Closing this out for #457 which solves this issue without modifying the `ReOpen` flag, which is too disruptive to existing behavior for me to be comfortable to merge

PR #458, opened 2022-05-02
Add ECS Metadata Integration Test for detect changes in ECS Container Agent Metadata Endpoint

khanhntd:
Fixed https://github.com/aws/amazon-cloudwatch-agent/issues/118 and continue of https://github.com/aws/amazon-cloudwatch-agent/pull/453

# Description of the issue
Adding an integration test which detect the changes in scrapping metadata URL when Container Agent Version changes.

# Description of changes
Able to detect unnoticed changes in URL with ECS Fargate or EC2.

# License

# Tests
- **Manually:** By using `terraform apply -auto-approve -lock=false  -var="test_dir=../../../test/ecs/ecs_metadata/"` to confirm there is a log group which has the `clustername`  which is [scrap from metadata](https://github.com/aws/amazon-cloudwatch-agent/blob/master/translator/translate/logs/metrics_collected/prometheus/ruleLogGroupName.go#L33). 
- **Github Action:**  https://github.com/khanhntd/amazon-cloudwatch-agent/actions/runs/2260465010

# Requirements
2. Run `make linter`


.github/workflows/integrationTest.yml
```diff
@@ -499,4 +557,57 @@ jobs:
         run: terraform init
 
       - name: Terraform destroy
-        run: terraform destroy --auto-approve
\ No newline at end of file
+        run: terraform destroy --auto-approve
+
+  ECSFargateLinuxIntegrationTest:
+    name: 'StopLocalStack'
```
SaxyPandaBear, line 564: I don't think that this is the right name for the job

.github/workflows/integrationTest.yml
```diff
@@ -499,4 +557,57 @@ jobs:
...
+        with:
+          aws-access-key-id: ${{ secrets.TERRAFORM_AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.TERRAFORM_AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-west-2
+
+      - name: Cache if success
+        id: ecs-fargate-linux-integration-test
+        uses: actions/cache@v2
+        with:
+          path: RELEASE_NOTES
```
SaxyPandaBear, line 585: Can we keep the path used here consistent? I see that above, we reference `go.mod` for the cache. I don't really care which one we use as long as we're consistent.

integration/terraform/ecs/linux/README.md
```diff
@@ -0,0 +1,14 @@
+Running integration tests
+=========================
+
+# Run tests in your AWS account
+By running `terraform apply -auto-approve -lock=false -var="test_dir=../../../test/ecs/ecs_metadata/"`, 
+you are agree to setup the following resources:
+* 1 IAM Task Role and 1 Execution Task Role (similar to [these IAM Roles](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/deploy_servicelens_CloudWatch_agent_deploy_ECS.html))
+* 2 SSM Parameter Store
+* 2 Task Definitions and 2 Services for those task definitions
+* 1 Security group which allows all inbound and outbound traffics.
```
SaxyPandaBear, line 10: Why do we need to allow **all** inbound and outbound traffic? It would be better if we know for sure what ports we need.

.github/workflows/integrationTest.yml
```diff
@@ -14,18 +14,73 @@ env:
   PASSPHRASE: ${{ secrets.PASSPHRASE }}
   GPG_KEY_NAME: ${{ secrets.GPG_KEY_NAME }}
   GPG_TTY: $(tty)
+  ECR_INTEGRATION_TEST_REPO: "cwagent-opensource-integration-test"
```
SaxyPandaBear, line 17: I think that this needs to be abstracted like everything else so that we can run this from our forks with some flexibility.

.github/workflows/integrationTest.yml
```diff
@@ -14,18 +14,73 @@ env:
...
+      - name: Login ECR
+        if: steps.build-docker-image.outputs.cache-hit != 'true'
+        id: login-ecr
+        uses: aws-actions/amazon-ecr-login@v1
+
+      - name: Set up Docker Buildx
+        if: steps.build-docker-image.outputs.cache-hit != 'true'
+        uses: docker/setup-buildx-action@v1
+
+      - name: Set up QEMU
```
SaxyPandaBear, line 63: Can you explain what this is for?

integration/terraform/ecs/linux/README.md
```diff
@@ -0,0 +1,14 @@
+Running integration tests
+=========================
+
+# Run tests in your AWS account
+By running `terraform apply -auto-approve -lock=false -var="test_dir=../../../test/ecs/ecs_metadata/"`, 
+you are agree to setup the following resources:
+* 1 IAM Task Role and 1 Execution Task Role (similar to [these IAM Roles](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/deploy_servicelens_CloudWatch_agent_deploy_ECS.html))
+* 2 SSM Parameter Store
+* 2 Task Definitions and 2 Services for those task definitions
```
SaxyPandaBear, line 9: I think this needs more information

integration/test/agent_util.go
```diff
@@ -1,21 +1,20 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: MIT
 
-//go:build linux && integration
-// +build linux,integration
```
SaxyPandaBear, line 5: Why are we removing this?

integration/test/ecs/ecs_metadata/ecs_metadata_test.go
```diff
@@ -0,0 +1,55 @@
...
+const (
+	RetryTime             = 10
+	// Log group format: https://github.com/aws/amazon-cloudwatch-agent/blob/master/translator/translate/logs/metrics_collected/prometheus/ruleLogGroupName.go#L33
+	ECSLogGroupNameFormat = "/aws/ecs/containerinsights/%s/prometheus" 
+)
+
+var clusterName = flag.String("clusterName", "", "Please provide the os preference, valid value: windows/linux.")
+
+func TestNumberMetricDimension(t *testing.T) {
+	// test for cloud watch metrics
```
SaxyPandaBear, line 31: This looks like copy paste

integration/test/ecs/ecs_metadata/ecs_metadata_test.go
```diff
@@ -0,0 +1,55 @@
...
+		}
+		describeLogGroupOutput, err := client.DescribeLogGroups(ctx, &describeLogGroupInput)
+
+		if err != nil {
+			t.Errorf("Error getting log group data %v", err)
+		}
+
+		if len(describeLogGroupOutput.LogGroups) > 0 {
+			break
+		}
```
SaxyPandaBear, line 50: So all we want to verify is that the log group gets created? how do we ensure that this is unique across integration test runs?

integration/test/agent_util.go
khanhntd: Forgot to added it back because of the constraints with my mac and I cannot run the test. However, added back with the newest commit.

integration/test/ecs/ecs_metadata/ecs_metadata_test.go
khanhntd: Forgot to fix this cause focusing on monitoring the integration test on github action.

integration/test/ecs/ecs_metadata/ecs_metadata_test.go
khanhntd: We will create a `unique random hex` every time [a test run](https://github.com/aws/amazon-cloudwatch-agent/pull/458/files/90951f325cfe1a47d9192747e42f25c580c6e011#diff-5cf3a68c413f5f031a2713f7488ab0c4668f7a4895120e1dfba706c50b28bba6R6-R8). Therefore, ensuring the test is unique across the integration test

.github/workflows/integrationTest.yml
khanhntd: Right.Changed to only ECSFargate though with the newest commit.

.github/workflows/integrationTest.yml
khanhntd: Yes, based on the [documents](https://martin-grigorov.medium.com/building-linux-packages-for-different-cpu-architectures-with-docker-and-qemu-d29e4ebc9fa5), 
```
QEMU emulates the machine’s processor through dynamic binary translation and provides a set of different hardware and device models for the machine, enabling it to run a variety of guest operating systems.
```
In another way of speaking, QEMU helps us in building multi platforms such as [arm64,amd64](https://github.com/aws/amazon-cloudwatch-agent/pull/458/files#diff-9b9f9a38b889c9dcb55ba0330baa69c8fdd62ac1100b6eb59b6fc44008f5d320R83) for our docker image.

.github/workflows/integrationTest.yml
khanhntd: I'm fine with both so yup.

.github/workflows/integrationTest.yml
khanhntd: I'm fine with being abstract though; but since it is a environment variable `ECR_INTEGRATION_TEST_REPO` so it can be changed anytime depends on the developer's opinion.

.github/workflows/integrationTest.yml
sethAmazon: This is required for buildx. Why not sure. We learned this was trail and error in adot.

.github/workflows/integrationTest.yml
```diff
@@ -14,18 +14,73 @@ env:
...
+        if: steps.build-docker-image.outputs.cache-hit != 'true'
+        uses: docker/setup-qemu-action@v1
+
+        #Build the cloudwatch agent image for two primary reasons:
+        #-Using the cloudwatch agent image to do the integration test (can be used for internal)
+        #-Export it for the future use in CD release pipeline
+        #Documentation: https://github.com/docker/build-push-action
+      - name: Build Cloudwatch Agent Image
```
sethAmazon, line 70: Not sure this works with the docker file since it curls the latest. It does not really build the deb from source under the hood. I think we need to talk this over. I don't think it does what it says.

sethAmazon: Since we can't merge this yet just blame it on me in the stand up saying that we dove deep and the current implementation does not match the docs.

Makefile
```diff
@@ -51,11 +51,11 @@ GOIMPORTS_OPT?= -w -local $(CW_AGENT_IMPORT_PATH)
 GOIMPORTS = $(TOOLS_BIN_DIR)/goimports
 SHFMT = $(TOOLS_BIN_DIR)/shfmt
 LINTER = $(TOOLS_BIN_DIR)/golangci-lint
-release: clean test build package-rpm package-deb package-win package-darwin
+release: clean  build package-rpm package-deb package-win package-darwin
```
sethAmazon, line 54: Can you change this back please.

.github/workflows/integrationTest.yml
khanhntd: The Dockerfile which curls from the latest binary is [this one](https://github.com/aws/amazon-cloudwatch-agent/blob/master/amazon-cloudwatch-container-insights/cloudwatch-agent-dockerfile/Dockerfile). This one use `build-for-docker-${TARGETARCH}` so no worries on that though.

integration/generator/test_case_generator.go
```diff
@@ -20,12 +20,17 @@ const (
...
 		"./integration/test/cloudwatchlogs",
-		"./integration/test/metrics_number_dimension"},
+		"./integration/test/metrics_number_dimension",
+	},
 	// @TODO add real tests
-	windows: {""},
-	mac:     {},
+	"ec2_windows": {""},
+	"ec2_mac":     {},
+	"ecs_fargate": {
```
sethAmazon, line 31: We should try to run this from the root dir so we can use ./

.github/workflows/integrationTest.yml
sethAmazon: Ignore this commit.

Makefile
khanhntd: Ouch. Forgot to change back. Update in newest commit.

.github/workflows/integrationTest.yml
khanhntd: I remember reading a medium doc about if `docker buildx` does not have QEMU or binfmt, it would only support `linux/amd64`; let's me try to find those documents since I have read those months ago.

SaxyPandaBear: Fair enough

integration/terraform/ecs/linux/README.md
khanhntd: Update the documents with more details regarding each resources' purpose

integration/terraform/ecs/linux/README.md
khanhntd: Its better to allow all traffics since this is a common terraform script for ECS test, including EC2 or Fargate. To be more specifically, I would not constraint the port since there would be extra apps with multiple ports and managing with port go to which test case is a pain though.

integration/generator/test_case_generator.go
khanhntd: I would agree at some point. However, it would pose a different question? Would the developers prefer to look at the root dir or from the current path since for ECS Fargate test, it use 
```
resource "null_resource" "validator" {
  provisioner "local-exec" {
    command = <<-EOT
    EOT
  }
}
```
instead of like EC2
```
resource "aws_instance" "integration-test" {
  provisioner "remote-exec" {
}
```
Therefore, would prefer this path over root_dir. One more addition point to support mine, since I use reduce the variables by using [local](https://github.com/aws/amazon-cloudwatch-agent/pull/458/files#diff-5cf3a68c413f5f031a2713f7488ab0c4668f7a4895120e1dfba706c50b28bba6R18-R23), so it would be also a pain for using the root dir if don't do something to the root dir.

Makefile
```diff
@@ -55,7 +55,7 @@ release: clean test build package-rpm package-deb package-win package-darwin
 
 nightly-release: release
 
-build: check_secrets cwagent-otel-collector amazon-cloudwatch-agent config-translator start-amazon-cloudwatch-agent amazon-cloudwatch-agent-config-wizard config-downloader
```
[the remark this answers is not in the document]
khanhntd: Forgot to return it back to normal after failing testing. Fixed it in the newest commit.

integration/test/ecs/ecs_metadata/ecs_metadata_test.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"flag"
+	"fmt"
+	"log"
+	"testing"
+	"time"
+	"github.com/aws/amazon-cloudwatch-agent/integration/test"
+)
+
+// Purpose: Detect the changes in metadata endpoint for ECS Container Agent https://github.com/aws/amazon-cloudwatch-agent/blob/master/translator/util/ecsutil/ecsutil.go#L67-L75
```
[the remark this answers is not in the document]
khanhntd: To be more specifically a bit, during the [scrapping  metadata endpoint](https://github.com/aws/amazon-cloudwatch-agent/blob/master/translator/util/ecsutil/ecsutil.go#L72-L79),  it would extract the Region, ClusterName, TaskARN. After extracting those variables, it would create a [log group based on that format that includes the ClusterName](https://github.com/aws/amazon-cloudwatch-agent/blob/master/translator/translate/logs/metrics_collected/prometheus/ruleLogGroupName.go#L33). So to speak, if the `clustername` which scrapped from `metadataendpointv4` does not exist, then it would fail the test. So to your point, yes. Also if the ECS cluster does have the v5 endpoint but we have not updated that, then it would fail the test based on the aforementioned points.

integration/generator/test_case_generator.go
khanhntd: If you guys prefer root module, then its fine by me also.

integration/terraform/ecs/linux/default_resources/default_extra_apps.tpl
```diff
@@ -0,0 +1,57 @@
+[
```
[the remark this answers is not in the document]
khanhntd: Reasonable. I added how ECS Fargate is setup in the [README.doc](https://github.com/aws/amazon-cloudwatch-agent/pull/458/files#diff-4069719fb55c19ddd4295f7df4131e72587210384df24ceb2841d266faaffba7R4-R14). Let's me know if its cover enough details for everyone.

SaxyPandaBear (approved):
lgtm

PR #469, opened 2022-05-17
Support setting retention to 6, 7, 8, 9 years

aateeqi:
https://github.com/aws/amazon-cloudwatch-agent/issues/470

From customer:

>While 6 of years of retention has been provided in console and CLI, this is missing in CloudWatch Agent.
>Please provide option to configure 6 years of retention policy.

Reference: https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Agent-Configuration-File-Details.html

From logs team:
> New Value: The number of days to retain the log events in the specified log group. Possible values are: 1, 3, 5, 7, 14, 30, 60, 90, 120, 150, 180, 365, 400, 545, 731, 1827, 2192, 2557, 2922, 3288, and 3653.

# Description of changes
Allows `retention_in_days` to also be set to 2192 ,2557, 2922, 3288

# Tests

Ran agent with this config:
image

and retention to 6, 7, 8, 9 years was set 
image

# License


translator/isValid.go
```diff
@@ -11,7 +11,7 @@ import (
 //ErrorMessages will provide detail error messages to user
 var ErrorMessages = []string{}
 var InfoMessages = []string{}
-var ValidRetentionInDays = []string{"-1", "1", "3", "5", "7", "14", "30", "60", "90", "120", "150", "180", "365", "400", "545", "731", "1827", "3653"}
+var ValidRetentionInDays = []string{"-1", "1", "3", "5", "7", "14", "30", "60", "90", "120", "150", "180", "365", "400", "545", "731", "1827", "2192", "2557", "2922", "3288", "3653"}
```
khanhntd, line 14: If it's good, please add a comment reference to this public doc https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_PutRetentionPolicy.html#CWL-PutRetentionPolicy-request-retentionInDays

khanhntd (commented):
If it's good, please show us in how to keep in sync with the CloudWatchLogs instead of letting the customer creates a ticket for us. Moreover, an interval of updating might be a good idea of aggressive planning.

translator/isValid.go
aateeqi: Thanks, added a comment referencing this. 

I think thats a good point and we shouldn't have to manually test it but instead we can extend integration tests to verify that each retention value is still valid

translator/isValid.go
```diff
@@ -11,7 +11,9 @@ import (
 //ErrorMessages will provide detail error messages to user
 var ErrorMessages = []string{}
 var InfoMessages = []string{}
-var ValidRetentionInDays = []string{"-1", "1", "3", "5", "7", "14", "30", "60", "90", "120", "150", "180", "365", "400", "545", "731", "1827", "3653"}
+
+// ValidRetentionInDays is based on what's supported by PutRetentionPolicy. See https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_PutRetentionPolicy.html#CWL-PutRetentionPolicy-request-retentionInDays
```
SaxyPandaBear, line 15: Just as an aside, we should also note that -1 is not a valid retention value. It's something that we use in the agent to denote no retention policy. Is this called out on public docs for the agent configuration separately?

SaxyPandaBear (commented):
I'm not scared about the changes, but I do think that this is an opportunity to add more straightforward integration tests without much additional effort.

Can you add an integration test that tries to create log groups with these new retention values and asserts that the log group was created successfully? We can pull up if you need anything about setting up the integration test itself.

translator/isValid.go
aateeqi: Im onboard with writing an integration test I don't think this is something i'll be able to add this sprint but retention in days as -1 is called out in the [docs](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Agent-Configuration-File-Details.html#CloudWatch-Agent-Configuration-File-Logssection).

> If you specify -1 or omit this field for a log group that the agent is creating now, the retention will be set to never expire. If the agent is sending logs to a log group that already exists, and you omit this field, the log group's retention is not changed.

But I do think we should edit the second sentence to say:
  If the agent is sending logs to a log group that already exists, and you omit this field [or set it to -1], the log group's retention is not changed.

SaxyPandaBear: I think that linking to https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Agent-Configuration-File-Details.html#CloudWatch-Agent-Configuration-File-Logssection is the better thing to point to in documentation here.

khanhntd:
Another thing that @ZhenyuTan-amz and @SaxyPandaBear  on the Slack Discussion, 
```
We should allow customer config whatever they want, and if it fails with invalid value, if we can distinguish that from response, we say “invalid retention value, for available values check this doc https://xxxx/” Therefore, we don't need to sync up with CWL, as long as they update the doc
```

khanhntd (approved):
LGTM.

SaxyPandaBear (changes requested):
It would be nice to add unit testing around accepting the new values in the schema, but I'm not going to block on that here. 

Based solely on this PR, it is not obvious whether the change is tested. Can you please build the agent from source, install it on an instance and configure a log group with 6 year retention via the agent? I'm not super concerned about the change, but I don't want to catch a bug related to setting 6,7,8,9 year retention after the fact.

aateeqi:
> It would be nice to add unit testing around accepting the new values in the schema, but I'm not going to block on that here.
> 
> Based solely on this PR, it is not obvious whether the change is tested. Can you please build the agent from source, install it on an instance and configure a log group with 6 year retention via the agent? I'm not super concerned about the change, but I don't want to catch a bug related to setting 6,7,8,9 year retention after the fact.

Test added in description

PR #475, opened 2022-05-24
Update EC2 integration test README to include Windows ports

SaxyPandaBear:
# Description of the issue
It's not clear in our documentation what ports need to be exposed in the security groups in order to allow Terraform to SSH into the test Windows EC2 instances. 

# Description of changes
* Updated table of ingress/egress for EC2 security groups in README
* Moved README out of linux subfolder, since the only real difference between linux and windows integration test set up is the security groups

I was originally going to consolidate the EC2 and ECS integration test READMEs, but after I did that, the doc felt too long and not directed enough. 

# License


SaxyPandaBear:
I think we need to update the build-and-test workflows on GitHub because there's no reason to run unit tests on this README change, and I thought we already accounted for that.

SaxyPandaBear:
The reason they ran this time is because we explicitly still run unit tests when there is an update to the unit test actions:
```
'!.github/workflows/build-*'
```

sethAmazon (approved):

integration/terraform/ec2/README.md
```diff
@@ -236,7 +246,7 @@ upload: ./terraform.tfstate to s3://***/integration-test/local-stack-terraform-s
 
 In this example, you should keep track of `ec2-35-87-254-148.us-west-2.compute.amazonaws.com`
 
-Start the linux integration tests:
+Start the linux integration tests (example):
```
khanhntd, line 249: IMO, I would see this as a `start the linux integration  test` more than an example because we don't provide any sample inputs. However, I would say as an example would be a good idea and comments for each input.

integration/terraform/ec2/README.md
```diff
@@ -112,12 +112,15 @@ require public access.
 
 The security group(s) that the integration tests use should include the following for ingress:
```
khanhntd, line 113: Personally, I would more favor in setting up the VPC and Security Group inside the terraform instead of setting them manually; moreover, I do not mind if we allows all ingress traffic since its in the VPC.

.github/workflows/build-test-linux.yml
```diff
@@ -6,7 +6,7 @@ on:
   push:
     branches: [master, main]
     paths-ignore:
-      - '**.md'
+      - '**/*.md'
```
khanhntd, line 9: While you are at it, I would consider adding a `not-sh, not-go files, not Makefile and definitely not-terraform files`; however, its not a blocker from me.

khanhntd (approved):

.github/workflows/build-test-linux.yml
SaxyPandaBear: We can add them, but also I would consider the likelihood of those being changed without something else being changed that would require running unit tests. I think Markdown files very typically can be updated independently of other things. I don't see many meaningful changes to a bash script or makefile or terraform file that wouldn't directly be associated with code that we need to build and unit test.

integration/terraform/ec2/README.md
SaxyPandaBear: You're welcome to update the EC2 terraform to create the security groups in a separate PR. I'm just trying to keep the document up to date

integration/terraform/ec2/README.md
SaxyPandaBear: I put "example" here because I moved this outside of the linux folder and don't want to confuse someone who is only concerned with running Windows tests (when those eventually get written)

PR #478, opened 2022-05-28
Add Sanity Check for Windows Integration Test 

khanhntd:
# Description of the issue
Based on the [stackoverflow](https://stackoverflow.com/questions/4055733/what-is-a-sanity-test-check), Sanity check is 
```
sanity test is just a casual term to mean that you're testing/confirming/validating something that should follow very clear and simple logic.
```

Applying to our agents, `sanity check` is confirming both CloudWatch Agent and ADOT Collector Agent are running independently and side-by-side as expected before our integration test.  However, as part of moving agent to Github Repository Epic and we releases the external agent before internal, we **could not know if our agent are still independent as usual and running as expected after each PR is merged**.  

Therefore, I propose to add a Sanity Check for Windows before running our integration test. 

# Description of Changes
By utilizing the existing [Terraform IAC for Windows Integration Test](https://github.com/aws/amazon-cloudwatch-agent/tree/master/integration/terraform/ec2/win), I add these changes:
* [Automatically add IAM Instance Profile](https://github.com/aws/amazon-cloudwatch-agent/pull/478/files#diff-dbec29f3463119e5aa1972e71e355b1f5028ecf2a4de111395ea7e7b7dc58cc3) instead of manually creating IAM Instance Profile which shares the same policies as the [README.doc](https://github.com/aws/amazon-cloudwatch-agent/blob/master/integration/terraform/ec2/README.md#ec2-instance-iam-role-permissions)
* [Automatically add ssh_key](https://github.com/aws/amazon-cloudwatch-agent/pull/478/files#diff-2f99e3ed4c00dce4989671d14671f7754f42b370edc1b6464d217e70b9130e6aR10-R25) instead of relying on manual key which developers can utilize to avoid creating one and also [reduce the environment variables for us to maintain during our integration test](https://github.com/aws/amazon-cloudwatch-agent/pull/478/files#diff-3ab46ee209a127470fce3c2cf106b1a1dbadbb929a4b5b13656a4bc4ce19c0b8L511).
* Added a GoLang function to run [PowerShell Script](https://github.com/aws/amazon-cloudwatch-agent/pull/478/files#diff-914f4c18aa0763094ec05c567b3d2a2481e25cd9e970fe0a3e081a118012de0eR98-R108) with these flags: NoProfile, NoExit, NonInteractive. For more information on these flags, you can read it here in the [following doc](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_powershell_exe?view=powershell-5.1)
* Run [Sanity Check](https://github.com/aws/amazon-cloudwatch-agent/pull/478/files#diff-2f99e3ed4c00dce4989671d14671f7754f42b370edc1b6464d217e70b9130e6aR56) before coming to our integration test 
* Add [how to run integration test on Windows](https://github.com/aws/amazon-cloudwatch-agent/pull/478/files#diff-e363c288be3ccc1d5f0a87e3bd4d65f07a780ef7c09ee17751402eef7db14deaR295-R333), delete [Security Groups for Windows](https://github.com/aws/amazon-cloudwatch-agent/pull/478/files#diff-e363c288be3ccc1d5f0a87e3bd4d65f07a780ef7c09ee17751402eef7db14deaL122-L123) in README.doc since we are not using `winrm` as a connection type and Terraform does not [recommended us to use WinRM (only as a last resolution)](https://www.terraform.io/language/resources/provisioners/file#destination-paths). Moreover, I also propose to delete Security Groups Ingress since the resources are created within the VPC.

# License

# Tests
After introduces the changes, I run the below command
`terraform apply -auto-approve github_repo={{my forked repo}} github_sha={{my commited sha in my forked repo}} 
s3_bucket={{my s3 bucket}}`

which gives me the follow result
```
null_resource.integration_test: Still creating... [1m30s elapsed]
null_resource.integration_test: Still creating... [1m40s elapsed]
null_resource.integration_test: Still creating... [1m50s elapsed]
null_resource.integration_test: Still creating... [2m0s elapsed]
null_resource.integration_test (remote-exec): === RUN   TestAgentStatus
null_resource.integration_test: Still creating... [2m10s elapsed]
null_resource.integration_test: Still creating... [2m20s elapsed]
null_resource.integration_test: Still creating... [2m30s elapsed]
null_resource.integration_test: Still creating... [2m40s elapsed]
null_resource.integration_test: Still creating... [2m50s elapsed]
null_resource.integration_test (remote-exec): --- PASS: TestAgentStatus (44.84s)
null_resource.integration_test (remote-exec): PASS
null_resource.integration_test (remote-exec): ok        github.com/aws/amazon-cloudwatch-agent/integration/test/sanity  45.203s
null_resource.integration_test: Creation complete after 2m52s [id=8591283884920986776]
```
 
and running `terraform destroy` afterwards to destroy the resources which are created by terraform on my account.

# Note
* The Build Check is unrelated to this change and because we have [known how flakey it is](https://github.com/aws/amazon-cloudwatch-agent/issues/360)

# Requirements
2. Run `make linter`


packaging/windows/amazon-cloudwatch-agent-ctl.ps1
```diff
@@ -331,15 +331,15 @@ Function ConfigAll() {
     )
 
     if ($OtelConfigLocation) {
-        Write-Output "****** processing cwagent-otel-collector ******"
+        Write-Output "****** Processing cwagent-otel-collector ******"
```
SaxyPandaBear, line 334: Why are we changing this? Does this mirror what the linux ctl script already does?

integration/terraform/ec2/win/main.tf
```diff
@@ -1,33 +1,71 @@
...
+}
+
+resource "null_resource" "integration_test" {
+  depends_on = [aws_instance.cwagent]
   provisioner "remote-exec" {
-    # @TODO when @ZhenyuTan-amz adds windows tests add "make integration-test"
-    # @TODO add export for AWS region from tf vars to make sure runner can use AWS SDK
     inline = [
+      # @TODO when @ZhenyuTan-amz adds windows tests add "make integration-test"
+      # @TODO add export for AWS region from tf vars to make sure runner can use AWS SDK
```
SaxyPandaBear, line 49: Why is it necessary to move the comment?

integration/terraform/ec2/win/main.tf
```diff
@@ -1,33 +1,71 @@
...
+resource "aws_key_pair" "aws_ssh_key" {
+  count      = var.ssh_key_name == "" ? 1 : 0
+  key_name   = "ec2-key-pair-${random_id.testing_id.hex}"
+  public_key = tls_private_key.ssh_key[0].public_key_openssh
+}
+
+locals {
+  ssh_key_name        = var.ssh_key_name != "" ? var.ssh_key_name : aws_key_pair.aws_ssh_key[0].key_name
+  private_key_content = var.ssh_key_name != "" ? var.ssh_key_value : tls_private_key.ssh_key[0].private_key_pem
+}
```
SaxyPandaBear, line 25: Why do we need to make a key pair in terraform? We already use a PEM file for SSH access to the hosts.

integration/test/sanity/sanity_test.go
```diff
@@ -1,17 +1,26 @@
...
 func TestAgentStatus(t *testing.T) {
-	test.RunShellScript("resources/verifyLinuxCtlScript.sh")
+	os := runtime.GOOS
+	
+	switch os {
+		case "windows":
+			test.RunPowerShellScript("resources/verifyWindowsCtlScript.ps1")
+		default: 
+			test.RunShellScript("resources/verifyLinuxCtlScript.sh")
+	}
```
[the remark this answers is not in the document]
khanhntd: Using a build constraint would be a good idea 👍 . But the script is small so I have not considered using it

packaging/dependencies/amazon-cloudwatch-agent-ctl
```diff
@@ -87,13 +87,13 @@ UsageString="
 start_all() {
      mode="${1:-}"
 
-     echo "****** processing cwagent-otel-collector ******"
+     echo "****** Processing cwagent-otel-collector ******"
```
SaxyPandaBear, line 90: Can you please explain why we are changing al of these?

SaxyPandaBear (changes requested):

packaging/dependencies/amazon-cloudwatch-agent-ctl
khanhntd: Yes. Some of the "echo" in linux are capitalizing the first word such as [invalid flags](https://github.com/aws/amazon-cloudwatch-agent/blob/master/packaging/dependencies/amazon-cloudwatch-agent-ctl#L510), [successfully fetch OTEL config](https://github.com/aws/amazon-cloudwatch-agent/blob/master/packaging/dependencies/amazon-cloudwatch-agent-ctl#L422) or in [Windows](https://github.com/aws/amazon-cloudwatch-agent/blob/master/packaging/dependencies/amazon-cloudwatch-agent-ctl#L422). Therefore, I want to be consistency between these echo commands

SaxyPandaBear: Ok but all of those were updated by you, and maybe we should have scrutinized it then. Why do we need to keep updating these? I don't see the value in adding all of these code changes.

khanhntd: The Linux is definitely not me :( (only Windows). I agree that there is not much value besides consistency so yeah (i'm still indecisive between revert back and keep it)

SaxyPandaBear: My bad. I was looking at the git blame and you were the last to update it. I checked back further and it was before your commit.

What I don't understand is how much value this adds. Do customers complain about this? Consistency is good, but this doesn't really add any value to the customer.

integration/terraform/ec2/win/main.tf
khanhntd: We make this to provide the option for dev on-board with integration-test does not needed to setup their resources (e.g only add another test case without worrying about set up a SSH Key). However, they can still provide their PEM file to ssh into the host.

integration/terraform/ec2/win/main.tf
khanhntd: No deep reason behind it. Forgot to changed it back 👍

integration/terraform/ec2/README.md
```diff
@@ -179,7 +175,7 @@ on creating the key pair.
 
 **Reminder: All AWS resources including EC2 key pair and ECR repository must be in the same region as the instances, so this assumes that they are created in the `us-west-2` region.**
 
-## Required parameters for Terraform to have handy
+# 2. Required parameters for Terraform to have handy
```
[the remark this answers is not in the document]
khanhntd: Yeah I guess the disadvantage outweight the advantage (mostly numbering headers help us separate the layer more clearer). Therefore, I like using numbering headers lot more.

integration/terraform/ec2/README.md
```diff
@@ -160,9 +158,7 @@ The security group(s) that the integration tests use should include the followin
 | HTTPS       | 443  | 0.0.0.0/0 |
 | HTTP        | 80   | 0.0.0.0/0 |
 | SSH         | 22   | 0.0.0.0/0 |
-| RDP         | 3389 | 0.0.0.0/0 |
-| WinRM-HTTP  | 5985 | 0.0.0.0/0 |
-| WinRM-HTTPS | 5986 | 0.0.0.0/0 |
```
SaxyPandaBear, line 165: Why are these being taken out of the doc? I'm pretty sure this is necessary for terraform.

integration/terraform/ec2/README.md
```diff
@@ -160,9 +158,7 @@ The security group(s) that the integration tests use should include the followin
...
+| RDP         | 3389 | 0.0.0.0/0 |Ω
```
SaxyPandaBear, line 161: There's an extra character here

.github/workflows/integrationTest.yml
```diff
@@ -507,10 +506,9 @@ jobs:
             -var="ssh_key=${PRIVATE_KEY}" -var="key_name=${KEY_NAME}" \
             -var="github_repo=${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}.git" \
             -var="github_sha=${GITHUB_SHA}" -var="ami=${{ matrix.arrays.ami }}" \
-            -var="test_name=cw-integ-test-${{ matrix.arrays.os }}" \
-            -var="iam_instance_profile=${IAM_ROLE}" \
-            -var="vpc_security_group_ids=${VPC_SECURITY_GROUPS_IDS}" \
-            -var="s3_bucket=${S3_INTEGRATION_BUCKET}" ; then terraform destroy -auto-approve
+            -var="test_name=${{ matrix.arrays.os }}" \
+            -var="s3_bucket=${S3_INTEGRATION_BUCKET}" ; then
```
sethAmazon, line 510: This works without a iam instance profile and vpc?

integration/test/sanity/sanity_test.go
```diff
@@ -1,17 +1,15 @@
...
 
 import (
 	"testing"
-
-	"github.com/aws/amazon-cloudwatch-agent/integration/test"
 )
 
 func TestAgentStatus(t *testing.T) {
-	test.RunShellScript("resources/verifyLinuxCtlScript.sh")
+	SanityCheck()
```
SaxyPandaBear, line 14: Nit - We should be using the `testing.T` struct to assert failure rather than relying on `log.Fatal`

sethAmazon (dismissed):

integration/test/agent_util.go
```diff
@@ -95,6 +95,17 @@ func RunShellScript(path string, args ...string) {
 	}
 }
 
+func RunPowerShellScript(path string, args ...string) {
+	ps, _ := exec.LookPath("powershell.exe")
```
SaxyPandaBear, line 99: I don't think that we should be ignoring an error here.

integration/terraform/ec2/README.md
khanhntd: its because we are RDP as a connection type instead of WinRM and Terraform only [recommendeds us to use WinRM (only as a last resolution)](https://www.terraform.io/language/resources/provisioners/file#destination-paths)

integration/test/agent_util.go
khanhntd: It should be safe to ignore error here since there is another [error gating when execution](https://github.com/aws/amazon-cloudwatch-agent/pull/478/files#diff-914f4c18aa0763094ec05c567b3d2a2481e25cd9e970fe0a3e081a118012de0eR104-R106)

integration/terraform/ec2/README.md
SaxyPandaBear: **Have you confirmed that you don't need WinRM?** Your security group configured here is fully open. I know when I was configuring my integration tests, I needed these in my security group.

integration/test/agent_util.go
```diff
@@ -77,11 +77,11 @@ func ReadAgentOutput(d time.Duration) string {
 	return string(out)
 }
 
-func RunShellScript(path string, args ...string) {
+func RunShellScript(path string, args ...string) error{
```
SaxyPandaBear, line 80: I'm pretty sure there are other places that use this function. Please verify that the integration tests still compile

khanhntd: Yes. I triggered an integration test on my fork

integration/terraform/ec2/README.md
khanhntd: I can double check this behavior again so yup

SaxyPandaBear:
@khanhntd there are merge conflicts with the main branch

khanhntd:
> @khanhntd there are merge conflicts with the main branch

Yup I'm testing the port so will merge it after the test.

integration/terraform/ec2/README.md
khanhntd: Nice catch.

integration/terraform/ec2/README.md
khanhntd: Yes double check and use this follow ingress to confirm we only use ssh instead of `WinRM`:
```
ingress {
    from_port   = "22"
    to_port     = "22"
    protocol    = "tcp"
    cidr_blocks = ["0.0.0.0/0"]
  }
```
It will result successfully. I can apply this port here when creating  SG for more security.
However, if I used the below port, it will reject me. Therefore, I believe we do not need these two ports for connection
```
ingress {
    from_port   = "5986"
    to_port     = "5986"
    protocol    = "tcp"
    cidr_blocks = ["0.0.0.0/0"]
  }
```

integration/test/agent_util.go
khanhntd: Here is the result for my integration test on my fork https://github.com/khanhntd/amazon-cloudwatch-agent/actions/runs/2956867720

SaxyPandaBear (approved):

PR #480, opened 2022-06-02
Reduce timeout for scrapping IMDS and give instruction when fail to scrape IMDS inside container

khanhntd:
# Description of the issue
Whenever customers turn on [EC2 Tagger](https://github.com/aws/amazon-cloudwatch-agent/tree/master/plugins/processors/ec2tagger), the customer would need to fulfill these two conditions [before able to scrap metadata from IMDS](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html):
* CloudWatch Agent needs to be run in EC2 Instances
* Enable [IMDSv2 and IMDSv1](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html)

However, for scrapping IMDSv2 and running in Docker Container, customers need to [increase their hop limit to 2](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html#imds-considerations) for general case and for certain case such as KOPS Vanilla, it would be 3. One thing to note here is for most AWS Resources, the default hop limit would be 2 for resources that manage container such as [EKS](https://aws.amazon.com/about-aws/whats-new/2020/08/amazon-eks-supports-ec2-instance-metadata-service-v2/). But if customers enable IMDSv1, the [AWS SDK Go will fall back to use IMDSv1](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html#imds-considerations) as default and CloudWatchAgent won't need to worry about Hop Limit. However, consider the security such as 
["Protecting against open layer 3 firewall and NATS"](https://aws.amazon.com/blogs/security/defense-in-depth-open-firewalls-reverse-proxies-ssrf-vulnerabilities-ec2-instance-metadata-service/), it would be best to keep in mind customers won't use IMDSv1 in the future and its not best practice to always use IMDSv1.

Therefore, we have considered these strategies for dealing with IMDSv2:
* Fail faster by reducing [4 minutes time out ( 4 retry and 1 retry/ minute)](https://github.com/aws/amazon-cloudwatch-agent/blob/master/cfg/aws/credentials.go#L94-L104) to [recommend timeout strategies](https://github.com/aws/aws-sdk-go/pull/3066) (same as [Terraform](https://github.com/hashicorp/vault/pull/10133)) and show corresponding course of action when customer runs in Docker container and only enable IMDSv2 with 1 Hop Limit.  
* CloudWatchAgent needs to modify hop limit under the hood. In order to modify the hop limit, the [API requires an instance ID](https://docs.aws.amazon.com/cli/latest/reference/ec2/modify-instance-metadata-options.html) and we can only get that from IMDS. However, we can scrap instance ID from certain scripts which are created when EC2 Instances are created such as in Linux Instances (`/var/lib/cloud/data/instance-id`) and we would need to use Docker Mount between container's volume and host's volume.
* Modify the TTL or Hop Limit in Token(package IP) from IMDSv2 ; however, it would not be recommended  to go against what IMDSv2 has been designed for security reasons.

Therefore, after discussing with the team, we would go with the first strategy.

# Description of changes
* Reduce timeout from [4 retry (1 retry / minute) to 2 retry ( 1 retry / second)](https://github.com/aws/amazon-cloudwatch-agent/pull/480/files#diff-6638b992018408bf44cac12361171fd050ff44fcc24c28f326988f4768a6dc2dR525) for fail faster and add [AWSDebugLogging for EC2Tagger( both IMDS and retrieve Tag, Volume)](https://github.com/aws/amazon-cloudwatch-agent/pull/480/files#diff-6638b992018408bf44cac12361171fd050ff44fcc24c28f326988f4768a6dc2dR524-R529)
* Add [EC2 Metadata Provider](https://github.com/aws/amazon-cloudwatch-agent/pull/480/files#diff-6638b992018408bf44cac12361171fd050ff44fcc24c28f326988f4768a6dc2dR520-R530), same as EC2 Client to separate the specific configuration for each Service SDK.
*  Reduce API Call when [scrapping IMDS Metadata](https://github.com/aws/amazon-cloudwatch-agent/pull/480/files#diff-dc70b68390b89e4d87ef2cec736f8587d0e736cecf1b032abf4891005ffa9432R110-R114) and create a separate function for only scrapping IMDS Metadata

# License

# Tests
I have tested CloudWatchAgent on two edge cases:
1. LXC Containers.
2. Vanilla Kops.

For specific test on second edge case:
* **Step 1:** Build KOPS EKS Cluster by following [this document(https://kops.sigs.k8s.io/getting_started/aws/)
* **Step 2:** Build Docker Image by using `make dockerized-build` and publish your image through ECR
* **Step 3:** Replace the ECR Image and the below config with [this template](https://github.com/aws/amazon-cloudwatch-agent/blob/master/amazon-cloudwatch-container-insights/k8s-yaml-templates/quickstart/cwagent-fluentd-quickstart.yaml)
```
{
	"agent": {
		"run_as_user": "root",
	},
	"metrics": {
		"append_dimensions": {
			"ImageId": "${aws:ImageId}",
		},
		"metrics_collected": {
			"disk": {
				"measurement": [
					"used_percent"
				],
				"resources": [
					"*"
				]
			}
		}
	}
}
```

* **Step 4:** See the expected following results
```
2022-06-07T03:36:11Z I! 
2022-06-07T03:36:11Z E! [processors.ec2tagger] ec2tagger: Unable to retrieve Instance Metadata Tags. This plugin must only be used on an EC2 instance.
2022-06-07T03:36:11Z E! [processors.ec2tagger] ec2tagger: Please increase hop limit to 2 by following this document https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-options.html#configuring-IMDS-existing-instances.
2022-06-07T03:36:11Z E! [telegraf] Error running agent: could not initialize processor ec2tagger: EC2MetadataRequestError: failed to get EC2 instance identity document
caused by: EC2MetadataError: failed to make EC2Metadata request
        status code: 401, request id: 
caused by:
```
# Requirements
2. Run `make linter`


plugins/processors/ec2tagger/ec2tagger.go
```diff
@@ -312,21 +246,21 @@ func (t *Tagger) Init() error {
...
-		t.Log.Errorf(msg)
-		return errors.New(msg)
+	md := t.ec2MetadataProvider()
+	doc, err := md.GetInstanceIdentityDocument()
+	if err != nil {
+		
+		t.Log.Errorf(metadataCheckStrInstanceDocumentFailure + metadataCheckStrEC2InstanceTagger)
+		if os.Getenv(config.RUN_IN_CONTAINER) == config.RUN_IN_CONTAINER_TRUE {
+			t.Log.Errorf(metadataCheckStrIncreaseHopLimit)
+		}
```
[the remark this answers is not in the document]
khanhntd: Yes. There are two ways currently to build Docker Image:
* Building from latest RPM which uses [RUN_IN_CONTAINER](https://github.com/aws/amazon-cloudwatch-agent/blob/master/amazon-cloudwatch-container-insights/cloudwatch-agent-dockerfile/Dockerfile#L26)
* Building from local RPM which uses [RUN_IN_CONTAINER](https://github.com/aws/amazon-cloudwatch-agent/blob/master/amazon-cloudwatch-container-insights/cloudwatch-agent-dockerfile/Dockerfile#L26).

The customers mostly uses the former Docker Image. While also note that [ADOT Collector uses this environment too](https://github.com/aws-observability/aws-otel-collector/blob/main/cmd/awscollector/Dockerfile#L74). Therefore, unless there are unexpected changes regarding how we detect the agents are running in container, , I won't see any problems in utilizing this to show the corresponding actions in container level

plugins/processors/ec2tagger/constants.go
```diff
@@ -0,0 +1,93 @@
...
+	mdKeyInstaneType     = "InstanceType"
+	ebsVolumeId          = "EBSVolumeId"
+)
+
+const (
+	metadataCheckStrStartInitialization         = "ec2tagger: EC2 IMDS has started initialization."
+	metadataCheckStrTagNotSupported             = "ec2tagger: Unsupported EC2 Metadata key: %s."
+	metadataCheckStrInstanceDocumentFailure     = "ec2tagger: Unable to retrieve Instance Metadata Tags."
+	metadataCheckStrEC2InstanceTagger           = "This plugin must only be used on an EC2 instance."
+	metadataCheckStrIncreaseHopLimit            = "ec2tagger: Please increase hop limit to 3 by following this document https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-options.html#configuring-IMDS-existing-instances."
```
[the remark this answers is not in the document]
khanhntd: 2 hops are recommended by [AWS](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html#imds-considerations). However, there are certain cases such as EKS Kops as @sethAmazon points out; moreover, I saw one use case of using 3 regarding hop limits when search on Slack Channel. 

Consider the security reason, 3 would not pose any impact as it would allow instance requests and the ones coming from containers while still blocking requests that might come from further away.

translator/util/sdkutil.go
```diff
@@ -90,17 +90,17 @@ func defaultECSRegion() string {
...
 	if region == "" && mode == config.ModeEC2 {
 		fmt.Println("I! Trying to detect region from ecs")
 		region = defaultECSRegion()
 	}
+
+	// For ec2, fallback to metadata when no region info found in credential profile.
+	if region == "" && mode == config.ModeEC2 {
+		fmt.Println("I! Trying to detect region from ec2")
+		region = defaultEC2Region()
+	}
```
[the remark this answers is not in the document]
khanhntd: For the concern, ECS Metadata only takes up to 4s based on the following log lines
```
2022/06/06 22:01:17 W! retry [0/3], unable to get http response from http://169.254.170.2/v2/metadata, error: unable to get response from http://169.254.170.2/v2/metadata, error: Get "http://169.254.170.2/v2/metadata": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
2022/06/06 22:01:18 W! retry [1/3], unable to get http response from http://169.254.170.2/v2/metadata, error: unable to get response from http://169.254.170.2/v2/metadata, error: Get "http://169.254.170.2/v2/metadata": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
2022/06/06 22:01:19 W! retry [2/3], unable to get http response from http://169.254.170.2/v2/metadata, error: unable to get response from http://169.254.170.2/v2/metadata, error: Get "http://169.254.170.2/v2/metadata": context deadline exceeded (Client.Timeout exceeded while awaiting headers)
2022/06/06 22:01:19 I! access ECS task metadata fail with response unable to get response from http://169.254.170.2/v2/metadata, error: Get "http://169.254.170.2/v2/metadata": context deadline exceeded (Client.Timeout exceeded while awaiting headers), assuming I'm not running in ECS.
```

I have three reasons why I prioritized ECS over EC2:
* CloudWatchAgent scraps only one ECS metadata endpoint which takes 4s while scraps [two IDMS endpoint which takes 8 second](https://github.com/aws/amazon-cloudwatch-agent/pull/480/files#diff-dc70b68390b89e4d87ef2cec736f8587d0e736cecf1b032abf4891005ffa9432R102-R116)
* ECS metadata is more easier to scraped more than IDMS when consider hop limits (for both EC2 and Fargate)
* IMO, we would move away from EC2 soon and utilize more EKS and ECS consider scaling problems; therefore, EC2 is only a surface level problem when testing/developing/GA. Therefore, container is needed in a long-term run.

Let me know if that makes sense.

translator/util/ec2util/ec2util.go
```diff
@@ -24,20 +25,22 @@ type ec2Util struct {
...
 
-const allowedRetries = 5
-
-var e *ec2Util
-var once sync.Once
+var (
+	ec2UtilInstance *ec2Util
+	once sync.Once
+)
+const allowedNetworkRetries = 5
```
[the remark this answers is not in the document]
khanhntd: Yes. IMDS can also be configured with Retry. However, consider the small impact of it, I have not added any retries yet. However, to keep [consistency what we have configured in EC2 Tagger](https://github.com/aws/amazon-cloudwatch-agent/pull/480/files#diff-6638b992018408bf44cac12361171fd050ff44fcc24c28f326988f4768a6dc2dR473), I have decided to add a constant in the newest commit.

.github/workflows/build-test-linux.yml
```diff
@@ -30,7 +30,7 @@ jobs:
         go-version: ~1.15.15
 
     - name: Check out code
-      uses: actions/checkout@v2
```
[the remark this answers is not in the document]
khanhntd: Its fine to separate in a separate PR since I only want to know if the `cache@v3` would [solve the cache our artifacts when building in Linux](https://github.com/aws/amazon-cloudwatch-agent/pull/480/checks)

plugins/processors/ec2tagger/ec2tagger_test.go
```diff
@@ -206,43 +214,35 @@ func (m *mockEC2Metadata) GetInstanceIdentityDocument() (ec2metadata.EC2Instance
...
-	tagger := Tagger{
-		Log:         testutil.Logger{},
-		ec2metadata: mockMetadata,
-	}
-	err := tagger.Init()
-	assert.NotNil(err)
-	assert.Contains(err.Error(), "ec2tagger: Unable to retrieve InstanceId. This plugin must only be used on an EC2 instance")
-}
-
-func TestInitFailWithNoInstanceIdentityDocument(t *testing.T) {
```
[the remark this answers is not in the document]
khanhntd: For the background a little bit: With the last EC2 Tagger, it uses [Available and GetInstanceIdentityDocument](https://github.com/aws/amazon-cloudwatch-agent/pull/480/files#diff-6638b992018408bf44cac12361171fd050ff44fcc24c28f326988f4768a6dc2dL319-L329). However, under the hood, `Available` only check if its able to get Instance-ID](https://github.com/aws/aws-sdk-go/blob/main/aws/ec2metadata/api.go#L216), it only improves slightly performance when comparing to `GetInstanceIdentityDocument`. However, it also provides drawback when CloudWatchAgent retries to scrap IMDS (extra 4 second with 1 retry/ second) and also without retry ( 1 extra second). Therefore, in the new PR, I only cover [GetInstanceIdentityDocument](https://github.com/aws/amazon-cloudwatch-agent/pull/480/files#diff-6638b992018408bf44cac12361171fd050ff44fcc24c28f326988f4768a6dc2dR256-R261). 

To main point, we have a test which covers [Instance Identity Document as nil](https://github.com/aws/amazon-cloudwatch-agent/blob/9adb253817dedb4d27a5a5e1f7329beb3465eb25/plugins/processors/ec2tagger/ec2tagger_test.go#L215-L233). Therefore, I would consider the mentioned test rather useless since its a duplication with the aforementioned test.

translator/util/ec2util/ec2util.go
```diff
@@ -24,20 +25,22 @@ type ec2Util struct {
 	AccountID  string
 }
 
-const allowedRetries = 5
-
-var e *ec2Util
-var once sync.Once
+var (
+	ec2UtilInstance *ec2Util
```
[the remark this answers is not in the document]
khanhntd: True. However, I only keep the variable name consistent with [ecsUtilInstance](https://github.com/aws/amazon-cloudwatch-agent/blob/9adb253817dedb4d27a5a5e1f7329beb3465eb25/translator/util/ecsutil/ecsutil.go#L35).

plugins/processors/ec2tagger/ec2tagger_test.go
```diff
@@ -206,43 +214,35 @@ func (m *mockEC2Metadata) GetInstanceIdentityDocument() (ec2metadata.EC2Instance
...
-		Log:         testutil.Logger{},
-		ec2metadata: mockMetadata,
+		Log:                 testutil.Logger{},
+		ec2MetadataProvider: ec2MetadataProvider,
 	}
 	err := tagger.Init()
+
 	assert.NotNil(err)
-	assert.Contains(err.Error(), "ec2tagger: Unable to retrieve InstanceId :")
+	assert.Contains(err.Error(), errors.New("No instance identity document").Error())
```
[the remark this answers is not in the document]
khanhntd: Yes `Error()` [stringify the error](https://pkg.go.dev/builtin#error). However, I only replicate the [error behavior](https://github.com/aws/amazon-cloudwatch-agent/pull/480/files#diff-6638b992018408bf44cac12361171fd050ff44fcc24c28f326988f4768a6dc2dR262). So there is no deep reason behind it.

plugins/processors/ec2tagger/constants.go
```diff
@@ -0,0 +1,93 @@
...
+	ec2TagAndVolumeCheckStrRetryFailure         = "ec2tagger: %v retry initial retrieval of tags and volumes"
+	ec2VolumeCheckStrRefreshFailure             = "ec2tagger: Error refreshing EC2 volumes, keeping old values : %+v"
+	ec2TagCheckStrRefreshFailure                = "ec2tagger: Error refreshing EC2 tags, keeping old values : %+v"
+	ec2TagAndVolumeCheckStrSuccess              = "ec2tagger: Finished initial retrieval of tags and volumes"
+)
+
+var (
+	defaultRefreshInterval = 180 * time.Second
+	// backoff retry for ec2 describe instances API call. Assuming the throttle limit is 20 per second. 10 mins allow 12000 API calls.
+	backoffSleepArray = []time.Duration{0, 1 * time.Minute, 1 * time.Minute, 3 * time.Minute, 3 * time.Minute, 3 * time.Minute, 10 * time.Minute}
```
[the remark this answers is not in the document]
khanhntd: Same idea initially. I also intend to re-structure the ec2 tagger interface but consider its out of scope; therefore, we could leave it in another PR.

plugins/processors/ec2tagger/ec2tagger.go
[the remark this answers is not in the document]
SaxyPandaBear: The `RUN_IN_CONTAINER` flag, by nature, is a workaround. I am not super comfortable with using it even more than we already do.

translator/util/sdkutil.go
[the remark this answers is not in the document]
SaxyPandaBear: > IMO, we would move away from EC2 soon and utilize more EKS and ECS consider scaling problems; therefore, EC2 is only a surface level problem when testing/developing/GA

I disagree. Do you have any data that would indicate that the percentage of customers using CloudWatch agent is trending more for EKS and ECS than regular EC2?

translator/util/ec2util/ec2util.go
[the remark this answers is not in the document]
SaxyPandaBear: What consistency are you talking about? I'm asking why the variable name for this was changed. This seems to just introduce noise to the PR with minimal value.

plugins/processors/ec2tagger/ec2tagger.go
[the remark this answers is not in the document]
khanhntd: Yes. RUN_IN_CONTAINER flag is a workaround. However, we can detect if we are running in a CONTAINER by using the code below
``
func (app *App) isRunningInDockerContainer() bool {
    // docker creates a .dockerenv file at the root
    // of the directory tree inside the container.
    // if this file exists then the viewer is running
    // from inside a container so return true
        
    if _, err := os.Stat("/.dockerenv"); err == nil {
        return true
    }
        
    return false
}
```

For more information on [the code](https://paulbradley.org/indocker/). What I meant is we can address the flag issue in another PR; for this PR, we can utilize it or waiting until the changes. However, consider its block some customers when running in Container, I would not waiting until the changes.

File: translator/util/sdkutil.go (hunk 3 above)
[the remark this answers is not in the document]
Comment (khanhntd) [reply translator/util/sdkutil.go:104]: > I disagree. Do you have any data that would indicate that the percentage of customers using CloudWatch agent is trending more for EKS and ECS than regular EC2?

No, that's why I only said its my personal opinion instead of reference any documents. However, I can get the data from CloudWatch Logs and Granite if it's good.

File: plugins/processors/ec2tagger/ec2tagger.go [hunk 10]
```diff
@@ -309,24 +243,23 @@ func (t *Tagger) Init() error {
...
 
-	doc, err := t.ec2metadata.GetInstanceIdentityDocument()
-	if nil != err {
-		msg := fmt.Sprintf("ec2tagger: Unable to retrieve InstanceId : %+v", err.Error())
-		t.Log.Errorf(msg)
-		return errors.New(msg)
+	md := t.ec2MetadataProvider()
+	doc, err := md.GetInstanceIdentityDocument()
+	if err != nil {
+		t.Log.Errorf(metadataCheckStrInstanceDocumentFailure + metadataCheckStrEC2InstanceTagger)
```
SaxyPandaBear, line 258: This is not readable to me. Just because the SSM agent code does it, doesn't mean that we have to follow it.

translator/util/sdkutil.go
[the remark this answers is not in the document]
SaxyPandaBear: This needs discussion internally. Even if it's your personal opinion, this PR would make it reality. I am not confident with approving this PR until we discuss design as a group.

SaxyPandaBear (changes requested):

translator/util/sdkutil.go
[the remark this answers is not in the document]
SaxyPandaBear: On top of the main issue, changing this behavior introduces unexpected warn log lines in the agent log. It introduces a false positive for customers who monitor the agent log with logs insights metrics. Although I don't think many people do that, I know some do, and this would impact their alarm thresholds. We should be able to resolve this issue without causing impact.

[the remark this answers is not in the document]
khanhntd: Agree. If that's the case, revert back to original behavior which is outside of this PR scope. Thanks for reminding me of this.

plugins/processors/ec2tagger/ec2tagger.go
khanhntd: Yes. I only want to introduce the changes and see everyone's comment. However, I have reduced the log to an [readable level](https://github.com/aws/amazon-cloudwatch-agent/pull/480/files#diff-6638b992018408bf44cac12361171fd050ff44fcc24c28f326988f4768a6dc2dR257-R261) which will show 
```
2022-06-07T03:36:11Z I! 
2022-06-07T03:36:11Z E! [processors.ec2tagger] ec2tagger: Unable to retrieve Instance Metadata Tags. This plugin must only be used on an EC2 instance.
2022-06-07T03:36:11Z E! [processors.ec2tagger] ec2tagger: Please increase hop limit to 3 by following this document https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-options.html#configuring-IMDS-existing-instances.
2022-06-07T03:36:11Z E! [telegraf] Error running agent: could not initialize processor ec2tagger: EC2MetadataRequestError: failed to get EC2 instance identity document
caused by: EC2MetadataError: failed to make EC2Metadata request
        status code: 401, request id: 
caused by:
```

Let's me know if it make sense.

plugins/processors/ec2tagger/constants.go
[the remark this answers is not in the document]
SaxyPandaBear: How is it out of scope? The purpose of the task is to reduce time to reach IMDS. This is relevant.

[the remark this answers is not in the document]
khanhntd: It does not relate since `backoffSleepArray` is for [ec2 API](https://github.com/aws/amazon-cloudwatch-agent/pull/480/files#diff-6638b992018408bf44cac12361171fd050ff44fcc24c28f326988f4768a6dc2dR298-R307) which uses describeTag and describe Volume which does not relate to IMDS while `imds` using [ec2metadata](https://github.com/aws/amazon-cloudwatch-agent/pull/480/files#diff-6638b992018408bf44cac12361171fd050ff44fcc24c28f326988f4768a6dc2dR255)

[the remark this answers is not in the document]
khanhntd: For more information on these two:
* EC2Metadata: https://github.com/aws/aws-sdk-go/blob/main/aws/ec2metadata/api.go
* EC2: https://github.com/aws/aws-sdk-go/tree/main/service/ec2

translator/util/ec2util/ec2util.go
```diff
@@ -24,20 +27,26 @@ type ec2Util struct {
...
+	once            sync.Once
+)
 
-var e *ec2Util
-var once sync.Once
+const (
+	allowedNetworkRetries = 5
+	allowedIMDSRetries = 2
```
SaxyPandaBear, line 37: This is duplicated from the constants file. Is the constant from that file used at all?

translator/util/ec2util/ec2util.go
```diff
@@ -67,46 +76,54 @@ func initEC2UtilSingleton() (newInstance *ec2Util) {
...
+		return err
 	}
 
-	if info, err := md.GetMetadata("hostname"); err == nil {
-		newInstance.Hostname = info
-	} else {
-		log.Println("E! getting hostname from EC2 metadata fail: ", err)
+	md := ec2metadata.New(ses,
+		&aws.Config{
+			Retryer: client.DefaultRetryer{NumMaxRetries: allowedIMDSRetries},
```
SaxyPandaBear, line 102: What is the frequency of retries/jitter for a `DefaultRetryer`? Is 2 retries sufficient?

translator/util/ec2util/ec2util.go
```diff
@@ -67,46 +76,54 @@ func initEC2UtilSingleton() (newInstance *ec2Util) {
...
-		return
-	}
+	return
+}
 
-	if info, err := md.GetMetadata("instance-id"); err == nil {
-		newInstance.InstanceID = info
-	} else {
-		log.Println("E! getting instance-id from EC2 metadata fail: ", err)
+func (e *ec2Util) getEC2MetadataFromIMDS() error {
```
SaxyPandaBear, line 93: Nitpick - using a `get` verb for this function makes me think that it will return something, but instead you are just setting values in the `ec2Util` struct. I would either change the name of the function to something like `deriveMetadata`, or I would suggest changing the function definition to return the values in a struct, and then the calling function can use the values to set on the ec2Util object itself.

plugins/processors/ec2tagger/constants.go
```diff
@@ -0,0 +1,93 @@
...
+	metadataCheckStrTagNotSupported             = "ec2tagger: Unsupported EC2 Metadata key: %s."
+	metadataCheckStrInstanceDocumentFailure     = "ec2tagger: Unable to retrieve Instance Metadata Tags."
+	metadataCheckStrEC2InstanceTagger           = "This plugin must only be used on an EC2 instance."
+	metadataCheckStrIncreaseHopLimit            = "ec2tagger: Please increase hop limit to 3 by following this document https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-options.html#configuring-IMDS-existing-instances."
+	ec2TagAndVolumeCheckStrInitRetrievalSuccess = "ec2tagger: Initial retrieval of tags succeeded"
+	ec2TagAndVolumeCheckStrStartInitialization  = "ec2tagger: EC2 tagger has started initialization."
+	ec2VolumeCheckStrInitRetrievalFailure       = "ec2tagger: Unable to describe ec2 volume for initial retrieval: %v"
+	ec2TagCheckStrInitRetrievalFailure          = "ec2tagger: Unable to describe ec2 tags for initial retrieval: %v"
+	ec2TagAndVolumeCheckStrStartRefresh         = "ec2tagger refreshing: EC2InstanceTags needed %v, retrieved: %v, ebs device needed %v, retrieved: %v"
```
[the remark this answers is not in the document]
SaxyPandaBear: I still disagree with these constant strings for logs, but will defer to others that review this PR.

plugins/processors/ec2tagger/constants.go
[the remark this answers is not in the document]
SaxyPandaBear: The recommendation is 2 hops. There are specific scenarios where 3 hops is necessary but those are not typical. Is there any reason we would want to go against ECS recommendations for 2 hops for our general customer base?

translator/util/ec2util/ec2util.go
khanhntd: I would not say its duplicated from constants files since it is separated between two package. For the second question, yes. However, the reason I created two constants would be: separate the TimeOut Strategies between Translator and also EC2Tagger. Lets me know if that make sense.

plugins/processors/ec2tagger/constants.go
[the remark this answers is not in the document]
khanhntd: Agree 👍 .

plugins/processors/ec2tagger/constants.go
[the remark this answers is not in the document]
khanhntd: Redraw my statements regarding setting hop limit to 3. After setting hop limit to 2 with KOPs, I have been able to scrape metadata from IMDSv2. Therefore, I have updated my PR for changing recommendation to 2. However, would need everyone to agree on this before final decision.

translator/util/ec2util/ec2util.go
khanhntd: Changing the naming convention would be a better idea since it aligns with ECS Util behavior. Thanks for the recommendation. 👍

plugins/processors/ec2tagger/constants.go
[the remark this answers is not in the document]
SaxyPandaBear: I just looked, and they are **NOT** constant strings that are used for log messages. Can you please explain your rationale for correlating their usage of constant strings with our log messages?

plugins/processors/ec2tagger/ec2tagger.go
SaxyPandaBear: Please revert this change. Based on the log output you have shown, I can't tell how to derive the log message from the two variables that you are using in this log line. Why is there a `[telegraf]` level error log introduced?

plugins/processors/ec2tagger/constants.go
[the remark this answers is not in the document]
khanhntd: Wdym @SaxyPandaBear?  The SSM Agent creates these constant strings that are used for log messages as shown below and there are [another boiler plate for create constant as a separate file](https://github.com/aws/amazon-ssm-agent/blob/mainline/agent/appconfig/constants.go)
```
const (
	maxTimeCallMetadataSeconds = 3
	metadataCheckStrName              = "EC2 IMDS"
	metadataCheckStrInstanceIdTimeout = "instance id request timeout"
	metadataCheckStrInstanceIdFailure = "failed to query instance id from IMDS: %v"
	metadataCheckStrRegionTimeout     = "region request timeout"
	metadataCheckStrRegionFailure     = "failed to query region from IMDS: %v"
	metadataCheckStrHasHybrid         = "Instance has hybrid registration"
	metadataCheckStrIMDSTimeout       = "IMDS request timeout"
	metadataCheckStrIMDSFailure       = "Failed to query IMDS: %v"
	metadataCheckStrSuccess           = "IMDS is accessible and has instance id %s in region %s"
)
```

translator/util/ec2util/ec2util.go
SaxyPandaBear: If the AWS SDK already changed the default behavior to 1 second timeout and 2 retries, why do we have to create our own retryer?

plugins/processors/ec2tagger/ec2tagger.go
khanhntd: Yes I reverted the changes yesterday as can be seen here https://github.com/aws/amazon-cloudwatch-agent/pull/480/files#diff-6638b992018408bf44cac12361171fd050ff44fcc24c28f326988f4768a6dc2dR257-R261 (only 1 constant string).

khanhntd: For the second question, its the default behavior when CloudWatchAgent fails to execute a daemon process that can be seen here https://github.com/aws/amazon-cloudwatch-agent/blob/master/cmd/amazon-cloudwatch-agent/amazon-cloudwatch-agent.go#L179-L182
```
err := runAgent(ctx, inputFilters, outputFilters)
if err != nil && err != context.Canceled {
			log.Fatalf("E! [telegraf] Error running agent: %v", err)
}
```

translator/util/ec2util/ec2util.go
khanhntd: 👍  Someone can argue with this: mostly for visibility and also consistent with the `EC2Tagger Retryer`. I'm fine with deleting this custom Retryer so non-block from me

plugins/processors/ec2tagger/constants.go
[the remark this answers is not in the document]
SaxyPandaBear: Can you please identify specific lines of code where those constants are used for log messages? They are all used either for error strings or return values.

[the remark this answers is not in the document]
SaxyPandaBear: None of the constants in https://github.com/aws/amazon-ssm-agent/blob/mainline/agent/appconfig/constants.go look like they are log message constants either.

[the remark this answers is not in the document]
khanhntd: I'm not sure I am understand your point, what is the difference between log messages using constant and [return note message with constant](https://github.com/aws/amazon-ssm-agent/blob/mainline/agent/cli/diagnostics/metadata_check.go#L161)?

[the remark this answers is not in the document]
khanhntd: The second one is the boiler plate to create a separate constant string file, it does not relate to use constant log message.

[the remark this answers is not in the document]
SaxyPandaBear: Six months from now, if someone is debugging an issue with the ec2tagger and needs to look through the logs, it will take longer to trace through the behavior of the plugin because the log messages are pulled out into a separate constants file. 

> They are using the same idea of using constant with log message

They are **not** the same. Reusing a constant string can make sense when it is a frequently used parameter. But these are diagnostic log messages. We should be able to very quickly identify where in the code that these log messages are getting emitted, and putting the log strings into a different file makes the log messages unreadable in the ec2tagger plugin source code. 

---
### Example
Pretend that we get a customer ticket and we see this when running on an EC2 instance:
```
2022-06-07T03:36:11Z E! [processors.ec2tagger] ec2tagger: Unable to retrieve Instance Metadata Tags. This plugin must only be used on an EC2 instance.
2022-06-07T03:36:11Z E! [processors.ec2tagger] ec2tagger: Please increase hop limit to 3 by following this document https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-options.html#configuring-IMDS-existing-instances.
```
The above log lines correspond to this in the code:
```go
	if err != nil {
		t.Log.Errorf(metadataCheckStrInstanceDocumentFailure)
		if context.CurrentContext().RunInContainer() {
			t.Log.Errorf(metadataCheckStrIncreaseHopLimit)
		}
		return err
	}
```
I only know that because I am reviewing this PR today. What are the steps that someone six months or a year from now would take in order to get to this part of the code? They would have to search the project for the contents of the log message, e.g.: `Unable to retrieve Instance Metadata Tags`, and that takes them to this constants file. 
```go
metadataCheckStrInstanceDocumentFailure = "ec2tagger: Unable to retrieve Instance Metadata Tags. This plugin must only be used on an EC2 instance."
```
This line alone does not provide **any** context at which point this log line gets emitted. Now the person looking into the issue has to search through the project **again** for where this variable `metadataCheckStrInstanceDocumentFailure` is used. 

Not every string needs to be a constant. Being a constant does not automatically make it cleaner or more readable/maintainable.

SaxyPandaBear (changes requested):

translator/util/ec2util/ec2util.go
SaxyPandaBear: Then I'll ask, why is the ec2tagger retryer being changed if they both use the same AWS SDK code, which based on the PR you linked, should both already be using the corrected default retry?

khanhntd: No the existing EC2 Tagger Retried Strategies used [our configuration](https://github.com/aws/amazon-cloudwatch-agent/blob/master/cfg/aws/credentials.go#L94-L104) 
```
      config := &aws.Config{
		Region:                        aws.String(c.Region),
		CredentialsChainVerboseErrors: aws.Bool(true),
		HTTPClient:                    &http.Client{Timeout: 1 * time.Minute},
		LogLevel:                      SDKLogLevel(),
		Logger:                        SDKLogger{},
	}
```

Therefore, the IMDS from Translation uses the default one while EC2 Tagger Retried used the custom one.

plugins/processors/ec2tagger/constants.go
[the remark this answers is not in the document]
khanhntd: 👍  Agree. Thought about it at some point but in exchange for readable. Sorry for this change and I have reverted back with normal log string message.

plugins/processors/ec2tagger/ec2tagger.go
```diff
@@ -309,25 +242,23 @@ func (t *Tagger) Init() error {
...
-	}
+	t.Log.Infof("ec2tagger: EC2 IMDS has started initialization.")
 
-	doc, err := t.ec2metadata.GetInstanceIdentityDocument()
-	if nil != err {
-		msg := fmt.Sprintf("ec2tagger: Unable to retrieve InstanceId : %+v", err.Error())
-		t.Log.Errorf(msg)
-		return errors.New(msg)
+	md := t.ec2MetadataProvider()
+	doc, err := md.GetInstanceIdentityDocument()
```
SaxyPandaBear, line 255: Nit: `md` only gets used once, and that's in the following line. There's really no reason to create the extra variable. This could be consolidated to:
```suggestion
	doc, err := t.ec2MetadataProvider().GetInstanceIdentityDocument()
```

plugins/processors/ec2tagger/ec2tagger.go
```diff
@@ -309,25 +242,23 @@ func (t *Tagger) Init() error {
...
+	if err != nil {
+		t.Log.Errorf("ec2tagger: Unable to retrieve Instance Metadata Tags. This plugin must only be used on an EC2 instance.")
```
SaxyPandaBear, line 257: > "This plugin must only be used on an EC2 instance"

Does the customer configure the ec2tagger? I thought that the CloudWatch agent does that under the hood. What is the point of calling this out if the customer doesn't have control over this?

translator/util/ec2util/ec2util.go
```diff
@@ -67,46 +71,51 @@ func initEC2UtilSingleton() (newInstance *ec2Util) {
...
-		log.Println("E! getting hostname from EC2 metadata fail: ", err)
+	md := ec2metadata.New(ses)
+
+	if !md.Available() {
+		return errors.New("EC2 metadata is not available.")
 	}
 
-	if info, err := md.GetMetadata("local-ipv4"); err == nil {
-		newInstance.PrivateIP = info
+	// Only need the API to scrape HostName
```
SaxyPandaBear, line 101: I don't see a lot of value in this comment. It is clear from the code that you are getting the `hostname` metadata value.

translator/util/ec2util/ec2util.go
```diff
@@ -67,46 +71,51 @@ func initEC2UtilSingleton() (newInstance *ec2Util) {
...
+		e.Hostname = hostname
 	} else {
-		log.Println("E! getting local-ipv4 from EC2 metadata fail: ", err)
+		log.Println("E! [EC2] Fetch hostname from EC2 metadata fail:", err)
 	}
 
-	if info, err := md.GetInstanceIdentityDocument(); err == nil {
-		newInstance.Region = info.Region
-		newInstance.AccountID = info.AccountID
+	// Only need the API to scrape Region, AccountId, PrivateIp, Instance ID
```
SaxyPandaBear, line 109: Similarly, I don't see a lot of value from this comment. I understand that there are other attributes that are returned in the identity document, but it is clear what is extracted from the response, directly from the code below. Parroting that information in this comment means that if we add/remove attributes that we store, we also have to update this comment, even if the code already provides enough clear context.

translator/util/ec2util/ec2util.go
```diff
@@ -67,46 +71,51 @@ func initEC2UtilSingleton() (newInstance *ec2Util) {
...
-		return
-	}
+	return
+}
 
-	if info, err := md.GetMetadata("instance-id"); err == nil {
-		newInstance.InstanceID = info
-	} else {
-		log.Println("E! getting instance-id from EC2 metadata fail: ", err)
+func (e *ec2Util) deriveEC2MetadataFromIMDS() error {
```
SaxyPandaBear, line 88: So the only two scenarios where we would return an actual error are if:
1. There is some error with creating a session
2.  IMDS is unavailable

We fail silently (with logs) in a case where there is an error trying to fetch specific data from the metadata endpoint. Why is that? There should be more consistency with the behavior of this function. Does it make sense to return a value here? If so, what are the different intended values, and what causes them to be returned?

translator/util/ec2util/ec2util.go
khanhntd: Agree. The code is straightforward enough. Therefore, I will remove it in the newest commit.

translator/util/ec2util/ec2util.go
khanhntd: Agree with the same though on previous one.

plugins/processors/ec2tagger/ec2tagger.go
SaxyPandaBear: We have the opportunity to address this to make the error message clearer. Can you explain your example? What does this configuration have to do with the ec2tagger specifically? If you can describe that, then we probably have enough data to reword the error message.

plugins/processors/ec2tagger/ec2tagger.go
```diff
@@ -309,25 +242,22 @@ func (t *Tagger) Init() error {
...
 		}
 	}
 
-	t.Log.Infof("ec2tagger: Check ec2 metadata")
-	if !t.ec2metadata.Available() {
-		msg := "ec2tagger: Unable to retrieve InstanceId. This plugin must only be used on an EC2 instance"
-		t.Log.Errorf(msg)
-		return errors.New(msg)
-	}
+	t.Log.Infof("ec2tagger: EC2 IMDS has started initialization.")
```
SaxyPandaBear, line 252: Nit:
1. This doesn't need to be `Infof()` because there is no format string used.
2. The log itself is confusing. It is not immediately obvious what "has started initialization" means in this context. This could be interpreted as the IMDS service is starting on the instance. 

What value does this log provide over the existing one?

translator/util/ec2util/ec2util.go
SaxyPandaBear: Expanding on this - does it make sense to not return an error if there is a non-nil error when trying to get the hostname from the metadata endpoint? Isn't that value required to be populated?

khanhntd: As you mentioned, we would return an actual error for both case. However, does that mean the error will impact the whole process? My answer is no (the same as existing behavior), the return error value only to consolidate the message for the customers. Based on my understanding, the existing behavior for CloudWatchAgent aware that it could be failing in getting the metadata (from host name or from get instance document). However, it does not return error because it can be replaced by the [placeHolder function](https://github.com/aws/amazon-cloudwatch-agent/blob/master/translator/translate/util/placeholderUtil.go#L65-L97). 
So long short story, CloudWatchAgent only expect the failure from those aforementioned case and will always return the values (even if it different intended values because of the replacement with placeholder function) so there are no reason why we would want to return the error when getting one of these metadata endpoint. 

**Note**: This does not apply to EC2Tagger

translator/util/ec2util/ec2util.go
```diff
@@ -67,46 +71,49 @@ func initEC2UtilSingleton() (newInstance *ec2Util) {
...
 
-	if info, err := md.GetInstanceIdentityDocument(); err == nil {
-		newInstance.Region = info.Region
-		newInstance.AccountID = info.AccountID
+	// More information on API: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html
+	if instanceIdentityDocument, err := md.GetInstanceIdentityDocument(); err == nil {
+		e.Region = instanceIdentityDocument.Region
+		e.AccountID = instanceIdentityDocument.AccountID
+		e.PrivateIP = instanceIdentityDocument.PrivateIP
+		e.InstanceID = instanceIdentityDocument.InstanceID
```
SaxyPandaBear, line 113: This is a good catch to consolidate the calls to IMDS to help avoid potential throttling. 

Have you confirmed in testing that using the identity document to get the local ipv4 and instance id show no regression/change compared to the existing way of getting that info?

khanhntd: Yes. There are no regression change between these two.
* **Before the changes:**
```
****** processing amazon-cloudwatch-agent ******
/opt/aws/amazon-cloudwatch-agent/bin/config-downloader --output-dir /opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.d --download-source file:./config.json --mode ec2 --config /opt/aws/amazon-cloudwatch-agent/etc/common-config.toml --multi-config default
Got Home directory: /root
I! Set home dir Linux: /root
I! SDKRegionWithCredsMap region:  us-west-2
Successfully fetched the config and saved in /opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.d/file_config.json.tmp
Start configuration validation...
/opt/aws/amazon-cloudwatch-agent/bin/config-translator --input /opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.json --input-dir /opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.d --output /opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.toml --mode ec2 --config /opt/aws/amazon-cloudwatch-agent/etc/common-config.toml --multi-config default
2022/06/10 03:23:20 Reading json config file path: /opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.d/file_config.json.tmp ...
2022/06/10 03:23:20 I! Valid Json input schema.
I! Detecting run_as_user...
Got Home directory: /root
I! Set home dir Linux: /root
I! SDKRegionWithCredsMap region:  us-west-2
2022/06/10 03:23:20 D! [EC2] Found active network interface
2022/06/10 03:23:20 Instance ID i-005113ad6a00082e0
2022/06/10 03:23:20 Hostname ip-10-189-44-37.amazon.com
2022/06/10 03:23:20 Private Ip 10.189.44.37
```
* **After the changes**
```
****** processing amazon-cloudwatch-agent ******
/opt/aws/amazon-cloudwatch-agent/bin/config-downloader --output-dir /opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.d --download-source file:./config.json --mode ec2 --config /opt/aws/amazon-cloudwatch-agent/etc/common-config.toml --multi-config default
Got Home directory: /root
I! Set home dir Linux: /root
I! SDKRegionWithCredsMap region:  us-west-2
Successfully fetched the config and saved in /opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.d/file_config.json.tmp
Start configuration validation...
/opt/aws/amazon-cloudwatch-agent/bin/config-translator --input /opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.json --input-dir /opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.d --output /opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.toml --mode ec2 --config /opt/aws/amazon-cloudwatch-agent/etc/common-config.toml --multi-config default
2022/06/10 03:56:00 Reading json config file path: /opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.d/file_config.json.tmp ...
2022/06/10 03:56:00 I! Valid Json input schema.
I! Detecting run_as_user...
Got Home directory: /root
I! Set home dir Linux: /root
I! SDKRegionWithCredsMap region:  us-west-2
2022/06/10 03:56:00 D! [EC2] Found active network interface
2022/06/10 03:56:00 Private Ip nay 10.189.44.37
2022/06/10 03:56:00 Instance ID nay i-005113ad6a00082e0
No csm configuration found.
Configuration validation first phase succeeded
/opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent -schematest -config /opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.toml
Configuration validation second phase succeeded
Configuration validation succeeded
Redirecting to /bin/systemctl stop amazon-cloudwatch-agent.service
Redirecting to /bin/systemctl restart amazon-cloudwatch-agent.service
```

plugins/processors/ec2tagger/ec2tagger.go
khanhntd: No deep reason behind it. I forgot to change back when merging from the telegraf upgrade. Nice catch on this 👍

plugins/processors/ec2tagger/constants.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+const (
+	allowedIMDSRetries = 2
+)
+
+var (
+	defaultIMDSTimeout     = 1 * time.Second
+	defaultRefreshInterval = 180 * time.Second
+	backoffSleepArray      = []time.Duration{0, 1 * time.Minute, 1 * time.Minute, 3 * time.Minute, 3 * time.Minute, 3 * time.Minute, 10 * time.Minute} // backoff retry for ec2 describe instances API call. Assuming the throttle limit is 20 per second. 10 mins allow 12000 API calls.
```
sethAmazon, line 78: This should be a random number imo. Think all systems send out the call at the same time. Then next round they all send at the same time. It needs to be random within an interval. This comment is probs out of scope for this pr but something to think about for best behaved agents. @straussb thoughts.

plugins/processors/ec2tagger/ec2tagger.go
```diff
@@ -309,25 +242,22 @@ func (t *Tagger) Init() error {
...
-	}
+	t.Log.Infof("ec2tagger: Check EC2 Metadata.")
 
-	doc, err := t.ec2metadata.GetInstanceIdentityDocument()
-	if nil != err {
-		msg := fmt.Sprintf("ec2tagger: Unable to retrieve InstanceId : %+v", err.Error())
-		t.Log.Errorf(msg)
-		return errors.New(msg)
+	doc, err := t.ec2MetadataProvider().GetInstanceIdentityDocument()
+	if err != nil {
```
sethAmazon, line 255: We do not know 100% this is timeout imo.

plugins/processors/ec2tagger/ec2tagger.go
```diff
@@ -309,25 +242,22 @@ func (t *Tagger) Init() error {
...
+		t.Log.Error("ec2tagger: Timeout when retrieving EC2 Metadata. This plugin must only be used on an EC2 instance.")
+		if context.CurrentContext().RunInContainer() {
+			t.Log.Warn("ec2tagger: Please increase hop limit to 2 by following this document https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-options.html#configuring-IMDS-existing-instances.")
```
sethAmazon, line 258: ec2tagger: timeout may occurred because hop limit is too small. Please increase hop limit to 2 by following this document https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-options.html#configuring-IMDS-existing-instances.

plugins/processors/ec2tagger/ec2tagger.go
```diff
@@ -237,20 +170,20 @@ func (t *Tagger) refreshLoop(refreshInterval time.Duration, stopAfterFirstSucces
...
 				if !refreshTags && !refreshVolumes {
-					t.Log.Infof("ec2tagger: Refresh is no longer needed, stop refreshTicker.")
+					t.Log.Info("ec2tagger: Refresh is no longer needed, stop refreshTicker.")
 					return
 				}
 			}
 
 			if refreshTags {
 				if err := t.updateTags(); err != nil {
-					t.Log.Warnf("ec2tagger: Error refreshing EC2 tags, keeping old values : +%v", err.Error())
```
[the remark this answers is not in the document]
khanhntd: For +%v, its a typo I believe. Here is an example of unspecified format
```
type Link struct { A, Z string }
l := Link{A: "n1:2", Z: "n2:1"}
fmt.Printf("%v\n", l)	// {n1:2 n2:1}
fmt.Printf("%+v\n", l)	// {A:n1:2 Z:n2:1}
fmt.Printf("%#v\n", l)	// pkg.Link{A:"n1:2", Z:"n2:1"}
```

plugins/processors/ec2tagger/ec2tagger.go
khanhntd: Agree on the HTTP Error Request. However, it will always return timeout even though the request returns a different error. Timeout is only a consolidate message for most of the errors unless we want to handle every single error.

plugins/processors/ec2tagger/ec2tagger.go
khanhntd: Agree 👍 . Thanks for the suggesting wording.

plugins/processors/ec2tagger/ec2tagger.go
```diff
@@ -418,44 +366,47 @@ func (t *Tagger) refreshLoopToUpdateTagsAndVolumes() {
 	}
 }
 
-// updateVolumes calls EC2 describe volume
-func (t *Tagger) updateVolumes() error {
-	if t.ebsVolume == nil {
-		t.ebsVolume = NewEbsVolume()
+/*
+	Retrieve metadata from IMDSv2,IMDSv1 and use these metadata to:
```
SaxyPandaBear, line 370: There is no point to call out that it uses IMDSv2 and IMDSv1 because you are not explicitly calling both of them and are letting the underlying SDK determine that for you.

plugins/processors/ec2tagger/ec2tagger.go
```diff
@@ -418,44 +366,47 @@ func (t *Tagger) refreshLoopToUpdateTagsAndVolumes() {
...
+	* Extract InstanceID, ImageID, InstanceType to create custom dimension for collected metrics
+	* Extract InstanceID to retrieve Instance's Volume and Tags
+	* Extract Region to create aws session with custom configuration
+	For more information on IMDSv2, please follow this document https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-service.html
+*/
+func (t *Tagger) initialRetrievalOfEC2Metadata() error {
```
SaxyPandaBear, line 376: This function name is confusing. What is the purpose of this function? It reaches out to IMDS and then stores some of the metadata so that we can use it internally. There is never really any reason to have `Of` in a function or method name. That is a clear indication that the naming is improper. The name should be changed to be more clear about the purpose. "Initial retrieval of EC2 Metadata" doesn't tell me what this does clearly. Why is it "initial"? 

```suggestion
func (t *Tagger) cacheEC2Metadata() error {
```

plugins/processors/ec2tagger/ec2tagger.go
```diff
@@ -418,44 +366,47 @@ func (t *Tagger) refreshLoopToUpdateTagsAndVolumes() {
...
-				Name:   aws.String("attachment.instance-id"),
-				Values: aws.StringSlice([]string{t.instanceId}),
-			},
-		},
+	t.Log.Infof("ec2tagger: Check EC2 Metadata.")
+	doc, err := t.ec2MetadataProvider().GetInstanceIdentityDocument()
+	if err != nil {
+		t.Log.Error("ec2tagger: Timeout when retrieving EC2 Metadata. This plugin must only be used on an EC2 instance.")
+		if context.CurrentContext().RunInContainer() {
+			t.Log.Warn("ec2tagger: Timeout may occurred because hop limit is too small. Please increase hop limit to 2 by following this document https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-options.html#configuring-IMDS-existing-instances.")
```
SaxyPandaBear, line 395: Nit-
```suggestion
			t.Log.Warn("ec2tagger: Timeout may have occurred because hop limit is too small. Please increase hop limit to 2 by following  https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/configuring-instance-metadata-options.html#configuring-IMDS-existing-instances.")
```

plugins/processors/ec2tagger/ec2tagger.go
```diff
@@ -418,44 +366,47 @@ func (t *Tagger) refreshLoopToUpdateTagsAndVolumes() {
...
-		Filters: []*ec2.Filter{
-			{
-				Name:   aws.String("attachment.instance-id"),
-				Values: aws.StringSlice([]string{t.instanceId}),
-			},
-		},
+	t.Log.Infof("ec2tagger: Check EC2 Metadata.")
+	doc, err := t.ec2MetadataProvider().GetInstanceIdentityDocument()
+	if err != nil {
+		t.Log.Error("ec2tagger: Timeout when retrieving EC2 Metadata. This plugin must only be used on an EC2 instance.")
```
SaxyPandaBear, line 393: I think Seth already called this out, but there's no guarantee that we get an error because of a timeout.

plugins/processors/ec2tagger/ec2tagger.go
```diff
@@ -418,44 +366,47 @@ func (t *Tagger) refreshLoopToUpdateTagsAndVolumes() {
...
-			for _, attachment := range volume.Attachments {
-				t.ebsVolume.addEbsVolumeMapping(volume.AvailabilityZone, attachment)
-			}
-		}
-		if result.NextToken == nil {
-			break
-		}
-		input.SetNextToken(*result.NextToken)
+	t.ec2MetadataRespond.region = doc.Region
+	t.ec2MetadataRespond.instanceId = doc.InstanceID
```
SaxyPandaBear, line 401: Why do we not check the lookup for instance ID like the other values below?

plugins/processors/ec2tagger/ec2tagger.go
```diff
@@ -4,103 +4,44 @@
...
 	imageId      bool
 	instanceType bool
 }
 
-type ec2ProviderType func(*configaws.CredentialConfig) ec2iface.EC2API
-
-type ec2Metadata interface {
-	Available() bool
-	GetInstanceIdentityDocument() (ec2metadata.EC2InstanceIdentityDocument, error)
+type ec2MetadataRespond struct {
```
SaxyPandaBear, line 35: Why are we putting all of this information inside of another struct? I don't see any benefit for pulling these out into a struct.

plugins/processors/ec2tagger/ec2tagger.go
```diff
@@ -118,20 +59,17 @@ type Tagger struct {
...
-	shutdownC      chan bool
-	tagFilters     []*ec2.Filter
-	metadataLookup metadataLookup
-	ebsVolume      *EbsVolume
+	ec2TagCache         map[string]string
+	started             bool
+	ec2Provider         ec2ProviderType
+	ec2API              ec2iface.EC2API
+	ec2MetadataProvider ec2MetadataProviderType
+	ec2MetadataRespond  ec2MetadataRespond
```
SaxyPandaBear, line 67: Nit - It's improper to have the name of the variable shadow the name of the type.

plugins/processors/ec2tagger/ec2tagger.go
```diff
@@ -184,6 +182,51 @@ func (t *Tagger) Apply(in ...telegraf.Metric) []telegraf.Metric {
 	return in
 }
 
+// Shutdown currently does not get called, as telegraf does not have a cleanup hook for Filter plugins
+func (t *Tagger) Shutdown() {
```
SaxyPandaBear, line 186: Moving all of the functions around in the file makes this a confusing diff so it'll take longer to review.

plugins/processors/ec2tagger/ec2tagger_test.go
```diff
@@ -275,26 +278,21 @@ func TestInitSuccessWithNoTagsVolumesUpdate(t *testing.T) {
 	assert.Equal(3, ec2Client.tagsCallCount)
 	assert.Equal(2, ec2Client.volumesCallCount)
 	//check tags and volumes
-	expectedTags := map[string]string{
-		"tagKey1":              "tagVal1",
-		"tagKey2":              "tagVal2",
-		"AutoScalingGroupName": "ASG-1",
-	}
+	expectedTags := map[string]string{tagKey1: tagVal1, tagKey2: tagVal2, "AutoScalingGroupName": tagVal3}
```
SaxyPandaBear, line 281: Why is there inconsistency for the map keys? `tagKey1` is created as a var/const at the top of the file but `"AutoScalingGroupName"` gets redefined in all of the tests.

plugins/processors/ec2tagger/ec2tagger.go
khanhntd: Agree. Resolving it in the newest commit.

plugins/processors/ec2tagger/ec2tagger.go
khanhntd: Agree. Should go with the initial name that you recommended me `deriveEC2MetadataFromIMDS` instead of `initialRetrievalOfEC2Metadata`. Thanks for reminding me.

plugins/processors/ec2tagger/ec2tagger.go
khanhntd: Agree. Should go with the initial log message instead of using timeout. 👍

plugins/processors/ec2tagger/ec2tagger.go
khanhntd: InstanceID is a required field, instead of **optional field** since InstanceID can be used to retrieve EC2 Instance Volumes and Tags (e.g [this line](https://github.com/aws/amazon-cloudwatch-agent/pull/480/files#diff-6638b992018408bf44cac12361171fd050ff44fcc24c28f326988f4768a6dc2dR102). Therefore, it is not optional to check the  lookup for instanceID

plugins/processors/ec2tagger/ec2tagger.go
khanhntd: Agree its take longer time and confusing. However, for better structure format (e.g following the [ecs_decorator](https://github.com/aws/amazon-cloudwatch-agent/blob/master/plugins/processors/ecsdecorator/ecsdecorator.go)), its worth to restructure it and better considerate for future reviewers.

plugins/processors/ec2tagger/ec2tagger.go
khanhntd: Its the same reason for [ec2MetadataLookup](https://github.com/aws/amazon-cloudwatch-agent/pull/480/files#diff-6638b992018408bf44cac12361171fd050ff44fcc24c28f326988f4768a6dc2dR29-R33):
* Increase consolidation since its consolidate all the EC2MetadataRespond
* Less variables to be aware of when looking at `EC2Tagger interface`

plugins/processors/ec2tagger/ec2tagger_test.go
khanhntd: This may explain why the inconsistency for ASG:
```
## Note: This plugin renames the "aws:autoscaling:groupName" EC2 Instance Tag key to be spelled "AutoScalingGroupName".
  ## This aligns it with the AutoScaling dimension-name seen in AWS CloudWatch.
  # ec2_instance_tag_keys = ["aws:autoscaling:groupName", "Name"]
```

SaxyPandaBear: I don't understand what this means

plugins/processors/ec2tagger/ec2tagger.go
```diff
@@ -418,44 +366,47 @@ func (t *Tagger) refreshLoopToUpdateTagsAndVolumes() {
...
+	* Extract InstanceID, ImageID, InstanceType to create custom dimension for collected metrics
+	* Extract InstanceID to retrieve Instance's Volume and Tags
+	* Extract Region to create aws session with custom configuration
+	For more information on IMDS, please follow this document https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html
+*/
+func (t *Tagger) deriveEC2MetadataFromIMDS() error {
+	for _, tag := range t.EC2MetadataTags {
+		switch tag {
+			case mdKeyInstanceId:
+				t.ec2MetadataLookup.instanceId = true
```
SaxyPandaBear, line 380: If instance ID is "required" then what was the original purpose of this lookup check? And why are we keeping it?

plugins/processors/ec2tagger/ec2tagger.go
SaxyPandaBear: How is it less variables to be aware of if you are introducing another variable? I don't think the reasons you presented are convincing enough for this change. Are we going to be passing around this "response" struct outside of this Tagger? If not, it just makes it more confusing and harder to maintain.

plugins/processors/ec2tagger/ec2tagger.go
khanhntd: As you already know, the EC2Tagger's purpose is to append the dimensions (which can be scraped by tags and metadata from IMDS) to the metrics. Therefore, the `ec2MetadataLookup` is to ensured the customer can append the [dimensions to their needs](https://github.com/aws/amazon-cloudwatch-agent/blob/master/plugins/processors/ec2tagger/ec2tagger.go#L167-L175).

**For example:** The below configuration will ignore `InstanceType` but will collect ImageID and add the dimension ImageId to the metrics, while InstanceId still be connected but only for retrieving tags and volumes.
```
"append_dimensions": {
			"ImageId": "${aws:ImageId}",
},
```
While if customer does have `"InstanceId": "${aws:InstanceId}",`, the CloudWatchAgent will also append the InstanceId and ImageId to the metrics as their dimension.

plugins/processors/ec2tagger/ec2tagger_test.go
khanhntd: Based on my understanding, whenever CWAgent retrieves EC2 Instance Tags (not volume or not IMDS), the EC2 Instance will have the key tag as `aws:autoscaling:groupName` for AutoScalingGroup; then we will scrapped it; converted to `AutoScalingGroupName` as a key (back and forth) and appends the AutoScalingGroup as a dimension to the collected metrics (to match the `AutoScalingGroupName` key in CWAgent JSON configuration)

So long short story, the EC2 has tag `aws:autoscaling:groupName` (and its value) for AutoScalingGroup, we collected the `aws:autoscaling:groupName` tag, converted to `AutoScalingGroup` as a dimension. 

Let's me know if that make sense

plugins/processors/ec2tagger/ec2tagger.go
khanhntd: Based on my understanding, no for both EC2MetadataLookup and EC2MetadataRespond need to be passed outside the Tagger. 
Long short story, these metadata are parsed from IMDS, so why it's more confusing and harder to maintain? This follows the definition of `type structure` (```It is used to group related data to form a single unit. ```) so I'm not sure the reason why not to change it.
```
	instanceId     string
	imageId        string // aka AMI
	instanceType   string
	region         string
```
,

plugins/processors/ec2tagger/ec2tagger.go
khanhntd: If that's the case, I will revert it back 👍

SaxyPandaBear (approved):

PR #481, opened 2022-06-08
Increase PMD batch & buffer sizes

movence:
# Description of the issue
The limit on datum size with `PutMetricData` has been increased to 1000, and CWA should be aligned with that change in order to send more datums per API call.

# Description of changes
Increase the default datum size to 1000.

# License

# Tests
Running the existing integration test

# Requirements
2. Run `make linter`


plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -30,7 +30,7 @@ import (
 )
 
 const (
-	defaultMaxDatumsPerCall        = 20     // PutMetricData only supports up to 20 data metrics per call by default
+	defaultMaxDatumsPerCall        = 1000   // PutMetricData only supports up to 1000 data metrics per call by default
```
khanhntd, line 33: No blocker from me: Are there any public documents regarding the PMD Metric Datum that we could put here for future reference? Another thing would be [PutMetricData documents](https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_PutMetricData.html) only show 20 MetricDatum and 10 dimensions  instead of 1000 MetricDatum and 30 dimensions. This brings me the contrast of referencing public documents when comparing to CloudWatchAgent default PMD calls.

khanhntd:
Another thing to notice would be the Unit Test is failing all over different OS and its does not relate to our usual flakiness test. Therefore, we would need to deep dive into this 
```
--- FAIL: TestWrite (3.01s)
    cloudwatch_test.go:368: 
        	Error Trace:	cloudwatch_test.go:368
        	Error:      	Not equal: 
        	            	expected: 2
        	            	actual  : 1
        	Test:       	TestWrite
        	Messages:   	Expected number of calls (2) does not match the actual number of calls (1).
    cloudwatch_test.go:368: 
        	Error Trace:	cloudwatch_test.go:368
        	Error:      	Should be true
        	Test:       	TestWrite
FAIL
coverage: 65.7% of statements
```

SaxyPandaBear:
Is the change already deployed on the backend for this? Is there a way for us to manually/automatically integration test just to be sure it works and there isn't any other bottleneck or limiter?

movence:
> Another thing to notice would be the **unit test for CWAgent is failing with different OS** and its does not relate to our usual flakiness test.

I don't believe it's tied to any specific OSes but a minor failed test case which should have been fixed.

> coverage percentage is rather small

Are you suggesting the overall code coverage should be bumped up?

khanhntd:
> I don't believe it's tied to any specific OSes but a minor failed test case which should have been fixed.

Yes it does not tied to any specific OS. Good job on fixing the test  👍 .  However, the GitHub blocks running `unit-test` workflow until resolve the conflict

> Are you suggesting the overall code coverage should be bumped up?

Yes. However, it's a no blocker from me. If we confident the existing unit tests has covered enough daily use case then it's fine by me.

Another non-block from me would be adding integration test, same as @SaxyPandaBear idea. For the idea, Seth has done similar to this (https://github.com/aws/amazon-cloudwatch-agent/tree/master/integration/test/metrics_number_dimension)

SaxyPandaBear:
What coverage are you referring to @khanhntd? The change is tiny. Are there specific test cases that  you think are missing? Otherwise I don't see what you really think we can improve on test coverage for a one variable change. The only possible thing I could see is integration testing, like what was done for the max dimensions change, but that's not unit test coverage.

khanhntd:
> What coverage are you referring to @khanhntd? The change is tiny. Are there specific test cases that you think are missing? Otherwise I don't see what you really think we can improve on test coverage for a one variable change. The only possible thing I could see is integration testing, like what was done for the max dimensions change, but that's not unit test coverage.

Yes. That's why I mentioned its a **non-blocker from me** consider only the variables change  and the test bumping coverage can be done in another PR 👍.

adam-mateen (approved):
approved

I see existing unit test - `TestIsFull()` tests `batch.isFull()`.
Unit tests do not make actual calls using the actual CloudWatch backend though.

SaxyPandaBear (dismissed):
Lgtm

plugins/outputs/cloudwatch/cloudwatch.go
khanhntd, line 33: If the `defaultMaxDatumsPerCall` is increased, I would assume the `defaultMaxValuesPerDatum` would change. Therefore, would consider removing the rules when applied with internal flag. For example these two files: [maxDatumsPerCall](https://github.com/aws/amazon-cloudwatch-agent/blob/master/translator/translate/metrics/ruleMaxDatumsPerCall.go)  and also [maxValumPerDatum](https://github.com/aws/amazon-cloudwatch-agent/blob/master/translator/translate/metrics/ruleMaxValuesPerDatum.go)

SaxyPandaBear:
Is the backend ready for this or are we still waiting?

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -30,9 +30,9 @@ import (
...
 	defaultMaxValuesPerDatum       = 150    // By default only these number of values can be inserted into the value list
-	bottomLinePayloadSizeToPublish = 200000 // Leave 9600B for the last datum buffer. 200KB payload size, 5:1 compression ratio estimate.
+	bottomLinePayloadSizeToPublish = 500000 // Leave 2400B for the last datum buffer. 500KB payload size, 5:1 compression ratio estimate.
```
khanhntd, line 35: Should sync up with internal: `bottomLinePayloadSizeToPublish = 999 000` (0.999 MB) and delete `5:1 compression ratio estimate`

adam-mateen: Can you confirm the API limit is actually 1 MiB, as in 1024*1024 Bytes?
If so then this limit could be raised to 1023 KiB (1023*1024 Bytes).
I don't think it is critical since a payload of 1,000 metric datums is unlikely to reach 1023 KiB.
May as well be as accurate as possible, right?

SaxyPandaBear:
Done in #523

PR #486, opened 2022-06-09
Add Oracle Linux 8.6 to integration test

movence:
# Description of the issue
- Need to support OL 8.6 with the integration test
- README for the integration test is missing a few IAM policies and a step to create ECR repository

# Description of changes
- Build a public CWA integration test AMI based off of OL 8.6 AMI then add the new test AMI to `ec2_linux_test_matrix.json`

# License

# Tests
Executing the integration test in Github Actions

# Requirements
2. Run `make linter`


integration/terraform/ec2/README.md
```diff
@@ -37,8 +37,45 @@ for how to easily generate a new policy.
...
+        "s3:PutObject",
         "sts:GetCallerIdentity",
-        "s3:PutObject"
+        "ssm:PutParameter",
+        "ssm:DeleteParameter",
+        "ssm:DescribeParameters",
+        "ssm:ListTagsForResource",
+        "ssm:GetParameters",
+        "ssm:GetParameter",
+        "ssm:DeleteParameters",
```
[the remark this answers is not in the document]
movence: The test was throwing a few SSM related errors while running `prometheus-integ-test-ssm-config`.

integration/terraform/ec2/README.md
```diff
@@ -88,7 +125,9 @@ for how to easily generate a new policy.
         "logs:PutLogEvents",
         "s3:GetObjectAcl",
         "s3:GetObject",
-        "s3:ListBucket"
+        "s3:ListBucket",
+        "s3:PutObject",
```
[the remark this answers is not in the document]
movence: Good catch! It must have been added to wrong section while juggling with policies on different roles.

integration/generator/resources/ec2_linux_test_matrix.json
```diff
@@ -96,5 +96,14 @@
     "caCertPath": "/etc/ssl/certs/ca-bundle.crt",
     "arc": "amd64",
     "binaryName": "amazon-cloudwatch-agent.rpm"
+  },
+  {
```
khanhntd, line 100: No blocker from me: Do we want to test on every different version of the same OS when we have an upper bound and lower bound? For example: oracle linux 8-2 is no different than ol 8-1 (not any breaking changes being introduced so far); therefore, only ol8-1 and ol8-6 would be enough for CWAgent (open suggestion here)

integration/terraform/ec2/README.md
```diff
@@ -130,8 +168,16 @@ See [docs](https://docs.aws.amazon.com/cli/latest/userguide/cli-services-ec2-key
 on creating the key pair.
 > Note: Store the private key in a secure location!
 
-**Reminder: the EC2 key pair must be in the same region as the instances, so this assumes that the key pair is created
-in the `us-west-2` region.**
+### Create ECR Repository
+|Field              |Value                   |
```
khanhntd, line 172: **No blocker from me**: the EC2 README requires to create this resource beforehand. Therefore, I would say it would be more effortless to create these resources in terraform script and also let developer follow our requirements. For example: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ecr_repository

integration/generator/resources/ec2_linux_test_matrix.json
movence: I think it should work that way as long as we totally understand what's changing between versions and any side effects.

integration/terraform/ec2/README.md
```diff
@@ -37,8 +37,45 @@ for how to easily generate a new policy.
         "ec2:ModifyInstanceAttribute",
         "ec2:RunInstances",
         "ec2:TerminateInstances",
+        "s3:ListBucket",
+        "s3:GetObject",
+        "s3:GetObjectAcl",
```
khanhntd, line 42: Does the Integration Test need this though? Correct me if I am wrong since the Acl is more public permission and GetObject is more for private which we are using the private s3.

integration/generator/resources/ec2_linux_test_matrix.json
khanhntd: Agree 👍 . Will suggest this in parking lot and see how its go from there.

integration/terraform/ec2/README.md
SaxyPandaBear: That should be addressed separately.

integration/terraform/ec2/README.md
```diff
@@ -186,6 +232,8 @@ secrets.
 3. Select the `Run Integration Tests` action
 4. Select `Run workflow`, and choose the branch to execute integration tests on
 
+> Note: The Amazon EC2 quota limit might need to be increased to run the integration test: request an increase via `All Standard (A, C, D, H, I, M, R, T, Z) Spot Instance Request`
```
SaxyPandaBear, line 235: Ideally the integration tests shouldn't need this quota increase and should be able to queue things and execute as instance capacity frees up, but that's a problem for another day. As part of this note, can you call out that each EC2 integration test suite gets run on a different EC2, per OS tested?

movence: Updated the note as suggested. Thanks!

khanhntd (approved):
LGTM.

PR #487, opened 2022-06-09
Nvidia windows

ZhenyuTan-amz:
# Description of the issue
Current Nvidia GPU metrics plugin only works on Linux

# Description of changes
Adding a config translation rule for Windows so that with proper configuration, Nvidia GPU metrics will be collected on Windows instances.

# License

# Tests
Manually tested the feature, unit test the config translation, will add integration test in another PR

### DIFF
--- a/translator/translate/metrics/util/measurementutil.go
+++ b/translator/translate/metrics/util/measurementutil.go
@@ -22,6 +22,8 @@ const measurement_rename = "rename"
 const measurement_unit = "unit"
 const nvidia_smi_plugin_name = "nvidia_smi"
 const tag_exclude_key = "tagexclude"
+const smi_bin_path = "bin_path"
+const default_windows_smi_path = "C:\\Program Files\\NVIDIA Corporation\\NVSMI\\nvidia-smi.exe"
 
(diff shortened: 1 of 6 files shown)


translator/translate/metrics/util/measurementutil.go
```diff
@@ -22,6 +22,8 @@ const measurement_rename = "rename"
 const measurement_unit = "unit"
 const nvidia_smi_plugin_name = "nvidia_smi"
 const tag_exclude_key = "tagexclude"
+const smi_bin_path = "bin_path"
+const default_windows_smi_path = "C:\\Program Files\\NVIDIA Corporation\\NVSMI\\nvidia-smi.exe"
```
SaxyPandaBear, line 26: Is the path something that we need to expose for people to be able to configure to execute from another location? Or is it just expected to be at this path?

ZhenyuTan-amz: Windows OS (except Windows 10) with Nvidia tool installed will have smi in this path by default, also this is mentioned in the README of this plugin: https://github.com/aws/telegraf/tree/master/plugins/inputs/nvidia_smi, not sure how visible this is to customer, since it's in the telegraf repo instead of this one.

SaxyPandaBear (approved):
Lgtm

PR #491, opened 2022-06-15
Log number of montiored files

### LINKED ISSUE #405: Delete not-update logs after a certain amount of time.
**Description:**
Currently, CloudWatch Agent only stop monitors logs file only if `auto_removal` was turned on and that applies to the case when  logs files were not updated for a certain amount of time. However, this would leads to an issue of increasing files that Cloudwatch Agent needs to monitor, which would end up `run out of file descriptor` (meaning cannot open any more logs files to monitor) in most of the OS (including Linux). An example would be  
```
2021-10-12T12:18:16Z E! [outputs.cloudwatchlogs] Aws error received when sending logs to ./audit: NoCredentialProviders: no valid providers in chain
caused by: EnvAccessKeyNotFound: failed to find credentials in the environment.
SharedCredsLoad: failed to load profile, .
EC2RoleRequestError: no EC2 instance role found
caused by: RequestError: send request failed
caused by: Get "http://169.254.169.254/latest/meta-data/iam/security-credentials/": dial tcp 169.254.169.254:80: socket: too many open files
```

**Solution**:
Detect the newest changes when tailing and compare current time. If there was nothing changes regarding the log file, kill the tail process. 
 
**Alternative solution:**
* [Increase max open file limit](http://woshub.com/too-many-open-files-error-linux/) on a particular OS
* Use [Log Sweeper](https://github.com/umbrellio/log_sweeper) to rotate and delete logs after a certain amount of time.

### LINKED ISSUE #432: Feature request: print out how many files the agent is currently monitoring on demand
It'd be nice if there was a way to print out or log how many files the agent is currently tailing at any given point in time. The agent allows for glob-like syntax to discover files to tail, so people can pretty easily provide a bad log configuration that explodes CPU usage.

Because the agent allows for glob-style syntax, it is possible to provide a configuration that enumerates a lot of files to tail at the same time, which can/will cause CPU spikes, and can also eat up all of the file descriptors allowed on the host (see #405).

Now, this information is kind of available via the logs already, but not in a very succinct way. You could analyze the agent log file for all of the info logs of the form `I! [logagent] piping log from %v/%v(%v) to %v with retention %v`, and just count all of the unique file names (the 4th param) in order to see how many files are being monitored, but is not super intuitive to do, and keeps the agent difficult to use/troubleshoot without intimate knowledge of what it's doing under the hood. 

---
Ideally this would work in a non-intrusive way. You shouldn't have to restart the agent or update configurations in order to see this debugging information. Luckily, the agent supports setting the debug log level via an environment variable, since #293 (released in v1.247350). So adding something that emits/logs how many files are being actively tailed should be pretty seamless. 

The first thing that comes to mind is adding that info to the internal profiler, but am open to other solutions. In `logs.go`, the `LogAgent` already keeps track of a map in memory, `destNames`, which maps a single tailed file to its destination (i.e.: `cloudwatchlogs`). The issue with the profiler is that it only allows metric addition, so if we added something rudimentary like `profiler.AddStats("num_logs_monitored", len(l.destNames))`, it would add that number to what is already in the profiler. The timing wouldn't match up - the log agent has a ticker with a 1 second frequency, but the profiler dumps metrics and clears on a 60 second frequency. As such, we would get 60x the real number with that approach.

khanhntd:
Close #405 
Close #432 

# Description of the issue
If CloudWatchAgent runs in an intensive log environment, CloudWatchAgent will monitor all the files and it can exhaust all the file descriptors that being allowed by the OS for a single process. Therefore, CloudWatchAgent needs to show the number of monitored files and show the number of allowed to be monitored for customer to have corresponding actions.

# Description of changes
* Create `file-descriptor` add-on for each OS and different OS can return different file descriptors limit.
* Create a TailerQueue ( could suggest a better name here) to centralize the tailers (which contains a buffer channeled), add tailer and release tailer

# License

# Reasons to choose buffered channel
* [Allocated small memory upfront](https://stackoverflow.com/questions/36460306/how-does-memory-allocation-work-for-buffered-channels). A small trade off with the second reason since customers does not want to increase their memory when using CloudWatchAgent
* Will able to centralize the tailer (e.g stop one tailer in exchange for other tailer to push the logs completely and start the dropped tailer again until the previous execution is done)

# Tests
There are two tests that I have done:
* Generate logs everyone 3 minutes with `auto_removal` features:
Sample app
```
import glob
import os
import logging
import time
import json

# get root logger
logger = logging.getLogger()
logger.setLevel(logging.INFO)

# rotate our log file every 3 minutes
handler = TimedRotatingFileHandler("logdest/logs.log", when="M", interval=3)
logger.addHandler(handler)

# log a message
logging.info(json.dumps({"Metric": "12345"*10}))
# sleep for 4 minutes so that file will rotate upon next log message
time.sleep(60*4)
# log another message (this one will not appear since byte length of message == byte length of old log file)
logging.info(json.dumps({"Metric": "09876"*10}))
# sleep for another 4 minutes so that file will rotate upon next log message
time.sleep(60*4)
# this message will be partially written
logging.info({"Metric": "1234567890"*10})
```

Sample CloudWatchAgent JSON configuration:
```
{
        "agent": {
                "run_as_user": "root",
                "debug": true
        },
	"logs": {
                "logs_collected": {
                        "files": {
                                "collect_list": [
                                        {
                                                "file_path":"/local/home/khanhntd/Code/CloudWatch/logdest/**",
                                                "auto_removal": true,
                                                "log_group_name": "test-amazon-cloudwatch-agent.log",
                                                "log_stream_name": "{instance_id}-CloudWatchAgent",
                                        },
                                        {
                                                "log_group_name": "amazon-cloudwatch-agent.log",
                                                "log_stream_name": "{instance_id}-CloudWatchAgent",
                                        },
                                        {
                                                "file_path": "/home/khanhntd/Code/MetricAgent.log",
                                                "log_group_name": "amazon-cloudwatch-agent.log",
                                                "log_stream_name": "{instance_id}-MetricAgent",
                                        }
                                ]
                        }
                }
        },
}
```

*
After running these two tests, confirm the monitored file by comparing the monitored files `lsof -p {{CloudWatchAgent process id}}` with logs in `CloudWatchAgent.logs`. An example of expected results:
```
2022-06-15T19:47:31Z I! [logagent] piping log from amazon-cloudwatch-agent.log/i-005113ad6a00082e0-MetricAgent(/home/khanhntd/Code/MetricAgent.log) to cloudwatchlogs with retention -1
2022-06-15T19:47:32Z I! [inputs.logfile] Number of current monitored log files / Allowed monitored log files: 3/65535
2022-06-15T19:47:32Z D! [outputs.cloudwatchlogs] Buffer fullness: 0 / 10000 metrics
2022-06-15T19:47:32Z D! [outputs.cloudwatch] Buffer fullness: 0 / 10000 metrics
2022-06-15T19:47:33Z I! [inputs.logfile] Number of current monitored log files / Allowed monitored log files: 3/65535
```

* Generate log rotation by using `logrotate`
Sample script
```
#!/bin/sh
#mode can only be ""(default), create, copy, copytruncate
mode=$1
pkill -9 pprofCol


export cwAgentDir="/opt/aws/amazon-cloudwatch-agent"

# start up a new cwagent with pprof enabled
if [ "$(/sbin/init --version 2>/dev/null | grep -c upstart)" = 1 ]; then
        stop amazon-cloudwatch-agent || true
else
    	service amazon-cloudwatch-agent stop || true
fi
sleep 10

(${cwAgentDir}/bin/amazon-cloudwatch-agent -config ${cwAgentDir}/etc/amazon-cloudwatch-agent.toml --pprof-addr :6060 > /dev/null 2>&1 &)

sleep 3

./pprofCol -textInterval=300 -profileInterval=3600 -truncateSize=10 > /dev/null 2>&1 &

mkdir soakTest

cat > minute_rotate.conf << EOF
/tmp/soakTest/minute_*.log {
        size 10k
        rotate 4
        compress
        notifempty
        ${mode}
}
EOF

cat > hour_rotate.conf << EOF
/tmp/soakTest/hour.log {
        size 10k
        rotate 4
        compress
        notifempty
        ${mode}
}
EOF

#set up cron job to run logrotate per minute/hour
sudo service crond start
crontab << EOF
* * * * * sudo logrotate -f /tmp/minute_rotate.conf
3 * * * * sudo logrotate -f /tmp/hour_rotate.conf
EOF

#generate logs continuously
(./generate_log.sh ${mode} > /dev/null 2>&1 &)
```

Sample Amazon CloudWatch Agent JSON configuration
```
{"agent":{"metrics_collection_interval":10},"metrics":{"append_dimensions":{"InstanceType":"${aws:InstanceType}"},"metrics_collected":{"cpu":{"measurement":["cpu_usage_idle","cpu_usage_iowait","cpu_usage_user","cpu_usage_system"],"totalcpu":false},"disk":{"measurement":["used_percent","inodes_free"]},"diskio":{"measurement":["io_time"]},"mem":{"measurement":["mem_used_percent"]},"statsd":{},"swap":{"measurement":["swap_used_percent"]},"procstat":[{"measurement":["cpu_usage","memory_rss"],"exe":"amazon-cloudwatch-agent"}]},"namespace":"CWAgent/LinuxSoakTest"},"logs":{"logs_collected":{"files":{"collect_list":[{"file_path":"/opt/aws/amazon-cloudwatch-agent/logs/amazon-cloudwatch-agent.log","log_group_name":"Linux-Soak-CWAgent.log","multi_line_start_pattern":"{timestamp_format}","timestamp_format":"%Y-%m-%dT%H:%M:%SZ","timezone":"UTC"},{"file_path":"/tmp/soakTest/memStats","log_group_name":"Linux-Soak-memStats","multi_line_start_pattern":"{timestamp_format}","timestamp_format":"%d %b %y %H:%M:%S","timezone":"UTC"},{"file_path":"/tmp/soakTest/goroutine","log_group_name":"Linux-Soak-goroutine","multi_line_start_pattern":"{timestamp_format}","timestamp_format":"%d %b %y %H:%M:%S","timezone":"UTC"},{"file_path":"/tmp/soakTest/tmp0.log","log_group_name":"LinuxSoaking0.log","multi_line_start_pattern":"{timestamp_format}","timestamp_format":"%d %b %y %H:%M:%S","timezone":"UTC"},{"file_path":"/tmp/soakTest/tmp1.log","log_group_name":"LinuxSoaking1.log","multi_line_start_pattern":"{timestamp_format}","timestamp_format":"%d %b %y %H:%M:%S","timezone":"UTC"},{"file_path":"/tmp/soakTest/tmp2.log","log_group_name":"LinuxSoaking2.log","multi_line_start_pattern":"{timestamp_format}","timestamp_format":"%d %b %y %H:%M:%S","timezone":"UTC"},{"file_path":"/tmp/soakTest/tmp3.log","log_group_name":"LinuxSoaking3.log","multi_line_start_pattern":"{timestamp_format}","timestamp_format":"%d %b %y %H:%M:%S","timezone":"UTC"},{"file_path":"/tmp/soakTest/tmp4.log","log_group_name":"LinuxSoaking4.log","multi_line_start_pattern":"{timestamp_format}","timestamp_format":"%d %b %y %H:%M:%S","timezone":"UTC"},{"file_path":"/tmp/soakTest/tmp5.log","log_group_name":"LinuxSoaking5.log","multi_line_start_pattern":"{timestamp_format}","timestamp_format":"%d %b %y %H:%M:%S","timezone":"UTC"},{"file_path":"/tmp/soakTest/tmp6.log","log_group_name":"LinuxSoaking6.log","multi_line_start_pattern":"{timestamp_format}","timestamp_format":"%d %b %y %H:%M:%S","timezone":"UTC"},{"file_path":"/tmp/soakTest/tmp7.log","log_group_name":"LinuxSoaking7.log","multi_line_start_pattern":"{timestamp_format}","timestamp_format":"%d %b %y %H:%M:%S","timezone":"UTC"},{"file_path":"/tmp/soakTest/tmp8.log","log_group_name":"LinuxSoaking8.log","multi_line_start_pattern":"{timestamp_format}","timestamp_format":"%d %b %y %H:%M:%S","timezone":"UTC"},{"file_path":"/tmp/soakTest/tmp9.log","log_group_name":"LinuxSoaking9.log","multi_line_start_pattern":"{timestamp_format}","timestamp_format":"%d %b %y %H:%M:%S","timezone":"UTC"},{"file_path":"/tmp/soakTest/structuredLogFile0.json","log_group_name":"linuxStructuredLogSoakTst0","multi_line_start_pattern":"^{"},{"file_path":"/tmp/soakTest/structuredLogFile1.json","log_group_name":"linuxStructuredLogSoakTst1","multi_line_start_pattern":"^{"}]}}},"csm":{"memory_limit_in_mb":20,"port":31000}}
```

# Requirements
2. Run `make linter`

# Performance benchmarking
1. **Before(v353) and after adding tailer semaphore ( numUsedFds) with monitoring 15 log files** (not sure why my version is more lighter)
image

2. **Comparing between Semaphore**
```
BenchmarkSemaphore_AcquireReleaseUnderLimitSimple-12          19336089                62.26 ns/op            0 B/op          0 allocs/op
BenchmarkSemaphore_AcquireReleaseUnderLimit-12                    1652306               726.5 ns/op             0 B/op          0 allocs/op
BenchmarkSemaphore_AcquireReleaseOverLimit-12                     32737             37202 ns/op               0 B/op          0 allocs/op
BenchmarkMarusamaSemaphore_AcquireReleaseUnderLimitSimple-12                         6038138               171.7 ns/op            96 B/op          1 allocs/op
BenchmarkMarusamaSemaphore_AcquireReleaseUnderLimit-12                                 479114              2268 ns/op             960 B/op         10 allocs/op
BenchmarkMarusamaSemaphore_AcquireReleaseOverLimit-12                                   50464             24728 ns/op            9600 B/op        100 allocs/op
BenchmarkAbiosoftSemaphore_AcquireReleaseUnderLimitSimple-12                 8135151               145.0 ns/op             0 B/op          0 allocs/op
BenchmarkAbiosoftSemaphore_AcquireReleaseUnderLimit-12                          432472              2731 ns/op               0 B/op          0 allocs/op
BenchmarkAbiosoftSemaphore_AcquireReleaseOverLimit-12                           51364             27126 ns/op               0 B/op          0 allocs/op
BenchmarkPivotalGolangSemaphore_AcquireReleaseUnderLimitSimple-12            3358042               355.3 ns/op           128 B/op          2 allocs/op
BenchmarkPivotalGolangSemaphore_AcquireReleaseUnderLimit-12                      149810              7881 ns/op            1200 B/op         20 allocs/op
BenchmarkPivotalGolangSemaphore_AcquireReleaseOverLimit-12                      14558             83574 ns/op           12000 B/op        200 allocs/op
BenchmarkXSyncSemaphore_AcquireReleaseUnderLimitSimple-12                   40687284                27.22 ns/op            0 B/op          0 allocs/op
BenchmarkXSyncSemaphore_AcquireReleaseUnderLimit-12                             1000000              1060 ns/op               0 B/op          0 allocs/op
BenchmarkXSyncSemaphore_AcquireReleaseOverLimit-12                              17275             69725 ns/op           15994 B/op        299 allocs/op

```


plugins/inputs/logfile/tail/tail_queue.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+/*
+	Centralize all the tailers which are scheduled for each file
+*/
+
+var defaultOpenFileLimit int = 1024
+
+type TailerEnqueue struct {
+	Queue chan bool
```
SaxyPandaBear, line 17: I strongly disagree with this. Why are we storing a boolean here? Do we need to check the value? I thought this was just a counter. 

Ultimately what you are implementing is a semaphore with a buffered channel. If all we want is a counting semaphore, we could just do `chan struct{}`. Otherwise we're exploding memory usage.

SaxyPandaBear (changes requested):

plugins/inputs/logfile/fdlimit/fdlimit_windows.go
```diff
@@ -0,0 +1,15 @@
...
+// SPDX-License-Identifier: MIT
+
+//go:build windows
+// +build windows
+
+package fdlimit
+
+//On Windows, Go uses the CreateFile API, which is limited to 16K files; therefore, non-changeable from within a running process
+//For more information on the API,please follow https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilea
+const hardLimitFileDescriptor = 16384
```
SaxyPandaBear, line 11: I skimmed the docs and I don't see where this value comes from. The value `16384` comes from `FILE_ATTRIBUTE_ENCRYPTED`

plugins/inputs/logfile/fdlimit/fdlimit_test.go
```diff
@@ -0,0 +1,21 @@
...
+
+	"github.com/aws/amazon-cloudwatch-agent/translator/config"
+)
+
+func TestFileDescriptorsLimit(t *testing.T) {
+	currentOpenFileLimit, err := CurrentOpenFileLimit()
+	assert.NoError(t, err)
+
+	if runtime.GOOS == config.OS_TYPE_WINDOWS {
+		assert.Equal(t, 16384, currentOpenFileLimit)
```
SaxyPandaBear, line 19: What happens if this changes on a newer version of Windows?

plugins/inputs/logfile/tail/tail.go
```diff
@@ -329,12 +332,14 @@ func (tail *Tail) tailFileSync() {
 		tail.Logger.Debugf("Seeked %s - %+v\n", tail.Filename, tail.Location)
 		if err != nil {
 			tail.Killf("Seek error on %s: %s", tail.Filename, err)
+			tail.TailerQueue.Dequeue()
```
SaxyPandaBear, line 335: There are a lot of places where this `Dequeue()` call gets added. I think it makes more sense to trace through and identify where all of these different code paths converge and then only add one place where you dequeue. Also back to my other issue, you're not actually using any value from the queue (whether it is true or false). This is really just an atomic counter / counting semaphore, so if we're going to include this in some capacity, I would expect the naming to change.

plugins/inputs/logfile/tailersrc_test.go
```diff
@@ -213,56 +201,30 @@ func TestOffsetDoneCallBack(t *testing.T) {
...
-			if err != nil {
-				t.Errorf("Failed to read state file: %v", err)
-			}
+			assert.NoError(t, err)
 			file_parts := bytes.Split(b, []byte("\n"))
-			log.Println("file_parts: ", file_parts)
 			file_string := string(file_parts[0])
-			log.Println("file_string: ", file_string)
 			offset, err := strconv.Atoi(file_string)
-			log.Println(offset)
```
SaxyPandaBear, line 242: Normally, I wouldn't want these log statements. But I do for tailersrc testing. What is the reason for removing these?

plugins/inputs/logfile/logfile.go
```diff
@@ -259,6 +266,10 @@ func (t *LogFile) FindLogSrc() []logs.LogSrc {
 		}
 	}
 
+	if allowedOpenFileLimit, err := fdlimit.CurrentOpenFileLimit(); err == nil {
+		t.Log.Infof("Number of current monitored log files / Allowed monitored log files: %v/%v", t.tailerQueue.Size(), allowedOpenFileLimit)
```
SaxyPandaBear, line 270: Realistically, we aren't tracking the total "allowable" number of log files to monitor. If they run out of file descriptors, we can tail the file, maybe, but can't do anything with it. 

I think this could be reworded:
```suggestion
		t.Log.Infof("Number of fds used by agent / Total allotted fds: %v/%v", t.tailerQueue.Size(), allowedOpenFileLimit)
```

plugins/inputs/logfile/tail/tail_queue.go
SaxyPandaBear: Or instead of using a channel, we could just use `atomic` operations on a global count. But I would probably just use a chan struct{}

plugins/inputs/logfile/tail/tail_queue.go
```diff
@@ -0,0 +1,43 @@
...
+}
+
+func NewTailerFifoQueue() *TailerEnqueue {
+	if allowedOpenFileLimit, err := fdlimit.CurrentOpenFileLimit(); err == nil {
+		defaultOpenFileLimit = allowedOpenFileLimit
```
SaxyPandaBear, line 22: Declaring a "default" value and then updating the default value is bad code smell. What is the purpose of having this default to begin with? If there is a failure with deriving the allowed file descriptors for a given host, we're just going to assume it is 1024? What is our experience debugging issues like this going to be?

plugins/inputs/logfile/logfile.go
```diff
@@ -259,6 +266,10 @@ func (t *LogFile) FindLogSrc() []logs.LogSrc {
 		}
 	}
 
+	if allowedOpenFileLimit, err := fdlimit.CurrentOpenFileLimit(); err == nil {
```
SaxyPandaBear, line 269: We should not keep looking up this file descriptor limit. We can get this information from the buffered channel and only look up the file descriptor limit once. 

`len(ch)` to see how many things are currently in the channel, and `cap(ch)` to see the allowed capacity for the channel.

plugins/inputs/logfile/tail/tail_queue.go
khanhntd: > Why are we storing a boolean here?

In future, TailerEnqueue could  change to store `tailer` instead of `struct`. Therefore, it would be best to prepare the behavior changes  for the customer `now`. However, if we only limit the scope to only log number of `monitored files`, then I agree with you regarding using `struct`.

plugins/inputs/logfile/tail/tail_queue.go
```diff
@@ -0,0 +1,43 @@
...
+	t.Queue <- true
+}
+
+func (t *TailerEnqueue) Dequeue() bool {
+	select {
+	case v := <-t.Queue:
+		return v
+	default:
+		return false
+	}
```
SaxyPandaBear, line 38: Does this account for the scenario where the channel is empty we try to dequeue?

SaxyPandaBear (commented):
I was hoping that this change would be less intrusive. I see that you've added output from manually testing functionality, but I don't see any data on performance. Personally, I'd like to see some metrics on CPU/mem monitoring different magnitudes of files, i.e.: monitor 1, 100, 1000 log files

plugins/inputs/logfile/tail/tail_queue.go
SaxyPandaBear: Did we discuss storing a duplicate tailer object in a channel before? I disagree. The entire purpose of this is counting open monitored files. Why do we need more information than a count?

khanhntd: 👍  Yes, maybe I'm worrying about too much about things out of scope. Will change back in the latest commit.

plugins/inputs/logfile/fdlimit/fdlimit_windows.go
khanhntd: The documents are for the API. For the discussion regarding the hard limit, my SQL has encountered the same issue https://bugs.mysql.com/bug.php?id=24509 and decided to go with 16384. Another example could be found with [go-ethereum](https://github.com/ethereum/go-ethereum/blob/8a134014b4b370b4a3632e32a2fc8e84ee2b6947/common/fdlimit/fdlimit_windows.go)

plugins/inputs/logfile/fdlimit/fdlimit_test.go
khanhntd: `16384` has been a hard limit from 2006 (based on MySQL issues, maybe before Windows 7) and has been increased from `2048`. Therefore, I expect Windows do not decrease the file descriptors and if it is increased, it does not bring any impacts to CloudWatchAgent and as a customer, `16384` is not feasible to reach and if it is reachable , I will worry about their memory more than the CloudWatchAgent

plugins/inputs/logfile/tailersrc_test.go
khanhntd: No deep reason against it since I don't see any value of keeping it.

plugins/inputs/logfile/logfile.go
khanhntd: Agree. I have not known in go channels we could use `cap(ch)`. First time dealing with `go channel`

plugins/inputs/logfile/tail/tail_queue.go
khanhntd: The purpose of having the default would be avoid `interrupting the onboarding CloudWatchAgent for customers`. I'm indecisive about these two: would we `log.panic` if we cannot get file descriptors due to OS issue or should we allow the use the 1024 (the general upper bound for CWAgent which is used by customers). I'm still indecisive regarding about that so open to suggestion here.

plugins/inputs/logfile/tail/tail_queue.go
khanhntd: Yes. Default `return false` always ensure even if the channel does not receive any message, the process still return false (since we don't use the value as you mentioned before) and will be garbage collected in some ways. Its the same as
```
if len(enqueue): return false
return true
```

khanhntd:
> I was hoping that this change would be less intrusive. I see that you've added output from manually testing functionality, but I don't see any data on performance. Personally, I'd like to see some metrics on CPU/mem monitoring different magnitudes of files, i.e.: monitor 1, 100, 1000 log files

Agree with you on this point since the customer only expects the CWAgent to be less intrusive and less cpu/memory consumption in the future version. However, in OH, Adam said let's focus with performance metrics later on so I have not done any progressing on these so far.

plugins/inputs/logfile/tail/tail.go
khanhntd: Yes. That's true. I have skimmed/trace through; however, I have not able to identify the `one place` to dequeue, that's why I have to spread out and deliver a solving solution and keep up the deadline.

khanhntd: Resolve the issue by adding `tail.Deuque()` with defer (forgot about that): https://github.com/aws/amazon-cloudwatch-agent/pull/491/files#diff-b4cbda30ad21153305e4658dca8a5844580e52b12544b5147237a3315d185284R314

plugins/inputs/logfile/tail/tail.go
```diff
@@ -309,6 +311,7 @@ func (tail *Tail) readlineUtf16() (string, error) {
 func (tail *Tail) tailFileSync() {
 	defer tail.Done()
 	defer tail.close()
+	defer tail.TailerQueue.Dequeue()
```
SaxyPandaBear, line 314: Not sure why we need to include another `defer` statement here. If dequeuing from this structure is part of cleanup, it should just be put in the `Done` or `close` function rather than adding an additional defer here.

plugins/inputs/logfile/fdlimit/fdlimit_windows.go
```diff
@@ -0,0 +1,16 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+//go:build windows
+// +build windows
+
+package fdlimit
+
+// On Windows, Go uses the CreateFile API, which is limited to 16K files; therefore, non-changeable from within a running process
+// MySQL has encountered the same issue https://bugs.mysql.com/bug.php?id=24509
```
SaxyPandaBear, line 10: Reading through that link, and looking around online, that is 16384 handles for network communication as a default on Windows machines, which can be changed. Not necessarily file handlers that can be open at once. 
Screenshot 2022-06-16 230800 

I have over 80k file handles open (for some reason)

Running `net config server` on my desktop shows:
```
Maximum open files per session        16384
```

I am fine with leaving it at 16k, but the reasoning was not concrete enough.

plugins/inputs/logfile/logfile.go
```diff
@@ -35,14 +35,17 @@ type LogFile struct {
...
 	started           bool
 }
 
 func NewLogFile() *LogFile {
+
 	return &LogFile{
 		configs:           make(map[*FileConfig]map[string]*tailerSrc),
 		done:              make(chan struct{}),
 		removeTailerSrcCh: make(chan *tailerSrc, 100),
+		tailerQueue:       tail.NewTailerFifoQueue(),
```
SaxyPandaBear, line 48: I thought the point was to keep track of all of the monitored log files. Why would we have a separate counter per log file configuration? If I configure to monitor 10 different paths with wildcards in them, this would not communicate across the other routines. It would be misrepresenting the data because if the fd limit is 100 and I have configured 10 different log file monitors, where each path picks up 10 different files, then there are going to be 100 handles opened, but the logging here would only indicate 10 are being used in the log file.

plugins/inputs/logfile/logfile.go
```diff
@@ -259,6 +265,10 @@ func (t *LogFile) FindLogSrc() []logs.LogSrc {
 		}
 	}
 
+	
+	t.Log.Infof("Number of file descriptors used by agent / Total allowed file descriptors: %v/%v", t.tailerQueue.Size(), t.tailerQueue.Capacity())
```
SaxyPandaBear, line 269: How many times is FindLogSrc() called? Do we want this at info level?

plugins/inputs/logfile/logfile.go
khanhntd: Based on my understanding, it's not a separate per log file configuration so the answer is no. In your case, it can be explained by looking through [this code](https://github.com/aws/amazon-cloudwatch-agent/blob/master/plugins/inputs/logfile/logfile.go#L153-L161)
```
for i := range t.FileConfig { // Loop through all in FileConfig
		fileconfig := &t.FileConfig[i]

		targetFiles, err := t.getTargetFiles(fileconfig) // Get all the file if specifies wildcard( e.g log_file_1,log_file_2) 
		if err != nil {
			t.Log.Errorf("Failed to find target files for file config %v, with error: %v", fileconfig.FilePath, err)
		}

		for _, filename := range targetFiles { // Put the tailer
```
However, `tailerQueue:       tail.NewTailerFifoQueue(),` which is in the [init function](https://github.com/aws/amazon-cloudwatch-agent/blob/master/plugins/inputs/logfile/logfile.go#L461-L465) which is created when CloudWatchAgent applies the plugin. Long short story, it is created before `FindLogSrc` and is not created between `FileConfig`

Let me know if that make sense

plugins/inputs/logfile/logfile.go
khanhntd: > How many times is FindLogSrc() called?

[1 per second](https://github.com/aws/amazon-cloudwatch-agent/blob/master/logs/logs.go#L93).

>  Do we want this at info level?

Yes. The reason may be in the design doc.
```
Requires the customer to monitor the files on a scheduled interval and can result in missing metrics/logs if not able to handle on time (expected data to be loss in an average of after 30 days)
```
Therefore, would it be good for the customers to detect the issues on time without relying on to restart the agent `with debug flag` (warning is the same level as info). From customer perspective again, would this be corrupt their logs ? The answer is no. I have thought to create another `Ticker`; however, I have thought of create another `time.Ticker` at 1 minute interval but not confidence in it since between log every second and fires a 1 minute interval ticker every second, I would choose the former in terms of memory consumption for customer

plugins/inputs/logfile/tail/tail_queue.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"log"
+	"github.com/aws/amazon-cloudwatch-agent/plugins/inputs/logfile/fdlimit"
+)
+
+/*
+	Centralize all the tailers which are scheduled for each file
+*/
+
+type TailerEnqueue struct {
```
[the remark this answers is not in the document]
khanhntd: FYI, there is another [NewBlockingFifoQueue](https://github.com/aws/amazon-cloudwatch-agent/blob/master/internal/publisher/blockingfifoqueue.go) in `./internal/publisher` which mostly shares the same functionality; however, we can refine these two in another `reformat PR` or now. I'm open to suggestion here.

plugins/inputs/logfile/tail/tail_queue.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	return &TailerEnqueue{Queue: make(chan struct{}, allowedOpenFileLimit)}
+}
+
+func (t *TailerEnqueue) Enqueue() {
+	t.Queue <- struct{}{}
+}
+
+func (t *TailerEnqueue) Dequeue() struct{} {
```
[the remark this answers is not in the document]
khanhntd: Instead of 
```
if len(enqueue) == 0 {
  log.format("Not enough to deque")
  return  
}
return 
```
, I would prefer the below for better format (maybe a better signal and confirm the test as you said https://github.com/aws/amazon-cloudwatch-agent/pull/491#discussion_r899722195)
```
func (t *TailerEnqueue) Dequeue() bool {
	select {
	case <-t.Queue:
		return true
	default:
		return false
	}
}
```

plugins/inputs/logfile/tailersrc_test.go
khanhntd: Not entirely not relate. The `tailSrc` also uses the `tailFile`, so its kinda connivence for me to change this to.

plugins/inputs/logfile/tail/tail_queue.go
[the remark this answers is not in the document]
khanhntd: Sounds to me we already have [it](https://github.com/aws/amazon-cloudwatch-agent/blob/master/internal/publisher/nonblockingfifoqueue.go) lol. However, good call out regarding `Semaphore` and a nice thing to learn

plugins/inputs/logfile/logfile.go
SaxyPandaBear: > Requires the customer to monitor the files on a scheduled interval and can result in missing metrics/logs if not able to handle on time (expected data to be loss in an average of after 30 days)

I don't understand this. Are we telling the customer to monitor something in the CloudWatch agent log file? I thought that the point of this logging was for helping us diagnose. 

---
What does this have to do with memory consumption? I'm asking about if we're willing to accept emitting a log line every second into a file that we do not manage log rotation on. The agent log file can grow unbound, out of the box.

plugins/inputs/logfile/fdlimit/fdlimit_windows.go
```diff
@@ -0,0 +1,16 @@
...
+// An example of how go-ethereum handle file descriptors https://github.com/ethereum/go-ethereum/blob/8a134014b4b370b4a3632e32a2fc8e84ee2b6947/common/fdlimit/fdlimit_windows.go
```
SaxyPandaBear, line 11: I don't think that any of the links provided actually point to why this is set to 16384 for log file handles that we count. I think all you need to say is:
```suggestion
// The default number of allowed file handles for network on Windows is 16384, so aligning with that
```

plugins/inputs/logfile/tail/tail_queue.go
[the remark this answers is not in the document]
SaxyPandaBear: That FIFO queue implementation wastes memory as it accepts an `interface{}` and is not a viable replacement. If you are saying you agree, can you please generalize this how I described above and update the PR

plugins/inputs/logfile/tail/tail_queue_test.go
```diff
@@ -0,0 +1,27 @@
...
+	"testing"
+)
+
+func TestTailerEnqueueAndDequeue(t *testing.T) {
+	tailerEnqueue := NewTailerFifoQueue()
+	tailerEnqueue.Enqueue()
+	tailerEnqueue.Enqueue()
+	assert.Equal(t, 2, tailerEnqueue.Size())
+
+	ok := tailerEnqueue.Dequeue()
```
SaxyPandaBear, line 17: Again, I strongly disagree with this because this is misusing a queue. We don't care what the value is. We care about the size and capacity of the counter.

plugins/inputs/logfile/tail/tail_queue_test.go
```diff
@@ -0,0 +1,27 @@
...
+	assert.Equal(t, 1, tailerEnqueue.Size())
+	assert.Equal(t, true, ok)
```
SaxyPandaBear, line 19: If you don't intend to change the interface for your counter, please simplify this to 
```suggestion
	assert.True(t, ok)
```

plugins/inputs/logfile/tail/tail_queue_test.go
khanhntd: Sorry I'm still in the middle of transitioning and have not planned to touch this in the morning.

plugins/inputs/logfile/tail/tail_queue.go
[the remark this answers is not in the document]
khanhntd: 👍 Yes. I planned to do that after updating the design docs. So will do it in the afternoon if that's good @SaxyPandaBear

plugins/inputs/logfile/logfile.go
khanhntd: > I thought that the point of this logging was for helping us diagnose.

No I believe this is for us and customer. That's why I put it in the `INFO` level instead of `DEBUG` level

>  I'm asking about if we're willing to accept emitting a log line every second into a file that we do not manage log rotation on

Long short answer, yes. However, if you are worrying about `log line every second`, I can manage to using `time.Ticker` to log every 1 minute.

plugins/inputs/logfile/tail/tail_queue_test.go
khanhntd: Sorry for this one too. I'm still in the middle of transitioning and have not planned to touch this in the morning.

logs/logs.go
```diff
@@ -16,7 +16,7 @@ var ErrOutputStopped = errors.New("Output plugin stopped")
 
 // A LogCollection is a collection of LogSrc, a plugin which can provide many LogSrc
 type LogCollection interface {
-	FindLogSrc() []LogSrc
+	FindLogSrc(ctx context.Context) []LogSrc
```
SaxyPandaBear, line 19: We should not be modifying this. I do not see a good reason to break this interface by adding a context parameter. Please revert this.

SaxyPandaBear (changes requested):

plugins/inputs/logfile/logfile.go
```diff
@@ -205,6 +216,11 @@ func (t *LogFile) FindLogSrc() []logs.LogSrc {
 				continue
 			}
 
+			if err := t.tailerSemaphore.Acquire(ctx, 1); err != nil {
+				t.Log.Errorf("Failed increase counter in semaphore:", err)
+				continue
+			}
```
SaxyPandaBear, line 222: So this is not the original intended behavior. The original purpose of the PR was just to monitor and count the number of log files being monitored. Now, this introduces a leak, because `tail.TailFile()` has already been called so the file is going to start being monitored, but not propagated to the list of LogSrcs.

internal/semaphore/semaphore.go
```diff
@@ -0,0 +1,164 @@
+// Copyright 2017 Maru Sama. All rights reserved.
```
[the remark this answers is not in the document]
khanhntd: Sorry @SaxyPandaBear. I'm still doing the benchmark between semaphore and only pushed it for committing a snapshot. Have not done the final PR yet.

internal/semaphore/semaphore_test.go
```diff
@@ -0,0 +1,315 @@
+// Copyright 2017 Maru Sama. All rights reserved.
+// Use of this source code is governed by the MIT license
+// that can be found in the LICENSE file.
```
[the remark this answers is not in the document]
khanhntd: Same previous comment.
```
I'm still doing the benchmark between semaphore and only pushed it for committing a snapshot. Have not done the final PR yet.
```

plugins/inputs/logfile/logfile.go
khanhntd: Same previous comment.
```
I'm still doing the benchmark between semaphore and only pushed it for committing a snapshot. Have not done the final PR yet.
```

logs/logs.go
khanhntd: Same with below  comment.
```
I'm still doing the benchmark between semaphore and only pushed it for committing a snapshot. Have not done the final PR yet.
```

plugins/inputs/logfile/logfile.go
```diff
@@ -35,14 +37,22 @@ type LogFile struct {
...
+	tailerSemaphore   *semaphore.Semaphore
 	started           bool
 }
 
 func NewLogFile() *LogFile {
+	fileDescriptorLimit, err := fdlimit.CurrentFileDescriptorLimit()
+
+	if err != nil {
+		log.Panicf("E! Failed to get file descriptor limit:", err)
+	}
```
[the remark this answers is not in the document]
khanhntd: No. As you have said, the file descriptors are only useful for us instead of customers. Therefore, this feature should be non-intrusive to customer side by utilizing 
```
return &LogFile{
			numUsedFds:   	   semaphore.NewSemaphore(0),
}
```

and 

```
if t.numUsedFds.GetLimit != 0 {
		t.Log.Debugf("Number of file descriptors used by agent / Total allowed file descriptors: %v/%v", t.numUsedFds.GetCount(), t.numUsedFds.GetLimit())
}
```

plugins/inputs/logfile/logfile.go
```diff
@@ -35,14 +37,22 @@ type LogFile struct {
 	configs           map[*FileConfig]map[string]*tailerSrc
 	done              chan struct{}
 	removeTailerSrcCh chan *tailerSrc
+	tailerSemaphore   *semaphore.Semaphore
```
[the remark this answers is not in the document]
khanhntd: Agree. It's still in the middle of transitioning so have not thought deep enough for variable name. However, thanks for the head up and thoughts.

plugins/inputs/logfile/tail/tail_test.go
```diff
@@ -99,66 +98,65 @@ func TestStopAtEOF(t *testing.T) {
...
+		Config{
+			Logger: &tl,
+			ReOpen: false,
+			Follow: true,
+		})
+	assert.NoError(t, err)
+
+	//Increase the tailer in queue by 1 to later confirmed if the tailer in queue has been released yet in line 181
+	tail.numUsedFds.Acquire()
+	assert.Equal(t, 1, tail.numUsedFds.GetCount())
```
SaxyPandaBear, line 150: This is not useful for an assertion. This just duplicates any testing on the `Acquire` function that you already have

plugins/inputs/logfile/tailersrc_test.go
```diff
@@ -59,11 +58,12 @@ func TestTailerSrc(t *testing.T) {
...
 
-	if err != nil {
-		t.Errorf("Failed to create tailer src for file %v with error: %v", file, err)
-		return
-	}
+	assert.NoError(t, err)
 
+	//Increase number of monitored files by 1
+	numUsedFds.Acquire()
+	assert.Equal(t, 1, numUsedFds.GetCount())
```
SaxyPandaBear, line 65: Not useful assertion

plugins/inputs/logfile/tailersrc_test.go
```diff
@@ -145,29 +145,30 @@ func TestTailerSrc(t *testing.T) {
...
-		t.Errorf("failed to remove log file '%v': %v", file.Name(), err)
-	}
+	err = os.Remove(file.Name())
+	assert.NoError(t, err)
+	
+	time.Sleep(time.Second)
+	assert.Equal(t, 0, numUsedFds.GetCount())
+	
 	<-done
+	numUsedFds.Done()
```
SaxyPandaBear, line 155: This means we never call `Done` if an exception occurs earlier. I'd prefer putting this call at the beginning in a defer

plugins/inputs/logfile/tailersrc_test.go
```diff
@@ -179,10 +180,11 @@ func TestOffsetDoneCallBack(t *testing.T) {
...
 
-	if err != nil {
-		t.Errorf("Failed to create tailer src for file %v with error: %v", file, err)
-		return
-	}
+	assert.NoError(t, err)
+
+	//Increase number of monitored files by 1
+	numUsedFds.Acquire()
+	assert.Equal(t, 1, numUsedFds.GetCount())
```
SaxyPandaBear, line 187: Not useful assertion

plugins/inputs/logfile/fdlimit/fdlimit_windows.go
khanhntd: > Reading through that link, and looking around online, that is 16384 handles for network communication as a default on Windows machines, which can be changed. Not necessarily file handlers that can be open at once.
> 
> I have over 80k file handles open (for some reason)
> 
> Running net config server on my desktop shows:
> 
> Maximum open files per session        16384
> I am fine with leaving it at 16k, but the reasoning was not concrete enough.

Based on my understanding, the 80k file handles has been distributed amongst 214 processors since the default number per processors is 16384. However, as you said, it can be changed but it won't be an **easy task**. Unfortunately, `golang` has not supported `_getmaxstdio` so yes, I'm hesitated to implement one. Therefore, the 16384 is meant for align the `maximum open files per process/session`.

plugins/inputs/logfile/tailersrc_test.go
```diff
@@ -179,10 +180,11 @@ func TestOffsetDoneCallBack(t *testing.T) {
...
 		})
 
-	if err != nil {
-		t.Errorf("Failed to create tailer src for file %v with error: %v", file, err)
-		return
-	}
+	assert.NoError(t, err)
+
+	//Increase number of monitored files by 1
+	numUsedFds.Acquire()
```
SaxyPandaBear, line 186: Do we need to test this fd counter in every test? 

If we do, then we need to remove this code duplication. Either we should pull out the entire part where we create the tailer, `tailer, err := tail.TailFile()` into a separate test function, where we instantiate the semaphore and increment it by one there, or just have a separate function that creates the semaphore where it has already acquired once and call that. There's not much reason to copy paste this `Acquire` call and accompanying comment into every test

plugins/inputs/logfile/tailersrc_test.go
SaxyPandaBear: @adam-mateen would want your opinion on if you want these log statements in here or not.

plugins/inputs/logfile/tailersrc_test.go
```diff
@@ -304,13 +280,14 @@ func TestOffsetDoneCallBack(t *testing.T) {
...
+	assert.NoError(t, err)
+	time.Sleep(time.Second)
+	assert.Equal(t, 0, numUsedFds.GetCount())
+
 	<-done
-	if i < 35 {
-		t.Errorf("Not enough logs have been processed, only %v are processed", i)
-	}
+	assert.GreaterOrEqualf(t, i, 35, "Not enough logs have been processed, only %v are processed", i)
+	numUsedFds.Done()
```
SaxyPandaBear, line 290: Do we need to introduce all of these changes?

plugins/inputs/logfile/tailersrc_test.go
```diff
@@ -325,9 +302,11 @@ func TestTailerSrcFiltersSingleLineLogs(t *testing.T) {
 	publishLogsToFile(resources.file, matchedLog, unmatchedLog, n, 0)
 
 	// Removal of log file should stop tailersrc
-	if err := os.Remove(resources.file.Name()); err != nil {
-		t.Errorf("failed to remove log file '%v': %v", resources.file.Name(), err)
-	}
+	err := os.Remove(resources.file.Name())
+	assert.NoError(t, err)
+	time.Sleep(time.Second)
```
SaxyPandaBear, line 307: Do we need to add a sleep here for a whole second?

plugins/inputs/logfile/tailersrc_test.go
```diff
@@ -357,9 +336,11 @@ func TestTailerSrcFiltersMultiLineLogs(t *testing.T) {
 	publishLogsToFile(resources.file, matchedLog, unmatchedLog, n, 100)
 
 	// Removal of log file should stop tailersrc
-	if err := os.Remove(resources.file.Name()); err != nil {
-		t.Errorf("failed to remove log file '%v': %v", resources.file.Name(), err)
-	}
+	err := os.Remove(resources.file.Name())
+	assert.NoError(t, err)
+	time.Sleep(time.Second)
```
SaxyPandaBear, line 341: Do we need this

internal/semaphore/semaphore.go
```diff
@@ -0,0 +1,70 @@
...
+		log.Panic("E! Semaphore size limit must not be negative")
+	}
+
+	sem := &semaphore{
+		slots: make(chan struct{}, limit),
+	}
+
+	for i := 0; i < cap(sem.slots); i++ {
+		sem.slots <- struct{}{}
+	}
```
SaxyPandaBear, line 42: I was mostly fine with what it was before where you started with an empty channel and pushed to it as files got monitored. Why did you invert it so that you allocate all of the "slots" ahead of time?

internal/semaphore/semaphore.go
```diff
@@ -0,0 +1,70 @@
...
+func (sem *semaphore) Acquire() {
+	<-sem.slots
+}
+
+// Release the acquired semaphore.
+func (sem *semaphore) Release() {
+	select {
+	case sem.slots <- struct{}{}:
+	default:
+		log.Panic("E! Semaphore released more than held")
```
SaxyPandaBear, line 56: Why are we panicking here? I thought this was supposed to be for debugging. I do not want this change to break the agent just for debugging.

internal/semaphore/semaphore.go
```diff
@@ -0,0 +1,70 @@
...
+	for i := 0; i < cap(sem.slots); i++ {
+		sem.slots <- struct{}{}
+	}
+	
+	return sem
+}
+
+func (sem *semaphore) Acquire() {
+	<-sem.slots
+}
```
SaxyPandaBear, line 49: So if the channel is empty, this blocks until something is added to the channel, which makes sense as a semaphore implementation, but consider how we're using it here. This still aligns with what I said before where we are changing this from being debug logging into becoming a limiter for log files that can be monitored. Is that what you intended for this change?

internal/semaphore/semaphore.go
```diff
@@ -0,0 +1,70 @@
...
+
+	//GetLimit returns semaphore limit.
+	GetLimit() int
+
+	//GetCount returns current number of occupied slots in semaphore.
+	GetCount() int
+
+	//Done closes down the semaphore
+	Done()
+}
```
SaxyPandaBear, line 25: Do we intend to reuse this interface elsewhere with a different underlying implementation? Do we need an interface here?

internal/semaphore/semaphore_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+func checkLimit(t *testing.T, sem Semaphore, expected int) {
+	limit := sem.GetLimit()
+	assert.Equal(t, expected, limit)
+}
+
+func checkCount(t *testing.T, sem Semaphore, expected int) {
+	count := sem.GetCount()
+	assert.Equal(t, expected, count)
+}
```
SaxyPandaBear, line 20: Neither of these functions get used outside of the `checkLimitAndCount` function. They are not useful to separate.

internal/semaphore/semaphore_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+	sem.Release()
+	checkLimitAndCount(t, sem, 2, 1)
+
+	sem.Release()
+	checkLimitAndCount(t, sem, 2, 0)
+}
+
+func TestSemaphore_ReleaseMoreThanAcquire(t *testing.T) {
+	sem := NewSemaphore(1)
```
SaxyPandaBear, line 49: One of the ways that you are using the semaphore is creating one with a zero limit, `NewSemaphore(0)`. Do we need test cases for that?

internal/semaphore/semaphore_test.go
```diff
@@ -0,0 +1,93 @@
...
+			sem.Release()
+			wg.Done()
+		}()
+	}
+
+	close(c) // start
+	wg.Wait()
+
+	checkLimitAndCount(t, sem, 10, 0)
+}
```
SaxyPandaBear, line 93: Not a blocker, but these are the same test with different parameters. This could be consolidated to a single test that has a set of test parameters that get executed with sub-tests via `t.Run()`

plugins/inputs/logfile/logfile.go
```diff
@@ -35,14 +37,27 @@ type LogFile struct {
...
+
+	if err != nil {
+		return &LogFile{
+			configs:           make(map[*FileConfig]map[string]*tailerSrc),
+			done:              make(chan struct{}),
+			removeTailerSrcCh: make(chan *tailerSrc, 100),
+			numUsedFds:        semaphore.NewSemaphore(0),
+		}
+	}
+
```
SaxyPandaBear, line 55: Why do we need to catch the error here? We aren't doing anything with the error and just use a default 0 value. This is equivalent to just removing the whole `if` block because the default value returned when the file descriptor check on Unix fails is zero.

internal/semaphore/semaphore_test.go
khanhntd: Yes. Thought about `table driven test`; however, not sure if the `table driven test` are independent to each other when its come to go channel.

internal/semaphore/semaphore.go
khanhntd: There is a slightly different regarding performance between these twos when scaling up to passed 1000 tailers. Here is an benchmark testing 
```
BenchmarkNewSemaphore_Acquire_ReleaseUnderLimitSimple-12          19538026                60.56 ns/op            0 B/op          0 allocs/op
BenchmarkNewSemaphore_Acquire_ReleaseUnderLimit-12                     8692            124471 ns/op               0 B/op          0 allocs/op
BenchmarkNewSemaphore_Acquire_ReleaseOverLimit-12                     32360             37397 ns/op               0 B/op          0 allocs/op
BenchmarkOldSemaphore_Acquire_ReleaseUnderLimitSimple-12        28253834                41.13 ns/op            0 B/op          0 allocs/op
BenchmarkOldSemaphore_Acquire_ReleaseUnderLimit-12                   9055            125437 ns/op               0 B/op          0 allocs/op
BenchmarkOldSemaphore_Acquire_ReleaseOverLimit-12                   31078             38852 ns/op               0 B/op          0 allocs/op
```

plugins/inputs/logfile/logfile.go
```diff
@@ -35,14 +36,27 @@ type LogFile struct {
...
+	if err != nil {
+		return &LogFile{
+			configs:           make(map[*FileConfig]map[string]*tailerSrc),
+			done:              make(chan struct{}),
+			removeTailerSrcCh: make(chan *tailerSrc, 100),
+			numUsedFds:   	   semaphore.NewSemaphore(0),
+		}
+	}
+
```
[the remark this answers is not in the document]
khanhntd: Agree. AWS SDK Go uses the same error handling ways; forgot about how to deal with when error is not usefulness 👍 .

plugins/inputs/logfile/tailersrc_test.go
khanhntd: FYi, there is a function for [setup tailer](https://github.com/aws/amazon-cloudwatch-agent/blob/master/plugins/inputs/logfile/tail/tail_test.go#L121-L153). The different between these two are for setting the `ts.SetOutput(func(evt logs.LogEvent)`

plugins/inputs/logfile/logfile.go
```diff
@@ -205,6 +219,10 @@ func (t *LogFile) FindLogSrc() []logs.LogSrc {
 				continue
 			}
 
+			if t.numUsedFds.GetLimit() != 0 {
+				t.numUsedFds.Acquire()
+			}
```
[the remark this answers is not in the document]
khanhntd: 👍  Agree. We could use `TryAcquire` instead of Acquire and give a `timeout` for this instead of holding down the Tailers. Good call on this

plugins/inputs/logfile/tailersrc_test.go
khanhntd: Same as previous comment though
```
Maybe not a second. However, we would still need `Sleep` for the tailer to be deleted.
```

internal/semaphore/semaphore_test.go
khanhntd: Yes. I have added [NewSemaphore(0)](https://github.com/aws/amazon-cloudwatch-agent/pull/491/files#diff-dabd66c7b47b8a390b2b5716a07ee02e854cf06cdb4a5288a1e10854b7d1ee19R43-R50). Thanks for reminding me

internal/semaphore/semaphore.go
khanhntd: > Do we intend to reuse this interface elsewhere with a different underlying implementation?

For ourselves, no. My intention for using interface only to centralize the Function for Semaphore. However, if there are customers using our semaphore, then that's fine too for having an interface here so no disadvantages when adding the interface.

internal/semaphore/semaphore.go
```diff
@@ -0,0 +1,86 @@
...
+func (sem *semaphore) Acquire(timeout time.Duration) bool {
+	if (timeout > 0) {
+		timer := time.NewTimer(timeout)
+		defer timer.Stop()
+		select {
+		case <-sem.slots:
+			return true
+		case <-timer.C:
+		}
+	}
```
SaxyPandaBear, line 57: We don't need to have a timer here. We can just use `time.After()` to fail through:
```suggestion
	if (timeout > 0) {
		select {
		case <-sem.slots:
			return true
		case <-time.After(timeout):
		}
	}
```

plugins/inputs/logfile/constant.go
```diff
@@ -0,0 +1,51 @@
...
+      ## Whether file is a named pipe
+      pipe = false
+      destination = "cloudwatchlogs"
+      ## Max size of each log event, defaults to 262144 (256KB)
+      max_event_size = 262144
+      ## Suffix to be added to truncated logline to indicate its truncation, defaults to "[Truncated...]"
+      truncate_suffix = "[Truncated...]"
+`
+
+const defaultTimeoutToAcquire = 1 * time.Second
```
SaxyPandaBear, line 51: One second is a long time. Are you sure this can't be like 100ms? I'm very concerned about blocking on a channel for this semaphore because we run the risk of putting the agent in a deadlock.

plugins/inputs/logfile/tailersrc_test.go
```diff
@@ -304,13 +282,13 @@ func TestOffsetDoneCallBack(t *testing.T) {
 	time.Sleep(2 * time.Second)
 
 	// Removal of log file should stop tailersrc
-	if err := os.Remove(file.Name()); err != nil {
-		t.Errorf("failed to remove log file '%v': %v", file.Name(), err)
-	}
+	err = os.Remove(file.Name())
+	assert.NoError(t, err)
+	time.Sleep(500 * time.Millisecond)
```
SaxyPandaBear, line 287: Why do we need a sleep here? Blocking on `<-done` is supposed to account for this without sleep timers.

plugins/inputs/logfile/tailersrc_test.go
```diff
@@ -325,9 +303,11 @@ func TestTailerSrcFiltersSingleLineLogs(t *testing.T) {
 	publishLogsToFile(resources.file, matchedLog, unmatchedLog, n, 0)
 
 	// Removal of log file should stop tailersrc
-	if err := os.Remove(resources.file.Name()); err != nil {
-		t.Errorf("failed to remove log file '%v': %v", resources.file.Name(), err)
-	}
+	err := os.Remove(resources.file.Name())
+	assert.NoError(t, err)
+	time.Sleep(500 * time.Millisecond)
+	assert.Equal(t, 0, resources.numUsedFds.GetCount())
```
SaxyPandaBear, line 309: Why do we need to assert and test for the fds counter in every test? It's redundant and doesn't increase test coverage.

internal/semaphore/semaphore.go
```diff
@@ -0,0 +1,86 @@
...
+
+package semaphore
+
+import (
+	"log"
+	"time"
+)
+
+type Semaphore interface {
+	//Acquire a slot in the semaphore with a timeout
```
adam-mateen, line 12: I think it is worth documenting that if timeout is <=0, then Acquire() will not block. I initially thought timeout=0 would behave as if timeout=infinity.

internal/semaphore/semaphore.go
SaxyPandaBear: > if there are customers using our semaphore

I highly doubt that anyone would modify the underlying implementation of this for their own needs, in their own forks. Adding this interface does not make it any more "maintainable" than omitting it.

plugins/inputs/logfile/logfile.go
```diff
@@ -205,6 +171,13 @@ func (t *LogFile) FindLogSrc() []logs.LogSrc {
 				continue
 			}
 
+			if t.numUsedFds.GetLimit() != 0 {
+				if ok := t.numUsedFds.Acquire(defaultTimeoutToAcquire); !ok {
+					t.Log.Debugf("Cannot increase counter of used file descriptors")
+				}
+			}
```
adam-mateen, line 178: The semaphore code was really well done!

I don't think it needs to be used though.
It would be simpler and use less memory to have a single counter (integer instead of channel) of the open files (aka files being tailed). Then just log the counter's current value periodically.

It is undesirable to change behavior here and block for up to 1 second.

khanhntd: For the information: there were others use semaphore as a counter (XSync, Marusama). However, after comparing the performance between them, I would say this version (or similar to DropBox), this would be the best option(for now) when decide the file descriptors as a upper bound.
```
BenchmarkSemaphore_AcquireReleaseUnderLimitSimple-12          19336089                62.26 ns/op            0 B/op          0 allocs/op
BenchmarkSemaphore_AcquireReleaseUnderLimit-12                    1652306               726.5 ns/op             0 B/op          0 allocs/op
BenchmarkSemaphore_AcquireReleaseOverLimit-12                     32737             37202 ns/op               0 B/op          0 allocs/op
BenchmarkMarusamaSemaphore_AcquireReleaseUnderLimitSimple-12                         6038138               171.7 ns/op            96 B/op          1 allocs/op
BenchmarkMarusamaSemaphore_AcquireReleaseUnderLimit-12                                 479114              2268 ns/op             960 B/op         10 allocs/op
BenchmarkMarusamaSemaphore_AcquireReleaseOverLimit-12                                   50464             24728 ns/op            9600 B/op        100 allocs/op
BenchmarkAbiosoftSemaphore_AcquireReleaseUnderLimitSimple-12                 8135151               145.0 ns/op             0 B/op          0 allocs/op
BenchmarkAbiosoftSemaphore_AcquireReleaseUnderLimit-12                          432472              2731 ns/op               0 B/op          0 allocs/op
BenchmarkAbiosoftSemaphore_AcquireReleaseOverLimit-12                           51364             27126 ns/op               0 B/op          0 allocs/op
BenchmarkPivotalGolangSemaphore_AcquireReleaseUnderLimitSimple-12            3358042               355.3 ns/op           128 B/op          2 allocs/op
BenchmarkPivotalGolangSemaphore_AcquireReleaseUnderLimit-12                      149810              7881 ns/op            1200 B/op         20 allocs/op
BenchmarkPivotalGolangSemaphore_AcquireReleaseOverLimit-12                      14558             83574 ns/op           12000 B/op        200 allocs/op
BenchmarkXSyncSemaphore_AcquireReleaseUnderLimitSimple-12                   40687284                27.22 ns/op            0 B/op          0 allocs/op
BenchmarkXSyncSemaphore_AcquireReleaseUnderLimit-12                             1000000              1060 ns/op               0 B/op          0 allocs/op
BenchmarkXSyncSemaphore_AcquireReleaseOverLimit-12                              17275             69725 ns/op           15994 B/op        299 allocs/op
```
However, the ` 1 second` is only an arbitrary value and it could be more or less. But as long as it within acceptable range (e.g do customer cares about the show number of monitored files being late in `1 second` or not? Maybe its exchange between source of truth and the fastness a little bit.

plugins/inputs/logfile/tailersrc_test.go
khanhntd: To account for the case whether the semaphore would drop by 1 if it stops tailing the file. This can be addressed in [tail_test](https://github.com/aws/amazon-cloudwatch-agent/blob/40bb174c0e2309da6bd2c6e1a36c501324b2d6b0/plugins/inputs/logfile/tail/tail_test.go) too which I had done it. However, there is a strong correlation between `tail` and `tailsrc` (tailsrc has a attribute of tailer). Therefore, I would introduce this to avoid any regression between these two

plugins/inputs/logfile/tailersrc_test.go
khanhntd: Ahhhh. I forgot we could use `<-done` to block. Forgot about that. Nice catch

plugins/inputs/logfile/constant.go
khanhntd: Yes. This can be 100ms or even `1ms` since `1 * time.Second` is only an arbitrary value so good point on this one.

plugins/inputs/logfile/logfile.go
khanhntd: Maybe I'm wrong but according to the `syns/atomic`
```
Package atomic provides low-level atomic memory primitives useful for implementing synchronization algorithms.
These functions require great care to be used correctly. Except for special, low-level applications, synchronization is better done with channels or the facilities of the sync package. 
```
So I'm still considering between the two. However, will compare performance before giving my final decision.

khanhntd: Compare a little bit when doing `atomic counter with mutex` 
```
BenchmarkSemaphore_Acquire_Release_under_limit-12           9673            124611 ns/op               0 B/op          0 allocs/op
BenchmarkAtomicCounterWithOutLock_Under-12                            19537             53280 ns/op               0 B/op          0 allocs/op
BenchmarkAtomicCounterWithLock_Under-12                              7071            247445 ns/op              18 B/op          0 allocs/op
```
So maybe I'm still standing by with using semapahore instead

adam-mateen: Sorry to derail, I wasn't as concerned about the implementation of counter vs semaphore.
Agreed that an atomic counter (no lock needed) is simpler and would suffice here.

More concerned that you are introducing a blocking function call on `Acquire()`.
I initially thought this PR would just add a variable to track the count of open files AND then periodically log it.
Not to check the counter value and change behavior based on the counter value.

internal/semaphore/semaphore.go
```diff
@@ -0,0 +1,84 @@
...
+	slots chan struct{}
+}
+
+func NewSemaphore(limit int) Semaphore {
+	if limit < 0 {
+		log.Panic("E! Semaphore size limit must not be negative")
+	}
+
+	sem := &semaphore{
+		slots: make(chan struct{}, limit),
```
adam-mateen, line 38: I was worried about limit being a big number (<16K) and `slots` taking up an unnecessary amount of memory.
From what I can see though empty struct is 0 bytes, and a channel of them is only 8 bytes no matter what the capacity is.

So I am not worried anymore.

plugins/inputs/logfile/logfile.go
```diff
@@ -205,6 +171,13 @@ func (t *LogFile) FindLogSrc() []logs.LogSrc {
 				continue
 			}
 
+			if t.numUsedFds.GetLimit() != 0 {
+				if ok := t.numUsedFds.Acquire(defaultTimeoutToAcquire); !ok {
+					t.Log.Debugf("Cannot increase counter of used file descriptors")
```
adam-mateen, line 176: Nitpick:
Can use `t.Log.Debug()` instead of `t.Log.Debugf()` if you have no format string.

plugins/inputs/logfile/logfile.go
adam-mateen: Yeah, I know it does not block indefinitely, but I consider it "blocking" even if it is blocks for 100ms.

*** I don't think it should block at all. I propose setting `defaultTimeoutToAcquire` to 0, or just removing the timeout parameter from `Acquire()`.***

SaxyPandaBear: Agreed. The timeout stops it from blocking indefinitely but is still potentially paused for up to 100ms. That's not great

internal/semaphore/semaphore.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+
+func (sem *semaphore) Acquire(timeout time.Duration) bool {
+	if (timeout > 0) {
+		select {
+		case <-sem.slots:
+			return true
+		case <-time.After(timeout):
+		}
```
SaxyPandaBear, line 55: Shouldn't this return false because it failed to claim before the timeout

plugins/inputs/logfile/constant.go
```diff
@@ -0,0 +1,51 @@
...
+      ## Whether file is a named pipe
+      pipe = false
+      destination = "cloudwatchlogs"
+      ## Max size of each log event, defaults to 262144 (256KB)
+      max_event_size = 262144
+      ## Suffix to be added to truncated logline to indicate its truncation, defaults to "[Truncated...]"
+      truncate_suffix = "[Truncated...]"
+`
+
+const defaultTimeoutToAcquire = 0 * time.Millisecond
```
SaxyPandaBear, line 51: Do we need to multiply by millisecond?

khanhntd:
Close since https://github.com/aws/amazon-cloudwatch-agent/pull/506 has a better solution and better performance

PR #493, opened 2022-06-16
Increase Timeout, Sometimes CA Bundle Test Needs More Time To Run

sethAmazon:
# Description of the issue
CA Bundle Time out 
https://github.com/aws/amazon-cloudwatch-agent/runs/6919650158?check_suite_focus=true

# Description of changes
Increase timeout

# Tests
N/A

Possible better solution would be to download the go mod on instances during the instance builder phase so this is cached in the ami. Since the go mod download took 10 minutes in this test.

### DIFF
--- a/.github/workflows/integrationTest.yml
+++ b/.github/workflows/integrationTest.yml
@@ -431,7 +431,7 @@ jobs:
         uses: nick-invision/retry@v2
         with:
           max_attempts: 3
-          timeout_minutes: 15
+          timeout_minutes: 30
           retry_wait_seconds: 5
           command: |
             cd integration/terraform/ec2/linux


.github/workflows/integrationTest.yml
```diff
@@ -431,7 +431,7 @@ jobs:
         uses: nick-invision/retry@v2
         with:
           max_attempts: 3
-          timeout_minutes: 15
+          timeout_minutes: 20
```
khanhntd, line 434: 25 minutes or 30 minutes would be fine by me. What would be the result of bad timeout? As far as I am concerning, none and the test seems not holding on something or leaking something.

sethAmazon: It is more I don't want to wait for a hanging test to finish. Remember in the otel project we would have tests hanging for 4 hours. I can change this to 30 if you want.

khanhntd: Based on what I remembered, its related to create AWS resources which are not on our part; however, that does not seem frequency enough to take a concern. Therefore, 30 minutes would be fine by me and bad case, only waiting for 1 hour and 30 minutes :)

PR #496, opened 2022-06-20
Basic Functionality for Performance Tracking's DB

okankoAMZ:
# There was no DynamoDB manager API for Performance Tracker
_We need to store data in dynamodb but there was no script that was interacting with dynamo_

#  Added basic table management for Performance Tracking
_This change allows us to create, remove, search list of table, upload items to the db_

# License

# Tests
_I have tested if we can create a table, remove a table, upload an element to a table, and if we can parses data collection data_

# Requirements
2. Run `make linter`


integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,237 @@
...
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/dynamodb"
+	"github.com/aws/aws-sdk-go/service/dynamodb/dynamodbattribute"
+	"github.com/aws/aws-sdk-go/service/dynamodb/dynamodbiface"
+)
+const(
+	TIME_CONSTANT = 300.0 // this const is in seconds , 5 mins
```
sethAmazon, line 16: Can we make this configurable?

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,237 @@
...
+)
+type TransmitterAPI struct{
+	DynamoDbClient dynamodbiface.DynamoDBAPI
+	DataBaseName      string // this is the name of the table when test is run
+}
+// this is the packet that will be sent converted to DynamoItem
+type Metric struct{
+	Average float64
```
sethAmazon, line 24: We could want p99 pMax pMin as pre calculated in the future. I do not think that would be required now. This struct is extendable good job.

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,237 @@
...
+	return transmitter
+
+}
+/*
+CreateTable()
+Desc: Will create a DynamoDB Table with given param. and config
+Params:
+Side Effects: Creates a dynamoDB table
+*/
+func (transmitter * TransmitterAPI) CreateTable() error{
```
sethAmazon, line 76: Like this since we will being this on more than 1 account. Good automation.

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+	return false,err
+
+}
+/*
+RemoveTable()
+Desc: Removes the table that was craeted with initialization. 
+Side effects: Removes a Dynamodb table
+*/
+func (transmitter * TransmitterAPI) RemoveTable() error{
```
sethAmazon, line 155: Why would we want to remove the table?

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+func (transmitter * TransmitterAPI) Parser(data []byte) (map[string]interface{},error){
+	dataHolder := collectorData{}
+	err := json.Unmarshal(data,&dataHolder)
+	if err !=nil{
+		return nil,err
+	}
+	packet := make(map[string]interface{})
+	//temp solution
```
sethAmazon, line 198: I would add @TODO since it will highlight in intelij.

sethAmazon:
I don't think this should be merged at the moment since it does not have functionality as is. Good start. @SaxyPandaBear comments.

integration/test/performance_tracker/transmitter/transmitter.go
okankoAMZ: I had it for test purposes, where I was creating a table for each test then cleaning it by removing it. However, it will not be used in performance tracker.

sethAmazon: This can be made into a clean gh action with the input table name. Don't do that for now since that will take a while. You can keep this here for now.

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+	return fmt.Sprintf("%v",item),err
+}
+/*
+TableExist()
+Desc: Checks if the the table exist and returns the value
+//https://github.com/awsdocs/aws-doc-sdk-examples/blob/05a89da8c2f2e40781429a7c34cf2f2b9ae35f89/gov2/dynamodb/actions/table_basics.go
+*/
+func (transmitter * TransmitterAPI) TableExist() (bool,error){
+	l,err := transmitter.ListTables()
```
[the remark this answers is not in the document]
sethAmazon: Would not have to loop through tables. Would be able to pull down only 1 table. Either way is fine but I don't like the loop. @SaxyPandaBear thoughts.

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,237 @@
...
+		fmt.Printf("Couldn't add item to table. Here's why: %v\n", err)
+	}
+	return fmt.Sprintf("%v",item),err
+}
+/*
+TableExist()
+Desc: Checks if the the table exist and returns the value
+//https://github.com/awsdocs/aws-doc-sdk-examples/blob/05a89da8c2f2e40781429a7c34cf2f2b9ae35f89/gov2/dynamodb/actions/table_basics.go
+*/
+func (transmitter * TransmitterAPI) TableExist() (bool,error){
```
SaxyPandaBear, line 140: Do we need to return an error here? Looks like we just need to return a bool

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,237 @@
...
+	l,err := transmitter.ListTables()
+	for i:=0; i< len(l); i++{
+		if transmitter.DataBaseName == l[i]{
+			return true,nil
+		}
+	}
```
SaxyPandaBear, line 146: Assuming there's no error, the for loop could be simplified, but I think we might need to sync up and talk about the database table management to see if we really want to do anything particular in this code about handling the database in code versus separately
```suggestion
	tables, err := transmitter.ListTables()
	for table := range tables {
	    if transmitter.DataBaseName == table {
	        return true
	    }
	}
```

integration/test/performance_tracker/transmitter/transmitter.go
SaxyPandaBear, line 155: Why do we want to delete the table?

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,237 @@
...
+	if err != nil {
+		fmt.Printf("Couldn't list tables. Here's why: %v\n", err)
+	} else {
+		for  i:=0; i< len(tables.TableNames); i++{
+			tableNames =  append(tableNames,*tables.TableNames[i])
+		}
+		
+	}
+	return tableNames, err
+}
```
SaxyPandaBear, line 179: Do we need to list tables outside of checking whether the table name exists? This function seems unnecessary to me.

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,237 @@
...
+func testFiller(testname string) * TransmitterAPI{
+	transmitter := new(TransmitterAPI)
+	//setup aws session
+	session := session.Must(session.NewSessionWithOptions(session.Options{
+		SharedConfigState: session.SharedConfigEnable,
+	}))
+	transmitter.DynamoDbClient =  dynamodb.New(session)
+	transmitter.DataBaseName = testname
+	return transmitter
+}
```
SaxyPandaBear, line 237: Doesn't this duplicate the code in the init function?

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,237 @@
...
+func (transmitter * TransmitterAPI) Parser(data []byte) (map[string]interface{},error){
+	dataHolder := collectorData{}
+	err := json.Unmarshal(data,&dataHolder)
+	if err !=nil{
+		return nil,err
+	}
+	packet := make(map[string]interface{})
+	//temp solution
+	packet["Hash"] = fmt.Sprintf("%d",time.Now().UnixNano())
+	/// will remove
```
SaxyPandaBear, line 200: Will you remove it?

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,237 @@
...
+		diffSum := 0.0
+		for _,val := range rawMetricData.Values{
+			diffSum = diffSum +  (avg - val)
+		}
+		
+		metric := Metric{
+			Average: avg,
+			StandardDev: math.Sqrt(math.Pow(diffSum,2)/float64(numDataPoints)),
+			Period: int(TIME_CONSTANT/(numDataPoints)),
+			Data:  rawMetricData.Values}
```
SaxyPandaBear, line 218: I don't think we need to do any of this math ourselves. Take a look at GetMetricStatistics https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/US_SingleMetricPerInstance.html We also discussed in the last review that we think that the p99 stat is more valuable than the mean.

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,237 @@
...
+	return err
+}
+// ListTables lists the DynamoDB table names for the current account.
+func (transmitter * TransmitterAPI) ListTables() ([]string, error) {
+	var tableNames []string
+	input := &dynamodb.ListTablesInput{}
+	tables, err := transmitter.DynamoDbClient.ListTables(
+		input)
+	if err != nil {
+		fmt.Printf("Couldn't list tables. Here's why: %v\n", err)
```
SaxyPandaBear, line 171: "Here's why" is not necessary here. 
```suggestion
		fmt.Printf("Error listing tables: %v\n", err)
```

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,237 @@
...
+	item, err := dynamodbattribute.MarshalMap(packet)
+	if err != nil {
+		panic(err)
+	}
+	DBitem := &dynamodb.PutItemInput{
+		Item:      item,
+		TableName: aws.String(transmitter.DataBaseName),
+	}
+	// fmt.Println(DBitem)
+	// return err
```
SaxyPandaBear, line 128: If you don't need this, can delete it

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,237 @@
...
+	// check if the dynamo table exist if not create it
+	tableExist, err:= transmitter.TableExist()
+	if err !=nil{
+		return nil
+	}
+	if !tableExist{
+		fmt.Println("Table doesn't exist")
+		err := transmitter.CreateTable()
+		if err != nil{
+			fmt.Println("Couldn't create table")
```
SaxyPandaBear, line 62: Isn't this print statement redundant? In the `CreateTable()` function, it prints out any error that occurs when trying to create the table. This just doubles up on logging and doesn't provide any additional context or value.

integration/test/performance_tracker/transmitter/transmitter.go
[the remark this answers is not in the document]
SaxyPandaBear: I'm also not thrilled with the loop, though it is very simple and clean. I'm not versed in DDB APIs so not sure how complicated it would be to try to search for table name with a filter

SaxyPandaBear:
> I don't think this should be merged at the moment since it does not have functionality as is.

@sethAmazon I think there's a rework that needs to be done with respect to the statistics that are being calculated on the fly here when we could just use GetMetricStatistics. Frankly, it's my own fault for knowing that we could call GetMetricData, but we could have just used GetMetricStatistics on the period.

integration/test/performance_tracker/transmitter/transmitter.go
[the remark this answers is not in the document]
khanhntd: Its different. For example, AWS SDK Go V2 has paginated which reduce loads (maybe GoV1 too) than pulling the whole table and loop through it. Therefore, would recommend to use FilterAPI as Seth suggested.

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,237 @@
...
+		// fmt.Printf("%+v\n",metric)
+		// packet.Metrics[rawMetricData.Label] = metric
+		packet[rawMetricData.Label] = metric
+
+		
+	}
+	return packet,nil
+}
+
+func testFiller(testname string) * TransmitterAPI{
```
khanhntd, line 228: Is this meant for testing? If it is, then maybe following the [Go API itself would be a good idea](https://github.com/aws/aws-sdk-go/blob/v1.30.15/aws/ec2metadata/api_test.go). However, it sounds counter `counterintuitive` for me to test what have been tested in the API itself.

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,237 @@
...
+}
+// ListTables lists the DynamoDB table names for the current account.
+func (transmitter * TransmitterAPI) ListTables() ([]string, error) {
+	var tableNames []string
+	input := &dynamodb.ListTablesInput{}
+	tables, err := transmitter.DynamoDbClient.ListTables(
+		input)
+	if err != nil {
+		fmt.Printf("Couldn't list tables. Here's why: %v\n", err)
+	} else {
```
khanhntd, line 172: **Nitpick**: Case guard here is better. For example:
```
if err != nil {
  return [],err
}

for  i:=0; i< len(tables.TableNames); i++{
			tableNames =  append(tableNames,*tables.TableNames[i])
}
return tableNames, err
```

integration/test/performance_tracker/transmitter/transmitter.go
SaxyPandaBear: The problem with terraform is we need this table to persist. We can't just create and recreate it.

SaxyPandaBear: I think it's best to leave this for now and revisit this later on

khanhntd: One of the way is to create a `separate` initialize resources folder which is meant only for creating resources for somebody onboarding to integration test. Therefore, avoid being involved in terraform life cycle and its a one-time (same how ADOT does)

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,237 @@
...
+Desc: Will create a DynamoDB Table with given param. and config
+Params:
+Side Effects: Creates a dynamoDB table
+*/
+func (transmitter * TransmitterAPI) CreateTable() error{
+	input := &dynamodb.CreateTableInput{
+		AttributeDefinitions: []*dynamodb.AttributeDefinition{
+			{
+				AttributeName: aws.String("Hash"),
+				AttributeType: aws.String("S"),
```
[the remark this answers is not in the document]
SaxyPandaBear: The hash key is the commit, which is alphanumeric. There is no reason to consider Number or Byte for the type of the commit hash.

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,237 @@
+package transmitter
+
+import (
+	"encoding/json"
+	"fmt"
+	"math"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
```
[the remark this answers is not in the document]
SaxyPandaBear: This uses v2. The usage of `aws-sdk-go/aws` is for creating the properly typed inputs for the API calls.

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,237 @@
...
+type TransmitterAPI struct{
+	DynamoDbClient dynamodbiface.DynamoDBAPI
+	DataBaseName      string // this is the name of the table when test is run
+}
+// this is the packet that will be sent converted to DynamoItem
+type Metric struct{
+	Average float64
+	StandardDev float64
+	Period	int 	//in seconds
+	Data []float64
```
SaxyPandaBear, line 27: Do we want to store the raw data points as part of the record? What value does it provide?

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,237 @@
...
+)
+const(
+	TIME_CONSTANT = 300.0 // this const is in seconds , 5 mins
+)
+type TransmitterAPI struct{
+	DynamoDbClient dynamodbiface.DynamoDBAPI
+	DataBaseName      string // this is the name of the table when test is run
+}
+// this is the packet that will be sent converted to DynamoItem
+type Metric struct{
```
SaxyPandaBear, line 23: I think that the structure for what gets stored in DDB might need some more thought. When we pull data out of Dynamo, how do we know the chronological ordering of the commits?

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+	fmt.Println("API ready")
+	return transmitter
+
+}
+/*
+CreateTable()
+Desc: Will create a DynamoDB Table with given param. and config
+Params:
+Side Effects: Creates a dynamoDB table
```
SaxyPandaBear, line 74: I think that this style of commenting is fine, but this is also really verbose and redundant. The name of the function is `CreateTable`, so I would expect that a side effect is that it creates a table.

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,237 @@
...
+			return true,nil
+		}
+	}
+	return false,err
+
+}
+/*
+RemoveTable()
+Desc: Removes the table that was craeted with initialization. 
+Side effects: Removes a Dynamodb table
```
SaxyPandaBear, line 153: Again, kind of redundant for a side effect. It's fine to keep it, just calling out that it is not necessary

integration/test/performance_tracker/transmitter/transmitter.go
SaxyPandaBear: We will not want to delete the table once it's live. It makes me think that we will need to automate backups of the database, but luckily that's pretty easy https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/CreateBackupAWS.html

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,237 @@
...
+	} else {
+		for  i:=0; i< len(tables.TableNames); i++{
+			tableNames =  append(tableNames,*tables.TableNames[i])
+		}
+		
+	}
+	return tableNames, err
+}
+
+func (transmitter * TransmitterAPI) SendItem(data []byte) (string,error) {
```
SaxyPandaBear, line 181: Send item to where? The function name is not very descriptive

integration/test/performance_tracker/transmitter/transmitter.go
SaxyPandaBear: If we move to use GetMetricStatistics, I don't think most of this is necessary.

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,237 @@
...
+	if err != nil {
+		panic(err)
+	}
+	DBitem := &dynamodb.PutItemInput{
+		Item:      item,
+		TableName: aws.String(transmitter.DataBaseName),
+	}
+	// fmt.Println(DBitem)
+	// return err
+	_, err = transmitter.DynamoDbClient.PutItem(DBitem)
```
SaxyPandaBear, line 129: Luckily, I think we're fine here because if we reuse a key, PutItem replaces the whole item, which is desirable for our usage of DDB.

integration/test/performance_tracker/transmitter/transmitter.go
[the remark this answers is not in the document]
khanhntd: No. `aws-sdk-go-v2/aws` also supports [String](https://github.com/aws/aws-sdk-go-v2/blob/main/aws/types.go). Moreover, the below dependencies are not using v2 right?
```
"github.com/aws/aws-sdk-go/service/dynamodb"
"github.com/aws/aws-sdk-go/service/dynamodb/dynamodbattribute"
"github.com/aws/aws-sdk-go/service/dynamodb/dynamodbiface"
```

integration/test/performance_tracker/transmitter/transmitter.go
khanhntd: Sounds other way around to me. Do we need to return a bool here? An error is fine by me 👍

integration/test/performance_tracker/transmitter/transmitter.go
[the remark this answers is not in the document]
SaxyPandaBear: The code you linked is not relevant. The code used from the `aws` package converts types to pointers to a type:
```
// String returns a pointer to the string value passed in.
func String(v string) *string {
	return &v
}
```
You are right that the actual DynamoDB imports are v1

integration/test/performance_tracker/transmitter/transmitter.go
SaxyPandaBear: No, the name of the function implies that the purpose of the function is to check whether or not the table exists. That means it makes sense to be a bool. If the table doesn't exist, want to try to create it, so it doesn't make much of a difference if we don't find the table or if there's an error trying to look it up. It could help short-circuit the logic so if there is an error we fail fast, but I'm honestly not really concerned about that at all.

.github/workflows/integrationTest.yml
```diff
@@ -615,4 +615,74 @@ jobs:
           max_attempts: 3
           timeout_minutes: 8
           retry_wait_seconds: 5
-          command: cd integration/terraform/ecs/linux && terraform destroy --auto-approve
\ No newline at end of file
+          command: cd integration/terraform/ecs/linux && terraform destroy --auto-approve
+
+  PerformanceTrackingTest:
```
SaxyPandaBear, line 620: I forgot, are we keeping this in here for merge?

SaxyPandaBear:
I think you're going to need to download the dynamodb dependencies in order to build.

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,236 @@
...
+
+func (transmitter * TransmitterAPI) Parser(data []byte) (map[string]interface{},error){
+	dataHolder := collectorData{}
+	err := json.Unmarshal(data,&dataHolder)
+	if err !=nil{
+		return nil,err
+	}
+	packet := make(map[string]interface{})
+	//@TODO: add git integration temp solution
+	packet["Hash"] = fmt.Sprintf("%d",time.Now().UnixNano())
```
khanhntd, line 210: **Nit pick**: Same for CommitDate

```suggestion
	packet["Hash"] = time.Now().UnixNano().String()
```

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,236 @@
...
+			sum += val
+		}
+		avg :=  sum /numDataPoints
+		// calculate diff between mean and values 
+		diffSum := 0.0
+		for _,val := range rawMetricData.Values{
+			diffSum = diffSum +  (avg - val)
+		}
+		//----------------
+		metric := Metric{
```
khanhntd, line 227: Maybe I'm wrong. @SaxyPandaBear do we already have the aggregator plugins that gave us more statistical values? So why do we want to aggregate the "aggregated" metrics again?

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,236 @@
...
+	"errors"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/aws/session"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb/dynamodbattribute"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb/dynamodbiface"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb/types"
+)
+const METRIC_PERIOD= 300.0 // this const is in seconds , 5 mins
```
khanhntd, line 18: Why would we want to initiate this variable to `float` and in later time we convert the [numDataPoints](https://github.com/aws/amazon-cloudwatch-agent/pull/496/files#diff-c3d3ee8804154aa85568c8778e4b9e58dcae04e6a2d03c416a727c4a3f4457fcR214) to float again?

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,236 @@
...
+
+type TransmitterAPI struct{
+	dynamoDbClient dynamodbiface.DynamoDBAPI
+	DataBaseName      string // this is the name of the table when test is run
```
khanhntd, line 22: **Nitpick**: Would this use elsewhere `DataBaseName`? If not then suggesting to use `dataBaseName`

integration/test/performance_tracker/transmitter/transmitter.go
```diff
@@ -0,0 +1,236 @@
...
+		fmt.Printf("Error adding item to table.  %v\n", err)
+	}
+	return fmt.Sprintf("%v",item),err
+}
+/*
+TableExist()
+Desc: Checks if the the table exist and returns the value
+//https://github.com/awsdocs/aws-doc-sdk-examples/blob/05a89da8c2f2e40781429a7c34cf2f2b9ae35f89/gov2/dynamodb/actions/table_basics.go
+*/
+func (transmitter * TransmitterAPI) TableExist() (bool,error){
```
khanhntd, line 134: **Not Nit pick**: Provide the `context.Background()` as the top level and pass down to input function would be a greater idea since `context` is used for avoid SLE case and wasteful memory resources which we will handle

integration/test/performance_tracker/transmitter/transmitter.go
SaxyPandaBear: I'm not sure what you're referring to. This isn't running inside the agent. This is part of the performance test. What this is doing is expecting a bunch of raw values and then is doing the calculations in memory. What we want to do is move away from this to just call GetMetricStatistics to cut out a lot of this work.

integration/test/performance_tracker/transmitter/transmitter.go
SaxyPandaBear: Now that #497 is merged, to integrate with this, we can pass in a `context.Context` parameter that gets instantiated in the performance test.

integration/test/performance_tracker/transmitter/transmitter.go
SaxyPandaBear: Maybe a different question, do we need to do floating point division and then round by casting to an int, or can we change it around?

The issue is around using the number of data points to calculate the average. You could just keep numDataPoints as an int, and then cast when calculating the average, e.g.:
```go
avg := sum / float64(numDataPoints)
```
which leaves numDataPoints as an int. Then we don't need to do the casting when doing METRIC_PERIOD / numDataPoints

.github/workflows/integrationTest.yml
```diff
@@ -402,6 +402,8 @@ jobs:
       matrix:
         arrays: ${{ fromJson(needs.GenerateTestMatrix.outputs.ec2_linux_matrix) }}
     steps:
+      - name: TEMP TEST LOOKUP
```
sethAmazon, line 405: Why?

integration/terraform/ec2/linux/main.tf
```diff
@@ -6,6 +6,7 @@ resource "aws_instance" "integration-test" {
   vpc_security_group_ids = var.vpc_security_group_ids
   provisioner "remote-exec" {
     inline = [
+      "echo sha ${var.sha}",
```
sethAmazon, line 9: should go after cloud init

integration/test/performancetest/performance_query_utils.go
```diff
@@ -19,7 +19,7 @@ const (
 	DimensionName = "InstanceId"
 	Stat = "Average"
 	Period = 30
-	configPath = "./resources/config.json"
+	// configPath = "./resources/config.json"
```
sethAmazon, line 22: Why the comment out.

integration/test/performancetest/performance_test.go
```diff
@@ -8,17 +8,17 @@ import(
...
 	"context"
-
 	"github.com/aws/amazon-cloudwatch-agent/integration/test"
 )
 
 const (
 	configPath = "resources/config.json"
 	configOutputPath = "/opt/aws/amazon-cloudwatch-agent/bin/config.json"
-	agentRuntimeMinutes = 20
+	agentRuntimeMinutes = 5 //20 def
```
sethAmazon, line 17: 20 def?

integration/test/performancetest/transmitter.go
```diff
@@ -0,0 +1,239 @@
...
+	"strconv"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/feature/dynamodb/attributevalue"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb/types"
+)
+
+const METRIC_PERIOD = agentRuntimeMinutes * 60 // this const is in seconds , 5 mins
```
sethAmazon, line 19: Name change this

integration/test/performancetest/transmitter.go
```diff
@@ -0,0 +1,239 @@
...
+		}else{
+			// @TODO:ADD GetMetricStatistics after merging with data collection code
+			sum :=0.0
+			for _,val := range rawMetricData.Values {
+				sum += val
+			}
+			avg =  sum /numDataPoints
+		}
+		//----------------
+		metric := Metric{
```
sethAmazon, line 229: Add a comment //@TODO

integration/test/performancetest/transmitter.go
```diff
@@ -0,0 +1,239 @@
...
+}
+
+/*
+InitializeTransmitterAPI
+Desc: Initializes the transmitter class
+Side effects: Creates a dynamodb table if it doesn't already exist
+*/
+func InitializeTransmitterAPI(DataBaseName string) *TransmitterAPI {
+	//setup aws session
+	cfg, err := config.LoadDefaultConfig(context.TODO(), config.WithRegion(os.Getenv("AWS_REGION")))
```
sethAmazon, line 52: AWS_REGION use const. 
does this not work by default?

integration/test/performancetest/transmitter.go
```diff
@@ -0,0 +1,239 @@
...
+				ReadCapacityUnits:  aws.Int64(10),
+				WriteCapacityUnits: aws.Int64(10),
+			},
+			TableName: aws.String(transmitter.DataBaseName),
+		}) // this is the config for the new table)
+	if err != nil {
+		fmt.Printf("Error calling CreateTable: %s", err)
+		return err
+	}
+	time.Sleep(10*time.Second)
```
sethAmazon, line 115: Try to see if there is a way to have wait for create instead of sleep maybe a polling for loop

integration/test/performancetest/transmitter.go
```diff
@@ -0,0 +1,239 @@
...
+	return fmt.Sprintf("%v", item), err
+}
+
+/*
+TableExist()
+Desc: Checks if the the table exist and returns the value
+//https://github.com/awsdocs/aws-doc-sdk-examples/blob/05a89da8c2f2e40781429a7c34cf2f2b9ae35f89/gov2/dynamodb/actions/table_basics.go
+*/
+func (transmitter *TransmitterAPI) TableExist() (bool, error) {
+	// l,err := transmitter.ListTables()
```
sethAmazon, line 151: Remove this

integration/test/performancetest/transmitter.go
```diff
@@ -0,0 +1,239 @@
...
+		exists = false
+	}
+	return exists, err
+}
+
+/*
+RemoveTable()
+Desc: Removes the table that was craeted with initialization.
+*/
+func (transmitter *TransmitterAPI) RemoveTable() error {
```
sethAmazon, line 179: I do not like code not being used.

sethAmazon: //@TODO add cleaner script

integration/test/performancetest/transmitter.go
```diff
@@ -0,0 +1,239 @@
...
+
+func (transmitter *TransmitterAPI) Parser(data []byte) (map[string]interface{}, error) {
+	dataHolder := collectorData{}
+	err := json.Unmarshal(data, &dataHolder)
+	if err != nil {
+		return nil, err
+	}
+	packet := make(map[string]interface{})
+
+	packet["Hash"] =  os.Getenv("SHA") //fmt.Sprintf("%d", time.Now().UnixNano())
```
sethAmazon, line 212: consts

integration/test/performancetest/performance_test.go
```diff
@@ -34,9 +34,9 @@ func PerformanceTest(t *testing.T) {
 	test.StopAgent()
 
 	//collect data
-	data, err := GetPerformanceMetrics(instanceId, agentRuntime, agentContext)
+	data, err := GetPerformanceMetrics(instanceId, agentRuntimeMinutes, agentContext)
```
sethAmazon, line 37: get stats in GetPerformanceMetrics method

integration/test/performancetest/transmitter.go
```diff
@@ -0,0 +1,239 @@
...
+	packet["CommitDate"],_ = strconv.Atoi(os.Getenv("SHA_DATE"))
+	/// will remove
+	for _, rawMetricData := range dataHolder {
+		numDataPoints := float64(len(rawMetricData.Timestamps))
+		var avg float64
+		if numDataPoints <= 0{
+			avg = 0.0
+		}else{
+			// @TODO:ADD GetMetricStatistics after merging with data collection code
```
sethAmazon, line 221: Yeah here not the place to get stats

SaxyPandaBear (dismissed):
good enough for now. can refine in another iteration

sethAmazon:
Please rebase and squash commits. git history too long for a pr

SaxyPandaBear:
looks like you need to `go mod tidy`

sethAmazon (approved):

PR #497, opened 2022-06-20
Performance Tracking Integration Test and Data Collection

gmealy1:
# Description of the issue
Needed integration test that gets benchmark performance on cloudwatch agent.

# Description of changes
Added an integration test that starts cloudwatch agent and then collects metrics on it. The data still needs to be passed to the database API transmitter in get_performance_metrics.go.

Commit messages that were condensed down into one commit for this pr:
added configuration file for CWA test

initial commit for data collection/GetMetricData portion of test

initial commit for main test driver that starts agent and then calls data collection

changed test error to fatal if error is returned by data collection

added comments

removed unnecessary code that prints output

# License

# Tests
Tested formatting queries for GetMetricData and ran tests making successful queries to the API.

# Requirements
2. Run `make linter`


khanhntd:
**No blocker from me**: Maybe you and Okan can create another branch from the repo (same as how Telegraf upgrade are deployed) and created a final PR from that branch

SaxyPandaBear:
>  Maybe you and Okan can create another branch from the repo (same as how Telegraf upgrade are deployed) and created a final PR from that branch

I think this is fine. I would rather make incremental progress on this because it does not disrupt anyone else's development. The telegraf upgrade was a large, breaking change so it made sense to fix everything in a separate branch prior to merging.

integration/test/performance_tracker/performance_query_utils.go
```diff
@@ -0,0 +1,78 @@
...
+	metricDataQueries := []types.MetricDataQuery{}
+	
+	//generate list of individual metric requests
+	for i := 0; i < len(ids); i++ {
+		metricDataQueries = append(metricDataQueries, ConstructMetricDataQuery(ids[i], Namespace, DimensionName, dimensionValue, metricNames[i], timeDiff))
+	}
+	
+	input := &cloudwatch.GetMetricDataInput{
+		EndTime:   aws.Time(time.Unix(time.Now().Unix(), 0)),
+		StartTime: aws.Time(time.Unix(time.Now().Add(time.Duration(-timeDiff)*time.Minute).Unix(), 0)),
```
SaxyPandaBear, line 50: Technically there would be a slight gap in calling `time.Now()` twice. It would make sense to call `time.Now()` once and then operate off of that.

integration/test/performance_tracker/performance_query_utils.go
```diff
@@ -0,0 +1,78 @@
...
+	GetMetricData(ctx context.Context, params *cloudwatch.GetMetricDataInput, optFns ...func(*cloudwatch.Options)) (*cloudwatch.GetMetricDataOutput, error)
+}
+
+// GetMetrics Fetches the cloudwatch metrics for your provided input in the given time-frame
+func GetMetrics(c context.Context, api CWGetMetricDataAPI, input *cloudwatch.GetMetricDataInput) (*cloudwatch.GetMetricDataOutput, error) {
+	return api.GetMetricData(c, input)
+}
+
+// GenerateGetMetricInputStruct generates the struct required to make a query request to cloudwatch's GetMetrics
+func GenerateGetMetricInputStruct(ids []string, metricNames []string, instanceId string, timeDiff int) (*cloudwatch.GetMetricDataInput, error) {
```
SaxyPandaBear, line 31: Nitpick - you can consolidate the type definitions if two adjacent ones are the same. I think.
```suggestion
func GenerateGetMetricInputStruct(ids, metricNames []string, instanceId string, timeDiff int) (*cloudwatch.GetMetricDataInput, error) {
```

integration/test/performance_tracker/performance_query_utils.go
```diff
@@ -0,0 +1,78 @@
...
+
+// CWGetMetricDataAPI defines the interface for the GetMetricData function
+type CWGetMetricDataAPI interface {
+	GetMetricData(ctx context.Context, params *cloudwatch.GetMetricDataInput, optFns ...func(*cloudwatch.Options)) (*cloudwatch.GetMetricDataOutput, error)
+}
+
+// GetMetrics Fetches the cloudwatch metrics for your provided input in the given time-frame
+func GetMetrics(c context.Context, api CWGetMetricDataAPI, input *cloudwatch.GetMetricDataInput) (*cloudwatch.GetMetricDataOutput, error) {
+	return api.GetMetricData(c, input)
+}
```
SaxyPandaBear, line 28: This is a pass-through function that I don't think adds a lot of value. All we do is take in the required inputs to call the `GetMetricData` API, and then return exactly what the client returns. This function is not necessary since we could just call the API via the SDK client directly.

integration/test/performance_tracker/performance_query_utils.go
```diff
@@ -0,0 +1,78 @@
...
+		return nil, errors.New("Must supply metric ids, metric names, instance id, and time to collect metrics")
+	}
+
+	dimensionValue := instanceId
+	metricDataQueries := []types.MetricDataQuery{}
+	
+	//generate list of individual metric requests
+	for i := 0; i < len(ids); i++ {
+		metricDataQueries = append(metricDataQueries, ConstructMetricDataQuery(ids[i], Namespace, DimensionName, dimensionValue, metricNames[i], timeDiff))
+	}
```
SaxyPandaBear, line 46: The `range` keyword returns the index and the value so we can change the for loop here, though I guess I'm not too picky in this scenario since we need to use the index twice.
```suggestion
	for i, id := range ids {
		metricDataQueries = append(metricDataQueries, ConstructMetricDataQuery(id, Namespace, DimensionName, dimensionValue, metricNames[i], timeDiff))
	}
```

integration/test/performance_tracker/performance_query_utils.go
```diff
@@ -0,0 +1,78 @@
...
+		EndTime:   aws.Time(time.Unix(time.Now().Unix(), 0)),
+		StartTime: aws.Time(time.Unix(time.Now().Add(time.Duration(-timeDiff)*time.Minute).Unix(), 0)),
+		MetricDataQueries: metricDataQueries,
+	}
+
+	return input, nil
+}
+
+// ConstructMetricDataQuery is a helper function for GenerateGetMetricInputStruct and constructs individual metric requests
+func ConstructMetricDataQuery(id string, namespace string, dimensionName string, dimensionValue string, metricName string, timeDiff int) (types.MetricDataQuery) {
```
SaxyPandaBear, line 58: Just pointing out that I think this function is fine, because it serves a different purpose than the `GetMetrics` function that I already called out as being unnecessary. This one is fine because it's taking a bunch of raw inputs and transforming it into a payload that we would then use for our request to the API. I am fine with it. Though, nitpick - we can consolidate the types
```suggestion
func ConstructMetricDataQuery(id, namespace, dimensionName, dimensionValue, metricName string, timeDiff int) (types.MetricDataQuery) {
```

SaxyPandaBear: That being said, I think that this might change a little if you move to GetMetricStatistics instead of GetMetricData.

integration/test/performance_tracker/performance_query_utils.go
```diff
@@ -0,0 +1,78 @@
...
+	Namespace = "CWAgent"
+	DimensionName = "InstanceId"
+	Stat = "Average"
+	Period = 30
+)
+
+// CWGetMetricDataAPI defines the interface for the GetMetricData function
+type CWGetMetricDataAPI interface {
+	GetMetricData(ctx context.Context, params *cloudwatch.GetMetricDataInput, optFns ...func(*cloudwatch.Options)) (*cloudwatch.GetMetricDataOutput, error)
+}
```
SaxyPandaBear, line 23: I don't think we need t his. We can just invoke the SDK client directly and cut out needing to define this interface just to pass the client in to the `GetMetrics` function.

integration/test/performance_tracker/get_performance_metrics.go
```diff
@@ -0,0 +1,53 @@
...
+
+	client := cloudwatch.NewFromConfig(cfg)
+
+	//declare metrics you want to gather from cloudwatch agent
+	ids := []string{"m1", "m2"}
+	metricNames := []string{"procstat_cpu_usage", "procstat_memory_rss"}
+
+	//give a 30 second buffer before metrics collection to allow for agent startup
+	runtimeSeconds -= 30
+	input, err := GenerateGetMetricInputStruct(ids, metricNames, instanceId, runtimeSeconds)
```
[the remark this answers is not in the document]
gmealy1: This would disregard the first 30 seconds after the agent starts. Do you think it is unnecessary and we should just collect everything?

integration/test/performance_tracker/performance_metrics_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+	instanceId := test.GetInstanceId()
+	log.Println("Instance ID used for performance metrics : %s", instanceId)
+
+	test.CopyFile(configPath, configOutputPath)
+
+	test.StartAgent(configOutputPath, true)
+
+	//let agent run before collecting performance metrics on it
+	time.Sleep(agentRuntime)
```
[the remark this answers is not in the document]
gmealy1: This was just to get started and get it working properly first

integration/test/performance_tracker/get_performance_metrics.go
```diff
@@ -0,0 +1,53 @@
...
+	cfg, err := config.LoadDefaultConfig(context.TODO(), config.WithRegion(region))
+	if err != nil {
+		return err
+	}
+
+	client := cloudwatch.NewFromConfig(cfg)
+
+	//declare metrics you want to gather from cloudwatch agent
+	ids := []string{"m1", "m2"}
+	metricNames := []string{"procstat_cpu_usage", "procstat_memory_rss"}
```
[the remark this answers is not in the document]
gmealy1: Changed it so the script grabs this data from the config itself

integration/test/performance_tracker/get_performance_metrics.go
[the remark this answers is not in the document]
gmealy1: Removed it

integration/test/performance_tracker/resources/config.json
```diff
@@ -0,0 +1,50 @@
...
+				"measurement": [
+					"mem_used_percent"
+				],
+				"metrics_collection_interval": 60
+			},
+			"statsd": {
+				"metrics_aggregation_interval": 60,
+				"metrics_collection_interval": 10,
+				"service_address": ":8125"
+			},
```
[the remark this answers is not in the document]
gmealy1: Removed this from config

integration/test/performance_tracker/get_performance_metrics.go
```diff
@@ -0,0 +1,53 @@
...
+	"encoding/json"
+
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/service/cloudwatch"
+)
+
+//integration tests run on us-west-2
+const region = "us-west-2"
+
+func GetPerformanceMetrics(instanceId string, runtimeSeconds int) (error) {
```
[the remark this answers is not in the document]
gmealy1: Consolidated into one file and made GetMetrics return the json data to the test function

integration/test/performance_tracker/performance_query_utils.go
gmealy1: Assigned it to a variable and use that

integration/test/performance_tracker/performance_query_utils.go
gmealy1: removed

integration/test/performance_tracker/performance_metrics_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"time"
+	"log"
+
+	"github.com/aws/amazon-cloudwatch-agent/integration/test"
+)
+
+const (
+	configPath = "resources/config.json"
+	configOutputPath = "/opt/aws/amazon-cloudwatch-agent/bin/config.json"
+	agentRuntime = 20 //minutes
```
SaxyPandaBear, line 17: I think that rather than doing ` // minutes` as a comment, we could update the variable name to reflect that the units is minutes. But trying to reason what to change the name to, it feels more awkward than leaving this. Idk. I'll think it over but not something I would block review on.

integration/test/performance_tracker/performance_metrics_test.go
```diff
@@ -0,0 +1,46 @@
...
+	instanceId := test.GetInstanceId()
+	log.Printf("Instance ID used for performance metrics : %s\n", instanceId)
+
+	test.CopyFile(configPath, configOutputPath)
+
+	test.StartAgent(configOutputPath, true)
+
+	//let agent run before collecting performance metrics on it
+	time.Sleep(agentRuntime * time.Minute)
+	log.Printf("Agent has been running for : %s\n", (agentRuntime * time.Minute).String())
```
SaxyPandaBear, line 31: We can define the `time.Duration` value once rather than calculating it twice
```suggestion
        agentRuntimeMin := agentRuntime * time.Minute
	time.Sleep(agentRuntimeMin)
	log.Printf("Agent has been running for : %s\n", agentRuntimeMin.String())
```

integration/test/performance_tracker/performance_query_utils.go
```diff
@@ -0,0 +1,143 @@
...
+)
+
+const (
+	Namespace = "CWAgent"
+	DimensionName = "InstanceId"
+	Stat = "Average"
+	Period = 30
+)
+// GenerateGetMetricInputStruct generates the struct required to make a query request to cloudwatch's GetMetrics
+func GenerateGetMetricInputStruct(ids, metricNames []string, instanceId string, timeDiff int) (*cloudwatch.GetMetricDataInput, error) {
```
SaxyPandaBear, line 24: From what I can tell, this is using GetMetricData and #497 is expecting that. I think the end result will make more sense to call GetMetricStatistics here, but it's okay to have this as is for now while you both explore how interacting with GetMetricStatistics would work. It's possible I'm just wrong and GetMetricStatistics doesn't help at all

integration/test/performance_tracker/performance_query_utils.go
```diff
@@ -0,0 +1,143 @@
...
+	//use the index to get the rest of the path
+	metricList := procstatList[cloudwatchIndex].(map[string]interface{})["measurement"].([]interface{})
+
+	//convert the resulting []interface{} to []string and create matching metric ids for each one
+	metricNames := make([]string, len(metricList))
+	ids := make([]string, len(metricList))
+	for i, metricName := range metricList {
+		metricNames[i] = "procstat_" + metricName.(string)
+		ids[i] = fmt.Sprint("m", i + 1)
+	}
```
SaxyPandaBear, line 122: I think this should be a separate function. There's a lot of logic in here, and hard coding the path to the config makes it harder to extend this to other configurations in the future (not by a lot but it's an easy change to add now).

integration/test/performance_tracker/performance_query_utils.go
```diff
@@ -0,0 +1,143 @@
...
+
+	//make input struct
+	input, err := GenerateGetMetricInputStruct(ids, metricNames, instanceId, agentRuntime)
+	if err != nil {
+		return nil, err
+	}
+
+	//call to cloudwatch agent API
```
SaxyPandaBear, line 130: nitpick, the agent doesn't have an API. The agent is just a normal client of CloudWatch like everyone else
```suggestion
	//call to CloudWatch API
```

integration/test/performance_tracker/performance_query_utils.go
```diff
@@ -0,0 +1,143 @@
...
+			Period: aws.Int32(int32(Period)),
+			Stat:   aws.String(Stat),
+		},
+	}
+
+	return query
+}
+
+func GetPerformanceMetrics(instanceId string, agentRuntime int) ([]byte, error) {
+	agentContext := context.TODO()
```
SaxyPandaBear, line 75: So, for how I've seen contexts used, they are usually created at the very beginning of whatever the main driver is (in this case, your integration test, and then passed to the functions that need a context. I would change this to accept a context.Context as an input, and update the test code to create the context there and pass it in.

integration/test/performance_tracker/performance_query_utils.go
```diff
@@ -0,0 +1,143 @@
...
+			},
+			Period: aws.Int32(int32(Period)),
+			Stat:   aws.String(Stat),
+		},
+	}
+
+	return query
+}
+
+func GetPerformanceMetrics(instanceId string, agentRuntime int) ([]byte, error) {
```
SaxyPandaBear, line 74: Any strong reason why we return a byte array here?

integration/test/performance_tracker/performance_metrics_test.go
```diff
@@ -0,0 +1,46 @@
+//go:build linux && integration
+// +build linux,integration
+
+package data_collector
```
SaxyPandaBear, line 4: I missed this originally:
1. The package name shouldn't have underscores: https://go.dev/blog/package-names
2. The package name makes sense to follow the directory name: https://go.dev/doc/effective_go#package-names
3. This is not something that _needs_ to be changed now, but I think it would make sense to change the directory name to `./integration/test/performance_test/`, and would be easier to change it sooner rather than later. 
```suggestion
package performancetest
```

integration/test/performance_tracker/resources/config.json
```diff
@@ -0,0 +1,45 @@
+{
```
SaxyPandaBear, line 1: One thing to consider going forward is if there is going to be collision when trying to get metrics from different performance tests that run concurrently. 

I ran into an issue when writing the CloudWatch Logs integration tests because I reused the log group and log stream name across all of the tests, so I had issues with getting consistent results from assertions because I was picking up data from multiple tests at the same time by accident. 

Not something that needs to be addressed right now, but should be something looked into in the next few weeks. The main focus is getting some performance test running and persisting data.

integration/test/performance_tracker/performance_query_utils.go
gmealy1: I will look more into GetMetricStatistics

integration/test/performance_tracker/performance_query_utils.go
gmealy1: It's returning a byte array because that is what json.MarshalIndent returns and I believe that is what the database api part takes as an initial input

SaxyPandaBear (approved):

PR #500, opened 2022-06-24
Support Debian ARM64 Integration Test

khanhntd:
# Description of the issue
Currently, CloudWatchAgent's binaries are able to compile on ARM OS. However, we would lack a mechanism to confirm if its able to run on those platform. Therefore, adding Debian ARM64 Integration Test for the mechanism. 

# Description of changes
There are two changes:
* Add generate IAM, SG, Private Key  automatically with terraform instead of manually create it beforehand.
* Add Debian ARM64 for integration test. 

# License

# Tests
By running `integration-test` on my fork, here is the result: https://github.com/khanhntd/amazon-cloudwatch-agent/actions/runs/2557741979

# Requirements
2. Run `make linter`


.github/workflows/integrationTest.yml
```diff
@@ -381,13 +380,9 @@ jobs:
         run: >
           echo run terraform and execute test code &&
           terraform apply --auto-approve
-          -var="ssh_key=${PRIVATE_KEY}"
```
sethAmazon, line 384: The reason this is hard coded is so we can ssh into env if need to

khanhntd: There would be a bottle neck with that case (e.g lifecycle of terraform resources - destroyed before able to ssh). Therefore, I would encourage dev to ssh to the env locally instead of racing against integration-test and also encourage people to less depends on these hard-code environment if they only want to test only (avoid setup up too much resources).

integration/generator/resources/ec2_linux_test_matrix.json
```diff
@@ -2,14 +2,27 @@
...
     "ami": "cloudwatch-agent-integration-test-ubuntu*",
     "caCertPath": "/etc/ssl/certs/ca-certificates.crt",
-    "arc": "amd64", "binaryName": "amazon-cloudwatch-agent.deb"
+    "arc": "amd64",
+    "binaryName": "amazon-cloudwatch-agent.deb"
+  },
+  {
+    "os": "debian-10-arm64",
+    "username": "admin",
+    "instanceType": "c6g.xlarge",
```
SaxyPandaBear, line 15: Why do you choose this instance type?

integration/generator/resources/ec2_linux_test_matrix.json
```diff
@@ -2,14 +2,27 @@
   {
     "os": "ubuntu",
     "username": "ubuntu",
+    "instanceType":"t3a.xlarge",
```
SaxyPandaBear, line 5: Why did you choose this instance type?

integration/terraform/ec2/linux/main.tf
```diff
@@ -1,18 +1,58 @@
...
   provisioner "remote-exec" {
     inline = [
       "cloud-init status --wait",
       "echo clone and install agent",
-      "export PATH=$PATH:/usr/local/go/bin",
       "git clone ${var.github_repo}",
       "cd amazon-cloudwatch-agent",
       "git reset --hard ${var.github_sha}",
       "aws s3 cp s3://${var.s3_bucket}/integration-test/binary/${var.github_sha}/linux/${var.arc}/${var.binary_name} .",
+      "sleep 10",
```
SaxyPandaBear, line 55: Why do we need to sleep? The `aws s3 cp` command is going to block until it completes.

integration/terraform/ec2/linux/main.tf
```diff
@@ -21,23 +61,37 @@ resource "aws_instance" "integration-test" {
...
+  }
+
+  #Run sanity check and integration test
+  provisioner "remote-exec" {
+    inline = [
+      "echo prepare environment",
       "export LOCAL_STACK_HOST_NAME=${var.local_stack_host_name}",
       "export AWS_REGION=${var.region}",
-      "echo run tests with the tag integration, one at a time, and verbose",
+      "export PATH=$PATH:/snap/bin:/usr/local/go/bin",
```
SaxyPandaBear, line 80: Why are we adding /snap/bin?

integration/terraform/ec2/linux/main.tf
```diff
@@ -21,23 +61,37 @@ resource "aws_instance" "integration-test" {
...
+      "echo run integration test",
       "cd ~/amazon-cloudwatch-agent",
-      "echo run sanity test && go test ./integration/test/sanity -p 1 -v --tags=integration",
+      "go test ./integration/test/sanity -p 1 -v --tags=integration",
```
SaxyPandaBear, line 83: Why are we introducing this change?

integration/terraform/ec2/linux/main.tf
```diff
@@ -1,18 +1,58 @@
...
+resource "aws_key_pair" "aws_ssh_key" {
+  count      = var.ssh_key == "" ? 1 : 0
+  key_name   = "ec2-key-pair-${random_id.testing_id.hex}"
+  public_key = tls_private_key.ssh_key[0].public_key_openssh
+}
+
+locals {
+  ssh_key_name        = var.ssh_key != "" ? var.key_name : aws_key_pair.aws_ssh_key[0].key_name
+  private_key_content = var.ssh_key != "" ? var.ssh_key : tls_private_key.ssh_key[0].private_key_pem
+}
```
SaxyPandaBear, line 27: Is this a necessary part of the change

integration/generator/resources/ec2_linux_test_matrix.json
```diff
@@ -2,14 +2,27 @@
...
     "installAgentCommand": "sudo dpkg -i -E ./amazon-cloudwatch-agent.deb",
     "ami": "cloudwatch-agent-integration-test-ubuntu*",
     "caCertPath": "/etc/ssl/certs/ca-certificates.crt",
-    "arc": "amd64", "binaryName": "amazon-cloudwatch-agent.deb"
+    "arc": "amd64",
+    "binaryName": "amazon-cloudwatch-agent.deb"
+  },
+  {
+    "os": "debian-10-arm64",
```
SaxyPandaBear, line 13: Why do we have to specify `arm64` in the os name here if the `arc` value already tells us that?

integration/generator/resources/ec2_linux_test_matrix.json
khanhntd: Debian for ARM does not support the default one `t3a.xlarge`. Therefore, I have to choose a similar one with sufficient hardware as possible.

integration/generator/resources/ec2_linux_test_matrix.json
khanhntd: This instance type is the [default one](https://github.com/aws/amazon-cloudwatch-agent/blob/master/integration/terraform/ec2/linux/variables.tf#L1-L4) for Linux AMD.

integration/generator/resources/ec2_linux_test_matrix.json
khanhntd: Mainly because of this reason: 
* The [cache key](https://github.com/aws/amazon-cloudwatch-agent/blob/master/.github/workflows/integrationTest.yml#L419) is using `os` instead of aarch ( we can change that). Therefore, if there is a similar os but in different aarch, it would return a similar key when one of the test executes successfully and return a false positive.

integration/terraform/ec2/linux/main.tf
khanhntd: Since the components to install go is snap instead of usual go (same as ubuntu). Therefore, we need to export the path `/snap/bin` for the system to know where to use go.

integration/terraform/ec2/linux/main.tf
khanhntd: IMO, not entirely. However, this help me reduces the necessary resources to create for when using integration test.

integration/terraform/ec2/linux/main.tf
khanhntd: The `sleep 10` is not entirely for `aws s3 cp`, mostly it is used for avoid `"sudo ${var.install_agent}"` because of  `dpkg frontend lock is locked by another process` (https://github.com/khanhntd/amazon-cloudwatch-agent/runs/6841609674?check_suite_focus=true)

integration/terraform/ec2/linux/main.tf
khanhntd: No deep reason behind it. Only format change and for cleaner code (separating stage).

integration/generator/resources/ec2_linux_test_matrix.json
SaxyPandaBear: Then doesn't it make more sense to update the cache key instead of this?

.github/workflows/integrationTest.yml
SaxyPandaBear: I disagree. The SSH key should not be maintained by Terraform. If we want to ssh onto a host in order to troubleshoot, all you need to do is comment out the destroy call.

integration/generator/resources/ec2_linux_test_matrix.json
khanhntd: Yes. I have changed the [cache key](https://github.com/aws/amazon-cloudwatch-agent/pull/500/files#diff-9b9f9a38b889c9dcb55ba0330baa69c8fdd62ac1100b6eb59b6fc44008f5d320R414) in the latest commit 👍

integration/terraform/ec2/linux/main.tf
SaxyPandaBear: Wasn't this part of the install_agent var before?

.github/workflows/integrationTest.yml
khanhntd: Should we have `comment out the destroy call` as a commit though? Moreover, I'm not sure about the reason `ssh onto a host` too since Terraform does help us to `ssh onto a host`.

SaxyPandaBear: If we want to troubleshoot an issue on an integration test host, making the SSH key managed by terraform means we don't have any access to SSH onto the host. We would not be commenting out the terraform destroy and merging it to the main branch. It would be for troubleshooting on your fork, or in a branch on the main repo. Please revert the ssh key change. This takes away operational value.

khanhntd: IMO, Terraform does indeed support `ssh onto a host` by using their creation of SSH key. Therefore, I would believe it still keeps the operational value (so I'm somehow still indecisive by this)

integration/terraform/ec2/linux/main.tf
khanhntd: Maybe I'm misunderstanding the questions but I don't think the `sleep` is part of the `install_agent` var before the changes

integration/terraform/ec2/linux/main.tf
SaxyPandaBear: We discussed separately about why the ssh key needs to be separate. We should revert this change

integration/terraform/ec2/linux/main.tf
SaxyPandaBear: @sethAmazon can you weigh in on this? I swore we accounted for the dpkg lock somewhere before previously

SaxyPandaBear: Ultimately I think it's fine, though 10 seconds seems like a lot of time to wait.

integration/terraform/ec2/linux/main.tf
khanhntd: I think its happen at GitHub Action level. If you decided not to use it locally, then its bring no harm. Moreover, if you decided not to use `ssh key` at Github Action, that should be fine too.

SaxyPandaBear (approved):
Assuming that this doesn't break our existing setup for the ssh key, lgtm

.github/workflows/integrationTest.yml
```diff
@@ -640,16 +636,16 @@ jobs:
...
-        id: sha
-        run: echo "::set-output name=sha_short::$(git rev-parse --short HEAD)"
+
       - name: Get git date
         id: sha_date
         run: echo "::set-output name=sha_date::$(git show -s --format=%ct ${{ steps.sha.outputs.sha_short }} )"
+
       - name: Check env
-        run: echo "SHA ${{ steps.sha.outputs.sha_short }} | Date ${{ steps.sha_date.outputs.sha_date }} "
+        run: echo "SHA ${{ GITHUB_SHA }} | Date ${{ steps.sha_date.outputs.sha_date }} "
```
SaxyPandaBear, line 648: Can you explain why you're changing this code that is unrelated to your PR?

integration/terraform/ec2/linux/variables.tf
```diff
@@ -1,98 +1,85 @@
...
   default = ""
 }
-variable "sha_date"{
+
+variable "github_repo" {
+  type    = string
+  default = ""
+}
+
+variable "github_sha_date"{
```
SaxyPandaBear, line 82: What are these changes for

khanhntd: Changing from "sha_date" to "github_sha_date" to grouping the "Github variables" and also delete the `sha` which is a shorter version of "github_sha" when considering the impacts as can be found in this comment https://github.com/aws/amazon-cloudwatch-agent/pull/500#discussion_r915235575

SaxyPandaBear (approved):

PR #501, opened 2022-06-24
Fix CloudWatch.publish() sometimes not jittering.

adam-mateen:
# Description of the issue
When the agent is configured to monitor StatsD, and there is a burst of 10,000 metrics collected and aggregated every minute (same issue if it is done every 5 minutes), the agent's internal buffer of batches will fill up and it will immediately try pushing to CloudWatch. And as soon as that is published the batch will fill up again, and the agent will push again, and again...

This becomes problematic if there are _many_ hosts, because they will all try pushing during the first second of each minute, and will repeat with little to no delay.

Inevitably this will reach the PutMetricData API throttle limit of 150 transaction per second.
 
Change CloudWatch.publish() to avoid bursting backend when datumBatchChan is full.
Change CloudWatch.backoffSleep() to use jitter.
Update tests.

# Description of changes
1. Change API retries to use randomized jitter.
2. Change what happens when the buffer of metricBatches is full. Instead of publishing immediately, gradually reduce the delay. Please note there are other buffers in the agent, so even if the buffer of metricBatches remains full for _awhile_ it doesn't mean metric will be dropped.

# License

# Tests
1. New test - `go test -v ./plugins/outputs/cloudwatch -count=1 -run TestPublish`
2. All tests - `make test`
3. Ran this python script against the current agent, and with my change. Verified the burst of uploads is reduced to a gradually more frequent set of uploads.
```
import statsd
import os
import sys
import time

"""
Each API call can include 20 metrics.
So a metricBatch hold 20 metrics.
CWA has a buffer/channel for these batches with a size of 50.
So it takes 50 batches of 20 metrics each to fill the buffer and cause CWA to 
upload  immediately. 1,000 metrics total.

When NUM_METRICS=1000, CWA hits the full condition and makes 50 API calls.
When NUM_METRICS=10,000, CWA makes 500 API calls.

Using the new publish() where the interval is cut in half, 10K metrics takes:
30 sec - first full batchBuffer, so 50*20=1,000 metrics.
15 sec - 1,000 metrics
7.5 sec - 1,000
.
.
.
30 + 15 + 7.5 + ... = 41 seconds for 10 full buffers of 50 batches = 500 API calls.
"""
NUM_METRICS = 10000
VALUES_PER_METRIC = 2
SLEEP_INTERVAL = 60 / VALUES_PER_METRIC

def send_metrics(statsd_client):
    for i in range(NUM_METRICS):
        name = "adam" + str(i)
        statsd_client.incr(name)

def main():
    statsd_client = statsd.StatsClient("localhost", 8125)
    count = 0
    while 1:
        count += 1
        print("loop", count)
        send_metrics(statsd_client)
        time.sleep(SLEEP_INTERVAL)
        
if __name__ == "__main__":
    main()
```

CWA config:
```
{
	"metrics": {
		"metrics_collected": {
			"statsd": {
				"metrics_aggregation_interval": 60,
				"metrics_collection_interval": 10,
				"service_address": ":8125"
			}
		}
	}
}
```

# Requirements
4. Run `make linter`

NOTE: I am skipping these requirements because it would seriously clutter this PR.
If required, I can do it in a separate PR.

### DIFF
--- a/plugins/outputs/cloudwatch/aggregator_test.go
+++ b/plugins/outputs/cloudwatch/aggregator_test.go
@@ -39,6 +39,8 @@ func TestAggregator_NoAggregationKeyFound(t *testing.T) {
 
 	assertNoMetricsInChan(t, metricChan)
 	close(shutdownChan)
+	// Cleanup
+	wg.Wait()
 }
 
 func TestAggregator_NotDurationType(t *testing.T) {
@@ -60,6 +62,8 @@ func TestAggregator_NotDurationType(t *testing.T) {
 
 	assertNoMetricsInChan(t, metricChan)
 	close(shutdownChan)
+	// Cleanup
+	wg.Wait()
 }
 
 func TestAggregator_ProperAggregationKey(t *testing.T) {
@@ -78,6 +82,8 @@ func TestAggregator_ProperAggregationKey(t *testing.T) {
 
 	assertNoMetricsInChan(t, metricChan)
 	close(shutdownChan)
+	// Cleanup
+	wg.Wait()
 }
 
 func TestAggregator_MultipleAggregationPeriods(t *testing.T) {
@@ -121,6 +127,8 @@ func TestAggregator_MultipleAggregationPeriods(t *testing.T) {
 
 	assertNoMetricsInChan(t, metricChan)
 	close(shutdownChan)
+	// Cleanup
+	wg.Wait()
 }
 
 func TestAggregator_ShutdownBehavior(t *testing.T) {
--- a/plugins/outputs/cloudwatch/cloudwatch.go
+++ b/plugins/outputs/cloudwatch/cloudwatch.go
@@ -5,7 +5,6 @@ package cloudwatch
 
 import (
 	"log"
-	"math"
 	"reflect"
 	"runtime"
 	"sort"
@@ -290,7 +289,6 @@ func getFirstPushMs(interval time.Duration) int64 {
 	if nextMs < nowMs {
 		nextMs += interval.Milliseconds()
 	}
-	log.Printf("I! cloudwatch: next - now, %v", nextMs - nowMs)
 	return nextMs
 }
 
@@ -322,6 +320,7 @@ func (c *CloudWatch) publish() {
 		nowMs := time.Now().UnixMilli()
 
 		if c.metricDatumBatchFull() {
+			log.Printf("I! cloudwatch: buffer of batches is full")
 			if !bufferFullOccurred {
 				// Set to false so this only happens once per push.
 				bufferFullOccurred = true
@@ -337,7 +336,7 @@ func (c *CloudWatch) publish() {
 		// Check if the interval has elapsed.
 		if nowMs >= nextMs {
 			shouldPublish = true
-			nextMs = nowMs + currentInterval.Milliseconds()
+			nextMs += currentInterval.Milliseconds()
 			// Restore interval if buffer did not fill up during this interval.
 			if !bufferFullOccurred {
 				log.Printf("I! cloudwatch: reset interval")
@@ -360,6 +359,7 @@ func (c *CloudWatch) metricDatumBatchFull() bool {
 }
 
 func (c *CloudWatch) pushMetricDatumBatch() {
+	log.Printf("I! cloudwatch: push batches")
 	for {
 		select {
 		case datumBatch := <-c.datumBatchChan:
@@ -371,16 +371,17 @@ func (c *CloudWatch) pushMetricDatumBatch() {
 	}
 }
 
-// backoffSleep sleeps some amount of time based on number of retries already done.
+// backoffSleep sleeps some amount of time based on number of retries done.
 func (c *CloudWatch) backoffSleep() {
 	var backoffInMillis int64 = 60 * 1000 // 1 minute
 	if c.retries <= defaultRetryCount {
-		backoffInMillis = int64(backoffRetryBase * math.Pow(2, float64(c.retries)))
+		backoffInMillis = int64(backoffRetryBase * 1 << c.retries)
 		// Adding at most 1 second on each retry for the sake of jitter.
 		backoffInMillis += publisJitterInt(1000)
 	}
 	sleepDuration := time.Millisecond * time.Duration(backoffInMillis)
-	log.Printf("W! %v retries, going to sleep %v before retrying.", c.retries, sleepDuration)
+	log.Printf("W! %v retries, going to sleep %v before retrying.", c.retries,
+		sleepDuration)
 	c.retries++
 	time.Sleep(sleepDuration)
 }
@@ -393,7 +394,6 @@ func (c *CloudWatch) WriteToCloudWatch(req interface{}) {
 	}
 	var err error
 	for i := 0; i < defaultRetryCount; i++ {
-		//log.Printf("I! cloudwatch: calling PutMetricData(), len %v", len(datums))
 		_, err = c.svc.PutMetricData(params)
 
 		if err != nil {
--- a/plugins/outputs/cloudwatch/cloudwatch.go
+++ b/plugins/outputs/cloudwatch/cloudwatch.go
@@ -300,7 +300,6 @@ func getFirstPushMs(interval time.Duration) int64 {
 func (c *CloudWatch) publish() {
 	currentInterval := c.ForceFlushInterval.Duration
 	nextMs := getFirstPushMs(currentInterval)
-	// Only allow shortening interval once per push.
 	bufferFullOccurred := false
 
 	for {
@@ -318,7 +317,7 @@ func (c *CloudWatch) publish() {
 
 		if c.metricDatumBatchFull() {
 			if !bufferFullOccurred {
-				// Set to false so this only happens once per push.
+				// Set to true so this only happens once per push.
 				bufferFullOccurred = true
 				// Keep interval above 1 second.
 				if currentInterval.Seconds() >=  2 {
@@ -375,7 +374,7 @@ func (c *CloudWatch) backoffSleep() {
 		backoffInMillis += getJitter(1000)
 	}
 	sleepDuration := time.Millisecond * time.Duration(backoffInMillis)
-	log.Printf("W! %v retries, going to sleep %v before retrying.", c.retries,
+	log.Printf("W! cloudwatch: %v retries, going to sleep %v before retrying.", c.retries,
 		sleepDuration)
 	c.retries++
 	time.Sleep(sleepDuration)
@@ -394,13 +393,13 @@ func (c *CloudWatch) WriteToCloudWatch(req interface{}) {
 		if err != nil {
 			awsErr, ok := err.(awserr.Error)
 			if !ok {
-				log.Printf("E! Cannot cast PutMetricData error %v into awserr.Error.", err)
+				log.Printf("E! cloudwatch: Cannot cast PutMetricData error %v into awserr.Error.", err)
 				c.backoffSleep()
 				continue
 			}
 			switch awsErr.Code() {
 			case cloudwatch.ErrCodeLimitExceededFault, cloudwatch.ErrCodeInternalServiceFault:
-				log.Printf("W! cloudwatch PutMetricData, error: %s, message: %s",
+				log.Printf("W! cloudwatch: PutMetricData, error: %s, message: %s",
 					awsErr.Code(),
 					awsErr.Message())
 				c.backoffSleep()
@@ -416,7 +415,7 @@ func (c *CloudWatch) WriteToCloudWatch(req interface{}) {
 		break
 	}
 	if err != nil {
-		log.Println("E! WriteToCloudWatch failure, err: ", err)
+		log.Println("E! cloudwatch: WriteToCloudWatch failure, err: ", err)
 	}
 }
 
--- a/plugins/outputs/cloudwatch/cloudwatch.go
+++ b/plugins/outputs/cloudwatch/cloudwatch.go
@@ -292,10 +292,10 @@ func getFirstPushMs(interval time.Duration) int64 {
 	return nextMs
 }
 
-// publish loops forever until a shutdown occurs.
+// publish loops until a shutdown occurs.
 // It periodically tries pushing batches of metrics (if there are any).
 // If thet batch buffer fills up the interval will be gradually reduced to avoid
-// bursting the backend and maintain some amount of jitter.
+// many agents bursting the backend.
 func (c *CloudWatch) publish() {
 	currentInterval := c.ForceFlushInterval.Duration
 	nextMs := getFirstPushMs(currentInterval)
@@ -318,17 +318,18 @@ func (c *CloudWatch) publish() {
 			if !bufferFullOccurred {
 				// Set to true so this only happens once per push.
 				bufferFullOccurred = true
-				// Keep interval above 1 second.
-				if currentInterval.Seconds() >=  2 {
-					// Cut the next interval in half.
+				// Keep interval above above 1 second.
+				if currentInterval.Seconds() >  1 {
 					currentInterval /= 2
+					if currentInterval.Seconds() < 1 {
+						currentInterval = 1 * time.Second
+					}
 					// Cut the remaining interval in half.
 					nextMs = nowMs + ((nextMs - nowMs) / 2)
 				}
 			}
 		}
 
-		// Check if the interval has elapsed.
 		if nowMs >= nextMs {
 			shouldPublish = true
 			// Restore interval if buffer did not fill up during this interval.
--- a/plugins/outputs/cloudwatch/cloudwatch.go
+++ b/plugins/outputs/cloudwatch/cloudwatch.go
@@ -638,7 +638,9 @@ func (c *CloudWatch) ProcessRollup(rawDimension []*cloudwatch.Dimension) [][]*cl
 		}
 
 	}
-	log.Printf("D! cloudwatch: Get Full dimensionList %v", fullDimensionsList)
+	if len(fullDimensionsList) > 0 && len(fullDimensionsList[0]) > 0 {
+		log.Printf("D! cloudwatch: Get Full dimensionList %v", fullDimensionsList)
+	}
 	return fullDimensionsList
 }
 
--- a/plugins/outputs/cloudwatch/cloudwatch.go
+++ b/plugins/outputs/cloudwatch/cloudwatch.go
@@ -292,18 +292,15 @@ func getFirstPushMs(interval time.Duration) int64 {
 	return nextMs
 }
 
-// publish will begin by sleeping for a random amount of jitter between 0 and
-// the configured forceFlushInterval. Previously, if the batch buffer filled up
-// it flushed immediately. Now it cuts the remaining interval in half,
-// and the next interval in half.
-// If the buffer continues to stay full then it will keep cutting down.
-// This avoids bursting the backend and maintains some amount of jitter.
+// publish loops forever until a shutdown occurs.
+// It periodically tries pushing batches of metrics (if there are any).
+// Previously, if the batch buffer filled up it flushed immediately.
+// Now it gradually reduces the interval to avoid bursting the backend and
+// maintain some amount of jitter.
 func (c *CloudWatch) publish() {
 	currentInterval := c.ForceFlushInterval.Duration
 	nextMs := getFirstPushMs(currentInterval)
 	// Only allow shortening interval once per push.
-	// This avoids bursting many pushes all at once, but still allows the
-	// push frequency to increase.
 	bufferFullOccurred := false
 
 	for {
@@ -320,15 +317,15 @@ func (c *CloudWatch) publish() {
 		nowMs := time.Now().UnixMilli()
 
 		if c.metricDatumBatchFull() {
-			log.Printf("I! cloudwatch: buffer of batches is full")
 			if !bufferFullOccurred {
 				// Set to false so this only happens once per push.
 				bufferFullOccurred = true
 				// Keep interval above 1 second.
 				if currentInterval.Seconds() >=  2 {
+					// Cut the next interval in half.
 					currentInterval /= 2
+					// Cut the remaining interval in half.
 					nextMs = nowMs + ((nextMs - nowMs) / 2)
-					log.Printf("I! cloudwatch: cut interval to %v", currentInterval)
 				}
 			}
 		}
@@ -336,12 +333,11 @@ func (c *CloudWatch) publish() {
 		// Check if the interval has elapsed.
 		if nowMs >= nextMs {
 			shouldPublish = true
-			nextMs += currentInterval.Milliseconds()
 			// Restore interval if buffer did not fill up during this interval.
 			if !bufferFullOccurred {
-				log.Printf("I! cloudwatch: reset interval")
 				currentInterval = c.ForceFlushInterval.Duration
 			}
+			nextMs += currentInterval.Milliseconds()
 		}
 
 		if shouldPublish {
@@ -359,7 +355,6 @@ func (c *CloudWatch) metricDatumBatchFull() bool {
 }
 
 func (c *CloudWatch) pushMetricDatumBatch() {
-	log.Printf("I! cloudwatch: push batches")
 	for {
 		select {
 		case datumBatch := <-c.datumBatchChan:
--- a/plugins/outputs/cloudwatch/cloudwatch.go
+++ b/plugins/outputs/cloudwatch/cloudwatch.go
@@ -38,7 +38,7 @@ const (
 	pushIntervalInSec              = 60 // 60 sec
 	highResolutionTagKey           = "aws:StorageResolution"
 	defaultRetryCount              = 5 // this is the retry count, the total attempts would be retry count + 1 at most.
-	backoffRetryBase               = 200
+	backoffRetryBase               = 200 * time.Millisecond
 	MaxDimensions                  = 30
 )
 
@@ -294,9 +294,8 @@ func getFirstPushMs(interval time.Duration) int64 {
 
 // publish loops forever until a shutdown occurs.
 // It periodically tries pushing batches of metrics (if there are any).
-// Previously, if the batch buffer filled up it flushed immediately.
-// Now it gradually reduces the interval to avoid bursting the backend and
-// maintain some amount of jitter.
+// If thet batch buffer fills up the interval will be gradually reduced to avoid
+// bursting the backend and maintain some amount of jitter.
 func (c *CloudWatch) publish() {
 	currentInterval := c.ForceFlushInterval.Duration
 	nextMs := getFirstPushMs(currentInterval)
@@ -343,8 +342,12 @@ func (c *CloudWatch) publish() {
 			c.pushMetricDatumBatch()
 			bufferFullOccurred = false
 		}
-		// Always sleep, so the fastest we push is every 1 second.
-		time.Sleep(time.Second)
+		// Sleep 1 second, unless the nextMs is less than a second away.
+		if nextMs - nowMs > time.Second.Milliseconds() {
+			time.Sleep(time.Second)
+		} else {
+			time.Sleep(time.Duration(nextMs - nowMs) * time.Millisecond)
+		}
 	}
 }
 
@@ -367,17 +370,15 @@ func (c *CloudWatch) pushMetricDatumBatch() {
 
 // backoffSleep sleeps some amount of time based on number of retries done.
 func (c *CloudWatch) backoffSleep() {
-	var backoffInMillis int64 = 60 * 1000 // 1 minute
+	d := 1 * time.Minute
 	if c.retries <= defaultRetryCount {
-		backoffInMillis = int64(backoffRetryBase * 1 << c.retries)
-		// Adding at most 1 second on each retry for the sake of jitter.
-		backoffInMillis += getJitter(1000)
+		d = backoffRetryBase * time.Duration(1 << c.retries)
 	}
-	sleepDuration := time.Millisecond * time.Duration(backoffInMillis)
-	log.Printf("W! cloudwatch: %v retries, going to sleep %v before retrying.", c.retries,
-		sleepDuration)
+	d = (d / 2) + publishJitter(d / 2)
+	log.Printf("W! cloudwatch: %v retries, going to sleep %v ms before retrying.",
+		c.retries, d.Milliseconds())
 	c.retries++
-	time.Sleep(sleepDuration)
+	time.Sleep(d)
 }
 
 func (c *CloudWatch) WriteToCloudWatch(req interface{}) {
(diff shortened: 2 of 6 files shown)


plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -279,45 +277,86 @@ func (c *CloudWatch) timeToPublish(b *MetricDatumBatch) bool {
...
-	if len(c.datumBatchChan) >= datumBatchChanBufferSize {
-		if len(c.datumBatchFullChan) == 0 {
-			c.datumBatchFullChan <- true
-		}
-		return c.datumBatchFullChan
-	}
-	return nil
+// metricDatumBatchFull returns true if the channel/buffer of batches if full.
+func (c *CloudWatch) metricDatumBatchFull() bool {
+	return len(c.datumBatchChan) >= datumBatchChanBufferSize
```
adam-mateen, line 359: This was not a necessary change.
It is simpler and clearer to just check this condition by calling a function instead of trying to recv from a channel.

plugins/outputs/cloudwatch/cloudwatch_test.go
```diff
@@ -323,45 +323,43 @@ func (svc *mockCloudWatchClient) PutMetricData(input *cloudwatch.PutMetricDataIn
...
 	cloudwatch := &CloudWatch{
 		svc:                svc,
-		ForceFlushInterval: internal.Duration{Duration: 1 * time.Second},
+		ForceFlushInterval: internal.Duration{Duration: forceFlushInterval},
 	}
 	cloudwatch.startRoutines()
 	return cloudwatch
 }
 
-func TestWrite(t *testing.T) {
```
adam-mateen, line 335: In case this diff is confusing, what I did was remove some common code from `TestWrite()` and `TestWriteError()` and put it in a new `makeMetrics()` function. And I use the new function in the new test case `TestPublish()` as well.

plugins/outputs/cloudwatch/util.go
```diff
@@ -46,11 +47,25 @@ const (
...
+// Global set once.
+var doOnce sync.Once
+var gRand *rand.Rand
+
+// publisJitterInt returns a random int64 between 0 and the given maxVal.
+func publisJitterInt(maxVal int64) int64 {
+	// Set seed once at startup
+	doOnce.Do(func() {
+		gRand = rand.New(rand.NewSource(time.Now().Unix()))
+	})
```
adam-mateen, line 59: In the past when using a `rand` like library, I just initialize/seed once.
I am happy to revert to the old way of always seeding on each call.
It wasn't much of an issue before since the function was called once when the CloudWatch output plugin was initialized.

adam-mateen:
This CWA log snippet shows uploads happening at the typical jittered offset, but then there is a burst of collected metrics at the start of the minute.
Notice when the next `push batches` message occurs (the 4th second after the minute, not the 7th).
```
2022-06-24T22:14:07Z I! cloudwatch: push batches
2022-06-24T22:15:07Z I! cloudwatch: push batches
2022-06-24T22:16:07Z I! cloudwatch: push batches
2022-06-24T22:17:07Z I! cloudwatch: push batches
2022-06-24T22:18:07Z I! cloudwatch: push batches
2022-06-24T22:19:07Z I! cloudwatch: push batches
2022-06-24T22:20:07Z I! cloudwatch: push batches
2022-06-24T22:21:07Z I! cloudwatch: push batches
2022-06-24T22:22:00Z I! cloudwatch: buffer of batches is full
2022-06-24T22:22:00Z I! cloudwatch: cut interval to 30s
2022-06-24T22:22:01Z I! cloudwatch: buffer of batches is full
2022-06-24T22:22:02Z I! cloudwatch: buffer of batches is full
2022-06-24T22:22:03Z I! cloudwatch: buffer of batches is full
2022-06-24T22:22:04Z I! cloudwatch: buffer of batches is full
2022-06-24T22:22:04Z I! cloudwatch: push batches
```

plugins/outputs/cloudwatch/util.go
SaxyPandaBear: yeah I don't think we stand to really benefit much from creating a new random every time.

plugins/outputs/cloudwatch/util.go
```diff
@@ -46,11 +47,25 @@ const (
...
-	r := rand.New(rand.NewSource(time.Now().Unix()))
-	jitter := r.Int63n(int64(publishInterval.Seconds()))
-	publishJitter = time.Duration(jitter) * time.Second
-	return
+// Global set once.
+var doOnce sync.Once
+var gRand *rand.Rand
+
+// publisJitterInt returns a random int64 between 0 and the given maxVal.
+func publisJitterInt(maxVal int64) int64 {
```
SaxyPandaBear, line 55: Typo, but maybe a better name would just be to call it something else entirely. At first glance, I thought we were doing another publish function that jitters with an int. But what this really does is it spits out random values to use as the jitter. I'm bad at names but maybe:
```suggestion
func getJitter(maxVal int64) int64 {
```

SaxyPandaBear (commented):
LGTM. I think the only concern I have is around the change with the force flush interval. Is it possible to set up some test hosts and monitor the agent and do some kind of soaking test to see if 1, there is any weird mem/cpu change with the agent, but more importantly 2, does the metric publication change drastically?

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -279,48 +276,90 @@ func (c *CloudWatch) timeToPublish(b *MetricDatumBatch) bool {
...
+// getFirstPushMs returns the time at which the first upload should occur.
+// It uses random jitter as an offset from the start of the given interval.
+func getFirstPushMs(interval time.Duration) int64 {
+	publishJitter := publishJitter(interval)
+	log.Printf("I! cloudwatch: publish with ForceFlushInterval: %v, Publish Jitter: %v",
+		interval, publishJitter)
+	nowMs := time.Now().UnixMilli()
+	// Truncate i.e. round down, then add jitter.
+	// If the rounded down time is in the past, move it forward.
+	nextMs := nowMs - (nowMs % interval.Milliseconds()) + publishJitter.Milliseconds()
```
khanhntd, line 288: Maybe I'm wrong but the first publish will off the the time for `ForceFlushInterval` and will cause wasteful API call if the random falls into first half of ` ForceFlushInterval`. Therefore, would be good to always add `interval.Milliseconds()/2` to allow the Agent to collect more metrics? If it's full then good its not wasteful

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -279,48 +276,90 @@ func (c *CloudWatch) timeToPublish(b *MetricDatumBatch) bool {
...
+				if currentInterval.Seconds() >=  2 {
+					currentInterval /= 2
+					nextMs = nowMs + ((nextMs - nowMs) / 2)
+					log.Printf("I! cloudwatch: cut interval to %v", currentInterval)
+				}
+			}
+		}
+
+		// Check if the interval has elapsed.
+		if nowMs >= nextMs {
```
khanhntd, line 337: If during the time buffer is full and we need the `nowMs >= nextMs`, will the buffer drop any metrics?

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -332,14 +371,17 @@ func (c *CloudWatch) pushMetricDatumBatch() {
 	}
 }
 
-//sleep some back off time before retries.
+// backoffSleep sleeps some amount of time based on number of retries done.
 func (c *CloudWatch) backoffSleep() {
 	var backoffInMillis int64 = 60 * 1000 // 1 minute
```
khanhntd, line 376: Are there any reason why we need to keep up the 1 minute but not as low as possible? For example, 30s or 15s (even though the higher value of backOff, the more distribution Backoff would be but I'm still seeing 30s would be enough to cover the distribution?

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -332,14 +371,17 @@ func (c *CloudWatch) pushMetricDatumBatch() {
...
 	if c.retries <= defaultRetryCount {
-		backoffInMillis = int64(backoffRetryBase * math.Pow(2, float64(c.retries)))
+		backoffInMillis = int64(backoffRetryBase * 1 << c.retries)
```
khanhntd, line 378: Nice on using shift operation !!

plugins/outputs/cloudwatch/cloudwatch_test.go
```diff
@@ -323,45 +323,43 @@ func (svc *mockCloudWatchClient) PutMetricData(input *cloudwatch.PutMetricDataIn
 	return args.Get(0).(*cloudwatch.PutMetricDataOutput), args.Error(1)
 }
 
-func newCloudWatchClient(svc cloudwatchiface.CloudWatchAPI) *CloudWatch {
+func newCloudWatchClient(svc cloudwatchiface.CloudWatchAPI, forceFlushInterval time.Duration) *CloudWatch {
 	cloudwatch := &CloudWatch{
 		svc:                svc,
-		ForceFlushInterval: internal.Duration{Duration: 1 * time.Second},
+		ForceFlushInterval: internal.Duration{Duration: forceFlushInterval},
```
khanhntd, line 329: I understand the input variables; however, all of the test case is using `time.Second`; therefore, I would say hard code constant would be a good idea unless the test are flakiness.

plugins/outputs/cloudwatch/cloudwatch_test.go
```diff
@@ -531,54 +539,38 @@ func TestBackoffRetries(t *testing.T) {
...
 		c.backoffSleep()
-		assert.True(t, math.Abs((time.Now().Sub(now)-sleeps[i]).Seconds()) < 1)
+		// Allow up 2 seconds difference due to 1 second random jitter, and 1
+		// second for Sleep() accuracy.
+		assert.True(t, math.Abs((time.Since(now)-sleeps[i]).Seconds()) < 2)
 	}
 	now := time.Now()
 	c.backoffSleep()
-	assert.True(t, math.Abs((time.Now().Sub(now)-time.Minute).Seconds()) < 1)
+	assert.True(t, math.Abs((time.Since(now)-time.Minute).Seconds()) < 2)
```
khanhntd, line 548:
```suggestion
	assert.Greater(t, 2, math.Abs((time.Since(now)-time.Minute).Seconds()))
```

plugins/outputs/cloudwatch/cloudwatch_test.go
```diff
@@ -531,54 +539,38 @@ func TestBackoffRetries(t *testing.T) {
 	for i := 0; i <= defaultRetryCount; i++ {
```
khanhntd, line 539: Maybe add `assert := assert.New(t)` to avoid using many `t` as recommended by [the package itself](https://pkg.go.dev/github.com/stretchr/testify/assert#hdr-Example_Usage)

plugins/outputs/cloudwatch/cloudwatch_test.go
```diff
@@ -531,54 +539,38 @@ func TestBackoffRetries(t *testing.T) {
...
-		assert.Fail(t, "program should not enter metricDatumBatchFull")
-	default:
-	}
-
+	assert.False(t, c.metricDatumBatchFull())
 }
 
 func TestBuildMetricDatums_SkipEmptyTags(t *testing.T) {
 	c := &CloudWatch{
 		datumBatchChan:     make(chan []*cloudwatch.MetricDatum, 0),
```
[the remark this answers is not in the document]
adam-mateen: Yes, I think my IDE suggested that change too.
I tried to minimize the code touched in this PR to make it easier to review.
I can do a separate PR just for lint-like suggestions.

plugins/outputs/cloudwatch/util.go
adam-mateen: will fix

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -279,48 +276,90 @@ func (c *CloudWatch) timeToPublish(b *MetricDatumBatch) bool {
...
+// it flushed immediately. Now it cuts the remaining interval in half,
+// and the next interval in half.
+// If the buffer continues to stay full then it will keep cutting down.
+// This avoids bursting the backend and maintains some amount of jitter.
 func (c *CloudWatch) publish() {
-	now := time.Now()
-	forceFlushInterval := c.ForceFlushInterval.Duration
-	publishJitter := publishJitter(forceFlushInterval)
-	log.Printf("I! cloudwatch: publish with ForceFlushInterval: %v, Publish Jitter: %v", forceFlushInterval, publishJitter)
-	time.Sleep(now.Truncate(forceFlushInterval).Add(publishJitter).Sub(now))
```
adam-mateen, line 287: Just a note, this old call to `time.Sleep()` could be called with a negative value.
If current time was `04:36:47` and the flush interval was `60s`, then it rounds down to `04:36:00` then add jitter.
If the jitter was was anything less than `47s` it wouldn't sleep at all.
So then the jitter is only based on when the agent process started.

plugins/outputs/cloudwatch/cloudwatch.go
adam-mateen: My thinking was that I don't want to make the Agent ever wait longer than the given `ForceFlushInterval`.
So if current time is `04:26:04`, `ForceFlushInterval` is 60s, and jitter is `55s`, I do not want the agent waiting until `04:27:55. I'd prefer to use `04:26:55`.

There is no wasteful API call since if there are no metrics in any batches, the call does nothing.

plugins/outputs/cloudwatch/cloudwatch.go
adam-mateen: Yes, the idea of this PR is to reduce "burstiness".
In other words, to gradually increase the upload frequency instead of uploading metrics as fast as possible.
This does increase the chance that some buffers fill up and another metric value cannot be collected.

Keep in mind there are other buffers.

(input plugin) --> telegraf_buffering --> (output plugin) --> 10,000 metrics in `metricChan` --> 50 batches of 20 datums each in `datumBatchChan` --> 10,000 API request buffered/queued

plugins/outputs/cloudwatch/cloudwatch.go
khanhntd: > I do not want the agent waiting until 04:27:55. I'd prefer to use 04:26:55

This only happens with initial push so I understand your concern. However, would the  jitter is `2s` be the number of collected metrics as 55s? However, after your confirmation, then its good to me

```
There is no wasteful API call since if there are no metrics in any batches, the call does nothing.
```

plugins/outputs/cloudwatch/cloudwatch_test.go
adam-mateen: Almost all test functions use the `time.Second` has 2nd parameter.
But the new `TestPublish()` function uses a different value (60 seconds).
That is why I added the 2nd parameter.

plugins/outputs/cloudwatch/cloudwatch_test.go
adam-mateen: will fix

plugins/outputs/cloudwatch/cloudwatch.go
khanhntd: > This does increase the chance that some buffers fill up and another metric value cannot be collected.

The chances are rarely but still, do we can know the percentage of metrics dropping to avoid impacting the current customers? Another question from me: does this impact high-resolution metric's statistic too?

plugins/outputs/cloudwatch/cloudwatch.go
adam-mateen: Correct, calling `c.pushMetricDatumBatch()` does this:
```
func (c *CloudWatch) pushMetricDatumBatch() {
	for {
		select {
		case datumBatch := <-c.datumBatchChan:
			c.publisher.Publish(datumBatch)
			continue
		default:
		}
		break
	}
}
```
So if there are no batches in the channel it does nothing.

And if you take a look at: `func (c *CloudWatch) pushMetricDatum() {` you will see it puts a batch on the channel if the batch is full or if the batch has existed for `>= c.ForceFlushInterval.Duration`.

plugins/outputs/cloudwatch/aggregator_test.go
```diff
@@ -60,6 +62,8 @@ func TestAggregator_NotDurationType(t *testing.T) {
 
 	assertNoMetricsInChan(t, metricChan)
 	close(shutdownChan)
+	// Cleanup
+	wg.Wait()
```
khanhntd, line 66: Does this need to be addressed now or in other PR and I don't see any wg.Add(1) or wg.Done() though?

plugins/outputs/cloudwatch/cloudwatch_test.go
```diff
@@ -528,57 +536,42 @@ func TestBackoffRetries(t *testing.T) {
...
 	now := time.Now()
 	c.backoffSleep()
-	assert.True(t, math.Abs((time.Now().Sub(now)-time.Minute).Seconds()) < 1)
+	assert.Greater(2.0, math.Abs((time.Since(now)-time.Minute).Seconds()))
 
 	c.retries = 0
 	now = time.Now()
 	c.backoffSleep()
-	assert.True(t, math.Abs((time.Now().Sub(now)-sleeps[0]).Seconds()) < 1)
+	assert.True(math.Abs((time.Since(now)-sleeps[0]).Seconds()) < 2)
```
khanhntd, line 554:
```suggestion
	assert.Greater(2.0,math.Abs((time.Since(now)-sleeps[0]).Seconds()))
```

plugins/outputs/cloudwatch/cloudwatch_test.go
```diff
@@ -528,57 +536,42 @@ func TestBackoffRetries(t *testing.T) {
...
 	sleeps := []time.Duration{time.Millisecond * 200, time.Millisecond * 400, time.Millisecond * 800,
 		time.Millisecond * 1600, time.Millisecond * 3200, time.Millisecond * 6400}
+	assert := assert.New(t)
 	for i := 0; i <= defaultRetryCount; i++ {
 		now := time.Now()
 		c.backoffSleep()
-		assert.True(t, math.Abs((time.Now().Sub(now)-sleeps[i]).Seconds()) < 1)
+		// Allow up 2 seconds difference due to 1 second random jitter, and 1
+		// second for Sleep() accuracy.
+		assert.True(math.Abs((time.Since(now)-sleeps[i]).Seconds()) < 2)
```
khanhntd, line 545:
```suggestion
		assert.Greater(2.0,math.Abs((time.Since(now)-sleeps[i]).Seconds()) )
```

plugins/outputs/cloudwatch/cloudwatch_test.go
```diff
@@ -375,40 +373,50 @@ func TestWriteError(t *testing.T) {
 	svc.On("PutMetricData", mock.Anything).Return(
 		&res,
 		serverInternalErr)
-	cloudWatchOutput := newCloudWatchClient(svc)
-	cloudWatchOutput.publisher, _ = publisher.NewPublisher(publisher.NewNonBlockingFifoQueue(10), 10, 2*time.Second, cloudWatchOutput.WriteToCloudWatch)
-	metrics := make([]telegraf.Metric, 0, 2)
-	measurement := "Test_namespace"
-	fields := map[string]interface{}{
```
khanhntd, line 382: What are the reasons to delete the tags and dimensions though?

plugins/outputs/cloudwatch/cloudwatch_test.go
```diff
@@ -323,45 +323,43 @@ func (svc *mockCloudWatchClient) PutMetricData(input *cloudwatch.PutMetricDataIn
...
-		"usage_nice":       100,
-		"usage_iowait":     100,
-		"usage_irq":        100,
-		"usage_softirq":    100,
-		"usage_steal":      100,
-		"usage_guest":      100,
-		"usage_guest_nice": 100,
 	}
 
-	tags := map[string]string{"dimension_name1": "dimension_value2"}
```
khanhntd, line 358: Any reason why to delete the dimensions?

plugins/outputs/cloudwatch/cloudwatch.go
adam-mateen: I tested a burst of 20,000 metrics using the script I posted in the description.
No metrics were dropped.
I will do more testing to see if at any point metrics will get dropped.

I can see that statsd has `defaultAllowPendingMessage = 10000`.
and
```
	// Number of messages allowed to queue up in between calls to Gather. If this
	// fills up, packets will get dropped until the next Gather interval is ran.
	AllowedPendingMessages int
```
Just in the output plugin there is buffering to allow for 10,000 metrics + 1,000 (50 batches of 20).
So for statsd you would need to burst >21,000 metrics in any given ForceFlushInterval.
This does not account for any additional buffering the Telegraf framework might do prior to the output plugin's `Write()` function.

straussb:
High-level comment on this.  Even though CWAgent has multiple levels of buffering internally, there must be some rate of metric production that this change would slow down, right?  To put it another way, if we can stand to wait half of the normal flush interval when the batch is full, could we just wait the full flush interval anyway?  If not, then what's to say that half the flush interval is OK?

plugins/outputs/cloudwatch/cloudwatch.go
adam-mateen: `backOffSleep()` starts with a low amount of sleep time (200 ms) and increases it.
I assume the intent is that the Agent never wait more than a minute between retries, anything above that is considered excessive. 

Why not set the max sleep to 30 seconds...?
There is no 1 right answer here. 30 seconds could be an acceptable upper limit.
I would prefer not changing that specific behavior just yet.

Keep in mind that number of (number_of_retries * sleep_time) will give the total time a request is retried, so lowering either of those numbers affects the overall retry time.

plugins/outputs/cloudwatch/aggregator_test.go
adam-mateen: Yes, it was needed due to intermittently failing tests.

plugins/outputs/cloudwatch/cloudwatch_test.go
adam-mateen: I don't think those were dimensions, just metric-names and metric-values.
I deleted it so that the the given `count` parameter is accurately the count of metrics.
Otherwise the input would have to be called `how_many_batches_of_10_metrics_would_you_like`. :)

plugins/outputs/cloudwatch/cloudwatch_test.go
adam-mateen: see the other comment and response on line 350.

plugins/outputs/cloudwatch/cloudwatch_test.go
adam-mateen: will fix

plugins/outputs/cloudwatch/cloudwatch_test.go
adam-mateen: will fix

adam-mateen:
> High-level comment on this. Even though CWAgent has multiple levels of buffering internally, there must be some rate of metric production that this change would slow down, right? To put it another way, if we can stand to wait half of the normal flush interval when the batch is full, could we just wait the full flush interval anyway? If not, then what's to say that half the flush interval is OK?

>there must be some rate of metric production that this change would slow down, right

Yes, the max metric production rate has slowed down _in theory_. Previously the `publish()` method could call `c.pushMetricDatumBatch()` as fast as possible / without delay. Now, the fastest it will get called is once a second.
Which means 1,000 metrics per second, or 60,000 metrics per minute. However, in practice the current way of uploading as fast as possible does not work because of API rate limiting and when consider there could be multiple hosts running CWA.

>could we just wait the full flush interval anyway? If not, then what's to say that half the flush interval is OK?

If CWA only ever waited the full flush interval, then the fastest it can upload with the default (60 second upload interval) would be 1,000 metrics per minute (50 batches of 20 metrics).

Keep in mind that the proposed change keeps shortening the interval from 60,30,15,...1

Considering the defaults, for many customers the collection/aggregation produces metric values once a minute at the start of the minute. So this is a burst of metrics. If the burst is small enough (<1,000) then CWA can just wait the full publish interval, but as the burst size increases, the upload rate must too.

plugins/outputs/cloudwatch/aggregator_test.go
```diff
@@ -39,6 +39,8 @@ func TestAggregator_NoAggregationKeyFound(t *testing.T) {
...
```
SaxyPandaBear, line 43: I definitely recall trying to do this to fix the flaky tests but it didn't completely fix it. In practice, I think this is what we need either way so I'm fine leaving it - just calling out I remember that this didn't fix it when I tried.. unless I did it wrong.

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -279,45 +276,82 @@ func (c *CloudWatch) timeToPublish(b *MetricDatumBatch) bool {
...
-			shouldPublish = true
 		default:
-			shouldPublish = false
 		}
+
+		nowMs := time.Now().UnixMilli()
+
+		if c.metricDatumBatchFull() {
+			if !bufferFullOccurred {
+				// Set to false so this only happens once per push.
```
SaxyPandaBear, line 321: nit - should this say set to "true"?

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -332,14 +366,17 @@ func (c *CloudWatch) pushMetricDatumBatch() {
...
 	var backoffInMillis int64 = 60 * 1000 // 1 minute
 	if c.retries <= defaultRetryCount {
-		backoffInMillis = int64(backoffRetryBase * math.Pow(2, float64(c.retries)))
+		backoffInMillis = int64(backoffRetryBase * 1 << c.retries)
+		// Adding at most 1 second on each retry for the sake of jitter.
+		backoffInMillis += getJitter(1000)
 	}
 	sleepDuration := time.Millisecond * time.Duration(backoffInMillis)
-	log.Printf("W! %v retries, going to sleep %v before retrying.", c.retries, sleepDuration)
+	log.Printf("W! %v retries, going to sleep %v before retrying.", c.retries,
```
SaxyPandaBear, line 378: since we're already here, would it make sense to include the `cloudwatch` plugin as part of the log message here?

SaxyPandaBear (approved):
lgtm

plugins/outputs/cloudwatch/cloudwatch.go
adam-mateen: Yes.
Will fix.

plugins/outputs/cloudwatch/aggregator_test.go
adam-mateen: Well shoot. I appreciate the call out.

SaxyPandaBear (dismissed):
LGTM

plugins/outputs/cloudwatch/cloudwatch.go
adam-mateen: I tested a burst of 50,000 statsd metrics with the current official release of CWA:
I saw this message in the log:
```
2022-07-06T17:33:00Z W! [outputs.cloudwatch] Metric buffer overflow; 29000 metrics have been dropped
```
So then 21,000 metrics were not dropped, which makes sense given my previous comment.

I did another burst and saw:
```
2022-07-06T17:50:31Z W! [outputs.cloudwatch] Metric buffer overflow; 19057 metrics have been dropped
2022-07-06T17:52:48Z W! [outputs.cloudwatch] Metric buffer overflow; 1656 metrics have been dropped
2022-07-06T17:52:48Z W! [outputs.cloudwatch] Metric buffer overflow; 7132 metrics have been dropped
2022-07-06T17:52:48Z W! [outputs.cloudwatch] Metric buffer overflow; 1126 metrics have been dropped
```
Which adds up to 28,971.

**Same behavior with this new change**
If there were more metrics per batch in the batch-buffer then less metrics would get dropped...

straussb:
Thanks for the response, that clears things up for me.  Now I see better how this spreads out the burst throughout the minute.

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -279,45 +276,81 @@ func (c *CloudWatch) timeToPublish(b *MetricDatumBatch) bool {
...
+		nextMs += interval.Milliseconds()
+	}
+	return nextMs
+}
+
+// publish loops forever until a shutdown occurs.
+// It periodically tries pushing batches of metrics (if there are any).
+// Previously, if the batch buffer filled up it flushed immediately.
+// Now it gradually reduces the interval to avoid bursting the backend and
+// maintain some amount of jitter.
```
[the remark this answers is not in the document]
adam-mateen: will fix.

plugins/outputs/cloudwatch/cloudwatch_test.go
```diff
@@ -528,57 +535,43 @@ func TestBackoffRetries(t *testing.T) {
...
 		time.Millisecond * 1600, time.Millisecond * 3200, time.Millisecond * 6400}
+	assert := assert.New(t)
 	for i := 0; i <= defaultRetryCount; i++ {
-		now := time.Now()
+		start := time.Now()
 		c.backoffSleep()
-		assert.True(t, math.Abs((time.Now().Sub(now)-sleeps[i]).Seconds()) < 1)
+		// Expect time since start is between sleeps[i]/2 and sleeps[i].
+		assert.Less(sleeps[i] / 2, time.Since(start))
+		assert.Greater(sleeps[i], time.Since(start))
```
adam-mateen, line 544: Well this is failing on github/macos:
```
 --- FAIL: TestBackoffRetries (49.26s)
    cloudwatch_test.go:544: 
        	Error Trace:	cloudwatch_test.go:544
        	Error:      	"200ms" is not greater than "251.13787ms"
        	Test:       	TestBackoffRetries
    cloudwatch_test.go:544: 
        	Error Trace:	cloudwatch_test.go:544
        	Error:      	"400ms" is not greater than "425.195568ms"
        	Test:       	TestBackoffRetries
FAIL
```
It consistently passes on my macos laptop:
```
go test -v -count=1 ./plugins/outputs/cloudwatch --failfast -run TestBackoffRetries               
=== RUN   TestBackoffRetries
2022/07/07 09:13:25 W! cloudwatch: 0 retries, going to sleep 153 ms before retrying.
2022/07/07 09:13:25 W! cloudwatch: 1 retries, going to sleep 250 ms before retrying.
2022/07/07 09:13:25 W! cloudwatch: 2 retries, going to sleep 563 ms before retrying.
2022/07/07 09:13:26 W! cloudwatch: 3 retries, going to sleep 885 ms before retrying.
2022/07/07 09:13:27 W! cloudwatch: 4 retries, going to sleep 1780 ms before retrying.
2022/07/07 09:13:29 W! cloudwatch: 5 retries, going to sleep 3392 ms before retrying.
2022/07/07 09:13:32 W! cloudwatch: 6 retries, going to sleep 53950 ms before retrying.
2022/07/07 09:14:26 W! cloudwatch: 0 retries, going to sleep 182 ms before retrying.
--- PASS: TestBackoffRetries (61.16s)
PASS
ok  	github.com/aws/amazon-cloudwatch-agent/plugins/outputs/cloudwatch	61.988s
```

khanhntd (approved):
LGTM.

PR #504, opened 2022-07-07
Statistics calculation and log monitoring for performance benchmark test

gmealy1:
# Description of the issue
The performance benchmarking test does not currently calculate statistics on the data properly. It also uses a weak CloudWatch Agent config when benchmarking that does not monitor log files.

# Description of changes
This adds statistics calculations to the test as well as changes the test configuration to monitor two log files. The test will now write lines to those log files as the agent is running to more accurately model a standard use case. The config now also collects more metrics for increased CWAgent usage.

# License

# Tests
Validated statistics calculations. Tested config for log monitoring.

# Requirements
2. Run `make linter`


integration/test/performancetest/performance_test.go
```diff
@@ -1,22 +1,28 @@
...
-	agentRuntimeMinutes = 5 //20min desired but 5mins for testing purposes 
-	DynamoDBDataBase = "CWAPerformanceMetrics"
-
+	configPath               = "resources/config.json"
+	configOutputPath         = "/opt/aws/amazon-cloudwatch-agent/bin/config.json"
+	agentRuntimeMinutes      = 5 //20 mins desired but 5 mins for testing purposes
+	DynamoDBDataBase         = "CWAPerformanceMetrics"
+	logOutputPath1           = "/tmp/test1.log"
+	logOutputPath2           = "/tmp/test2.log"
+	transactionRatePerSecond = 10
```
SaxyPandaBear, line 25: Nitpick - we don't need to be so verbose with the name. I think tps is acceptable

integration/test/performancetest/performance_test.go
```diff
@@ -42,18 +64,44 @@ func TestPerformance(t *testing.T) {
...
 	}
 
-	//------Placeholder to put data into database------//
-	//useless code so data get used and compiler isn't mad
 	if data == nil {
 		t.Fatalf("No data")
 	}
-	//data base 
+
+	//data base
```
SaxyPandaBear, line 71: Didn't notice this. Normally you'd initialize these connections at the beginning of the function

integration/test/performancetest/performance_test.go
```diff
@@ -42,18 +64,44 @@ func TestPerformance(t *testing.T) {
...
+
+	log.Printf("Writing lines to %s with %d transactions per second", filePath, transactionRatePerSecond)
+
+	startTime := time.Now()
+
+	//loop until the test duration is reached
+	for currTime := startTime; currTime.Sub(startTime) < durationMinutes; currTime = time.Now() {
+
+		//assume this for loop runs instantly for purposes of simple throughput calculation
+		for i := 0; i < transactionRatePerSecond; i++ {
```
SaxyPandaBear, line 98: In anticipation of providing multiple different tps values, I think that this should be passed in as a function parameter

integration/test/performancetest/resources/config.json
```diff
@@ -1,6 +1,6 @@
 {
 	"agent": {
-		"metrics_collection_interval": 60,
+		"metrics_collection_interval": 1,
```
SaxyPandaBear, line 3: Collecting at this interval means that we're getting 60x the number of metric values, and we're persisting all of the data into DynamoDB, right? I think it's fine - just pointing it out

integration/test/performancetest/resources/config.json
```diff
@@ -29,17 +59,56 @@
...
+			  {
+				"file_path": "/tmp/test1.log",
+				"log_group_name": "{instance_id}",
+				"log_stream_name": "{instance_id}",
+				"timezone": "UTC"
+			  },
+			  {
+				"file_path": "/tmp/test2.log",
+				"log_group_name": "{instance_id}",
+				"log_stream_name": "{instance_id}",
```
SaxyPandaBear, line 107: I would refrain from publishing both of them to the same log stream cause it runs the risk of hitting a rate limit. I would add some arbitrary string to differentiate them, like 
```
log_stream_name: "{instance_id}/tmp2"
```

integration/test/performancetest/performance_test.go
```diff
@@ -29,8 +35,24 @@ func TestPerformance(t *testing.T) {
...
+		defer logWaitGroup.Done()
+		writeToLogs(t, logOutputPath1, agentRunDuration)
+	}()
+	go func() {
+		defer logWaitGroup.Done()
+		writeToLogs(t, logOutputPath2, agentRunDuration)
+	}()
+
+	//wait until writing to logs finishes
+	logWaitGroup.Wait()
```
SaxyPandaBear, line 54: I would pull all of this out of the main function since we don't really need the wait group anywhere else other than to control synchronous writing to the log files for the duration of the test

integration/test/performancetest/transmitter.go
```diff
@@ -210,39 +211,56 @@ func (transmitter *TransmitterAPI) Parser(data []byte) (map[string]interface{},
...
+			Std: 	 stdDev,
 			Period:  int(METRIC_PERIOD / (numDataPoints)),
 			Data:    rawMetricData.Values}
 		packet[rawMetricData.Label] = metric
 	}
 	return packet, nil
+}
+
+//CalcStats takes in an array of data and returns the average, min, max, p99, and stdev of the data
+func CalcStats(data []float64) (float64, float64, float64, float64, float64) {
```
SaxyPandaBear, line 232: Couldn't this function return a Metrics struct instead?

integration/test/performancetest/transmitter.go
```diff
@@ -210,39 +211,56 @@ func (transmitter *TransmitterAPI) Parser(data []byte) (map[string]interface{},
...
+	max := dataCopy[length - 1]
+	
+	sum := 0.0
+	for _, value := range dataCopy {
+		sum += value
+	}
+
+	avg := sum / float64(length)
+
+	p99Index := length - 1 - (length / 99)
```
SaxyPandaBear, line 256: Interesting. If there are 500 values, then 500/99 would integer divide to 5, right? So we'd sort and then get the 94th index value?

gmealy1: 500/99 = 5 -> 500 - 1 - (5) = 500 - 6 = 494. But the array is zero-indexed so this would be the 495th value and 99% of 500 is 495.
But I agree that this is kind of convoluted. I will rewrite it to cast to a float, take 99% (length * .99) and then cast back to int and subtract one to get index. More readable that way.

SaxyPandaBear: Oh that makes more sense. That's my bad.

integration/test/performancetest/performance_test.go
gmealy1: Moved to beginning

integration/test/performancetest/performance_test.go
gmealy1: Updated

gmealy1:
Wait - my changes on the test itself didn't go through. I will fix

integration/test/performancetest/performance_test.go
```diff
@@ -1,36 +1,49 @@
...
-	agentRuntimeMinutes = 5 //20min desired but 5mins for testing purposes 
-	DynamoDBDataBase = "CWAPerformanceMetrics"
-
+	configPath               = "resources/config.json"
+	configOutputPath         = "/opt/aws/amazon-cloudwatch-agent/bin/config.json"
+	agentRuntimeMinutes      = 5 //20 mins desired but 5 mins for testing purposes
+	DynamoDBDataBase         = "CWAPerformanceMetrics"
+	logOutputPath1           = "/tmp/test1.log"
+	logOutputPath2           = "/tmp/test2.log"
+	tpsMed 					 = 10
```
SaxyPandaBear, line 25: nitpick - formatting looks a little weird here for the spacing. not too concerned though

integration/test/performancetest/performance_test.go
```diff
@@ -42,18 +55,58 @@ func TestPerformance(t *testing.T) {
...
+		//assume this for loop runs instantly for purposes of simple throughput calculation
+		for i := 0; i < tps; i++ {
+			_, err = f.WriteString(fmt.Sprintf("%s - #%d This is a log line.\n", currTime.Format(time.StampMilli), i))
+			if err != nil {
+				t.Logf("Error occurred writing log line: %v", err)
+			}
+		}
+
+		time.Sleep(1 * time.Second)
+	}
```
SaxyPandaBear, line 111: Part of me wants to rewrite the outer for loop here.. but I'll have to think about it. What I kind of want to do is to change it to a `for-select`, so like:
```go
for {
    case <-time.After(durationMinutes * time.Minute): 
         return
}
```
but don't know if the case where we write is cleaner that way or not.

integration/test/performancetest/performance_test.go
gmealy1: weird, this looks different in my vscode but fixing it via vim seems to work

integration/test/performancetest/performance_test.go
gmealy1: Implemented

integration/test/performancetest/transmitter.go
```diff
@@ -205,44 +206,62 @@ func (transmitter *TransmitterAPI) Parser(data []byte) (map[string]interface{},
...
+
+//CalcStats takes in an array of data and returns the average, min, max, p99, and stdev of the data in a Metric struct
+func CalcStats(data []float64) Metric {
+	length := len(data)
+	if length == 0 {
+		return Metric{}
+	}
+
+	//make a copy so we aren't modifying original
+	dataCopy := make([]float64, length)
```
khanhntd, line 229: We are modifying the data locally so why we don't modify the original ? Is the original data being used by somewhere else?

integration/test/performancetest/performance_test.go
```diff
@@ -42,18 +55,64 @@ func TestPerformance(t *testing.T) {
...
 	_, err = dynamoDB.SendItem(data)
-	if err !=nil{
+	if err != nil {
 		t.Fatalf("Error: couldnt upload metric data to table")
 	}
-}
\ No newline at end of file
+}
+
+func startLogWrite(t *testing.T, filePath1, filePath2 string, agentRunDuration time.Duration, tps int) {
```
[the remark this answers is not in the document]
SaxyPandaBear: I would not do that either. The problem is that this is not scalable so it doesn't account for a variable number of log files that we want to write to. 

What we want to really test out is how cpu and memory change as we add logs to monitor. This makes me think that we should either:
## Decide on a strict path naming convention and then just specify a number of files to write to:
```go
func startLogWrite(... numFiles int) {
    for i := 0; i < numFiles; i++ {
        filepath := fmt.Sprintf("/some/file/path/something%d", i)
    }
}
```

## Pass in a list of file paths:
```go
func startLogWrite(... filePaths []string) {
    for _, fp := range filePaths {
    }
}
```

---
I'm mostly indifferent on which approach to take. They're pretty similar. 

This is also important for adding to the wait group, because we aren't always going to want to monitor two files.

integration/test/performancetest/performance_test.go
```diff
@@ -42,18 +55,64 @@ func TestPerformance(t *testing.T) {
...
-	//data base 
-	dynamoDB := InitializeTransmitterAPI(DynamoDBDataBase) //add cwa version here
-	if dynamoDB == nil{
-		t.Fatalf("Error: generating dynamo table")
-	}
+	
 	_, err = dynamoDB.SendItem(data)
-	if err !=nil{
+	if err != nil {
 		t.Fatalf("Error: couldnt upload metric data to table")
```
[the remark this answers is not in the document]
SaxyPandaBear: This is unrelated and could be added separately.

integration/test/performancetest/performance_test.go
```diff
@@ -42,18 +55,64 @@ func TestPerformance(t *testing.T) {
...
+
+
+	ticker := time.Ticker(1 * time.Second)
+	defer ticker.Stop()
+
+	//loop until the test duration is reached
+	for {
+		select {
+		case <-ticker.C:
+			for i := 0; i < tps; i++ {
```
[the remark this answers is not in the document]
SaxyPandaBear: It is not meant to be a constant.

integration/test/performancetest/performance_test.go
```diff
@@ -42,18 +55,64 @@ func TestPerformance(t *testing.T) {
...
+				_, err = f.WriteString(fmt.Sprintf("%s - #%d This is a log line.\n", currTime.Format(time.StampMilli), i))
+				if err != nil {
+					t.Logf("Error occurred writing log line: %v", err)
+				}
+			}
+		
+		case <-time.After(durationMinutes):
```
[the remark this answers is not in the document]
SaxyPandaBear: That's not really relevant. The old code worked too.

integration/test/performancetest/transmitter.go
```diff
@@ -205,44 +206,62 @@ func (transmitter *TransmitterAPI) Parser(data []byte) (map[string]interface{},
 	}
 	packet := make(map[string]interface{})
 	packet[PARTITION_KEY] = time.Now().Year()
-	packet[HASH] =  os.Getenv(SHA_ENV) //fmt.Sprintf("%d", time.Now().UnixNano())
+	packet[HASH] = os.Getenv(SHA_ENV) //fmt.Sprintf("%d", time.Now().UnixNano())
```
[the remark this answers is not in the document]
SaxyPandaBear: Please clarify your comment. What are you suggesting? I only see `os.Getenv(SHA_ENV)` used once. What is the relevance here?

integration/test/performancetest/transmitter.go
```diff
@@ -205,44 +206,62 @@ func (transmitter *TransmitterAPI) Parser(data []byte) (map[string]interface{},
...
+	p99Val := dataCopy[p99Index]
+
+	stdDevSum := 0.0
+	for _, value := range dataCopy {
+		stdDevSum += math.Pow(avg - value, 2)
+	}
+
+	stdDev := math.Sqrt(stdDevSum / float64(length))
+
+	metrics := Metric{
```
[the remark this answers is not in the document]
SaxyPandaBear: We are not using GetMetricStatistics.

integration/test/performancetest/performance_test.go
```diff
@@ -1,36 +1,49 @@
...
-	agentRuntimeMinutes = 5 //20min desired but 5mins for testing purposes 
-	DynamoDBDataBase = "CWAPerformanceMetrics"
-
+	configPath               = "resources/config.json"
+	configOutputPath         = "/opt/aws/amazon-cloudwatch-agent/bin/config.json"
+	agentRuntimeMinutes      = 5 //20 mins desired but 5 mins for testing purposes
+	DynamoDBDataBase         = "CWAPerformanceMetrics"
+	logOutputPath1           = "/tmp/test1.log"
+	logOutputPath2           = "/tmp/test2.log"
+	tpsMed 			 = 10
```
SaxyPandaBear, line 25: One thing I don't think this addresses exactly is how we know which test corresponds to which use-case scenario/TPS/etc. Not something that needs to be tackled in this PR, but warrants discussion. 

For the `logOutputPath1` and logOutputPath2`, I think my other comment below addresses my concern with them, but I don't think we should have them hardcoded like this because we intend to have different numbers of log files monitored for performance testing.

integration/test/performancetest/transmitter.go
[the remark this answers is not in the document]
khanhntd: For example, should we export "SHA_ENV" even though only performance benchmarking is the only one who uses it? Environment variable is not good for single application ( so considering passing it as an input flag)

integration/test/performancetest/performance_test.go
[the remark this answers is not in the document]
khanhntd: I'm fine with both approach. The reason why I let `filePath := []string{"filepath1", "filepath2"}` in the function was:
* Only the performance is using it. Therefore, would consider tuning in the function itself (and it would indeed scalable in its function)
* We can also replace the `filePath := []string{"filepath1", "filepath2"}` to `numberOfFiles` so its easier in adding a files

integration/test/performancetest/performance_test.go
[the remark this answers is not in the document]
khanhntd: Yeah I'm confused myself a little bit with the constant [tpsMed](https://github.com/aws/amazon-cloudwatch-agent/pull/504/files#diff-480bc091bc017151d24a4a67c30cfa47a97583d4146022f8160d34bc9025ac9bR25)

PR #505, opened 2022-07-07
Fix backoffSleep() so it jitters the sleep time.

adam-mateen:
# Description of the issue
When a burst of metrics is forwards to the output plugin and the number of metrics exceeds the internal buffer size, CWA will push the metrics immediately. If there are many hosts running CWA and they all get a burst of metrics at the beginning of each minute, then they will all try uploading at the beginning of the minute. The burst of uploads is likely to surpass API rate limits, and will need to be retried.

Currently when PutMetricData API calls fail, they are retried after deterministic delays instead of jittered.

# Description of changes
Fix backoffSleep() so it jitters the sleep time.
Fix publish() so it does not use a negative sleep time.

# License

# Tests
- Updated unit tests.
- Manually tested a burst of 10K, 20K, 30K statsd metrics.

# Requirements
2. Run `make linter`

(skipped formatting to keep PR small)

### DIFF
--- a/plugins/outputs/cloudwatch/cloudwatch_test.go
+++ b/plugins/outputs/cloudwatch/cloudwatch_test.go
@@ -525,25 +525,32 @@ func buildCloudWatchFromToml(contents string) (*CloudWatch, error) {
 
 func TestBackoffRetries(t *testing.T) {
 	c := &CloudWatch{}
-	sleeps := []time.Duration{time.Millisecond * 200, time.Millisecond * 400, time.Millisecond * 800,
-		time.Millisecond * 1600, time.Millisecond * 3200, time.Millisecond * 6400}
+	sleeps := []time.Duration{
+		time.Millisecond * 200,
+		time.Millisecond * 400,
+		time.Millisecond * 800,
+		time.Millisecond * 1600,
+		time.Millisecond * 3200,
+		time.Millisecond * 6400}
 	a := assert.New(t)
+	// tolerance is necessary because GitHub/MacOs does not have an accurate sleep().
+	tolerance := 200 * time.Millisecond
 	for i := 0; i <= defaultRetryCount; i++ {
 		start := time.Now()
 		c.backoffSleep()
 		a.LessOrEqual(sleeps[i] / 2, time.Since(start))
-		a.GreaterOrEqual(sleeps[i], time.Since(start))
+		a.GreaterOrEqual(sleeps[i] + tolerance, time.Since(start))
 	}
 	start := time.Now()
 	c.backoffSleep()
 	a.LessOrEqual(time.Minute / 2, time.Since(start))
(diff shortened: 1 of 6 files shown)


plugins/outputs/cloudwatchlogs/pusher.go
```diff
@@ -316,10 +316,10 @@ func (p *pusher) send() {
...
 func retryWait(n int) time.Duration {
 	const base = 200 * time.Millisecond
-	const max = 1 * time.Minute
-	d := base * time.Duration(1<<int64(n))
-	if n > 5 {
-		d = max
+	// Max wait time is 1 minute (jittered)
+	d := 1 * time.Minute
+	if n < 5 {
+		d = base * time.Duration(1<<int64(n))
```
adam-mateen, line 322: No change in behavior.
Just a formatting change.

plugins/outputs/cloudwatch/util_test.go
```diff
@@ -18,10 +18,16 @@ import (
...
-	assert.True(t, publishJitter >= 0)
-	assert.True(t, publishJitter < time.Minute)
+	// Loop an arbitrary number of times.
+	last := time.Duration(-1)
+	for i := 0; i < 100; i++ {
+		publishJitter := publishJitter(time.Minute)
+		log.Printf("Got publisherJitter %v", publishJitter)
+		assert.GreaterOrEqual(t, publishJitter, time.Duration(0))
+		assert.Less(t, publishJitter, time.Minute)
+		assert.NotEqual(t, publishJitter, last)
```
adam-mateen, line 28: There is effectively 0 change that 2 random numbers can be the same considering the range is 0 to 60,000,000,000.
If someone is opposed I can remove the `NotEqual` check.

plugins/outputs/cloudwatch/cloudwatch_test.go
```diff
@@ -401,14 +400,14 @@ func TestWriteError(t *testing.T) {
...
 	for i := 0; i < defaultRetryCount; i++ {
-		sum = sum + math.Pow(2, float64(i))
+		sum += 1 << i
 	}
-	time.Sleep(time.Duration(backoffRetryBase*int64(sum)) * time.Millisecond)
-
+	time.Sleep(backoffRetryBase * time.Duration(sum))
 	assert.True(t, svc.AssertNumberOfCalls(t, "PutMetricData", 5))
-
+	cloudWatchOutput.Close()
```
adam-mateen, line 410: No real change in `TestWriteError()`.
The type of `backoffRetryBase` was changed elsewhere.
The extra call to `cloudWatchOutput.Close()` is just for cleanup.

adam-mateen:
Closed this PR since it is just a subset of this larger PR:
https://github.com/aws/amazon-cloudwatch-agent/pull/501
Which will be merged shortly.

PR #506, opened 2022-07-08
Log the open file count once a minute to aid debugging.

### LINKED ISSUE #432: Feature request: print out how many files the agent is currently monitoring on demand

adam-mateen:
# Description of the issue
Close #432 

# Description of changes
Add atomic counter which is modified when files are opened and closed.
Counter is logged once a minute at info level.

# License

# Tests
Manual test on a Windows 2016 EC2 instance.
Sample log:
```
2022-07-08T20:48:08Z I! [logagent] open file count, 2
2022-07-08T20:49:08Z I! [logagent] open file count, 2
2022-07-08T20:50:08Z I! [logagent] open file count, 2
2022-07-08T20:51:09Z I! [logagent] open file count, 2
2022-07-08T20:51:20Z I! [logagent] piping log from banana.log/i-086864b800a839c94(C:\adam\banana.log.2) to cloudwatchlogs with retention 30
2022-07-08T20:52:10Z I! [logagent] open file count, 3
```

# Requirements
2. Run `make linter`

### DIFF
--- a/internal/atomiccounter/atomiccounter.go
+++ b/internal/atomiccounter/atomiccounter.go
@@ -17,7 +17,7 @@ func (ac *AtomicCounter) Increment() {
 	atomic.AddInt64(&ac.val, 1)
 }
 
-func (ac *AtomicCounter) Decremenet() {
+func (ac *AtomicCounter) Decrement() {
 	atomic.AddInt64(&ac.val, -1)
 }
 
--- a/internal/atomiccounter/atomiccounter.go
+++ b/internal/atomiccounter/atomiccounter.go
@@ -1,0 +1,28 @@
+package atomiccounter
+
+import (
+	"sync/atomic"
+)
+
+type AtomicCounter struct {
+	val int64
+}
+
+// NewAtomicCounter returns a new counter with the default value of 0.
+func NewAtomicCounter() AtomicCounter {
+	return AtomicCounter{}
+}
+
+func (ac *AtomicCounter) Increment() {
+	atomic.AddInt64(&ac.val, 1)
+}
+
+func (ac *AtomicCounter) Decremenet() {
+	atomic.AddInt64(&ac.val, -1)
+}
+
+// Get is not safe to use for synchronizing work between goroutines.
+// It is just for logging the current value.
+func (ac *AtomicCounter) Get() int64 {
+	return ac.val
+}
--- a/logs/logs.go
+++ b/logs/logs.go
@@ -93,11 +93,10 @@ func (l *LogAgent) Run(ctx context.Context) {
 
 	t := time.NewTicker(time.Second)
 	defer t.Stop()
-	minuteTicker := time.NewTicker(time.Minute)
-	defer minuteTicker.Stop()
 	for {
 		select {
 		case <-t.C:
+			log.Printf("D! [logagent] open file count, %v", tail.OpenFileCount.Get())
 			for _, c := range l.collections {
 				srcs := c.FindLogSrc()
 				for _, src := range srcs {
@@ -113,8 +112,6 @@ func (l *LogAgent) Run(ctx context.Context) {
 					go l.runSrcToDest(src, dest)
 				}
 			}
-		case <-minuteTicker.C:
-			log.Printf("I! [logagent] open file count, %v", tail.OpenFileCount.Get())
 		case <-ctx.Done():
 			return
 		}
--- a/logs/logs.go
+++ b/logs/logs.go
@@ -93,12 +93,9 @@ func (l *LogAgent) Run(ctx context.Context) {
 
 	t := time.NewTicker(time.Second)
 	defer t.Stop()
-	last := time.Now()
+	minuteTicker := time.NewTicker(time.Minute)
+	defer minuteTicker.Stop()
 	for {
-		if time.Since(last) >= 1 * time.Minute {
-			last = time.Now()
-			log.Printf("I! [logagent] open file count, %v", tail.OpenFileCount())
-		}
 		select {
 		case <-t.C:
 			for _, c := range l.collections {
@@ -116,6 +113,8 @@ func (l *LogAgent) Run(ctx context.Context) {
 					go l.runSrcToDest(src, dest)
 				}
 			}
+		case <-minuteTicker.C:
+			log.Printf("I! [logagent] open file count, %v", tail.OpenFileCount.Get())
 		case <-ctx.Done():
 			return
 		}
--- a/logs/logs.go
+++ b/logs/logs.go
@@ -9,6 +9,7 @@ import (
 	"log"
 	"time"
 
+	"github.com/aws/amazon-cloudwatch-agent/plugins/inputs/logfile/tail"
 	"github.com/influxdata/telegraf/config"
 )
 
@@ -92,7 +93,12 @@ func (l *LogAgent) Run(ctx context.Context) {
 
 	t := time.NewTicker(time.Second)
 	defer t.Stop()
+	last := time.Now()
 	for {
+		if time.Since(last) >= 1 * time.Minute {
+			last = time.Now()
+			log.Printf("I! [logagent] open file count, %v", tail.OpenFileCount())
+		}
 		select {
 		case <-t.C:
 			for _, c := range l.collections {
--- a/profiler/profiler.go
+++ b/profiler/profiler.go
@@ -27,10 +27,6 @@ func (p *profiler) AddStats(key []string, value float64) {
 	p.Lock()
 	defer p.Unlock()
 	k := strings.Join(key, "_")
-
-	if _, ok := p.stats[k]; !ok {
-		p.stats[k] = 0
-	}
 	p.stats[k] += value
 }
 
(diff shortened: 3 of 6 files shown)


profiler/profiler.go
```diff
@@ -27,10 +27,6 @@ func (p *profiler) AddStats(key []string, value float64) {
 	p.Lock()
 	defer p.Unlock()
 	k := strings.Join(key, "_")
-
-	if _, ok := p.stats[k]; !ok {
-		p.stats[k] = 0
-	}
```
adam-mateen, line 33: Unrelated change.
No need to initialize a new key's value to zero.
If the key is not in the map, then golang will do exactly what this removed code did.

logs/logs.go
```diff
@@ -92,7 +93,12 @@ func (l *LogAgent) Run(ctx context.Context) {
 
 	t := time.NewTicker(time.Second)
 	defer t.Stop()
+	last := time.Now()
 	for {
+		if time.Since(last) >= 1 * time.Minute {
+			last = time.Now()
+			log.Printf("I! [logagent] open file count, %v", tail.OpenFileCount())
+		}
```
SaxyPandaBear, line 101: Part of me feels like this would be cleaner by introducing another timer at a minute frequency. Thoughts?

SaxyPandaBear (commented):
How easy would it be to add unit tests around this? The change is simple enough and you added the manual test output so I'm not super concerned, but would help ensure we don't break this

logs/logs.go
adam-mateen, line 101: Which do you prefer:
1. Log once a minute at INFO level
2. Log once a minute at DEBUG level
3. Log once a second at DEBUG level

Please let me know!

khanhntd: Consider your suggestions/comments before 1 is good by me (4 or 5 maybe overkill with atomic counter and it still useful for the customer to aware the number of monitored files every interval)

adam-mateen: Agreed.
Will fix.

SaxyPandaBear: Personally I vote 3, debug every second regardless of change. 

Firstly, and more importantly, we search for files every second. So it is theoretically possible to burst with hundreds of files and close them all in the same minute (who would do that?) but even ignoring the weird edge case, I think it's more intuitive to have the log reflect the most accurate state. Not everyone is going to need it, so that's why I'd lean towards debug (also it bloats the log file a lot). 

Secondly, and slightly related, I don't know how I feel about using the same for loop but using a different frequency in it (1s timer vs 60s custom logic). If we want to emit it minutely, I'd prefer to spawn a separate goroutine for that. Cause I think having two timers of different frequencies in the same loop is confusing and idk how deterministic Go is in choosing which one to select.

logs/logs.go
```diff
@@ -110,6 +113,8 @@ func (l *LogAgent) Run(ctx context.Context) {
 					go l.runSrcToDest(src, dest)
 				}
 			}
+		case <-minuteTicker.C:
```
SaxyPandaBear, line 116: Brought this up in the other comment, but I'm not sure about this edge case: is there strict determinism for when two timers of different frequency should fire at the same time? Do they both fire, does one always take precedence? My concern would be if we never hit this timer chan

internal/atomiccounter/atomiccounter.go
```diff
@@ -0,0 +1,28 @@
...
+func (ac *AtomicCounter) Increment() {
+	atomic.AddInt64(&ac.val, 1)
+}
+
+func (ac *AtomicCounter) Decrement() {
+	atomic.AddInt64(&ac.val, -1)
+}
+
+// Get is not safe to use for synchronizing work between goroutines.
+// It is just for logging the current value.
```
SaxyPandaBear, line 25: Do we want to add a mutex lock on this to ensure safety? I'm honestly not super concerned

khanhntd: If we are agreeing on adding a `mutex`, I would consider use channel for better performance. I have talked with Adam about this but he would give a better explanation than me.

khanhntd (commented):
Besides adding [a max file descriptors from OS](https://github.com/aws/amazon-cloudwatch-agent/pull/491/files#diff-21527397653ffd5f62a8c7c82fe243e2d5f5ee39e3fb0dc04f0e1b7011e3f128) and add unit test for "auto_removal", everything else is good by me. (maybe little concern about the ticker frequency as Andrew points it out so will take a look at that)

internal/atomiccounter/atomiccounter.go
```diff
@@ -0,0 +1,28 @@
+package atomiccounter
```
khanhntd, line 1: Would be best to add a unit test in similar to this test case that I did for my semaphore benchmarking?
```
for _, c := range cases {
		t.Run(c.testName, func(t *testing.T) {
			sem := NewSemaphore(c.limit)

			channel := make(chan struct{})
			wg := sync.WaitGroup{}
			for i := 0; i < c.numberOfFiles; i++ {
				wg.Add(1)
				go func() {
					<-channel
					ok := sem.Acquire(time.Second)
					assert.True(t, ok)
					sem.Release()
					wg.Done()
				}()
			}

			close(channel) // start
			wg.Wait()

			checkLimitAndCount(t, sem, c.limit, 0)
		})
	}
```

logs/logs.go
adam-mateen: It is definitely getting hit on my windows 2016 test host.
Every 60th second both tickers will be "runnable".

https://developpaper.com/detailed-explanation-of-select-multiple-selection-in-golang-development/
>Select will randomly execute a runnable case. If there is no case to run, goroutine will block until there is a case to run.

So it is not deterministic which case will run, but as long as the "work" under the 1 second ticker completes in less than a second, the `for` loop wraps around and runs `select` again in which case the 1 minute ticker is the only thing "runnable".

logs/logs.go
adam-mateen: Agreed on 1 second at DEBUG.
Will fix.

adam-mateen:
> How easy would it be to add unit tests around this? The change is simple enough and you added the manual test output so I'm not super concerned, but would help ensure we don't break this

Yes, I just remembered we have another test case which verifies logging output.
So I can add that on Monday.

internal/atomiccounter/atomiccounter.go
adam-mateen: Added 2 unit test cases.

logs/logs.go
SaxyPandaBear: Good enough to me

internal/atomiccounter/atomiccounter.go
SaxyPandaBear: I don't think it's necessary to add a locking mechanism for this. I'm indifferent

adam-mateen: No, IMO atomic inc/dec is good enough.
Just wanted to clear that just because this _thing_ works for this specific use case, doesn't mean it should be used for synchronization.

There is no behavior/decision based on the value of the counter, just logging it to file so even if by the time a user sees the value in the file the actual count has changed it isn't a big deal.

SaxyPandaBear: @adam-mateen agreed

adam-mateen: Specifically don't use it like this:
```
var lock = atomiccounter.NewAtomicCounter()

// foo tries to do some EXCLUSIVE work every second
func foo() {
    for {
        if lock.Get() == 0 {
            lock.Increment()
            // do work
            .
            .
            .
        }
        time.Sleep(time.Second)
    }
}
```
if 2 or more goroutines call `foo()`, _you're going to have a bad time_.

get() plus increment() is not the same as `atomic.CompareAndSwapInt64`.
And if that is what you want, you should just use a mutex!

PR #508, opened 2022-07-09
Use github_sha_date instead of sha_date for consistency 

khanhntd:
# Description of the issue
Changing to use github_sha_date instead of sha_date for consistency with this PR https://github.com/aws/amazon-cloudwatch-agent/commit/a1919cd173e4b691048eeae87464fcb68ce2111b

# License

# Tests
Before the changes: https://github.com/khanhntd/amazon-cloudwatch-agent/runs/7260722999?check_suite_focus=true
After the changes: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/2639000632

# Requirements
2. Run `make linter`


integration/terraform/ec2/linux/iam.tf
```diff
@@ -44,7 +44,8 @@ data "aws_iam_policy_document" "user-managed-policy-document" {
       "logs:PutLogEvents",
       "s3:GetObjectAcl",
       "s3:GetObject",
-      "s3:ListBucket"
+      "s3:ListBucket",
+      "dynamodb:*"
```
SaxyPandaBear, line 48: I would rather figure out which DDB actions we need than put * for the terraform creation

.github/workflows/integrationTest.yml
```diff
@@ -667,10 +667,8 @@ jobs:
               -var="binary_name=amazon-cloudwatch-agent.rpm" \
               -var="local_stack_host_name=${{ needs.StartLocalStack.outputs.local_stack_host_name }}" \
               -var="s3_bucket=${S3_INTEGRATION_BUCKET}" \
-              -var="vpc_security_group_ids=${VPC_SECURITY_GROUPS_IDS}" \
               -var="key_name=${KEY_NAME}" \
               -var="test_name=cw-integ-test-al2" \
-              -var="iam_instance_profile=${IAM_ROLE}" \
```
SaxyPandaBear, line 673: I thought this was already updated in a previous pr. Did I miss this?

khanhntd: Maybe the `iam_instance_profile` that we thought was in integration test. This is for the performance tracking test after the merge which I forgot to update it.

.github/workflows/integrationTest.yml
```diff
@@ -128,14 +128,6 @@ jobs:
...
         run: make build package-rpm package-deb package-win package-darwin
 
-      - uses: olafurpg/setup-gpg@v3
-
-      - name: Sign packages
-        if: steps.cached_binaries.outputs.cache-hit != 'true'
-        run: |
-          echo "${GPG_PRIVATE_KEY}" | gpg --batch --import -
-          for f in $(find build/bin/); do if [ ! -d $f ]; then echo "Signing file $f" && echo "${PASSPHRASE}" | gpg --detach-sign  --passphrase-fd 0 --batch --default-key "${GPG_KEY_NAME}" $f ; fi ; done
-
```
SaxyPandaBear, line 138: Why are we deleting this? What does this have to do with your changes?

.github/workflows/integrationTest.yml
```diff
@@ -524,36 +516,6 @@ jobs:
           retry_wait_seconds: 5
           command: cd cd integration/terraform/ec2/win && terraform destroy --auto-approve -var="ami=${{ matrix.arrays.ami }}"
 
-  StopLocalStack:
```
SaxyPandaBear, line 527: What is this change for?

.github/workflows/integrationTest.yml
khanhntd: My bad. I'm testing so its not finished yet.

.github/workflows/integrationTest.yml
khanhntd: The `PerformanceTrackingTest` is also using `localstack` to perform in GovCloud (not sure why). So `StopLocalStack` should have the `PerformanceTrackingTest` as a precondition before executing it. Therefore, I move the `StopLocalStack` to the end to wrap the `PerformanceTrackingTest` too.

SaxyPandaBear: There is no reference to govcloud

khanhntd: Let's me changing the wording a little bit. Our EC2 Linux terraform script uses `localstack`; so Performance Test also uses them. Therefore, there are two changes:
* Added the `PerformanceTrackingTest` as a precondition before executing `StopLocalStack`
* Moving the `StopLocalStack` to wrap around `PerformanceTrackingTest`

SaxyPandaBear: How does this relate to changing the sha date variable?

khanhntd: Not much relates to the PR but its a small thing so I thought I can introduce it also in this PR too.

SaxyPandaBear: This is a majority of the changes introduced in this PR. "small" is a relative term. We shouldn't need localstack for the performance tests and we can look at that separately. Please revert this change out of the PR unless it directly relates to the sha date to fix our CI.

SaxyPandaBear (approved):
lgtm

PR #513, opened 2022-07-13
Performance Benchmarking Log monitor and TPS tests

gmealy1:
# Description of the issue
The CW Agent performance test does not currently monitor log files when it is being benchmarked.

# Description of changes
Adds functionality to the test so the performance test actually becomes a test matrix. Currently, the test is set up to monitor 10, 100, and 1000 log files and each of those configurations is run with 10, 100, and 1000 tps.

# License

# Tests
Ran this integration test several times to ensure it works as expected.

# Requirements
2. Run `make linter`


sethAmazon:
Why no just write 1 log line and have it generate the jsons?

integration/test/performancetest/performance_test.go
```diff
@@ -1,59 +1,158 @@
...
+	//data base
+	dynamoDB := InitializeTransmitterAPI(DynamoDBDataBase) //add cwa version here
+	if dynamoDB == nil {
+		t.Fatalf("Error: generating dynamo table")
+	}
+
+	//run tests
+	for _, configPath := range configFiles {
+		for _, tps := range tpsVals {
+			t.Run(fmt.Sprintf("config file location: %s", configPath), func(t *testing.T) {
```
SaxyPandaBear, line 50: I think there's going to be a lot of overlap in these test names. We should include some information on the TPS as well. Maybe something like 
```suggestion
			t.Run(fmt.Sprintf("config file location: %s, TPS: %d", configPath, tps), func(t *testing.T) {
```

integration/test/performancetest/performance_test.go
```diff
@@ -1,59 +1,158 @@
...
+				if err != nil {
+					log.Println("Error: ", err)
+					t.Fatalf("Error: %v", err)
+				}
+	
+				if data == nil {
+					t.Fatalf("No data")
+				}
+				
+				_, err = dynamoDB.SendItem(data)
```
SaxyPandaBear, line 74: this is all data that goes to the same hash/sort key, isn't it? How are we persisting this data so that we can differentiate between the different tests?

SaxyPandaBear (commented):
will review this again tomorrow

integration/test/performancetest/performance_test.go
```diff
@@ -1,59 +1,158 @@
...
-	configPath = "resources/config.json"
 	configOutputPath = "/opt/aws/amazon-cloudwatch-agent/bin/config.json"
-	agentRuntimeMinutes = 5 //20min desired but 5mins for testing purposes 
+	agentRuntimeMinutes = 5 //20 mins desired but 5 mins for testing purposes
 	DynamoDBDataBase = "CWAPerformanceMetrics"
-
 )
 
 func TestPerformance(t *testing.T) {
+	configFiles := []string {
```
khanhntd, line 25:
```suggestion
	testcases := []struct{
	    testName             string
            tps                         int
	    agentConfig         string
	}
```
since the test case is small and also provides me a centralize place to know regarding the different test cases.

integration/test/performancetest/performance_test.go
```diff
@@ -1,59 +1,158 @@
...
+
+	//run tests
+	for _, configPath := range configFiles {
+		for _, tps := range tpsVals {
+			t.Run(fmt.Sprintf("config file location: %s", configPath), func(t *testing.T) {
+				test.CopyFile(configPath, configOutputPath)
+	
+				test.StartAgent(configOutputPath, true)
+	
+				agentRunDuration := agentRuntimeMinutes * time.Minute
```
khanhntd, line 55: Would it better to initialize the 
```
agentRuntimeMinutes := 5 * time.Minute
```
instead of initialize it here? To avoid creating shallow variables and recreate the variable all the time (pointers and memory perhaps)

integration/test/performancetest/performance_test.go
```diff
@@ -1,59 +1,158 @@
...
+				StartLogWrite(t, agentRunDuration, configPath, tps)
+	
+				log.Printf("Agent has been running for : %s\n", (agentRunDuration).String())
+				test.StopAgent()
+	
+				//collect data
+				data, err := GetPerformanceMetrics(instanceId, agentRuntimeMinutes, agentContext, configPath)
+				//@TODO check if metrics are zero remove them and make sure there are non-zero metrics existing
+				if err != nil {
+					log.Println("Error: ", err)
```
khanhntd, line 66: Only `t.Fatalf("Error: %v", err)` instead of having `log.Println("Error: ", err)`

integration/test/performancetest/performance_test.go
khanhntd: Same question for me 👍

integration/test/performancetest/performance_test.go
```diff
@@ -1,59 +1,158 @@
...
+					t.Fatalf("Error: couldnt upload metric data to table")
+				}
+			})
+		}
+	} 
+}
+
+//StartLogWrite starts go routines to write logs to each of the logs that are monitored by CW Agent according to
+//the config provided
+func StartLogWrite(t *testing.T, agentRunDuration time.Duration, configFilePath string, tps int) {
```
khanhntd, line 85: If `agentRunDuration time.Duration` is a constant, why should we pass it as an input variable though?

integration/test/performancetest/performance_test.go
```diff
@@ -1,59 +1,158 @@
...
+	//create wait group so main test thread waits for log writing to finish before stopping agent and collecting data
+	var logWaitGroup sync.WaitGroup
```
khanhntd, line 87: If you declare var then it maybe putting it outside of the function instead of putting it here.

integration/test/performancetest/performance_test.go
```diff
@@ -1,59 +1,158 @@
...
-		t.Fatalf("No data")
+	ticker := time.Ticker(1 * time.Second)
+	defer ticker.Stop()
+
+	//loop until the test duration is reached
+	for {
+		select {
+		case <-ticker.C:
+			for i := 0; i < tps; i++ {
+				_, err = f.WriteString(fmt.Sprintf("%s - #%d This is a log line.\n", currTime.Format(time.StampMilli), i))
```
khanhntd, line 125: If `f.WriteString` fails the first time, should it fail the next try too? I can't imaging the scenario which one fails and others pass. Therefore, would it be better to `t.Fatalf`?

integration/test/performancetest/resources/config10.json
```diff
@@ -0,0 +1,162 @@
+{
+ "agent": {
+  "metrics_collection_interval": 1,
```
khanhntd, line 3: Same question as before: I see the value of using collecting high-resolution metrics for the metrics_collected but I'm not sure if the trade off is worth it? 10% different in values and higher cost (with a 60x collection rate). Therefore, I propose we should keep the 60 and also I am not sure why we would change the metrics_collection_interval for agent at an interval of 1

integration/test/performancetest/performance_test.go
```diff
@@ -1,59 +1,158 @@
...
+		"./resources/config100Logs.json",
+		"./resources/config1000Logs.json"
+	}
+
+	tpsVals := []int {
+		10,
+		100,
+		1000
+	}
+
```
khanhntd, line 36: Would be better to add a `defer deleteLogGroup(instanceID)` (another function since your test case will push a lots of logs into different log stream)?

khanhntd:
Would be better to show the performance works as expect on your fork? Another code change that outside of this change would be 
```
//within procstat metrics, find cloudwatch-agent process
	cloudwatchIndex := -1
	for i, process := range procstatList {
		if process.(map[string]interface{})["exe"].(string) == "cloudwatch-agent" {
			cloudwatchIndex = i
		}
	}
	//check to see if the process was not found
	if  cloudwatchIndex == -1 {
		return nil, nil, errors.New("cloudwatch-agent process not found in cloudwatch agent config")
	}
	//use the index to get the rest of the path
	metricList := procstatList[cloudwatchIndex].(map[string]interface{})["measurement"].([]interface{})
```
to
```
//within procstat metrics, find cloudwatch-agent process
	var metricList (map[string]interface{})["measurement"].([]interface{})
	for index, process := range procstatList {
		if process.(map[string]interface{})["exe"].(string) == "cloudwatch-agent" {
			metricList := procstatList[index]
		}
	}
	//check to see if the process was not found
	if  metricList == nil {
		return nil, nil, errors.New("cloudwatch-agent process not found in cloudwatch agent config")
	}
```

integration/test/performancetest/performance_test.go
gmealy1: This is getting changed in my next commit so it will be different

integration/test/performancetest/performance_test.go
gmealy1: I think leaving it as a constant at the top is better because it makes it easier to change and propagate through the code, especially because the 5 is temporary right now

integration/test/performancetest/performance_test.go
gmealy1: Changing this

integration/test/performancetest/performance_test.go
gmealy1: I will change this

integration/test/performancetest/resources/config10.json
gmealy1: Metric collection does not have a high impact on agent resource usage, so I figured that increasing metric collection interval and adding more metrics collected would make for a better benchmark of the agent's resources and cover more use cases for the average customer.

integration/test/performancetest/performance_test.go
khanhntd: It's the same though. `const agentRuntimeMinutes = 20 * time.Minute` at the top is better than initialize inside during the run so I'm not convinced it would be better to change the variable inside the run.

integration/test/performancetest/performance_test.go
khanhntd: I mostly want to address this 
```
Would it better to initialize the

agentRuntimeMinutes := 5 * time.Minute
instead of initialize it here? To avoid creating shallow variables and recreate the variable all the time (pointers and memory perhaps)
```
and this comment would follow that issue. Unless you are changing something with the variable.

integration/test/performancetest/resources/config10.json
khanhntd: Yes it does. More collected metrics, more metrics need to aggregate even though the batch of MetricDatum is the same. Therefore, it could cause a spike in the moment of collection. In exchange between more noise in the performance metric and  10% different in value, lower cost, I am stand by 60 though.

SaxyPandaBear:
> I don't like that we collect metric and logs inside the same test.

I see where you're coming from, but the CPU/mem usage for metrics is so minimal in comparison to log collection that I don't care that we do both. Though, now that you mention it, it does kind of inflate metrics that we would publish on log collection performance benchmarks, even if by a little 🤔 idk I'll think about it more. 

> do you think this should be another test package or another test case

No, I don't. The main thing I'm looking for with performance benchmarking is log collection. If given the choice of either metrics benchmarks or logs, I would choose logs.

gmealy1:
I can remove most of the metric collection from the config and only use procstat

integration/test/performancetest/performance_test.go
```diff
@@ -1,59 +1,269 @@
...
 	configOutputPath = "/opt/aws/amazon-cloudwatch-agent/bin/config.json"
-	agentRuntimeMinutes = 5 //20min desired but 5mins for testing purposes 
+	agentRuntimeMinutes = 5 //20 mins desired but 5 mins for testing purposes
 	DynamoDBDataBase = "CWAPerformanceMetrics"
-
+	testLogNum = "PERFORMANCE_NUMBER_OF_LOGS"
 )
 
+type LogInfo struct {
+	FilePath string `json:"file_path"`
```
SaxyPandaBear, line 28: I don't think we need `json:` annotations here since we are creating structs ourselves down below this. But I could be wrong about that.

integration/test/performancetest/resources/config10.json
gmealy1: Screen Shot 2022-07-18 at 10 16 40 AM
I tested this on an ec2 instance, the difference between a 1s collection interval on all metrics and a 60s collection interval on all metrics was ~2% cpu on a t2.micro instance

integration/test/performancetest/performance_test.go
```diff
@@ -1,59 +1,269 @@
...
+			if err != nil {
+				t.Fatalf("Error: %v", err)
+			}
+
+			if data == nil {
+				t.Fatalf("No data")
+			}
+			
+			//append test metadata so we can differentiate between tests
+			data, err = AppendTestMetadata(data, logNum, tps)
```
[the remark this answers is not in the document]
gmealy1: I thought about doing that, but the data is returned in a struct which needs to be marshalled into a json before the metadata gets added so that I don't need to define and create an entire new custom struct. I left it in the test file because that's where the metadata is generated so I felt it was more relevant

integration/test/performancetest/performance_test.go
```diff
@@ -1,59 +1,269 @@
...
-
+	testLogNum = "PERFORMANCE_NUMBER_OF_LOGS"
 )
 
+type LogInfo struct {
+	FilePath string `json:"file_path"`
+	LogGroupName string `json:"log_group_name"`
+	LogStreamName string `json:"log_stream_name"`
+	Timezone string `json:"timezone"`
+}
```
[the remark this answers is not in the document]
gmealy1: Tested this, the log portion of the config isn't generated properly when importing FileConfig directly, so the test throws and exception. I will add a comment that it is derived from FileConfig

integration/test/performancetest/performance_test.go
SaxyPandaBear: yeah no that makes sense since we have to marshal it back into json

integration/test/performancetest/performance_test.go
```diff
@@ -1,59 +1,246 @@
...
 	}
 
-	//------Placeholder to put data into database------//
-	//useless code so data get used and compiler isn't mad
-	if data == nil {
-		t.Fatalf("No data")
+	//defer deleting log group and streams
+	for _, logStream := range logStreams {
+		defer test.DeleteLogGroupAndStream(instanceId, logStream)
+	}
```
SaxyPandaBear, line 56: I think this is fine, but an enhancement to this is to reduce the number of bad calls. The logs are being published to the same log group but different log streams. If we have 1000 log streams, in this we are making 1000 calls to delete a single log group. I would consider either modifying the function to take in a variable number of log stream names:
```go
func foo(group string, ...streams string) 
```
I think. 

Or, maybe we just want to separate it out so there is a function for deleting log groups, a function for deleting log streams, and then we gut the `DeleteLogGroupAndStream` to call both of those functions, and then we call them separately like:
```go
test.DeleteLogGroup(foo)
for _, logStream := range logStreams {
  test.DeleteLogStream(logStream)
}
```

integration/test/performancetest/performance_test.go
SaxyPandaBear: I don't see a reason to. We only use the WaitGroup for waiting on logs to be written

integration/test/performancetest/performance_test.go
```diff
@@ -1,59 +1,246 @@
...
+}
+
+/* GenerateConfig takes the number of logs to be monitored and applies it to a default config (at ./resources/config.json)
+* it writes logs to be monitored of the form /tmp/testNUM.log where NUM is from 1 to number of logs requested to
+* ./resources/configNUM.json where NUM is number of logs
+* DEFAULT CONFIG MUST BE SUPPLIED WITH AT LEAST ONE LOG BEING MONITORED 
+* (log being monitored will be overwritten - it is needed for json structure)
+*/
+func GenerateConfig(logNum int) (string, []string, error) {
```
SaxyPandaBear, line 117: Do you need to return the list of file paths, or can we cut that out and just call `GetLogFilePaths()` when needed?

integration/test/performancetest/performance_test.go
gmealy1: Just to clarify, are you proposing that I modify functions in or append functions to the cwl_util.go file?

integration/test/performancetest/performance_test.go
gmealy1: It's a list of log stream names, not file paths

integration/test/performancetest/performance_test.go
```diff
@@ -1,59 +1,253 @@
...
+		
+		case <-endTimeout:
+			return nil
+		}
+	}
+}
+
+//GetLogFilePaths parses the cloudwatch agent config at the specified path and returns a list of the log files that the 
+//agent will monitor when using that config file
+func GetLogFilePaths(configPath string) ([]string, error) {
```
SaxyPandaBear, line 234: Not worth blocking this PR for, but I think that this function can be pulled out of this test cod and into a utility.

SaxyPandaBear (approved):
LGTM

PR #522, opened 2022-07-26
Drop Support for ADOT Collector

khanhntd:
# Description of the issue
Drop support for ADOT Collector and delete equivalent submodules.

# Description of changes 
* Delete unnecessary folders creation for following OS: Debian, Linux, Windows
* Delete [AWS-OTEL-Collector from the THIRD-PARTY-LICENSE](https://github.com/aws/amazon-cloudwatch-agent/pull/522/files#diff-35fb5cbcccfd11b8078a48d04c510b2bc00ae80eeea9b682c0f2c9bb36462333)
* Delete unnecessary build stages for AWS Otel Collector before bundle up with CloudWatchAgent
* Remove flag and also corresponding function for OTEL Collector in the packaging
* [Delete submodules and keep the full clone (instead of shallow clone) in Integration Test when building CWAgent](https://github.com/aws/amazon-cloudwatch-agent/pull/522/files#diff-9b9f9a38b889c9dcb55ba0330baa69c8fdd62ac1100b6eb59b6fc44008f5d320R91)

# License

# Tests
For backward compatibility, I have run the start-up commands in the previous OS and CloudWatchAgent runs normal to me (including MacOS)

**Image for installation of OTEL-Collector bundle with CWAgent before and after the change**
**CloudWatchAgent Installation on Windows:**
* Before the change
image
image

* After the change
image
image

**CloudWatchAgent Installation on Linux:**
* Before the change
image

* After the change
image

**CloudWatchAgent Installation on Debian:**
* Before the change
image

* After the change
image
image

# Requirements
2. Run `make linter`


packaging/debian/preinst
```diff
@@ -17,15 +17,3 @@ if ! id cwagent >/dev/null 2>&1; then
      echo "create user cwagent, result: $?"
 fi
 
-if ! grep "^aoc:" /etc/group >/dev/null 2>&1; then
-     groupadd -r aoc >/dev/null 2>&1
-     echo "create group aoc, result: $?"
-fi
-
-if ! id aoc >/dev/null 2>&1; then
```
khanhntd, line 25: It should be safe to remove this since we had triggered the roll-back and also there has not been any customers who report this issue.

sethAmazon (approved):

khanhntd:
Will on hold the merge until the final decision has been made .

sethAmazon (approved):

packaging/linux/amazon-cloudwatch-agent.spec
```diff
@@ -100,19 +86,6 @@ if ! id cwagent >/dev/null 2>&1; then
     echo "create user cwagent, result: $?"
 fi
 
-if ! grep "^aoc:" /etc/group >/dev/null 2>&1; then
```
SaxyPandaBear, line 103: What we should do here is remove the aoc user if it exists since we don't need it anymore. We shouldn't just leave the user there on the host

SaxyPandaBear (changes requested):
1. It's been over a month since the PR was first opened. Can you confirm that this still works as expected and provide evidence on Windows and Linux, minimum, MacOS if possible?

packaging/linux/amazon-cloudwatch-agent.spec
ymtaye: Thanks for the suggestion! I will make changes and also validate changes work as expected

ymtaye:
> 1. It's been over a month since the PR was first opened. Can you confirm that this still works as expected and provide evidence on Windows and Linux, minimum, MacOS if possible?
image

image

SaxyPandaBear (approved):
LGTM

SaxyPandaBear:
I think what we should do eventually as a follow up is remove the statuses for CWOC in the control script, but probably don't need to do that right now.

khanhntd:
> I think what we should do eventually as a follow up is remove the statuses for CWOC in the control script, but probably don't need to do that right now.


Hmm I removed that and remove some sanity check relates to that status but I guess we have decided in other way in QS.

PR #526, opened 2022-07-27
Added concurrent testing and release updates

okankoAMZ:
# Description of the issue
* Different configuration test were unable to run
* Performance Tests were not concurrently consistent
* Releases were not identified

# Description of changes
* Able to run different configurations as different tests using test_generator and json config. 
* Concurrently running performance tests are concurrently consistent using dynamo's optimistic lock with a retry method 
* Added new release workflow file that updates a tag's item's isRelease attribute on dynamo.

# License

# Tests
* Tested with multiple thread that start at the same time 
* Tested with integration test
* Manually running releases to see if it is updated.

# Requirements
2. Run `make linter`


integration/test/performancetest/demo.sh
```diff
@@ -0,0 +1,14 @@
+for commitCount in {0..12}
```
sethAmazon, line 1: Do we need this in the commit?

.github/workflows/integrationTest.yml
```diff
@@ -638,52 +651,52 @@ jobs:
...
-
       - name: Check env
-        run: echo "SHA ${GITHUB_SHA} | Date ${{ steps.sha_date.outputs.sha_date }}"
-
+        run: echo "SHA ${{ steps.sha.outputs.sha_short }} | Date ${{ steps.sha_date.outputs.sha_date }} "
+      - name: Verify Terraform version
+        run: terraform --version
       - name: Terraform apply
-        if: steps.performance-tracking.outputs.cache-hit != 'true'
+        if: steps.ec2-linux-integration-test.outputs.cache-hit != 'true'
```
khanhntd, line 665: **Nit pick**: Keep `performance-tracking` would be a better idea

integration/terraform/ec2/linux/variables.tf
```diff
@@ -82,4 +82,16 @@ variable "github_repo" {
 variable "github_sha_date"{
   type = string
   default = ""
-}
\ No newline at end of file
+}
+variable "sha" {
```
khanhntd, line 86: Seems duplicate to 
```
variable "github_sha" {
  type    = string
  default = ""
} and 
variable "github_sha_date"{
  type = string
  default = ""
} 
```

integration/test/performancetest/performance_test.go
```diff
@@ -35,11 +35,10 @@ type LogInfo struct {
 func TestPerformance(t *testing.T) {
 	//get number of logs for test from github action
 	//@TODO
-	//logNum, err := strconv.Atoi(os.Getenv(testLogNum)) //requires a commit from Okan that updates the workflow file so the log tests will run concurrently
-	// if err != nil {
-	// 	t.Fatalf("Error: cannot convert test log number to integer, %v", err)
-	// }
-	logNum := 10 //THIS IS TEMPORARY SO CODE RUNS
+	logNum, err := strconv.Atoi(os.Getenv(testLogNum)) //requires a commit from Okan that updates the workflow file so the log tests will run concurrently
```
khanhntd, line 38: Still standby opinions that these variables should be passed as flag instead of env

integration/test/performancetest/demo.sh
okankoAMZ: Oh im sorry I though that was in my gitignore

integration/test/performancetest/transmitter.go
```diff
@@ -153,12 +197,194 @@ func (transmitter *TransmitterAPI) TableExist() (bool, error) {
...
+	var err error
+	var ae *types.ConditionalCheckFailedException
+	attributes := map[string]types.AttributeValue{
+		"isRelease":&types.AttributeValueMemberBOOL{Value: true},
+	}
+	for {//concurrency retry
+		/*solving parallel updates using an optimistic lock and retry,
+		this may result in a livelock;
+		however, random sleeps makes this possiblity nearly impossible*/
+		err = transmitter.UpdateItem(hash,attributes) //try to update the item
```
khanhntd, line 353: For metrics, I understand there is a need for `optimistic lock`. However, would it be the same when updating the `isRelease` flag? Do I care the second attemp to update to `true` if the first attempt update success to `true` ? So seems like `pessimistic lock` is more useful for me in this case.

integration/test/performancetest/transmitter.go
```diff
@@ -153,12 +197,194 @@ func (transmitter *TransmitterAPI) TableExist() (bool, error) {
...
+		//this may be overwritten by other test threads, in that case it will return a specific error
+		if errors.As(err,&ae){ //check if our call got overwritten
```
khanhntd, line 355: The comparison seems duplicate to [this code](https://github.com/okankoAMZ/amazon-cloudwatch-agent/blob/9676dd418451dbbfcb4edaf938a39293f6d818f2/integration/test/performancetest/transmitter.go#L333-L335) and other code. Therefore, would it be best for other function to return the error and the corresponding function who uses the function would handle how error instead? Since I favor more in centralized handling error

integration/test/performancetest/demo.sh
khanhntd, line 1: Would it be better to remove `demo.sh` and keep it in private one for you only?

integration/test/performancetest/transmitter.go
```diff
@@ -153,12 +197,194 @@ func (transmitter *TransmitterAPI) TableExist() (bool, error) {
...
+		this may result in a livelock;
+		however, random sleeps makes this possiblity nearly impossible*/
+		currentItemList,err = transmitter.Query(packet[HASH].(string)) //get the item's most up to date version
+		newAttributes,err := transmitter.TestCasePackager(packet,currentItemList[0])
+		if(err!=nil){
+			return "",err
+		}
+		err = transmitter.UpdateItem(packet[HASH].(string),newAttributes) //try to update the item
+		//this may be overwritten by other test threads, in that case it will return a specific error
+		if errors.As(err,&ae){ //check if our call got overwritten
```
khanhntd, line 239: Would be better to create another general function for handling optimism lock for DynamoDB

.github/workflows/releaseTest.yml
```diff
@@ -0,0 +1,53 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
khanhntd, line 1: I understand the need of creating a separate workflow but its seems wasteful for me only to update the RELEASE_TAG? Sure it could be done in the CD (maybe in the future) but for now, would it be better to pass as a input flag such as `isRelease` for testing?

integration/test/performancetest/transmitter.go
okankoAMZ: So dynamo doesn't have pessimistic locks, and I have that optimistic lock there because if we update release the moment the test ends, the isRelease flag can get overwritten (I had this bug while running previous releases).

.github/workflows/releaseTest.yml
okankoAMZ: I think this is a better practice, because you don't know if a commit will become a release, so it will be challenging to assign a flag to a commit.  Meanwhile, with this method and time a release is made the doesn't have to worry about updating anything. I also don't understand why we should do it in the future since it already is completed and working

integration/test/performancetest/transmitter.go
khanhntd: I'm not sure the use case here. Why we need to update the release the moment the test ends though? There would be a span amount of time in merging PR and deploying new version so I'm still hard to understand why we need `optimistic lock`?

integration/test/performancetest/transmitter.go
okankoAMZ: I am only using in two places thats why I though it would be okay. However, if you want I can move it inside UpdateItem() directly.

.github/workflows/releaseTest.yml
khanhntd: Step back a little bit: it seems I have not clearly explained my point here. In the future, we will have a separate CI (in charge of integration test and release candidate) and CD ( in charge of validation test and release final version). Therefore, since this workflow is only for updating release tag and for *testing*, would it be better to make it as a temporary script in the `integrationTest` workflow? Its not a one-way door so its up to the others

.github/workflows/integrationTest.yml
```diff
@@ -638,52 +651,52 @@ jobs:
...
+              -var="user=${{ matrix.arrays.username }}" \
+              -var="ami=${{ matrix.arrays.ami }}" \
+              -var="ca_cert_path=${{ matrix.arrays.caCertPath }}" \
+              -var="arc=${{ matrix.arrays.arc }}" \
+              -var="binary_name=${{ matrix.arrays.binaryName }}" \
               -var="local_stack_host_name=${{ needs.StartLocalStack.outputs.local_stack_host_name }}" \
               -var="s3_bucket=${S3_INTEGRATION_BUCKET}" \
               -var="key_name=${KEY_NAME}" \
-              -var="test_name=cw-integ-test-al2" \
+              -var="github_sha=${GITHUB_SHA}" \
```
khanhntd, line 685: **Nit pick**: Its duplication here https://github.com/aws/amazon-cloudwatch-agent/pull/526/files#diff-9b9f9a38b889c9dcb55ba0330baa69c8fdd62ac1100b6eb59b6fc44008f5d320R676

integration/test/performancetest/transmitter.go
okankoAMZ: It doesn't have to end at the same time. So when a release is published this function is called in order to mark the (already existing at that point) commit the release is connected to as isReleased. Since the commit is already existing I need to use the update function. The optimistic lock is there to prevent if a releases happened while performance test is still uploading. This is an edge case but an edge case I ran it to when I was running this for previous release, where I updated releases right after the performance test. In addition to overcoming this edge case, this optimistic lock also ensures that we will never overwrite an commit's data. You can imagine this as a safety feature and--from my understanding--an industry standard practice.

.github/workflows/releaseTest.yml
okankoAMZ: I think this script can stay until the separate CI is develop. I am not sure I understand how you can move it to integrationTest, since how would it be triggered on a release?

integration/test/performancetest/transmitter.go
```diff
@@ -153,12 +197,194 @@ func (transmitter *TransmitterAPI) TableExist() (bool, error) {
...
+            ":hash": &types.AttributeValueMemberS{Value: hash},
+        },
+        ExpressionAttributeNames: map[string]string{
+            "#hash": "Hash",
+        },
+        ScanIndexForward: aws.Bool(true), // true or false to sort by "date" Sort/Range key ascending or descending
+    })
+	if err != nil {
+        panic(err)
+    }
```
SaxyPandaBear, line 386: why are we panicking here?

SaxyPandaBear (commented):
I didn't want to spend time commenting on all the nitpicks, but in general, I think you should revisit a lot of the variable and function names because some of them are not descriptive or give enough indication of what they are / how they're used. Also, the file formatting is off in a bunch of places so try formatting the files. I might have missed something because I was trying to skim through it pretty quickly.

integration/test/performancetest/performance_query_utils.go
```diff
@@ -177,16 +179,19 @@ func GetPerformanceMetrics(instanceId string, agentRuntime, logNum, tps int, age
 	packet[PARTITION_KEY] = time.Now().Year()
 	packet[HASH] = os.Getenv(SHA_ENV) //fmt.Sprintf("%d", time.Now().UnixNano())
 	packet[COMMIT_DATE],_ = strconv.Atoi(os.Getenv(SHA_DATE_ENV))
-
+	packet["isRelease"] = false
 	//add test metadata
-	packet["NumberOfLogsMonitored"] = logNum
-	packet["TPS"] = tps
+	packet[TEST_HASH] = uuid.New().String()
```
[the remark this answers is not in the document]
okankoAMZ: I explained this more in the docs, but testHash is used for version control. We achieve optimistic lock by giving each packet a unique id (UIUD) as test hash, so whenever an item is updated or added with that packet, the item will have the that packet’s test hash.  This test hash then used to see if an item has changed when updating an item. If you would prefer I can rename it to version.

integration/test/performancetest/performance_test.go
```diff
@@ -78,8 +77,10 @@ func TestPerformance(t *testing.T) {
 	}
 
 	//run tests
+	//@ASK If we should move tps values to seperate config testCases
 	for _, tps := range tpsVals {
 		t.Run(fmt.Sprintf("TPS run: %d", tps), func(t *testing.T) {
+			os.Setenv("TPS",fmt.Sprintf("%d",tps))
```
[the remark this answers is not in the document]
okankoAMZ: I wanted to use env vars, instead of passing in to make it easier to build upon. However, if this looks complicated I can change it to a parameter.

integration/test/performancetest/performance_test.go
```diff
@@ -105,7 +106,7 @@ func TestPerformance(t *testing.T) {
 			if data == nil {
 				t.Fatalf("No data")
 			}
-
+			fmt.Printf("%v \n",data)
```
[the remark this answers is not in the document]
okankoAMZ: It is mostly for debugging but if there is an error in the code, it is essential to understand what went wrong. Personally, I would keep it.

integration/test/performancetest/performance_test.go
SaxyPandaBear: There isn't enough benefit for making that change.

.github/workflows/integrationTest.yml
```diff
@@ -161,14 +162,19 @@ jobs:
...
 
       - name: Echo test plan matrix
         run: |
           echo ${{ steps.set-matrix.outputs.ec2_linux_matrix }}
+          echo ${{ steps.set-matrix.outputs.ec2_performance_matrix}}
           echo ${{ steps.set-matrix.outputs.ec2_windows_matrix }}
           echo ${{ steps.set-matrix.outputs.ecs_fargate_matrix }}
+      - name: Echo performance matrix
+        run: |
+          echo ${{ steps.set-matrix.outputs.ec2_performance_matrix}}
```
SaxyPandaBear, line 177: this prints out the perf test matrix twice

.github/workflows/integrationTest.yml
```diff
@@ -638,52 +648,51 @@ jobs:
...
             fi
 
       #This is here just in case workflow cancel
       - name: Terraform destroy
-        if: ${{ cancelled() && steps.performance-tracking.outputs.cache-hit != 'true' }}
+        if: ${{ cancelled() && steps.ec2-linux-integration-test.outputs.cache-hit != 'true' }}
         uses: nick-invision/retry@v2
         with:
-          max_attempts: 3
+          max_attempts: 1
```
SaxyPandaBear, line 695: If we're only allowing one attempt, why even have the retry? Why are we reducing this to 1? This is also for just the terraform destroy, so I don't see the issue.

integration/test/performancetest/transmitter.go
```diff
@@ -153,12 +197,182 @@ func (transmitter *TransmitterAPI) TableExist() (bool, error) {
...
+/*
+PacketMerger()
+Desc:
+	This function updates the currentPacket with the unique parts of newPacket and returns in dynamo format
+Params:
+	newPacket: this is the agentData collected in this test
+	currentPacket: this is the agentData stored in dynamo currently
+*/
+func (transmitter * TransmitterAPI) PacketMerger(newPacket map[string]interface{}, currentPacket map[string]interface{}) (map[string]types.AttributeValue,error){
+	testSettings := fmt.Sprintf("%s-%s",os.Getenv(PERFORMANCE_NUMBER_OF_LOGS),os.Getenv(TPS))
```
SaxyPandaBear: This is the only place that uses `os.Getenv(TPS)`. This string should just be passed in as a parameter. We don't have any reason to persist the TPS or number of log files as an env var.

integration/test/performancetest/performance_test.go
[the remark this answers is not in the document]
SaxyPandaBear: `GetPerformanceMetrics` passes in tps and numLogs as parameters. This overcomplicates the problem, and is more error prone because it depends on nothing else in the environment/process overriding the variable. This should just be passed in as a parameter.

integration/test/performancetest/performance_test.go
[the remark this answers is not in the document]
SaxyPandaBear: Can you elaborate on when this specific print would be helpful for debugging?

integration/test/performancetest/transmitter.go
```diff
@@ -4,16 +4,29 @@ import (
...
 
+
+const (
+	UPDATE_DELAY_THRESHOLD = 60// this is how long we want to wait for random sleep in seconds
+	MAX_ATTEMPTS = 5 // number of attemps before we stop retrying to update
+	/*
+	!Warning: if this value is less than 25 there is a risk of testCases being lost.
+	This will only happen if all test threads and at the same time and get the same 
+	sleep value after first attempt to add ITEM
+	*/
```
SaxyPandaBear: This comment block should be moved **above** `UPDATE_DELAY_THRESHOLD`. When I read this, I don't know what it refers to.

integration/test/performancetest/transmitter.go
```diff
@@ -153,12 +197,182 @@ func (transmitter *TransmitterAPI) TableExist() (bool, error) {
...
+		return "",err
+	}
+	fmt.Println("SendItem Completed")
 	return sentItem, err
+
 }
+/*
+PacketMerger()
+Desc:
+	This function updates the currentPacket with the unique parts of newPacket and returns in dynamo format
```
SaxyPandaBear, line 239: Typically, you wouldn't combine functionality like this. When you do a data transformation like this, you don't want to also convert it into some other format. The function that calls this one should be responsible with translating it into how that parent function needs to send it where it needs to go. 

This function has nothing to do with DynamoDB until the very end when you convert it into the AttributeValue map.

integration/test/performancetest/transmitter.go
```diff
@@ -153,12 +197,182 @@ func (transmitter *TransmitterAPI) TableExist() (bool, error) {
...
+		n_expression := len(targetAttributes)
+		i :=0
+		for attribute, value := range targetAttributes{
+			expressionName := ":" +strings.ToLower(attribute)
+			expression += fmt.Sprintf("set %s = %s",attribute,expressionName)
+			expressionAttributeValues[expressionName] = value
+			if(n_expression -1 >i){
+				expression += "and"
+			}
+			i++
```
SaxyPandaBear: BTW you can use `strings.Join` to concatenate a bunch of strings.
```go
exprs := []string{"a", "b", "c"}
res := strings.Join(exprs, " and ")
```
https://go.dev/play/p/I3mTZzbyQ_6

I don't really mind how it's written now since the issue is that it's a map, but wanted to bring it up in case you didn't know that the `Join` function exists.

integration/test/performancetest/transmitter.go
SaxyPandaBear: This is such a small bit of code there's no reason to make a separate function

integration/test/performancetest/transmitter.go
okankoAMZ: I wanted to have it that so that it will be easier other configs, by just adding another os.Getenv into that sprintf

integration/test/performancetest/transmitter.go
khanhntd: No the idea was too avoid using `errors.As` less as possible. Therefore, only return err and let's other function dealing with the specific error instead of panic the error here.

SaxyPandaBear: The point of this code is to capture and handle an expected error. I'm not sure why it would be appropriate to return an error here.

SaxyPandaBear: We don't need to populate `ae` with `errors.As`. We only need to verify that the conditional check failure happened. You should change the check to `errors.Is`

.github/workflows/releaseTest.yml
SaxyPandaBear: When will that CI/CD workflow be worked on? It's not hard to switch from one way or the other. There's no reason to block this PR on the eventuality of something that we say we're going to do in the future

integration/test/performancetest/transmitter.go
SaxyPandaBear: Nvm. you still need `errors.As`

khanhntd: Maybe its only my design so I leave the choice to you guys.

SaxyPandaBear (dismissed):

.github/workflows/integrationTest.yml
```diff
@@ -638,52 +645,51 @@ jobs:
...
+              -var="performance_number_of_logs=${{ matrix.arrays.performance_number_of_logs}}"\
+              -var="test_dir=${{ matrix.arrays.test_dir }}" ; then terraform destroy -auto-approve
             else
               terraform destroy -auto-approve && exit 1
             fi
 
       #This is here just in case workflow cancel
       - name: Terraform destroy
-        if: ${{ cancelled() && steps.performance-tracking.outputs.cache-hit != 'true' }}
+        if: ${{ cancelled() && steps.ec2-linux-integration-test.outputs.cache-hit != 'true' }}
```
khanhntd, line 689:
```suggestion
        if: ${{ cancelled() && steps.performance-tracking.outputs.cache-hit != 'true' }}
```

go.mod
```diff
@@ -252,17 +239,13 @@ require (
...
 	github.com/xeipuuv/gojsonpointer v0.0.0-20180127040702-4e3ac2762d5f // indirect
 	github.com/xeipuuv/gojsonreference v0.0.0-20180127040603-bd5ef7bd5415 // indirect
-	github.com/yuin/gopher-lua v0.0.0-20220413183635-c841877397d8 // indirect
 	github.com/yusufpapurcu/wmi v1.2.2 // indirect
 	go.opencensus.io v0.23.0 // indirect
-	go.opentelemetry.io/collector/model v0.48.0 // indirect
 	go.uber.org/atomic v1.9.0 // indirect
 	go.uber.org/goleak v1.1.12 // indirect
-	go.uber.org/multierr v1.7.0 // indirect
+	go.uber.org/multierr v1.6.0 // indirect
```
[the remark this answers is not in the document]
khanhntd: +1

sethAmazon (dismissed):

sethAmazon (approved):

PR #540, opened 2022-08-04
Performance Tracking convert memory byte to megabyte

gmealy1:
# Description of the issue
The data returned for memory used is returned in bytes, which can be difficult to interpret easily.

# Description of changes
Converts the memory values returned from GetMetricData from bytes to megabytes.

# License

# Tests
Ran as integration test.

# Requirements
2. Run `make linter`

### DIFF
--- a/integration/test/performancetest/performance_query_utils.go
+++ b/integration/test/performancetest/performance_query_utils.go
@@ -198,6 +198,12 @@ func GetPerformanceMetrics(instanceId string, agentRuntime, logNum, tps int, age
 
 	//add actual test data with statistics
 	for _, result := range metrics.MetricDataResults {
+		//convert memory bytes to MB
+		if (*result.Label == "procstat_memory_rss") {
+			for i, val := range(result.Values) {
+				result.Values[i] = val / (1000000)
+			}
+		}
 		stats:= CalcStats(result.Values)
 		testMetricResults[*result.Label] = stats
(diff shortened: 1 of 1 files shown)


khanhntd (approved):
LGTM.

integration/test/performancetest/performance_query_utils.go
```diff
@@ -198,6 +198,12 @@ func GetPerformanceMetrics(instanceId string, agentRuntime, logNum, tps int, age
 
 	//add actual test data with statistics
 	for _, result := range metrics.MetricDataResults {
+		//convert memory bytes to MB
+		if (*result.Label == "procstat_memory_rss") {
+			for i, val := range(result.Values) {
+				result.Values[i] = val / (1000000)
```
khanhntd, line 204: Should this be 1024 * 1024 = 1048576?

gmealy1: I believe they are both valid depending on the context, but for most things `1000^2` is used. https://en.wikipedia.org/wiki/Megabyte
And then for some things like operating systems, 1024^2 is used. 🤷 
But we decided to use MB because it is a more meaningful measure of memory used and easier to know at a glance roughly how much memory the agent is using rather than trying to count how many zeros there are to see if it is 1 MB, 10MB or 100MB, etc.

PR #542, opened 2022-08-05
Update cwa ctl agent stop to use systemctl instead of service cmd

sky333999:
# Description of the issue
Currently, CWA ctl stop cmd uses `service` cmd instead of `systemctl`. For older linux kernels that come bundled with an older version of `service` cmd, it looks for the `.service` file in a diff location and is unable to properly redirect to `systemctl`.

# Description of changes
To avoid relying on `service` cmd properly redirecting to `systemctl`, this change updates to directly use `systemctl`. This brings the stop action in sync with other actions such as disable and runStatus.

# License

# Tests
Tested `install CWA -> start CWA -> uninstall CWA` on the following EC2s with and without change:
* Amazon Linux 2 Kernel 5.10 AMI 2.0.20220719.0 x86_64 HVM gp2
* Amazon Linux 2 Kernel 4.14 AMI 2.0.20220719.0 x86_64 HVM gp2
* Red Hat Enterprise Linux 8

Output of uninstall before change:
```
$ sudo rpm -e --nodeps amazon-cloudwatch-agent
cwagent-otel-collector has already been stopped
Redirecting to /bin/systemctl stop amazon-cloudwatch-agent.service
Removed symlink /etc/systemd/system/multi-user.target.wants/amazon-cloudwatch-agent.service.
```
Output of uninstall after change:
```
$ sudo rpm -e --nodeps amazon-cloudwatch-agent
cwagent-otel-collector has already been stopped
Removed symlink /etc/systemd/system/multi-user.target.wants/amazon-cloudwatch-agent.service.
```

# Requirements
2. Run `make linter`

### DIFF
--- a/packaging/dependencies/amazon-cloudwatch-agent-ctl
+++ b/packaging/dependencies/amazon-cloudwatch-agent-ctl
@@ -158,7 +158,7 @@ agent_stop() {
      fi
 
      if [ "${SYSTEMD}" = 'true' ]; then
-          service "${agent_name}" stop || return
+          systemctl stop "${agent_name}.service" || return
      else
           stop "${agent_name}" || return
      fi


khanhntd (approved):
LGTM.

packaging/dependencies/amazon-cloudwatch-agent-ctl
```diff
@@ -158,7 +158,7 @@ agent_stop() {
      fi
 
      if [ "${SYSTEMD}" = 'true' ]; then
-          service "${agent_name}" stop || return
```
SaxyPandaBear, line 161: Which operating systems rely on `service`? Do we still officially support those versions?

SaxyPandaBear (dismissed):

packaging/dependencies/amazon-cloudwatch-agent-ctl
sky333999: To confirm, is the question which of our officially supported operating systems rely _solely_ on `service` and not on systemd's `systemctl` out of the box?

SaxyPandaBear: Solely

sky333999: Got it. I could check but I don't think those would be impacted by this change? The current code path for the line being edited is only if `"${SYSTEMD}" = 'true'`.
So those operating systems if any are likely already using the else path and will continue to do so.

SaxyPandaBear: But the check looks for upstart (the old one) first https://github.com/aws/amazon-cloudwatch-agent/blob/a98d8a2e1040ae8c3c0dd80d367da0486651a8d3/packaging/dependencies/amazon-cloudwatch-agent-ctl#L480-L483 unless I'm misconstruing upstart with something else.

sky333999: Thats correct. However when systemd is the actual service manager configured, /sbin/init is a symlink to systemd even though upstart's service command could be available.
```
$ ls -ltr /sbin/init
lrwxrwxrwx 1 root root 22 Jul 21 19:46 /sbin/init -> ../lib/systemd/systemd
$ which service
/usr/sbin/service
$ /sbin/init --version 2>/dev/null | grep -c upstart
0
``` 

But better to be safe so would the best way to go about this be: For the linux distributions in the list of supported operating systems [here](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/Install-CloudWatch-Agent.html), I spin up an ec2 and confirm if /sbin/init is pointing to upstart or systemd?

SaxyPandaBear: That's what I would do, yeah

sky333999: | OS | Is systemd default? | Is stop command recogninzed? |
| --- | --- | --- |
| Amazon Linux version 2014.03.02 or later | No | Yes |
| Amazon Linux 2 | Yes | N/A |
| Ubuntu Server versions 20.04, 18.04, 16.04 | Yes | N/A |
| Ubuntu Server version 14.04 | No | Yes |
| CentOS version 8 stream and version 7 | Yes | N/A |
| Red Hat Enterprise Linux (RHEL) versions 8, 7.7, 7.6, 7.5, 7.4, 7.2, and 7.0 | Yes | N/A |
| Debian version 10 | Yes | N/A |
| SUSE Linux Enterprise Server (SLES) version 15 and version 12 | Yes | N/A |
| Oracle Linux versions 7.8, 7.6, and 7.5 | Yes | N/A |

I also came across https://en.wikipedia.org/wiki/Systemd#Adoption which gives a good overview of when systemd was released as a default for an OS.

Keeping the following in mind, https://github.com/aws/amazon-cloudwatch-agent/blob/a98d8a2e1040ae8c3c0dd80d367da0486651a8d3/packaging/dependencies/amazon-cloudwatch-agent-ctl#L480-L483 if I were to summarize the impact:
* Though it checks for upstart first in the `if` clause, for OSes that have systemd as their default, /sbin/init is a symlink to systemd and hence the `if` clause would evaluate to false ===> For OSes with systemd as their default, `SYSTEMD` will be true.
* For OSes where systemd is not their default, the `if` clause will evaluate to true and `SYSTEMD` will be false. So the `stop` command (i.e. /sbin/stop) would be used instead.
* In the table above, I verified that either systemd is the default and for the few where it isnt, confirmed that the stop command is recognized.

SaxyPandaBear (approved):

Issue #554, opened 2022-08-12
"retention_in_days" potentially does not work


adam-mateen:
**Describe the bug**
CWA allows "retention_in_days" to be set for each item in the "collect_list".
Example:
```
.
.
.
    "logs_collected": {
      "files": {
        "collect_list": [
          {
            "file_path": "/tmp/foo",
            "log_group_name": "MyLogGroup1",
            "log_stream_name": "MyLogStreamFoo",
            "retention_in_days": 14
          },
          {
            "file_path": "/tmp/bar",
            "log_group_name": "MyLogGroup1",
            "log_stream_name": "MyLogStreamBar",
            "retention_in_days": 14
          }
        ]
      }
    },
.
.
.
```

However after translation to TOML only 1 of the log-files with have `retention_in_days = 14` and the other will have `retention_in_days = -1`.

If neither file gets created, or if the file with `= 14` never actually gets created, the agent will not make the API call to set the retention for this log group.

**Steps to reproduce**
Create a JSON config for the agent containing the snippet above.

**What did you expect to see?**
I expected retention_in_days to get set for my log group.

Ideally, whenever the agent is started it should set the retention for the configured log groups regardless of whether files exist matching the file_paths.

**What did you see instead?**
Retention was unchanged (default = never expire).

**What version did you use?**
`v1.247352.0`

**What config did you use?**
See above

**Environment**
OS: AL2

**Additional context**

SaxyPandaBear:
Fixed in #556

PR #556, opened 2022-08-15
Fix duplicate retention case

aateeqi:
# Description of the issue
[todo] https://github.com/aws/amazon-cloudwatch-agent/issues/554

# Description of changes
When the same retention is set multiple times for a log group, retention will be set as long as any file_path is created for that log group.

# License

# Tests
Unit tests + manual tests linked below

# Requirements
2. Run `make linter`
3. 

image

Screen Shot 2022-08-15 at 10 22 41 AM

### DIFF
--- a/logs/logs.go
+++ b/logs/logs.go
@@ -56,6 +56,7 @@ type LogAgent struct {
 	backends    map[string]LogBackend
 	destNames   map[LogDest]string
 	collections []LogCollection
+	retentionAlreadySet map[string]bool
 }
 
 func NewLogAgent(c *config.Config) *LogAgent {
@@ -63,6 +64,7 @@ func NewLogAgent(c *config.Config) *LogAgent {
 		Config:    c,
 		backends:  make(map[string]LogBackend),
 		destNames: make(map[LogDest]string),
+		retentionAlreadySet: make(map[string]bool),
 	}
 }
 
@@ -106,7 +108,15 @@ func (l *LogAgent) Run(ctx context.Context) {
 						log.Printf("E! [logagent] Failed to find destination %v for log source %v/%v(%v) ", dname, src.Group(), src.Stream(), src.Description())
 						continue
 					}
-					dest := backend.CreateDest(src.Group(), src.Stream(), src.Retention())
+					retention := src.Retention()
+					if retention > 0 && l.retentionAlreadySet[src.Group()] == true {
+						log.Printf("D! [logagent] Retention already set for log group %v/%v retention:%v) ", src.Group(), src.Stream(), src.Retention())
+						retention = -1
+					} else if retention > 0 {
+						log.Printf("I! First time setting retention for log group %v/%v, update map to avoid setting twice) ", src.Group(), src.Stream())
+						l.retentionAlreadySet[src.Group()] = true
+					}
+					dest := backend.CreateDest(src.Group(), src.Stream(), retention)
 					l.destNames[dest] = dname
 					log.Printf("I! [logagent] piping log from %v/%v(%v) to %v with retention %v", src.Group(), src.Stream(), src.Description(), dname, src.Retention())
 					go l.runSrcToDest(src, dest)
--- a/logs/logs.go
+++ b/logs/logs.go
@@ -52,19 +52,19 @@ type LogDest interface {
 
 // LogAgent is the agent handles pure log pipelines
 type LogAgent struct {
-	Config              *config.Config
-	backends            map[string]LogBackend
-	destNames           map[LogDest]string
-	collections         []LogCollection
-	retentionAlreadySet map[string]bool
+	Config                    *config.Config
+	backends                  map[string]LogBackend
+	destNames                 map[LogDest]string
+	collections               []LogCollection
+	retentionAlreadyAttempted map[string]bool
 }
 
 func NewLogAgent(c *config.Config) *LogAgent {
 	return &LogAgent{
-		Config:              c,
-		backends:            make(map[string]LogBackend),
-		destNames:           make(map[LogDest]string),
-		retentionAlreadySet: make(map[string]bool),
+		Config:                    c,
+		backends:                  make(map[string]LogBackend),
+		destNames:                 make(map[LogDest]string),
+		retentionAlreadyAttempted: make(map[string]bool),
 	}
 }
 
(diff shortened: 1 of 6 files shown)


plugins/outputs/cloudwatchlogs/pusher.go
```diff
@@ -383,7 +383,7 @@ func (p *pusher) putRetentionPolicy() {
 				p.Log.Errorf("Unable to put retention policy for log group %v: %v ", p.Group, err)
 			}
 		} else {
-			p.Log.Debugf("successfully updated log retention policy for log group %v", p.Group)
+			p.Log.Debugf("successfully updated log retention policy for log group %v with log stream %v", p.Group, p.Stream)
```
SaxyPandaBear, line 386: I don't think that logging the stream is useful because the update happens to the overarching log group

aateeqi: agreed, used this for testing

logs/logs.go
```diff
@@ -56,13 +56,15 @@ type LogAgent struct {
...
 	collections []LogCollection
+	retentionAlreadySet map[string]bool
 }
 
 func NewLogAgent(c *config.Config) *LogAgent {
 	return &LogAgent{
 		Config:    c,
 		backends:  make(map[string]LogBackend),
 		destNames: make(map[LogDest]string),
+		retentionAlreadySet: make(map[string]bool),
```
khanhntd, line 67: I'm not fond of use cache with map since the more entries we add, the [more bytes](https://stackoverflow.com/questions/15313105/memory-overhead-of-maps-in-go) we have to consume unless we have any mechanism to delete the entries the map after certain amount of time. Therefore, I'm still standby create the resources beforehand. (correct me if I am wrong)

aateeqi: @khanhntd I think that it would be a concern if we're storing a lot of data but we are only storing the log group name and a boolean. In a sample `map[string]int` of 100 initial entries the memory usage is [4kb](https://stackoverflow.com/questions/46278003/how-much-memory-do-golang-maps-reserve#:~:text=A%20map%20of%20type%20map,7%20if%20not%20specified%20explicitly.). We should be okay in terms of memory usage.

khanhntd: If that's the case then this is fine. Maybe its only my design perspective but this would introduce special code for only this use case (in this case, the `retentionAlreadySet`) but yeah, if everyone is fine with the current approach then I'm good with it.

SaxyPandaBear:
Is there backwards compatibility with the old way that we configured?

SaxyPandaBear (commented):
We can't afford to break backwards-compatibility. What happens if we take an example config where there's multiple configs pointing to the same log group and only one of the configs has a retention policy (existing behavior)? Can you build this and manually verify that behavior works?

aateeqi:
> We can't afford to break backwards-compatibility. What happens if we take an example config where there's multiple configs pointing to the same log group and only one of the configs has a retention policy (existing behavior)? Can you build this and manually verify that behavior works?


Yes backwards compatibility still maintained: if the customer has multiple streams for a log_group and retention only configured in one, it will still successfully update the retention. Note the 2 configs pointing to the same log group `amazon-cloudwatch-agent.log` with only 1 of them having retention in days.
image
image

logs/logs.go
```diff
@@ -106,7 +108,15 @@ func (l *LogAgent) Run(ctx context.Context) {
 						log.Printf("E! [logagent] Failed to find destination %v for log source %v/%v(%v) ", dname, src.Group(), src.Stream(), src.Description())
 						continue
 					}
-					dest := backend.CreateDest(src.Group(), src.Stream(), src.Retention())
+					retention := src.Retention()
+					if retention > 0 && l.retentionAlreadySet[src.Group()] == true {
```
SaxyPandaBear, line 112:
```suggestion
					if retention > 0 && l.retentionAlreadySet[src.Group()] {
```
Don't need to assert true == true

SaxyPandaBear (dismissed):

logs/logs.go
```diff
@@ -52,17 +52,19 @@ type LogDest interface {
...
 
 func NewLogAgent(c *config.Config) *LogAgent {
 	return &LogAgent{
-		Config:    c,
-		backends:  make(map[string]LogBackend),
-		destNames: make(map[LogDest]string),
+		Config:              c,
+		backends:            make(map[string]LogBackend),
+		destNames:           make(map[LogDest]string),
+		retentionAlreadySet: make(map[string]bool),
```
khanhntd, line 67: **Nit** Should we better to use struct{} instead of bool here @SaxyPandaBear ?

logs/logs.go
```diff
@@ -106,7 +108,15 @@ func (l *LogAgent) Run(ctx context.Context) {
 						log.Printf("E! [logagent] Failed to find destination %v for log source %v/%v(%v) ", dname, src.Group(), src.Stream(), src.Description())
 						continue
 					}
-					dest := backend.CreateDest(src.Group(), src.Stream(), src.Retention())
+					retention := src.Retention()
+					if retention > 0 && l.retentionAlreadySet[src.Group()] {
+						log.Printf("D! [logagent] Retention already set for log group %v/%v retention:%v) ", src.Group(), src.Stream(), src.Retention())
```
khanhntd, line 113: **Nit pick**: Same comment as Andrew before. I don't think log stream is useful for logging since retention_in_days only happen at log group level

logs/logs.go
SaxyPandaBear: It doesn't make enough difference for me to suggest changing it. Changing this to a `map[string]struct{}` sacrifice what amounts to a couple bytes of memory used to hold a boolean in a map for readability. 
```go
m := make(map[string]struct{})
m["a"] = struct{}{}
if _, ok := m["b"]; ok {
    fmt.Println("hello")
} else if _, ok := m["a"]; ok {
    fmt.Println("hello there")
} else {
    fmt.Println("bye")
}
```
Switching to a `struct{}` complicates the code.

logs/logs.go
SaxyPandaBear, line 113: Minor thing - I think `%v` is a good catch-all, but we know the types of these values so we can be more direct:
```suggestion
						log.Printf("D! [logagent] Retention already set for log group %s/%s retention:%d) ", src.Group(), src.Stream(), src.Retention())
```

logs/logs.go
```diff
@@ -106,7 +108,15 @@ func (l *LogAgent) Run(ctx context.Context) {
 						log.Printf("E! [logagent] Failed to find destination %v for log source %v/%v(%v) ", dname, src.Group(), src.Stream(), src.Description())
 						continue
 					}
-					dest := backend.CreateDest(src.Group(), src.Stream(), src.Retention())
+					retention := src.Retention()
```
SaxyPandaBear, line 111: Minor thing - There are a bunch of calls to `src.Group()`, `src.Stream()` and `src.Retention()`, and we could limit that by defining the group and stream up here like retention. I doubt that the difference would be noticeable at all between one or the other, but it's also not a difficult thing to swap out.

SaxyPandaBear (approved):

logs/logs.go
```diff
@@ -101,14 +103,26 @@ func (l *LogAgent) Run(ctx context.Context) {
...
 					}
-					dest := backend.CreateDest(src.Group(), src.Stream(), src.Retention())
+
+					if retention > 0 && l.retentionAlreadySet[logGroup] {
+						log.Printf("D! [logagent] Retention already set for log group %s, current retention %d", logGroup, retention)
+						retention = -1
+					} else if retention > 0 {
+						log.Printf("I! First time setting retention for log group %s, update map to avoid setting twice", logGroup)
+						l.retentionAlreadySet[logGroup] = true
+					}
```
SaxyPandaBear, line 122: I know it's late to ask for this, but I'd feel better if we added a unit test specifically for this? Not as part of the translation, but specifically this part where we are modifying a map in memory. I would pull out this chunk of code into a separate function. 

It's pretty straightforward so I can live without it, but on the flip side, it wouldn't be a lot of extra work to include tests around it.

SaxyPandaBear (approved):

SaxyPandaBear (approved):

PR #566, opened 2022-08-17
update statefile cleanup error log

SaxyPandaBear:
# Description of changes
An error from removing this file is benign, so it doesn't make sense to make it at error level, but should still distinguish it from regular logs since it is not expected to have a failure. Forgot to add the error in the log that gets emitted, so added that too.

# License

### DIFF
--- a/plugins/inputs/logfile/tailersrc.go
+++ b/plugins/inputs/logfile/tailersrc.go
@@ -76,6 +76,7 @@ type tailerSrc struct {
 	startTailerOnce sync.Once
 	cleanUpFns      []func()
 }
+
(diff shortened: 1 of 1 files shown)


aateeqi (approved):
LGTM, its a little odd that theres only a pre-log (L314) for deleting the state file but none for saving/final save. Wondering if there would be any value in adding that but i'll leave that to you.

SaxyPandaBear:
Saving the state happens so often, and doesn't provide any useful logging unless an issue occurs.

adam-mateen (approved):

PR #567, opened 2022-08-18
Performance Benchmark Website Internal Authentication

gmealy1:
# Description of the issue
The website is not ready to be deployed publicly but it is also not currently available internally.

# Description of changes
Adds authentication to lock the website out unless you have the internal beta-testing password. Every database request must be authenticated with the password.

# License

# Tests
Build and test locally to ensure the website functions properly.

# Requirements
2. Run `make linter`

### DIFF
--- a/.github/workflows/deploy-to-gh-pages.yml
+++ b/.github/workflows/deploy-to-gh-pages.yml
@@ -22,6 +22,7 @@ jobs:
         env:
           REACT_APP_GATEWAY: ${{ secrets.REACT_APP_GATEWAY }}
           REACT_APP_GATEWAY_API_KEY: ${{ secrets.REACT_APP_GATEWAY_API_KEY }}
+          REACT_APP_SALT: ${{ secrets.REACT_APP_SALT }}
 
 
       - name: Deploy
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "react-scripts": "^5.0.1",
     "react-scroll-section": "^3.1.0",
     "recharts": "^2.1.12",
+    "sjcl": "^1.0.8",
     "styled-components": "^5.3.5",
     "util": "^0.12.4",
     "web-vitals": "^2.1.4"
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-icons": "^4.4.0",
-    "react-scripts": "^5.0.1",
+    "react-scripts": "4.0.3",
     "react-scroll-section": "^3.1.0",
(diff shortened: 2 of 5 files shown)


.github/workflows/deploy-to-gh-pages.yml
```diff
@@ -16,7 +16,8 @@ jobs:
 
       - name: Install and Build
         run: |
-          npm install -f 
+          npm config set legacy-peer-deps true
```
sethAmazon, line 19: legacy?

package.json
```diff
@@ -19,9 +19,10 @@
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
     "react-icons": "^4.4.0",
-    "react-scripts": "^5.0.1",
+    "react-scripts": "4.0.3",
```
sethAmazon, line 22: Version down?

sethAmazon:
Do we really need all this? Could we just use a static pass. Like what is the worse thing if someone gets into this site?

.github/workflows/deploy-to-gh-pages.yml
gmealy1: Worked with Ameen for this change. It fixes the need to force install because of a conflict with required and currently used versions of React. IIRC current React version is higher than a version MUI needs, so it was erroring and needed to force install everything.

package.json
gmealy1: There's a build error due to an unresolved path that is because of a change in the newer react scripts, so it was not able to find the crypto library in sjcl. I tried a bunch of solutions but this was the last and only one I could get to work.

src/helpers/reciever.js
```diff
@@ -270,7 +274,13 @@ class Receiver {
         return response.data.body;
       })
       .catch(function (error) {
-        //console.log(error);
+        if (error.code === "ERR_BAD_REQUEST") {
+          localStorage.clear();
+          alert(
+            "Incorrect Password. This for internal use only. Please contact CloudWatch Agent team for access."
```
SaxyPandaBear, line 280: Should take this wording out. This isn't a beta access thing where we'd be granting access.

src/helpers/reciever.js
```diff
@@ -341,6 +351,13 @@ class Receiver {
   getLastUpdate() {
     return JSON.parse(localStorage.getItem(LAST_UPDATE));
   }
+  auth() {
+    let input = prompt("Please enter your beta testing password");
+    const salt = "LCnF&$^66VtWQxG#2wTk";
```
SaxyPandaBear, line 356: 1. Change the salt
2. Put it in secrets.

SaxyPandaBear (approved):
lgtm

PR #194, opened 2021-03-23
prom: Save job and instance label before relabel for look up metric type

pingleig:
# Description of the issue

Fix #193 

# Description of changes

Similar to #191  save label using relabel because they can get relabeled during relabel.

- remove old job relabel look up logic

# License

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

# Tests

- [ ] internal integ test
- [x] manual test with relabeled job and instance

```yaml
      - source_labels: [__meta_kubernetes_pod_name]
        action: replace
        target_label: instance
      - action: replace
        source_labels:
        - __meta_kubernetes_pod_controller_name
        target_label: job
```

PR #191, opened 2021-03-22
prom: Save __name__ before relabel for look up metric type

pingleig:
# Description of the issue

Fix #190 

# Description of changes

Save name before relabel using relabel.

# License

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

# Tests

- [x] manually
- [x] unit test
- [x] internal integ test

PR #568, opened 2022-08-19
Reduce memory usage for prometheus plugins

khanhntd:
# Description of the issue
After [introducing save name before relabel](https://github.com/aws/amazon-cloudwatch-agent/pull/191) and save [instance, job before relabel](https://github.com/aws/amazon-cloudwatch-agent/pull/194). There has been an increase in memory consumption and also CPU cores since it adds extra `relabel_configs` and `metric_relabel_config` (200MB for normal case when using job, instance before relabel and 500MB for normal case without any other metric_relabel_config or relabel_config with more than 10,000 metrics). However, process for creating label for [each metric within a scrape loop is a single thread](https://github.com/prometheus/prometheus/blob/main/scrape/scrape.go#L1533-L1535).

# Description of changes
* [Dropping unknown type for internal metric for avoid increase metrics processing](https://github.com/aws/amazon-cloudwatch-agent/pull/568/files#diff-65a97ef6671030fef079f66274ceca7db8498d40a2e1d9aa2bbc7ef488c5a73bR118-R120)
* Instead of saving `relabel` metrics for each Prometheus Metric, moving looking for metric type when appending metrics into batch 
*  [Utilize metadata in context for OpenTelemetry](https://github.com/prometheus/prometheus/blob/8b863c42dd956d35d18a7a0b39c89c86adf7cebf/scrape/scrape.go#L1154-L1161) for better memory consumption (since the context for holding metadata store is created within each Scrape Loop and is garbaged collect afterwards). For previous behavior, we have to [use `relabel_config` to save job and instance before relabel to look up metadata type](https://github.com/khanhntd/amazon-cloudwatch-agent/blob/master/plugins/inputs/prometheus_scraper/metrics_type_handler.go#L121-L134) 
* [Search up for metadataStore in context only once during the entire batch](https://github.com/aws/amazon-cloudwatch-agent/pull/568/files#diff-65a97ef6671030fef079f66274ceca7db8498d40a2e1d9aa2bbc7ef488c5a73bR75-R82)
* [Only use metric relabel name when target label is __name__](https://github.com/aws/amazon-cloudwatch-agent/pull/568/files#diff-3c7d6f237c9178a7dd0f586f6b377443db53b827854ee3869d7669741b3d83beR296-R307)

# License

# Note 
* Ensure backward compatibility  for the current CWAgent and my PR
* The number of metrics is quite small (<30). Therefore, it might not show the whole picture

# Tests
**For my PR**
* **Without relabel name and instance and job**: result in 50% decrease in CPU Utilization from start (with less memory utilization during runtime) and also 5% - 10% in memory consumption (2.05% maximum, average is around 1.95% but the minimum always down to 1.86%)
```
        global:
          scrape_interval: 1m
          scrape_timeout: 10s
        scrape_configs:
          - job_name: cwagent-ecs-file-sd-config
            sample_limit: 10000
            file_sd_configs:
              - files: [ "/tmp/cwagent_ecs_auto_sd.yaml" ]
```
image

* **With relabel name and instance and job**: mostly same pattern with current CWAgent for memory utilization (with lesser maximum memory utilization pattern); however, decrease 20% in CPU Utilization during initialization
```        
       global:
          scrape_interval: 1m
          scrape_timeout: 10s
        scrape_configs:
          - job_name: cwagent-ecs-file-sd-config
            sample_limit: 10000
            file_sd_configs:
              - files: [ "/tmp/cwagent_ecs_auto_sd.yaml" ]
            relabel_configs:
              - source_labels: [__address__]
                replacement: job_replacement
                target_label: job
            metric_relabel_configs:
              - source_labels: [ __address__ ]
                replacement: instance_replacement
                target_label: instance
              - source_labels: [__name__]
                regex: "memcached_commands_total(.*)"
                target_label: __name__
                replacement: "memcached_commands"
```

image
image
image

**For the current CWAgent performance:**
image
image
image
image

# Requirements
2. Run `make linter`


plugins/inputs/prometheus_scraper/metrics_receiver.go
```diff
@@ -106,7 +64,71 @@ func (ma *metricAppender) Rollback() error {
...
+	// Each new scrape will create a context hold metadataStore. Therefore, the same context will be used
+	// by all metrics in the same batch. So we only need to fetch the metadataStore once
+	if ma.isNewBatch {
+		metadataCache, err := getMetadataCache(ma.ctx)
+		if err != nil {
+			return err
+		}
+		ma.isNewBatch = false
+		ma.mc = metadataCache
+	}
```
SaxyPandaBear, line 82: Why not just update the `Appender` function to get the metadata cache when creating the appender?

plugins/inputs/prometheus_scraper/metrics_receiver.go
```diff
@@ -106,7 +64,71 @@ func (ma *metricAppender) Rollback() error {
...
+		metricTags = ls.WithoutLabels(model.MetricNameLabel, savedScrapeNameLabel).Map()
+		metricMetadata = metadataForMetric(metricNameBeforeRelabel, ma.mc)
+	} else {
+		metricTags = ls.WithoutLabels(model.MetricNameLabel).Map()
+		metricMetadata = metadataForMetric(metricName, ma.mc)
+	}
+
+	if metricMetadata.Type == textparse.MetricTypeUnknown {
+		// The internal metrics sometimes will return with type unknown and we would not consider it as valid metric (only support Gauge, Counter, Summary)
+		// https://github.com/khanhntd/amazon-cloudwatch-agent/blob/master/plugins/inputs/prometheus_scraper/metrics_filter.go#L21-L48
```
SaxyPandaBear, line 116: same nit here, and this points to your fork

plugins/inputs/prometheus_scraper/metrics_receiver_test.go
```diff
@@ -1,112 +1,172 @@
-// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
-// SPDX-License-Identifier: MIT
+// Copyright The OpenTelemetry Authors
```
SaxyPandaBear, line 1: So this code is taken directly from the OTel receiver?

plugins/inputs/prometheus_scraper/start.go
```diff
@@ -288,44 +283,29 @@ func reloadConfig(filename string, logger log.Logger, rls ...func(*config.Config
...
-		metricRelabelConfigs := []*relabel.Config{
-			// __name__
-			{
-				Action:       relabel.Replace,
-				Regex:        relabel.MustNewRegexp("(.*)"),
-				Replacement:  "$1",
-				TargetLabel:  savedScrapeNameLabel,
-				SourceLabels: model.LabelNames{"__name__"},
-			},
+		if relabelMetricName {
```
SaxyPandaBear, line 296: So if any of the metrics have `__name__`, then all of the metrics get relabeled?

plugins/inputs/prometheus_scraper/metric_prometheus.go
```diff
@@ -0,0 +1,41 @@
...
+	return pm.metricType == string(textparse.MetricTypeGauge)
+}
+
+func (pm *PrometheusMetric) isHistogram() bool {
+	return pm.metricType == string(textparse.MetricTypeHistogram)
+}
+
+func (pm *PrometheusMetric) isSummary() bool {
+	return pm.metricType == string(textparse.MetricTypeSummary)
+}
```
SaxyPandaBear, line 41: Where are these functions used?

SaxyPandaBear (commented):
Great deep dive on the issue. I can't tell what the performance difference is, based on the screenshots though. Can you explain them more?

plugins/inputs/prometheus_scraper/metrics_receiver_test.go
khanhntd: Its part of copy paste for developing unit test. I learn some of their unit test and Prometheus unit test before able to write down this unit test so it is not taken directly from the OTEL receiver.

plugins/inputs/prometheus_scraper/start.go
khanhntd: Yes. To be more exactly, if the customer provides their configuration with target label as __name__, we would an extra label (a magic label) to store the original name before relabeling. OTEL does not allow the customers to [modify the metric name](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/prometheusreceiver/config.go#L198-L204) and this seems to be an issue in Prometheus too (since prometheus updates the  [metadata cache is only updated before apply relabeling](https://github.com/prometheus/prometheus/blob/e5478983c43cb839a77c144f953b9ea25783b995/scrape/scrape.go#L1280) so we have no control over it)

plugins/inputs/prometheus_scraper/metric_prometheus.go
khanhntd: It is used in the [Filter](https://github.com/khanhntd/amazon-cloudwatch-agent/blob/master/plugins/inputs/prometheus_scraper/metrics_filter.go#L21-L48) which drops the unsupported metrics and also used in [calculation](https://github.com/khanhntd/amazon-cloudwatch-agent/blob/master/plugins/inputs/prometheus_scraper/calculator.go#L24-L53)

**For more context in the delta calculation**: The reason why we would perform some calculation (e.g Delta with Counter Prometheus Type) since there is no support for Counter Type in CWMetrics. The only way we can do is calculate the difference and send it over to CW Backend.

plugins/inputs/prometheus_scraper/metrics_metadata.go
```diff
@@ -0,0 +1,79 @@
...
+
+	return &mCache{
+		metadata: metadata,
+	}, nil
+}
+
+func metadataForMetric(metricName string, mc MetadataCache) *scrape.MetricMetadata {
+	// Two ways to get metric type through metadataStore:
+	// * Use instance and job to get metadataStore. If customer relabel job or instance, it will fail
+	// * Use Context that holds metadataStore which is created within each scrape loop https://github.com/prometheus/prometheus/blob/main/scrape/scrape.go#L1154-L1161
```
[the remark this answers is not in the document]
khanhntd: Nice catch. Thanks for pointing it out ! I don't know if this documentation is easy to understand for new folks ( I hope it does)

khanhntd:
> I can't tell what the performance difference is, based on the screenshots though. Can you explain them more?

Sure and thanks too! As you already know, the OOM is caused because of [saving metrics name before relabel](https://github.com/aws/amazon-cloudwatch-agent/pull/191). Therefore, with my PR, I bench marking with these test case:
* The customer case without relabeling metrics name
* Relabeling metrics name

For the first test case which is the customer's use case, I notice the following,
* The CPU Utilization decreases from 2.15% to 1.11% (approximately 50% reduction when initialization CloudWatchAgent and small decrease during run time)
* The Memory Utilization is within the range of 1.86% to 2.05% (1.95% in average) while comparing with the current CloudWatchAgent is within the range of 1.95% to 2.15% (average of 2.05%) - be note that this is only a small number of metrics so it might different in customer environment
* Confirm with the customers that with my PR, their memory utilization  is reduced from 9GB to 2.5GB (which addition 500MB since iirc, their original states is 3GB)

For the second test case which is why we introduce the PR that causes OOM, I notice the following
* The CPU Utilization decreases from 2.15% to 1.70%  and a small decrease during run time). 
* The Memory Utilization is within the range of 1.95% to 2.15% (1.95% in average but less pattern in 2.15%) - be also note that this is only a small number of metrics so it might different in customer environment.

** Note**: The decrease in CPU Utilization might help the most since when trouble shooting with customers, increase CPU cores make the customer's pod stables around 9GB (since relabel metric might be a single thread and increase CPU would help Prometheus to access the metrics in memory more faster)

khanhntd:
Another update for internal customers
```
The original image with v346 (without being impacted by the OOM issue) and having 1 core is and 3GB memory limit
The original image with v347 (with being impacted by the OOM issue) and having 2 core is and 9GB memory limit
The fixed image without relabel rules and having 2 cores, 4GB memory limit and ~2.1GB memory usage
The fixed image with relabel rules and having 2 cores, 4GB memory limit and memory usage is ~20% more than without relabel rules.
```

ashevtsov-wawa:
@khanhntd I'm wondering if there are any plans to get this PR updated and merged any time soon?

PR #569, opened 2022-08-22
Merge cleaning aws resources

khanhntd:
# Description of the issue
Merge cleaning aws resources into one single wokrflow.

# License

# Tests
N/A

# Requirements
2. Run `make linter`


.github/workflows/clean-aws-resources.yml
```diff
@@ -0,0 +1,79 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT
+
+name: AMI Monthly Resources Cleaner
+
+on:
+  schedule:
+    - cron: "0 0 * * *" # Run Every Day At Midnight
```
SaxyPandaBear, line 8: This means we're cleaning the test AMIs every day

sethAmazon: I think under the hood. The code looks if the ami is older than 30 days if I'm not mistaken. @khanhntd please comment.

.github/workflows/clean-aws-resources.yml
```diff
@@ -0,0 +1,79 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT
+
+name: AMI Monthly Resources Cleaner
```
sethAmazon, line 4: This needs to be renamed.

.github/workflows/clean-aws-resources.yml
khanhntd: I don't see any-down side for removing AMIs every day except increase API calls and small costs. Overall, its acceptable for consolidate the whole workflow into one to less awareness of other workflow.

.github/workflows/clean-aws-resources.yml
khanhntd: Nice catch. Resolving it since itts in the latest commit.

.github/workflows/clean-aws-resources.yml
khanhntd: @sethAmazon Its 60 days https://github.com/aws/amazon-cloudwatch-agent/blob/master/integration/clean/clean_ami/clean_ami.go#L22-L23

SaxyPandaBear: > I don't see any-down side for removing AMIs with 60 day olds everyday except increase API calls and small costs.

Because they aren't related. Deleting AMIs should not impact terminating instances and vice versa.

khanhntd: Yes. True that's why even though we trigger the `clean-aws-resources` everyday, the AMI would only be deleted when its over 60 days old so it won't have any impact on the EC2 Instances or any integration test. Any other suggestion would be appreciated !

sethAmazon: I think this is okay. lgtm. @SaxyPandaBear if you disagree please comment

sethAmazon (approved):

.github/workflows/clean-aws-resources.yml
SaxyPandaBear: The jobs should have `continue-on-error: true`. Otherwise if one of these jobs fails, it will cancel the subsequent ones. 

I still don't think there is **ANY** value with merging the AMI cleaning job with the EC2 instance cleaning job, but the "cost" is negligible.

khanhntd: I would avoid using `continue-on-error` since the only different between the matrix array is region. If the code fails, then it does not make sense to continue on other region and I don't think of any scenario where one region does fail while other does not (unless its a new build region).

Other than that, I agree it would not provide much "impact value" other than cleaning up.

SaxyPandaBear: >  I don't think of any scenario where one region does fail while other does not

Throttling. Does our code gracefully handle API throttles? 

I suggested continue-on-error because if the AMI job fails, it will cancel the resource cleaning jobs.

khanhntd:
```
Does our code gracefully handle API throttles?
```
If talking about throttle, then its even harder though. If we consider the code for deleting as a multiple thread, then yes. Since its a single thread (with a single loop) for every deletion, then its hard to throttle. 

Example: Consider these cases:
- Clean-AMI: rarely throttle since its a single region (without matrix) and a single thread
- Clean dedicated-host: Same with AMI (single region)
- Clean dedicated host internal: multiple region but the hosts are described and released in single call; moreover, it is used with `fail-fast:false`. Therefore, it should not affect others job based on my understanding.

PR #573, opened 2022-08-24
No Need To Test Multiple Sub Version Of Linux OS

sethAmazon:
# Description of the issue
No need to test on multiple minor os version

# Description of changes
Remove minor versions

# Tests
N/A

# Requirements
2. Run `make linter`


ZhenyuTan-amz (approved):
LGTM

khanhntd:
I may have discussed this with the team but its better to use lower bound and upper bound of the OS !

PR #577, opened 2022-08-26
Update dependencies for nats-io/jwt and opencontainers/runc

taohungyang:
# Description of the issue
We have old `nats-io/jwt` version and old `opencontainers/runc` in our `go.sum` that need to be updated.

# Description of changes
- Bump the version of `opencontainers/runc` to v1.1.4
- The root cause of the old version of `nats-io/jwt` is from the `/go-kit/kit` pkg. While they recently merged the code change for fixing this, they haven't cut a release yet. Thus multiple pkgs depending on this has not been upgraded as well. We will replace this pkg with the hash that contains the code change for now until a new release from them is available.
- Made sure we have the required versions in the `go.sum` after this change.

# License

# Tests
make build && make release

# Requirements
2. Run `make linter`


ymtaye (approved):
LGTM!

go.mod
```diff
@@ -43,12 +43,15 @@ replace github.com/karrick/godirwalk v1.16.1 => github.com/karrick/godirwalk v1.
...
 replace github.com/docker/distribution => github.com/docker/distribution v2.8.1+incompatible
 
-replace github.com/opencontainers/runc => github.com/opencontainers/runc v1.1.0
+replace github.com/opencontainers/runc => github.com/opencontainers/runc v1.1.4
 
 // Prometheis messed up their library naming convention. v0.35.1 matches 2.35.1 prometheus version
 // Go says this is a downgrade, but this is the latest release as of 05/25/2022
 replace github.com/prometheus/prometheus => github.com/prometheus/prometheus v0.35.1-0.20220503184552-2381d7be5731
 
+// go-kit has the fix for nats-io/jwt/v2 merged but not released yet. Replacing this version for now until next release.
```
SaxyPandaBear, line 52: How are we keeping track of this change being released?

SaxyPandaBear (commented):
Just to be sure, did you run `go mod tidy`?

taohungyang:
> Just to be sure, did you run `go mod tidy`?

Yes. A lot of the change here were added/removed by `go mod tidy`

go.mod
taohungyang: That's a good question. How do we typically track the release of each package?

(Just some context here that other than the import we make directly, there are many other packages depending on this so that I didn't choose to put this version in the `require` block but replace it. I can imagine even after `go-kit` make the official release, it still depends on how soon each package owner bump the version so that we can get the old jwt version completely removed from go.sum without replacing it.

SaxyPandaBear: I don't think we've defined a strategy for how to track releases of dependencies. This also was an issue with cadvisor releases. Maybe something to discuss internally

SaxyPandaBear (approved):

PR #579, opened 2022-08-28
Add NVIDIA GPU Integration Test and Security Checking for binaries

khanhntd:
# Description of the issue
Add Security Check for Nvidia SMI on both Linux and Windows  and add NVIDIA GPU Integration Test for Linux and Windows

# Description of changes
* [Add Security Protection for NVIDIA binaries in Windows](https://github.com/aws/amazon-cloudwatch-agent/pull/579/files#diff-322362a6517808b82d63bc394a1107eff59e32f7964262191d14e065b7ca5f9eR17-R109) by confirm Administrator and Local System Account has the permission to protect those binaries
* [Add Security Protection for NVIDIA binaries in Linux](https://github.com/aws/amazon-cloudwatch-agent/pull/579/files#diff-6bb2c96113346ae579cb1dea1d4391f0b459cc9c4ef117d761c74170b461f8f4R14-R33) by confirm the owner has permission to write, read (for file) and write, read, execute.
* [Add util for running agent, stop agent, and execute custom command for agent in windows](https://github.com/aws/amazon-cloudwatch-agent/pull/579/files#diff-009a6a15d86c3d5020409c663e401e19ec3e50ec3086abf78cf98371ee0c4943)
* [Add Nvidia GPU in list of integration test for both windows and linux](https://github.com/aws/amazon-cloudwatch-agent/pull/579/files#diff-3cf6443aa9be684f83bdf5a8d8a34463d7f3a91c60ee7475da7fe5792cabc623R23-R25)
* [Add the configuration for both CPU metrics and NVIDIA_GPU ](https://github.com/aws/amazon-cloudwatch-agent/pull/579/files#diff-5cd87ac08bfa8625a546b31cca5101849c01acd6c262c5310daa40e0102a2cf3)runs in high resolution: to confirm there is no blocking resources, and no delaying in collecting metrics 

# License

# Tests
By executing integration test on my fork: https://github.com/khanhntd/amazon-cloudwatch-agent/actions/runs/3003711396
* Here is the result of terraform for Windows
image

* Here is the result of terraform for Linux
image

* Here is the expected metrics in CW

image
 
# Requirements
2. Run `make linter`


integration/terraform/ec2/win/main.tf
```diff
@@ -36,41 +36,65 @@ resource "aws_instance" "cwagent" {
   vpc_security_group_ids      = [aws_security_group.ec2_security_group.id]
   associate_public_ip_address = true
   get_password_data           = true
+  user_data                   = <<EOF
+<powershell>
+Write-Output "Install OpenSSH and Firewalls which allows port 22 for connection"
```
ZhenyuTan-amz, line 41: Curious, is this the only way to allow port 22 connection? Can this be setup upon instance launching?

SaxyPandaBear: @sethAmazon didn't you update something on the AMI to allow for this?

integration/terraform/ec2/win/variables.tf
```diff
@@ -5,22 +5,22 @@ variable "region" {
...
 variable "ec2_instance_type" {
   type    = string
-  default = "t3a.xlarge"
+  default = "g4dn.xlarge"
 }
 
 variable "ami" {
   type    = string
-  default = "cloudwatch-agent-integration-test-win-2022*"
+  default = "Windows_Server-2019-English-Deep-Learning*"
```
SaxyPandaBear, line 13: Do we need to use this AMI? The integration test AMI already fixes OpenSSH

integration/terraform/ec2/win/vpc.tf
```diff
@@ -20,9 +21,9 @@ resource "aws_security_group" "ec2_security_group" {
   }
 
   ingress {
-    from_port   = "22"
-    to_port     = "22"
-    protocol    = "tcp"
+    from_port   = 0
+    to_port     = 0
+    protocol    = "-1"
```
SaxyPandaBear, line 26: so this is completely open now?

integration/terraform/ecs/linux/README.md
```diff
@@ -1,7 +1,9 @@
 Running ECS Fargate Integration Tests
 =========================
+## Prerequisite
+* ECR Repository with the docker image
```
SaxyPandaBear, line 4: Do we have any instructions on how to set up the ECR repository?

integration/test/nvidia_gpu/metrics_nvidia_gpu_linux_test.go
```diff
@@ -0,0 +1,69 @@
...
+		}
+
+		if err := CheckFileOwnerRights(agentLinuxLogPath); err != nil {
+			t.Fatalf("CloudWatchAgent does not have right to CWA's log: %v", err)
+		}
+
+	})
+}
+
+func CheckFileOwnerRights(filePath string) error {
```
SaxyPandaBear, line 56: This should be put in the other utility file for security checks.

internal/util/security/windows_permission.go
```diff
@@ -0,0 +1,109 @@
...
+			if compareIsLocalSystem || compareIsAdministrators {
+				return fmt.Errorf("File %s has deny access for Administrators and Local System", filePath)
+			}
+		}
+
+		if pAce.AccessMask == FILE_ALL_ACCESS {
+			if compareIsLocalSystem {
+				hasFileAllAccessLocalSystem = true
+			}
+			if compareIsLocalSystem {
```
SaxyPandaBear, line 97: we don't need to check `compareIsLocalSystem` twice

integration/terraform/ec2/win/main.tf
khanhntd: Yes. The only time we allow port 22 connection before launching instance would be: interact with the AMI. There are two ways for us to do that: 
* Use ImagePipeline to build the AMI and install some of the software beforehand. (1)
* Install those software when spinning up instances (3)
(1) would be not feasible for us since there are "timeout" issue (I have built it and it only concerns Deep Learning AMI for windows but not for linux). However, (3) is not the long-term solution since we rely on ImageBuilder but its share the same effect with (1); but (3) is the temporary solution I have been able to think of.

integration/terraform/ec2/win/variables.tf
khanhntd: Yes. Since we have a separate pipeline for it https://github.com/aws/amazon-cloudwatch-agent/blob/master/.github/workflows/integrationTest.yml#L457-L511

internal/util/security/windows_permission.go
khanhntd: Agree. Nice catch on this !

integration/terraform/ecs/linux/README.md
khanhntd: We do [have the instructions](https://docs.aws.amazon.com/AmazonECR/latest/userguide/getting-started-console.html)

integration/terraform/ec2/win/vpc.tf
khanhntd: It should not be. OpenSSH should opens port 22 for this.

integration/test/nvidia_gpu/metrics_nvidia_gpu_linux_test.go
khanhntd: I put it here since its a unique case for the test and its not a general one. However, I don't mind putting it to the same folder for security checking.

ZhenyuTan-amz (approved):

PR #604, opened 2022-10-04
Increase agentRuntime in TestBundle

ymtaye:
# Description of the issue
`TestBundle` failing after removal of ADOT
Screen Shot 2022-10-04 at 7 32 04 AM

# Description of changes
Increase `agentRuntime` in TestBundle from 1 minute to 2.

# License

# Tests
[Run GitHub Actions on personal fork
](https://github.com/ymtaye/amazon-cloudwatch-agent/actions/runs/3181891326)

# Requirements
2. Run `make linter`

### DIFF
--- a/integration/test/ca_bundle/ca_bundle_test.go
+++ b/integration/test/ca_bundle/ca_bundle_test.go
@@ -22,8 +22,8 @@ const configJSON = "/config.json"
 const commonConfigTOML = "/common-config.toml"
 const targetString = "x509: certificate signed by unknown authority"
 
-//Let the agent run for 1 minutes. This will give agent enough time to call server
-const agentRuntime = 1 * time.Minute
+//Let the agent run for 2 minutes. This will give agent enough time to call server
+const agentRuntime = 2 * time.Minute
 
 type input struct {
 	findTarget bool


khanhntd (commented):
The integration test are failing because of no query AMI found. Therefore, would be best to verify E2E and would you able to share the frequency of success over failure?

sethAmazon (approved):

integration/test/ca_bundle/ca_bundle_test.go
```diff
@@ -22,8 +22,8 @@ const configJSON = "/config.json"
 const commonConfigTOML = "/common-config.toml"
 const targetString = "x509: certificate signed by unknown authority"
 
-//Let the agent run for 1 minutes. This will give agent enough time to call server
-const agentRuntime = 1 * time.Minute
+//Let the agent run for 2 minutes. This will give agent enough time to call server
+const agentRuntime = 2 * time.Minute
```
SaxyPandaBear, line 26: Is the failure because of test flakiness, or is the agent really taking a longer time to start up?

ymtaye: It's definitely flaky since even this 2 minute increase fails for one of the debain jobs. Even before the fix, that TestBundle was passing (2/4) & (3/4) in different jobs

adam-mateen: IMO, 2/4 or 3/4 is not good enough.
I want tests that pass 100/100 times.

Can we increase the time?
Any other options?

sethAmazon: Have you tried manually running this flow on hosts? I wonder if the failing hosts after the increase in time are not supporting ssl x509 enforcement properly. At 2 minutes I think this should remove start up time as an issue since we always will flush metrics at 1 minute it is possible we do not flush metrics if start up time is long.

ChenaLee: I don't know what this test exactly do but for my metric value test, 3 minutes was the minimum for it to constantly pass all the time. It seemed like depending on exactly which second in a minute agent starts, if I stop agent after 2 minutes values don't get reported. With 3 minutes it always passes. Just fyi in case it's relevant here too

adam-mateen: Frustratingly I do not see an SLA on the "put" to "get" latency between PutMetricData and GetMetricData.
- https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_PutMetricData.html
- https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_GetMetricData.html

I expect if metric collection interval is 60 seconds, then running the agent for 61 seconds should be enough to guarantee 1 collection occurred.

And if I understand the shutdown behavior correctly, CWA will wait for metric data to be flushed (i.e. uploaded) before the program exits.

So maybe you have discovered a bug? I suggest diving deep to understand why the test sometimes fails with 2 minutes.

ymtaye: I think the two minute agentRuntime should be sufficient in terms of time. 

Since it's currently failing only on the Debian instance `Run echo run on ec2 instance os debian-10 arc arm64 test dir ./integration/test/ca_bundle`  I'll run the test manually run the workflow on that host further root cause what is causing this. 

Interestingly, for other previously failing jobs this has resolved TestBundle failure and was passing with the agentRuntime increase.

Thanks for all the detail in your comments! :)

SaxyPandaBear: Have we figured out if 1, 2 , or 3 minutes is the right wait time yet? Not sure if this still needs review/approval

khanhntd:
We can close the PR since we have move the package and we can continue for the test package

PR #605, opened 2022-10-04
Clean Internal Pipeline SSM Parameter Older Than 60 Days.

sethAmazon:
# Description of the issue
We have 10k ssm param

# Description of changes
Delete ssm param over 2 months old in us-west-2

# Tests
https://github.com/aws/amazon-cloudwatch-agent/actions/runs/3185074152/jobs/5194217790

# Requirements
2. Run `make linter`


integration/clean/clean_ssm_parameter/clean_ssm_parameter.go
```diff
@@ -0,0 +1,102 @@
...
+	parameterInput := ssm.DescribeParametersInput{ParameterFilters: []types.ParameterStringFilter{nameFilter}, MaxResults: aws.Int32(50)}
+	// we are getting throttled for too many calls to ssm thus put a max on the amount of names to delete per run
+	for nextToken != nil && len(names) < 250 {
+		// see https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_DescribeParameters.html for max results
+		parameterOutput, err := ssmClient.DescribeParameters(ctx, &parameterInput)
+		if err != nil {
+			return nil, err
+		}
+		for i := range parameterOutput.Parameters {
+			log.Printf("Parametereter : %v was last modified %v", *parameterOutput.Parameters[i].Name, *parameterOutput.Parameters[i].LastModifiedDate)
```
SaxyPandaBear: typo

integration/clean/clean_ssm_parameter/clean_ssm_parameter.go
```diff
@@ -0,0 +1,102 @@
...
+			if expirationSSMParameter.After(*parameterOutput.Parameters[i].LastModifiedDate) {
+				log.Printf("Parametereter : %v was last modified : %v which is before 60 days : %v thus try to delete", *parameterOutput.Parameters[i].Name, *parameterOutput.Parameters[i].LastModifiedDate, expirationSSMParameter)
```
SaxyPandaBear: typo

integration/clean/clean_ssm_parameter/clean_ssm_parameter.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	log.Printf("Parameter names to remove %v", names)
+	return names, nil
+}
+
+func deleteSSMParameter(ctx context.Context, ssmClient *ssm.Client, names []string) error {
+	// must batch into size of 10 due to error
+	// "at 'names' failed to satisfy constraint: Member must have length less than or equal to 10"
+	for i := 0; i < len(names); i = i + 10 {
+		batch := make([]string, 0)
```
SaxyPandaBear: Not super important in the grand scheme of things, but is there a way to specify that the maximum size that the slice will reach is 10, without prepopulating the slice with empty string values?

integration/clean/clean_ssm_parameter/clean_ssm_parameter.go
```diff
@@ -0,0 +1,102 @@
...
+	cxt := context.Background()
+	ssmClient, err := getSSMClient(cxt)
+	if err != nil {
+		return err
+	}
+
+	ssmParameter, err := getSSMParameter(cxt, ssmClient, expirationSSMParameter)
+	if err != nil || len(ssmParameter) == 0 {
+		return err
+	}
```
SaxyPandaBear, line 40: If the length of ssmParameter is zero, err would be nil I think. I think it's probably fine, but not sure if that was intentional

sethAmazon: Yes if the len is 0 return nil err

SaxyPandaBear (approved):
lgtm

integration/clean/clean_ssm_parameter/clean_ssm_parameter.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/service/ssm"
+	"github.com/aws/aws-sdk-go-v2/service/ssm/types"
+	"log"
+	"time"
+)
+
+// delete ssm parameter older than 14 days
+const keepDurationSSMParameter = -1 * time.Hour * 24 * 14
```
khanhntd, line 14:
```suggestion
const (
   keepDurationSSMParameter = -1 * time.Hour * 24 * 14
   parameterSearchKey = "Name"
   parameterSearchValue = "AmazonCloudWatch"
   parameterSearchOption = "Contains"
)
```

integration/clean/clean_ssm_parameter/clean_ssm_parameter.go
```diff
@@ -0,0 +1,102 @@
...
+	nameFilter := types.ParameterStringFilter{Key: aws.String(parameterSearchKey),
+		Option: aws.String(parameterSearchOption),
+		Values: []string{
+			parameterSearchValue,
+		}}
+
+	nextToken := aws.String("")
+	names := make([]string, 0)
+	parameterInput := ssm.DescribeParametersInput{ParameterFilters: []types.ParameterStringFilter{nameFilter}, MaxResults: aws.Int32(50)}
+	// we are getting throttled for too many calls to ssm thus put a max on the amount of names to delete per run
```
khanhntd, line 65: Would this be a problem when the others SSM Config stack up before able to reach to AmazonCloudWatchAgent Parameter Store?

integration/clean/clean_ssm_parameter/clean_ssm_parameter.go
```diff
@@ -0,0 +1,102 @@
...
+		}
+		for i := range parameterOutput.Parameters {
+			log.Printf("Parameter : %v was last modified %v", *parameterOutput.Parameters[i].Name, *parameterOutput.Parameters[i].LastModifiedDate)
+			if expirationSSMParameter.After(*parameterOutput.Parameters[i].LastModifiedDate) {
+				log.Printf("Parameter : %v was last modified : %v which is before 14 days : %v thus try to delete", *parameterOutput.Parameters[i].Name, *parameterOutput.Parameters[i].LastModifiedDate, expirationSSMParameter)
+				names = append(names, *parameterOutput.Parameters[i].Name)
+			}
+		}
+		nextToken = parameterOutput.NextToken
+		parameterInput = ssm.DescribeParametersInput{ParameterFilters: []types.ParameterStringFilter{nameFilter}, MaxResults: aws.Int32(50), NextToken: nextToken}
```
khanhntd, line 80: We don't need to describe another parameters input since we can set the Next Token attributes  with the previous parameter input. Example https://github.com/evalphobia/aws-sdk-go-v2-wrapper/blob/b5915d66f6a5f4841d42e141979e32d5d477ac76/ssm/client_op_describe_parameters.go#L34-L48

integration/clean/clean_ssm_parameter/clean_ssm_parameter.go
```diff
@@ -0,0 +1,102 @@
...
+
+	nextToken := aws.String("")
+	names := make([]string, 0)
+	parameterInput := ssm.DescribeParametersInput{ParameterFilters: []types.ParameterStringFilter{nameFilter}, MaxResults: aws.Int32(50)}
+	// we are getting throttled for too many calls to ssm thus put a max on the amount of names to delete per run
+	for nextToken != nil && len(names) < 250 {
+		// see https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_DescribeParameters.html for max results
+		parameterOutput, err := ssmClient.DescribeParameters(ctx, &parameterInput)
+		if err != nil {
+			return nil, err
```
khanhntd, line 70:
```suggestion
			return []string{}, err
```

Small nitpicks since I against rely nil so much though

integration/clean/clean_ssm_parameter/clean_ssm_parameter.go
adam-mateen: nice

integration/clean/clean_ssm_parameter/clean_ssm_parameter.go
adam-mateen: If throttling is the issue, then can't we just rate-limit ourselves?
Like sleep 1 millisecond per loop?

Or add a retry/backoff?
Maybe the AWS SDK can do retry/backoff for us?

Maybe this is not an issue once the SSM parameters are pared down to a "reasonable" number...
What do you think?

adam-mateen (approved):

sky333999:
Have we considered attaching an [expiration policy](https://docs.aws.amazon.com/systems-manager/latest/userguide/parameter-store-policies.html) at the time of creation [here](https://code.amazon.com/packages/CWAgent-integ-tests/blobs/a53fadfacd37e3449cde954792a113d13bda84b4/--/src/com/amazon/cloudwatch/agent/wrappers/SsmClientWrapper.java#L457)? That way we dont have the throttling issue and dont need to manage cleanup ourselves.

SaxyPandaBear:
> Have we considered attaching an [expiration policy](https://docs.aws.amazon.com/systems-manager/latest/userguide/parameter-store-policies.html) at the time of creation?

Agreed. That being said, I don't see why we can't have both. Idk which is easier to incorporate but since we're already most of the way there with this PR, might as well use this as a failsafe for things that fall through the cracks for the expiration policy

SaxyPandaBear (approved):

PR #610, opened 2022-10-12
Add Retry Install To Deb Integration Tests

sethAmazon:
# Description of the issue
dpkg: error: dpkg frontend lock is locked by another process
It is a random occurrence. There are many reasons a package lock can be held. It is probably a process trying to update.
This is how we solve this issue in our internal pipeline and adot team solves the same problem. 

# Description of changes
Add retry for install deb package

# Tests
https://github.com/sethAmazon/amazon-cloudwatch-agent/actions/runs/3236445078/jobs/5302425581

# Requirements
2. Run `make linter`


integration/deb/install_deb.go
```diff
@@ -0,0 +1,33 @@
...
+	"time"
+)
+
+const (
+	numberRetry = 5
+	waitTime    = time.Minute
+)
+
+// Try to install amazon cloud watch agent deb. If you can't install wait 1 minute
+func main() {
```
adam-mateen, line 20: Just a note, if you find yourself adding "retry" logic to other commands then I suggest turning this into a function like:
```
func ExecuteWithRetry(command []string, retryCount int, retryDelay time.Duration) error
```

integration/generator/resources/ec2_linux_test_matrix.json
```diff
@@ -3,7 +3,7 @@
     "os": "ubuntu-20.04",
     "username": "ubuntu",
     "instanceType":"t3a.xlarge",
-    "installAgentCommand": "sudo dpkg -i -E ./amazon-cloudwatch-agent.deb",
```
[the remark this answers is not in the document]
sethAmazon: It is a random occurrence. There are many reasons a package lock can be held. It is probs a process trying to update.

integration/deb/install_deb.go
sethAmazon: Going to add this.

adam-mateen (approved):

integration/generator/resources/ec2_linux_test_matrix.json
```diff
@@ -3,7 +3,7 @@
...
+    "installAgentCommand": "go run integration/deb/install_deb.go",
```
SaxyPandaBear, line 6: Would it make more sense to write a bash script and call that rather than wrapping the shell commands inside of a go program?

integration/deb/install_deb.go
```diff
@@ -0,0 +1,33 @@
...
+	for i := 0; i < numberRetry; i++ {
+		log.Print("Attempt to install cloud watch agent deb number " + strconv.Itoa(i+1))
+		out, err := exec.Command("bash", "-c", "sudo apt install -y ./amazon-cloudwatch-agent.deb").Output()
+		if err != nil {
+			log.Print(fmt.Sprint(err) + string(out))
```
SaxyPandaBear, line 25: nit: 
```suggestion
			log.Printf("%v %s", err, string(out))
```

integration/deb/install_deb.go
```diff
@@ -0,0 +1,33 @@
...
+		} else {
+			log.Print("Agent Installed")
+			return
+		}
```
SaxyPandaBear, line 29: nit: we could rearrange the logic here so that instead of `if err != nil`, we just break early and remove the `else`:
```go
if err == nil {
    log.Println("agent installed")
   return
}

time.Sleep(waitTime)
```

PR #611, opened 2022-10-14
Refactor benchmark test package for future tests

ChenaLee:
# Description of the issue
When CPU value sanity test was added, the test package wasn't fully prepared to guide future test addition.

# Description of changes
-  Refactoring & organization for future tests
-  Created test suite so that summary for all the future tests can be printed together

## Out of scope
Print format isn't the most ideal, but I will leave improvement for that to anyone who has a good idea. 

# License

# Tests
Screen Shot 2022-10-14 at 08 58 08

# Requirements
2. Run `make linter`

^ Done

### DIFF
--- a/integration/test/agent_util_linux.go
+++ b/integration/test/agent_util_linux.go
@@ -32,6 +32,19 @@ func CopyFile(pathIn string, pathOut string) {
 	log.Printf("File : %s copied to : %s", pathIn, pathOut)
 }
 
+func DeleteFile(filePathAbsolute string) error {
+	log.Printf("Delete file %s", filePathAbsolute)
+	out, err := exec.Command("bash", "-c", "sudo rm "+filePathAbsolute).Output()
+
+	if err != nil {
+		log.Fatal(fmt.Sprint(err) + string(out))
+		return err
+	}
+
+	log.Printf("Removed file: %s", filePathAbsolute)
+	return nil
+}
+
 func StartAgent(configOutputPath string, fatalOnFailure bool) error {
 	out, err := exec.
 		Command("bash", "-c", "sudo /opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent-ctl -a fetch-config -m ec2 -s -c file:"+configOutputPath).
@@ -72,7 +85,7 @@ func ReadAgentOutput(d time.Duration) string {
 	return string(out)
 }
 
-func RunShellScript(path string, args ...string) error{
+func RunShellScript(path string, args ...string) error {
 	out, err := exec.Command("bash", "-c", "chmod +x "+path).Output()
 
 	if err != nil {
@@ -90,7 +103,7 @@ func RunShellScript(path string, args ...string) error{
 		log.Printf("Error occurred when executing %s: %s | %s", path, err.Error(), string(out))
 		return err
 	}
-	
+
 	return nil
 }
 
--- a/integration/test/metric_value_benchmark/base_test.go
+++ b/integration/test/metric_value_benchmark/base_test.go
@@ -1,0 +1,74 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+//go:build linux && integration
+// +build linux,integration
+
+package metric_value_benchmark
+
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/aws/amazon-cloudwatch-agent/integration/test"
+	"github.com/aws/amazon-cloudwatch-agent/integration/test/status"
+)
+
+const configOutputPath = "/opt/aws/amazon-cloudwatch-agent/bin/config.json"
+const agentConfigDirectory = "agent_configs"
+const agentConfigFileName = "/base_config.json"
+const minimumAgentRuntime = 3 * time.Minute
+
+type ITestRunner interface {
+	validate() status.TestGroupResult
+	getTestName() string
+	getAgentConfigFileName() string
+	getAgentRunDuration() time.Duration
+}
+
+type TestRunner struct {
+	testRunner ITestRunner
+}
+
+func (t *TestRunner) Run(s *MetricBenchmarkTestSuite) {
+	testName := t.testRunner.getTestName()
+	log.Printf("Running %v", testName)
+	testGroupResult, err := t.runAgent()
+	if err == nil {
+		testGroupResult = t.testRunner.validate()
+	}
+	s.AddToSuiteResult(testGroupResult)
+	if testGroupResult.GetStatus() != status.SUCCESSFUL {
+		log.Printf("%v test group failed", testName)
+	}
+}
+
+func (t *TestRunner) runAgent() (status.TestGroupResult, error) {
+	testGroupResult := status.TestGroupResult{
+		Name: t.testRunner.getTestName(),
+		TestResults: []status.TestResult{
+			{
+				Name:   "Starting Agent",
+				Status: status.SUCCESSFUL,
+			},
+		},
+	}
+
+	agentConfigPath := agentConfigDirectory + t.testRunner.getAgentConfigFileName()
+	log.Printf("Starting agent using agent config file %s", agentConfigPath)
+	test.CopyFile(agentConfigPath, configOutputPath)
+	err := test.StartAgent(configOutputPath, false)
+
+	if err != nil {
+		testGroupResult.TestResults[0].Status = status.FAILED
+		return testGroupResult, fmt.Errorf("Agent could not start")
+	}
+
+	runningDuration := t.testRunner.getAgentRunDuration()
+	time.Sleep(runningDuration)
+	log.Printf("Agent has been running for : %s", runningDuration.String())
+	test.StopAgent()
+
+	return testGroupResult, nil
+}
--- a/integration/test/metric_value_benchmark/base_test.go
+++ b/integration/test/metric_value_benchmark/base_test.go
@@ -70,5 +70,11 @@ func (t *TestRunner) runAgent() (status.TestGroupResult, error) {
 	log.Printf("Agent has been running for : %s", runningDuration.String())
 	test.StopAgent()
 
+	err = test.DeleteFile(configOutputPath)
+	if err != nil {
+		testGroupResult.TestResults[0].Status = status.FAILED
+		return testGroupResult, fmt.Errorf("Failed to cleanup config file after agent run due to: %v", err.Error())
+	}
+
 	return testGroupResult, nil
 }
--- a/integration/test/metric_value_benchmark/base_test.go
+++ b/integration/test/metric_value_benchmark/base_test.go
@@ -62,7 +62,7 @@ func (t *TestRunner) runAgent() (status.TestGroupResult, error) {
 
 	if err != nil {
 		testGroupResult.TestResults[0].Status = status.FAILED
-		return testGroupResult, fmt.Errorf("Agent could not start")
+		return testGroupResult, fmt.Errorf("Agent could not start due to: %v", err.Error())
 	}
 
 	runningDuration := t.testRunner.getAgentRunDuration()
--- a/integration/test/metric_value_benchmark/cpu_test.go
+++ b/integration/test/metric_value_benchmark/cpu_test.go
@@ -15,7 +15,8 @@ import (
 func (suite *MetricBenchmarkTestSuite) TestCPUValues() {
 	log.Printf("Testing Cpu values...")
 	suite.RunAgent(agentConfigFileName, minimumAgentRuntime)
-	suite.addToSuiteResult(validateCpuMetrics())
+	testGroupResult := validateCpuMetrics()
+	suite.addToSuiteResult(testGroupResult)
 	suite.Assert().Equal(status.SUCCESSFUL, testGroupResult.GetStatus(),
 		"Cpu test failed to validate that every metric value is greater than zero")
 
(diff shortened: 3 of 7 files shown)


integration/test/metric_value_benchmark/metrics_value_benchmark_test.go
```diff
@@ -10,83 +10,53 @@ import (
...
+func TestMetricValueBenchmarkSuite(t *testing.T) {
+	suite.Run(t, new(MetricBenchmarkTestSuite))
+}
 
-		if !isAllValuesGreaterThanZero(metricName, values) {
-			continue
-		}
+func (suite *MetricBenchmarkTestSuite) RunAgent(agentConfigFileName string, runningDuration time.Duration) {
+	agentConfigPath := agentConfigDirectory + agentConfigFileName
+	log.Printf("Starting agent using agent config file %s", agentConfigPath)
```
SaxyPandaBear, line 49: wrt the open question about printing, `testing.T` has a `Log/Logf` function that can be used. Not sure if that helps at all

integration/test/status/test_result.go
```diff
@@ -0,0 +1,66 @@
...
+func (r TestGroupResult) GetStatus() TestStatus {
+	for _, result := range r.TestResults {
+		if result.Status == FAILED {
+			return FAILED
+		}
+	}
+	return SUCCESSFUL
+}
+
+func (r TestGroupResult) Print() {
```
SaxyPandaBear, line 52: What's the difference between a test group and a test suite? From what I can gather, we will only ever have one test suite, and each group in the suite is the set of tests that get executed to validate a feature, e.g.: cpu?

integration/test/status/test_result.go
```diff
@@ -0,0 +1,66 @@
...
+	for _, result := range r.TestGroupResults {
+		result.Print()
+	}
+	log.Printf(">>>>>>>>>>>>>>><<<<<<<<<<<<<<<")
+}
+
+type TestGroupResult struct {
+	Name        string
+	TestResults []TestResult
+}
```
SaxyPandaBear, line 41: This makes me think that it could be pulled out so that `TestSuiteResult` is just a map.

```go
type TestSuiteResult struct {
    Name string
    TestGroupResults map[string][]TestResults
}
```
I think that doing it this way would break some of the abstractions you have like `GetStatus()`, but this feels too nested.

integration/test/metric_value_benchmark/placeholder_test.go
```diff
@@ -0,0 +1,11 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+//go:build linux && integration
+// +build linux,integration
+
+package metric_value_benchmark
+
+func (suite *MetricBenchmarkTestSuite) TestDummy() {
```
SaxyPandaBear, line 9: Rather than have this function, would it make more sense to put more context clues in the MetricBenchmarkTestSuite struct? IMO the methodology we should take with this testing "framework" (not sure if you would label this a framework or not but I think it's working towards that) is that it should be clear enough what needs to be done to implement the plumbing for a test. 

Like, we have this currently:
```go
	log.Printf("Testing Cpu values...")
	suite.RunAgent(agentConfigFileName, minimumAgentRuntime)
	testGroupResult := validateCpuMetrics()
	suite.AddToSuiteResult(testGroupResult)
	suite.Assert().Equal(status.SUCCESSFUL, testGroupResult.GetStatus(),
		"Cpu test failed to validate that every metric value is greater than zero")
```
I think we could take one step further by adding something to the test suite definition to encapsulate the real test validation. At a high level, we want to define (per test) some body of work that eventually results in a `status.TestResult`. So we could add something like:
```go
func (s *MetricBenchmarkTestSuite) Validate(f func() status.TestResult) {
    res := f()
    s.Assert().Equal(status.SUCCESSFUL, res)
}
```
and that would cut out the notion of adding the result of the test to a group... which.. maybe you don't want? Idk. Just food for thought. I'll think about it more

SaxyPandaBear (dismissed):

integration/test/status/test_result.go
ChenaLee: yeah that's the hope. Feel free to change the naming

integration/test/status/test_result.go
ChenaLee: I'd like to keep it this way. Just because it's not a newly defined type doesn't mean it's not nested imo! And also I like knowing that the string is a name for each test result

integration/test/metric_value_benchmark/placeholder_test.go
ChenaLee: How about this instead? I considered doing the following and just didn't in order to not go overboard and spend too much time but.. essentially a template method pattern:


```
type TestRunner interface {
   RunTest() 
   runAgent()
   validate() TestGroupResult
   
}
type baseTest struct {
}
func (t *baseTest) RunTest(s* MetricBenchmarkTestSuite) {
   t.runAgent() //to be implemented by each test.. etc cpu test
   t.validate() // to be implemented by each test .. etc cpu test
   s.addToSuiteResult(..) 
}
type CPUTest struct {
}
func (c *CPUTest) runAgent( ){
 .. call shared RunAgent() function with arguments for file path
}
func (c *CPUTest) validate() {
}


var testRunners := []TestRunner // cputest, memtest... etc

func (s *MetricBenchmarkTestSuite) TestSuite() {
    // iterate over all testRunners and call RunTest() on each of them. 
    
    // some assertion for the entire test success/failure
}

```

SaxyPandaBear: I like it. If we're already trying to refactor this for flexibility, might as well go all out now. I think my one gripe is that everything falls under this one test suite, whereas right now we execute tests individually on a host (e.g.: the metric dimensions test runs on a different host than the cloudwatch logs test). 

With this approach, would we be running multiple tests on the same host? Maybe this points to figuring out what our definition of benchmark is for this. Is it just a sanity test to make sure that we emit **some** value, or are we measuring performance/SLAs/etc.? I'm not _super_ concerned about testing multiple plugins on the same host/test run, but it is something to keep in mind.

ChenaLee: yeah on the same host. I was thinking maybe we could even run agent once to get most of the metric data and validate values instead of running multiple times. Yeah if it becomes a problem we'll figure it out

integration/test/metric_value_benchmark/metrics_value_benchmark_test.go
ChenaLee: Might try and see! Thanks. I'm not yet aware of the difference between all these different ways of logging

integration/test/metric_value_benchmark/placeholder_test.go
ChenaLee: Alright let me.. have fun and go overboard then haha. Will re-request review after I'm done

integration/test/metric_value_benchmark/cpu_test.go
```diff
@@ -0,0 +1,77 @@
...
+		Name:   metricName,
+		Status: status.FAILED,
+	}
+
+	fetcher, err := metric.GetMetricFetcher(metricName)
+	if err != nil {
+		return testResult
+	}
+
+	values, err := fetcher.Fetch(namespace, metricName, metric.AVERAGE)
```
sky333999, line 63: Assuming this is actually making a call to CW, would making one call to fetch all metrics be better as opposed to fetching one at a time?

integration/test/metric_value_benchmark/base_test.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/aws/amazon-cloudwatch-agent/integration/test"
+	"github.com/aws/amazon-cloudwatch-agent/integration/test/status"
+)
+
+const configOutputPath = "/opt/aws/amazon-cloudwatch-agent/bin/config.json"
```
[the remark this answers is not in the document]
ChenaLee: I don't believe so..? Why do you thinks so?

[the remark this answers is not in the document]
sky333999: Given its something we are creating as part of the test, ideally we dont want it lying around once the test is done.. also if its shared between tests accidentally, behavior could bleed across tests

integration/test/metric_value_benchmark/base_test.go
```diff
@@ -0,0 +1,74 @@
...
+	getAgentRunDuration() time.Duration
+}
+
+type TestRunner struct {
+	testRunner ITestRunner
+}
+
+func (t *TestRunner) Run(s *MetricBenchmarkTestSuite) {
+	testName := t.testRunner.getTestName()
+	log.Printf("Running %v", testName)
```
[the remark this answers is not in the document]
ChenaLee: I thought that wouldn't support formatting?

integration/test/metric_value_benchmark/base_test.go
```diff
@@ -0,0 +1,74 @@
...
+		testGroupResult.TestResults[0].Status = status.FAILED
+		return testGroupResult, fmt.Errorf("Agent could not start")
+	}
+
+	runningDuration := t.testRunner.getAgentRunDuration()
+	time.Sleep(runningDuration)
+	log.Printf("Agent has been running for : %s", runningDuration.String())
+	test.StopAgent()
+
+	return testGroupResult, nil
```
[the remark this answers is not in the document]
ChenaLee: It can return error. line 65?

integration/test/metric_value_benchmark/cpu_test.go
ChenaLee: Sacrificing api cost for logic/code cleanness. If we want to we can just use the provided interface (which takes a lot of query requests at once) but it makes code harder to understand/harder to think of each test independently. Given that api cost isn't a big deal for this test package, I think we can change this if this becomes a problem

integration/test/metric_value_benchmark/base_test.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	agentConfigPath := agentConfigDirectory + t.testRunner.getAgentConfigFileName()
+	log.Printf("Starting agent using agent config file %s", agentConfigPath)
+	test.CopyFile(agentConfigPath, configOutputPath)
+	err := test.StartAgent(configOutputPath, false)
+
+	if err != nil {
+		testGroupResult.TestResults[0].Status = status.FAILED
+		return testGroupResult, fmt.Errorf("Agent could not start")
```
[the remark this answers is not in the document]
ChenaLee: done

integration/test/metric_value_benchmark/base_test.go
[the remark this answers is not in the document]
ChenaLee: Hmm.. but new tests are supposedly overwriting this config right? Are you saying other tests might forget to overwrite the config?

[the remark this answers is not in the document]
sky333999: Yes thats one possible room for error.. and additionally, the last test would just leave this lying around (in this specific case, just one file on an instance thats about to be torn down wont hurt but I feel its good practice to always clean up what tests create)

integration/test/metric_value_benchmark/base_test.go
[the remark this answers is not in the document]
sky333999: Ah yea missed that!

integration/test/metric_value_benchmark/cpu_test.go
sky333999: Would the logic change a lot? Id imagine only the initial call to backend would be done once but the rest of the stuff like validating each metric can remain the same... just that it would leverage what was already retrieved from the backend at that point.

ChenaLee: well it wouldn't read the same. right now logic's clean with validate() call simply calling all validateCPUMetric with each metric. It's a higher level of abstraction. What is exactly the concern? Performance? How concerned are we about test run's performance?

sky333999: Yeah probably only a small bit from performance standpoint and test runtime since its more network calls. But this isnt a blocker if others are fine with it.

integration/test/agent_util_linux.go
```diff
@@ -32,6 +32,19 @@ func CopyFile(pathIn string, pathOut string) {
 	log.Printf("File : %s copied to : %s", pathIn, pathOut)
 }
 
+func DeleteFile(filePathAbsolute string) error {
```
sky333999, line 35: Perhaps we could leverage https://ieftimov.com/posts/testing-in-go-clean-tests-using-t-cleanup?

ChenaLee: Can't because I'm running only a one test *testing.T function in order to create a nice summary. If that happens we only clean up once for the entire suite which is not the behavior you wanted

sky333999 (commented):
LGTM

sky333999 (approved):

integration/test/metric_value_benchmark/metrics_value_benchmark_test.go
```diff
@@ -10,83 +10,46 @@ import (
...
-		values, err := fetcher.Fetch(namespace, metricName, metric.AVERAGE)
-		if err != nil {
-			continue
-		}
+func (suite *MetricBenchmarkTestSuite) TestAllInSuite() {
+	for _, testRunner := range testRunners {
+		testRunner.Run(suite)
+	}
+	suite.Assert().Equal(status.SUCCESSFUL, suite.result.GetStatus(), "Metric Benchmark Test Suite Failed")
+}
```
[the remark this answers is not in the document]
ChenaLee: Added to my personal todo I'm tracking

PR #619, opened 2022-10-20
Fix flaky aggregator shutdown test.

adam-mateen:
# Description of the issue
This test is flaky.
It passes quite often on my dev machine, but fails frequently when github actions runs the unit tests.
The point of this test case is to shutdown before aggregation has "naturally" occurred.
The problem is mainly because the `aggregating()` method has this line:
```
time.Sleep(now.Truncate(durationAgg.aggregationDuration).Add(durationAgg.aggregationDuration).Sub(now))
```
Whch means when a new metric name is first seen and a `durationAggregator` is started, it could take between `aggregationInterval` and `2*aggregationInterval` for the first Aggregation.

You would think the test shouldn't care, it can just shutdown right after `AddMetric` is called and expect the shutdown/flush to happen ASAP.

You'd be wrong because of the randomness of selecting when 2 cases are "ready":
- `case m := <-durationAgg.aggregationChan:`
- `case <-durationAgg.shutdownChan:`

# Description of changes
The test case submits 1 metric and waits _sufficiently long_ for that metric to be "naturally" aggregated.
Then it submits the same metric-name again and immediately shuts down.
We could delay between `AddMetric()` and shutdown, but it is not necessary.

# License

# Tests
```
go test -count=100 -v --failfast ./plugins/outputs/cloudwatch -run ShutdownBehavior
```

# Requirements
2. Run `make linter`

### DIFF
--- a/plugins/outputs/cloudwatch/aggregator_test.go
+++ b/plugins/outputs/cloudwatch/aggregator_test.go
@@ -147,8 +147,7 @@ func TestAggregator_ShutdownBehavior(t *testing.T) {
 	// The Aggregator creates a new durationAggregator for each metric.
 	// And there is a delay when each new durationAggregator begins.
 	// So submit a metric and wait for the first aggregation to occur.
-	time.Sleep(aggregationInterval + time.Second)
-	assertMetricContent(t, metricChan, 1*time.Second, m, expectedFieldContent{
+	assertMetricContent(t, metricChan, 3*aggregationInterval, m, expectedFieldContent{
 		"value", 1, 1, 1, 1, "", []float64{1.0488088481701516}, []float64{1}})
 	assertNoMetricsInChan(t, metricChan)
 	// Now submit the same metric and it should be routed to the existing
@@ -156,8 +155,7 @@ func TestAggregator_ShutdownBehavior(t *testing.T) {
 	timestamp = time.Now()
 	m = metric.New(metricName, tags, fields, timestamp)
 	aggregator.AddMetric(m)
-	// Shutdown before the aggregationInterval
-	time.Sleep(aggregationInterval / 2)
+	// Shutdown before the 2nd aggregationInterval completes.
 	close(shutdownChan)
 	wg.Wait()
 	assertMetricContent(t, metricChan, 1*time.Second, m, expectedFieldContent{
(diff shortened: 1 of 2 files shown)


plugins/outputs/cloudwatch/aggregator_test.go
```diff
@@ -136,19 +136,32 @@ func TestAggregator_ShutdownBehavior(t *testing.T) {
...
 	m := metric.New(metricName, tags, fields, timestamp)
 	aggregator.AddMetric(m)
-
-	//give some time to aggregation to do the work
-	time.Sleep(time.Second * 2)
-
+	// The Aggregator creates a new durationAggregator for each metric.
+	// And there is a delay when each new durationAggregator begins.
+	// So submit a metric and wait for the first aggregation to occur.
+	time.Sleep(aggregationInterval + time.Second)
```
SaxyPandaBear, line 150: Okay, I think I see the problem. It's because we wait long enough to potentially have 2 aggregation intervals occur when we only want one? 

Intuitively, does that mean we want to sleep for aggregationInterval < x < 2 * aggregationInterval?

What I'm mildly concerned about is if someone in the future tweaks the interval above but doesn't update this and it becomes flakey again (if agg interval changes to 1 second, then the problem reoccurs). Should this sleep time be a magnitude of aggregation interval, like what happens on line 160?

SaxyPandaBear (approved):

plugins/outputs/cloudwatch/aggregator_test.go
adam-mateen: The point of this test case is to shutdown before aggregation has "naturally" occurred.
The problem is mainly because the `aggregating()` method has this line:
```
time.Sleep(now.Truncate(durationAgg.aggregationDuration).Add(durationAgg.aggregationDuration).Sub(now))
```
Whch means when a new metric name is first seen and a `durationAggregator` is started, it could take between `aggregationInterval` and `2*aggregationInterval` for the first Aggregation.

You would think the test shouldn't care, it can just shutdown right after `AddMetric` is called and expect the shutdown/flush to happen ASAP.

You'd be wrong because of the randomness of selecting when 2 cases are "ready":
- `case m := <-durationAgg.aggregationChan:`
- `case <-durationAgg.shutdownChan:`

After my change, the requirement is that:
1. the first sleep is between aggregationInterval and 2*aggregationInterval.
2. The next sleep... I just realized is not needed. I will remove it...

adam-mateen: This may be an edge case to address.
At runtime if CWA has a duration aggregation interval of 60 seconds.
It is possible that if a new metric name is first seen just before shutdown it will delay shutdown for up to `aggregationDuration`.

ymtaye (approved):

SaxyPandaBear (approved):

taohungyang (approved):

PR #622, opened 2022-10-25
E2e test package separation

ChenaLee:
# Description of the issue
Test package is being separated, which meant that changes were needed in our workflow definition files

# Description of changes
Make changes to all github workflow steps to make tests run with the new test package

# License

# Tests
Ran integration test workflow in my fork using the new test package https://github.com/aws/amazon-cloudwatch-agent-test

# Requirements
2. Run `make linter`

### DIFF
--- a/.github/workflows/integrationTest.yml
+++ b/.github/workflows/integrationTest.yml
@@ -637,12 +637,13 @@ jobs:
     needs: [ StartLocalStack, EC2LinuxIntegrationTest ]
     defaults:
       run:
-        working-directory: integration/terraform/ec2/localstack
+        working-directory: terraform/ec2/localstack
     permissions:
       id-token: write
       contents: read
     steps:
       - uses: actions/checkout@v2
+        with: aws/amazon-cloudwatch-agent-test
 
       - name: Configure AWS Credentials
         uses: aws-actions/configure-aws-credentials@v1
--- a/.github/workflows/integrationTest.yml
+++ b/.github/workflows/integrationTest.yml
@@ -12,6 +12,8 @@ env:
   GPG_KEY_NAME: ${{ secrets.GPG_KEY_NAME }}
   GPG_TTY: $(tty)
   ECR_INTEGRATION_TEST_REPO: "cwagent-integration-test"
+  CWA_GITHUB_TEST_REPO_NAME: "aws/amazon-cloudwatch-agent-test"
+  CWA_GITHUB_TEST_REPO_URL: "https://github.com/aws/amazon-cloudwatch-agent-test.git"
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


SaxyPandaBear (approved):

.github/workflows/integrationTest.yml
```diff
@@ -767,11 +785,11 @@ jobs:
           timeout_minutes: 30
           retry_wait_seconds: 5
           command: |
-            cd integration/terraform/ec2/linux
+            cd terraform/ec2/linux
             terraform init
             if terraform apply --auto-approve \
-              -var="ssh_key_value=${PRIVATE_KEY}" -var="github_repo=${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}.git" \
-              -var="github_sha=${GITHUB_SHA}" -var="install_agent=${{ matrix.arrays.installAgentCommand }}" \
+              -var="ssh_key_value=${PRIVATE_KEY}" -var="github_test_repo=https://github.com/aws/amazon-cloudwatch-agent-test.git" \
```
khanhntd, line 791: Would this be better to add this as an environment variables since this is not the only place being changed?

.github/workflows/integrationTest.yml
```diff
@@ -158,6 +160,8 @@ jobs:
       contents: read
     steps:
       - uses: actions/checkout@v2
+        with:
+          repository: aws/amazon-cloudwatch-agent-test
```
khanhntd, line 164: Would it be better to use `path` option which would reduce less changes and raise more awareness when using the integration test repository?

ChenaLee: What is that good for?

.github/workflows/integrationTest.yml
ChenaLee: That is a good idea. Thanks!

.github/workflows/integrationTest.yml
khanhntd: Its create a [relative path for the checkout repository](https://github.com/actions/checkout#usage). For example, by using `path:integration`, you don't need to delete all the integration path and it differentiate between checkout the main repo and the integration test repo

ChenaLee: integration directory is already deleted in the test package

ChenaLee: Is there any value in re-making changes to revert them all both in here and in test package?

khanhntd: Not much besides raise awareness but should not block the PR from getting merged. So I'm good with either options, only want to provide options.

khanhntd (approved):
LGTM.

SaxyPandaBear (approved):

PR #625, opened 2022-10-27
Delete integration directory after test package separation

ChenaLee:
# Description of the issue
Test package got separated, and integration test code should no longer live here. Verified tests are succeeding using the new test package.

# Description of changes
Delete the entire integration directory

# License

# Tests
N/A

# Requirements
2. Run `make linter`


SaxyPandaBear:
Can you fix the go mod? Looks like it's failing for *Nix.

Should just be a `go mod tidy`

SaxyPandaBear:
Also what is the flow going to be for testing now? Maybe I still haven't wrapped my head around that. I understand there's a new package that houses our integration tests, but haven't pieced together how those get kicked off now. Also, there's likely going to be issues we need to resolve with the performance test measurements/workflows

ChenaLee:
> Also what is the flow going to be for testing now? Maybe I still haven't wrapped my head around that. I understand there's a new package that houses our integration tests, but haven't pieced together how those get kicked off now. Also, there's likely going to be issues we need to resolve with the performance test measurements/workflows

Something's weird. Yeah I'm holding off on merging this. I will explain how this all works together after I figure out what made perf tests fail after I nuke this, even though I thought perf tests were already using the new test package

ChenaLee:
> Also what is the flow going to be for testing now? Maybe I still haven't wrapped my head around that. I understand there's a new package that houses our integration tests, but haven't pieced together how those get kicked off now. Also, there's likely going to be issues we need to resolve with the performance test measurements/workflows

terminologies:
agent repo = aws/amazon-cloudwatch-agent (or your local fork)
test repo = aws/amazon-cloudwatch-agent-test (can use your local fork if you need to)

Github actions flow is defined in the agent repo. That github action triggers for each new commit in the agent repo OR for each manual run you trigger. That github action definitions are set to do some set up work, and checkout the test repo to run the tests defined in the test. What part is confusing?

integration/clean/clean_ami/clean_ami.go
```diff
@@ -1,79 +0,0 @@
-// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
khanhntd, line 1: For the clean AWS Resources, it would be better to keep it here since we also clean internal resources and various resources not relate  to integration test

ChenaLee:
@Andrew: Test succeeded https://github.com/ChenaLee/amazon-cloudwatch-agent/actions/runs/3338220556/jobs/5579613459

integration/clean/clean_ami/clean_ami.go
ChenaLee: Hmm doesn't seem like it's used anywhere else in the code (tests ran after this deletion)? Can you show me where this is used?

khanhntd: For example: [clean_ami](https://github.com/aws/amazon-cloudwatch-agent/blob/master/.github/workflows/clean-aws-resources.yml#L28), [clean_dedicated_host](https://github.com/aws/amazon-cloudwatch-agent/blob/master/.github/workflows/clean-aws-resources.yml#L46)

khanhntd (commented):
Other than the clean integration test, it would be more good to remove the dependencies in `go.mod` that relates to integration test (e.g aws sdk go v2 and go mod tidy again). Moreover, [the sanity test](https://github.com/aws/amazon-cloudwatch-agent/pull/625/checks) are failing too. I suspect it relates to not using dependencies even though they are declared. Other than that, we can fall back so there is no harm in rushing/merging it.

integration/clean/clean_ami/clean_ami.go
ChenaLee: oh I see, good catch! Will check if integ test needs it and make appropriate changes

go.mod
```diff
@@ -55,21 +55,15 @@ require (
...
-	github.com/aws/aws-sdk-go-v2 v1.16.5
-	github.com/aws/aws-sdk-go-v2/config v1.15.3
-	github.com/aws/aws-sdk-go-v2/feature/dynamodb/attributevalue v1.9.4
-	github.com/aws/aws-sdk-go-v2/feature/ec2/imds v1.12.3
-	github.com/aws/aws-sdk-go-v2/service/cloudwatch v1.18.1
-	github.com/aws/aws-sdk-go-v2/service/cloudwatchlogs v1.15.4
-	github.com/aws/aws-sdk-go-v2/service/dynamodb v1.15.7
-	github.com/aws/aws-sdk-go-v2/service/ec2 v1.29.0
-	github.com/aws/smithy-go v1.11.3
+	github.com/aws/aws-sdk-go-v2 v1.16.2
```
khanhntd, line 58: Any reasons why we have a degrade for these dependencies?

khanhntd (approved):
Other than the dependencies degration, LGTM

SaxyPandaBear (approved):

PR #630, opened 2022-11-01
Comment out performance test that is buggy and no public release

ChenaLee:
# Description of the issue
DDB primary index shouldn't be YEAR+CommitDate as that makes the workflow fail if there were two commits on the same day

# Description of changes
Commented out as discussed with the team

# License

# Tests
Tested github workflow and verified the performance test step is gone

# Requirements
2. Run `make linter`


jefchien (approved):

khanhntd (approved):
LGTM.

PR #645, opened 2022-11-29
Use Systemctl Command To Restart Agent

sethAmazon:
# Description of the issue
On ol9 service alias not starting agent. 

# Description of changes
Use systemctl command directly instead to start agent

# License

# Tests
[_Describe what tests you have done._](https://github.com/sethAmazon/amazon-cloudwatch-agent/actions/runs/3570899143/jobs/6002383229)

# Requirements
2. Run `make linter`

### DIFF
--- a/packaging/dependencies/amazon-cloudwatch-agent-ctl
+++ b/packaging/dependencies/amazon-cloudwatch-agent-ctl
@@ -96,7 +96,7 @@ agent_start() {
      if [ "${SYSTEMD}" = 'true' ]; then
           systemctl daemon-reload || return
           systemctl enable "${agent_name}.service" || return
-          service "${agent_name}" restart || return
+          systemctl restart "${agent_name}.service" || return
      else
           start "${agent_name}" || return
           sleep 1


SaxyPandaBear (approved):

straussb:
Are there other OSs where this would not work?  (Do we have tests for this code path that get run on all OSs?)

sethAmazon:
We still have the internal tests to make sure this works on other os. This works on all os tested in external tests.

SaxyPandaBear:
My reasoning for approving was that if the other commands worked, like `systemctl enable`, then this has to work. If it doesn't work on a currently supported OS, then we should have seen an issue by now

adam-mateen (approved):

adam-mateen (approved):

adam-mateen:
Looks like we already have dependencies on `systemctl`, so I see no issue with this change.
I am surprised all the Linux distros that CWA supports have `systemctl`.
Android and Alpine Linux do not... https://www.google.com/search?q=which+operating+systems+do+not+have+systemd&sa=X&ved=2ahUKEwjHsvCl19P7AhUgMlkFHf2DDNUQ7xYoAHoECAUQAQ&biw=1472&bih=1004

SaxyPandaBear:
> I am surprised all the Linux distros that CWA supports have systemctl

Doesn't the ctl script fall back to upstart if systemctl isn't on the host?

PR #646, opened 2022-11-30
Change LogFile.restoreState() so it does not return negative offset.

adam-mateen:
# Description of the issue
Avoid returning a negative offset so that CWA never tries seeking to a negative offset.

# Description of changes
add a conditional block.

# License

# Tests
1. Added to the existing test case
```
go test -count=1 -v ./plugins/inputs/logfile -run TestRestoreState

=== RUN   TestRestoreState
2022/11/30 15:53:50 Reading from offset 9323 in /tmp/logfile.log
--- PASS: TestRestoreState (0.00s)
PASS
ok  	github.com/aws/amazon-cloudwatch-agent/plugins/inputs/logfile	0.739s
```

# Requirements
3. Run `make linter`

I just ran `go fmt` on the 2 modified files.

### DIFF
--- a/plugins/inputs/logfile/logfile.go
+++ b/plugins/inputs/logfile/logfile.go
@@ -139,7 +139,7 @@ func (t *LogFile) Stop() {
 	close(t.done)
 }
 
-//Try to find if there is any new file needs to be added for monitoring.
+// Try to find if there is any new file needs to be added for monitoring.
 func (t *LogFile) FindLogSrc() []logs.LogSrc {
 	if !t.started {
 		return nil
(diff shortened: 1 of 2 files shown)


taohungyang:
I think we should also add the check in wineventlog input plugin
https://github.com/aws/amazon-cloudwatch-agent/blob/ce2454b9d524aee74d1df21c1d3a7a4cbdb80730/plugins/inputs/windows_event_log/wineventlog/wineventlog.go#L359

jefchien (approved):

adam-mateen:
> 

Good idea, will open another PR.

adam-mateen:
> I think we should also add the check in wineventlog input plugin

Good idea, will open another PR.

PR #652, opened 2022-12-14
Fix DetectAgentMode so auto resolves to onPrem if not ec2

jefchien:
# Description of the issue
In the `DetectAgentMode` function, if `auto` is passed in as the mode and it detects it as on premise, it should return `onPrem` or `onPremise`. Currently, it will return the value passed in which is `auto`. This causes the `SetMode` function in the context to panic (https://github.com/aws/amazon-cloudwatch-agent/blob/main/translator/context/context.go#L109-L120)

# Description of changes
Changed the function to return `onPrem` and added a simple unit test.

# License

# Tests
Added a unit test to cover the auto cases.

# Requirements
2. Run `make lint`

### DIFF
--- a/translator/util/sdkutil.go
+++ b/translator/util/sdkutil.go
@@ -22,8 +22,10 @@ const (
 	DEFAULT_PROFILE = "AmazonCloudWatchAgent"
 )
 
-var DetectRegion func(mode string, credsConfig map[string]string) string = detectRegion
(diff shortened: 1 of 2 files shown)


SaxyPandaBear (approved):
LGTM

khanhntd (approved):
LGTM

ymtaye (approved):

PR #653, opened 2022-12-15
cleanup weird file path and doc comment

SaxyPandaBear:
Pretty straightforward. The log path in the sample test was unintuitive. Separately, the comment for the cloudwatch plugin retry is misleading, because we make up to 5 call attempts, but the comment indicated that we do 6. The API call happens inside of the retry loop, so it doesn't make sense.


khanhntd (approved):
LGTM.

PR #662, opened 2022-12-23
Remove usage of deprecated ioutil package

### LINKED ISSUE #648: Remove usage of deprecated ioutil package
Since Go 1.16, `ioutil` has been deprecated. We never noticed because we were on Go 1.15.15 for so long. Now that we're on Go 1.18+, we should get rid of this usage and use the corresponding `os` and `io` equivalents. 

Example:
https://github.com/aws/amazon-cloudwatch-agent/blob/39372d5008e58c9d31f78027ecb4f1030f757c23/cfg/agentinfo/info_test.go#L47-L49
should be changed to:
```go
if err = os.WriteFile(vfp, []byte(expectedVersion), 0644); err != nil {
	t.Fatalf("failed to write version file at %v: %v", vfp, err)
}
```

matoblac:
# Description of the issue
_This change resolves #648 by removing the ioutil usage and corresponding os and io equivalents._

# Description of changes
_removing the ioutil usage and corresponding os and io equivalents, according to the information seen in #648._

# License
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice. - Agreed.

# Tests
_None because, I received this message: go: cannot run *test.go files. If needed, please let me know how to run tests. As this is my second commit ever and I used pipelines to conduct builds and testing in the past._

Running ... /home/ec2-user/environment/workspace/amazon-cloudwatch-agent/cfg/agentinfo/info.go
package command-line-arguments is not a main package

Running ... /home/ec2-user/environment/workspace/amazon-cloudwatch-agent/cfg/agentinfo/info.go
package command-line-arguments is not a main package

# Requirements
2. Run `make linter` - had trouble with this step explained below:

Sorry in advance if I am missing a basic concept here but, I tried running "make linter" and I got the following error:

fatal: No names found, cannot describe anything.
make: *** No rule to make target `linter'.  Stop.

Then, I tried "make lint" and here is a following output:

"WARN [runner] The linter 'golint' is deprecated (since v1.41.0) due to: The repository of the linter has been archived by the owner.  Replaced by revive." 

followed by multiple errors afterwards.

### DIFF
--- a/cfg/agentinfo/info_test.go
+++ b/cfg/agentinfo/info_test.go
@@ -44,7 +44,7 @@ func TestReadVersionFile(t *testing.T) {
 	vfp := filepath.Join(filepath.Dir(ex), versionFilename)
 	expectedVersion := "TEST_VERSION"
 
-	if err = ioutil.WriteFile(vfp, []byte(expectedVersion), 0644); err != nil {
+	if err = os.WriteFile(vfp, []byte(expectedVersion), 0644); err != nil {
 		t.Fatalf("failed to write version file at %v: %v", vfp, err)
 	}
 	defer os.Remove(vfp)


SaxyPandaBear (changes requested):
Thanks for contributing, but there's a lot of stuff that still needs to be addressed. 

```
amazon-cloudwatch-agent andhuy$ find . -type f -exec grep -E 'io/ioutil' {} + | wc -l
      44
```
There are 44 different imports of `io/ioutil`. This PR:
1. Does not remove the `io/ioutil` import from this file (since it should be unused)
2. Does not address usage in 43 other files.

khanhntd:
Thanks for the contribution @matoblac. I know its a very hard effort to create a PR and [we have another contributors](https://github.com/aws/amazon-cloudwatch-agent/pull/663) that align with your idea too. We have merged the PR for fixing the deprecated packages.  Therefore, will close this one !

PR #665, opened 2023-01-03
Unify all buid/test to PR build and reduce time for building

khanhntd:
# Description of the issue
By the amount of saving we have,  it would reduce from 3 to 1 actions / PR commits and for removing the PR builds when merging, it would 0 actions/ PR. Moreover, we are sharing the same benefits for publics and others decentralized VSC (e.g fork repo). Therefore, it would double the amount . For example, we have 6 PRs in Dec with an average of 3 commits/ PR. Therefore, we are reducing from ( 3 actions/ Pr commits * 3 commits + 3 actions / PR )* 6 * 2 = 144 actions to 12 actions and 13 minutes per PR and we are reducing the build time which increase productivity for developer and reduce the limited time that are limited to a single repository (e.g public only allows 2000 actions minute/ month
image

# Changes
* [Unify all the building binaries/running unit test on all platform ](https://github.com/aws/amazon-cloudwatch-agent/pull/665/files#diff-55d0bbcea83bbf16e4f050b69fe7bca6164d5c0925906448317f910529d34dee)
* Use caches with [fail fasts  equal to false ](https://github.com/aws/amazon-cloudwatch-agent/pull/665/files#diff-55d0bbcea83bbf16e4f050b69fe7bca6164d5c0925906448317f910529d34deeR64) to make the build / test faster and also independant to each other. If one fail, the [others won't affect each other and continue to run](https://github.com/aws/amazon-cloudwatch-agent/actions/runs/3834028001/jobs/6526028399) (solves the original issues and caches rarely fails - at least not on all platforms) 
*  [Remove Build PR when merging to main since when dev creates PR and merges it](https://github.com/aws/amazon-cloudwatch-agent/pull/665/files#diff-05316c183bd98fba8683fc40698ff3a6099c9f3a089b0bc45083d45bb92203a2L6-L12), we already accept the risks/benefits from the PR 
* Delete [releasePerformance](https://github.com/aws/amazon-cloudwatch-agent/pull/665/files#diff-b23812c865aadf36bd17fe486f9e86b66ec2a66a968c9a8583c3e52488eb1e56) as part of `performance testing`.

# License

# Tests
Running with my fork: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/3833731302

# Requirements
2. Run `make linter`


SaxyPandaBear:
> [Remove Build PR when merging to main since when dev creates PR and merges it](https://github.com/aws/amazon-cloudwatch-agent/pull/665/files#diff-05316c183bd98fba8683fc40698ff3a6099c9f3a089b0bc45083d45bb92203a2L6-L12), we already accept the risks/benefits from the PR

What risk? What are we gaining from removing it

.github/workflows/PR-build.yml
```diff
@@ -0,0 +1,144 @@
...
+        verbose: true
+
+    - name: Build
+      if: steps.cached_binaries.outputs.cache-hit != 'true'
+      run: make build
+  
+  windows-unittest:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
```
SaxyPandaBear, line 64: Are we failing fast or not? The description says fail-fast is true. This says false, and is the only place where `fail-fast` is set.

.github/workflows/PR-build.yml
```diff
@@ -0,0 +1,144 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT
+
+name: PR Build
+on:
+  pull_request:
+    branches: 
+      - main 
+    types: 
+      - opened
```
SaxyPandaBear, line 10: do we need to execute this on opened if we already have it on ready_for_review

.github/workflows/PR-build.yml
```diff
@@ -0,0 +1,144 @@
...
+      - synchronize
+      - reopened
+      - ready_for_review
+    paths-ignore:
+      - '.github/**'
+      - '!.github/workflows/CI.yml'
+      - '!.github/workflows/CI-Operator.yml'
```
SaxyPandaBear, line 17: We don't have either of these files

khanhntd:
> What risk? What are we gaining from removing it
Every time we create a PR, the PR will run the build/unit test and we always use that as an evidence for avoid any regression.  and serve as a gate before merging the PR. Therefore, running ` build/unit test ` after merging the PR is duplicating the effort and contest the resources for others workflow that run simultaneously (e.g Integration Test workflow). Small benefit for us would be increase the time for the repo to use [Github Workflow since there is a limit ](https://docs.github.com/en/billing/managing-billing-for-github-actions/viewing-your-github-actions-usage)

SaxyPandaBear (commented):
What problem does this solve? How does this save time for building? Does it address the issues initially raised in https://github.com/aws/amazon-cloudwatch-agent/pull/354 which is why we split these workflows out to begin with?

SaxyPandaBear:
> Therefore, running build/unit test after merging the PR is duplicating the effort and contest the resources for others workflow that run simultaneously

How much are we really saving?

.github/workflows/PR-build.yml
khanhntd: NIce catch. I have edited the fail fast to false. However, it still upholds the statement:
```
 If one fail, the others won't affect each other and continue to run
```
Yes since its the only [place we use matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#handling-failures). However, I can also include others jobs (e.g linux and mac-os) as a part of Windows matrix to share the same benefit. The only reason for separating is for clear separation only .

khanhntd:
For unification, it would reduce from 3 to 1 actions / PR commits and for removing the PR builds when merging, it would 0 actions/ PR. Moreover, we are sharing the same benefits between the public/private. Therefore, it would double the amount . For example, we have 6 PRs in Dec with an average of 3 commits/ PR. Therefore, we are reducing from ( 3 actions/ Pr commits * 3 commits + 3 actions / PR )* 6 * 2 = 144 actions to  12 actions.

.github/workflows/PR-build.yml
khanhntd: Opened is used for draft PR based on my understanding. However, I'm fine with using `ready_for_review` since it shows the final state and is the final gate before merging the PR.

.github/workflows/PR-build.yml
khanhntd: Nice catch. Remove all of theses and add the related files again ....

.github/workflows/PR-build.yml
```diff
@@ -0,0 +1,142 @@
...
+      - main 
+    types: 
+      - synchronize
+      - reopened
+      - ready_for_review
+    paths-ignore:
+      - '**/*.md'
+      - '.github/**'
+      - '!.github/workflows/PR-build.yml'
+  workflow_dispatch:
```
SaxyPandaBear, line 17: Why are we adding workflow_dispatch?

khanhntd: In case we want to verify if the build with the main branch is whether success or not and if not success, what is the return error. Since the logs in Github Action will be [1-90 days for public repo ](https://docs.github.com/en/organizations/managing-organization-settings/configuring-the-retention-period-for-github-actions-artifacts-and-logs-in-your-organization)

SaxyPandaBear: Why? We don't need this if we run build and test on merge to main.

SaxyPandaBear (changes requested):
If you want to consolidate the workflows into a single file, that's fine. But we should not remove building and testing on merges.

.github/workflows/PR-build.yml
khanhntd: Would it better to always making sure the PR get up to dates with main branch though (e.g through [auto-update](https://github.com/marketplace/actions/auto-update), etc). In this way, we will always get the PR build on branch and avoid duplication build on main branch when merge

khanhntd: I have added the options to "update branch" so developers can always merge branch when its out of date.
```
Whenever there are new changes available in the base branch, present an “update branch” option in the pull request.
```

.github/workflows/PR-build.yml
```diff
@@ -0,0 +1,141 @@
...
+    branches: 
+      - main 
+    types: 
+      - synchronize
+      - reopened
+      - ready_for_review
+    paths-ignore:
+      - '**/*.md'
+      - '.github/**'
+      - '!.github/workflows/PR-build.yml'
```
SaxyPandaBear, line 16: I am not going to approve a PR that turns off building and unit testing on the main branch after a merge. If other people want to, that's fine.

khanhntd: Question from me: would it be better to change the `integrationTest.yml` to `CI.yml` and add the unit testing on the `CI.yml` ? @SaxyPandaBear

chadpatel:
Do you have data/anecdotes on how much faster this is? Or will we not know until we merge

chadpatel:
Why is this still in review after 2 months? Is this still relevant?

Makefile
```diff
@@ -65,46 +57,44 @@ copy-version-file: create-version-file
...
 	$(LINUX_ARM64_BUILD)/amazon-cloudwatch-agent-config-wizard github.com/aws/amazon-cloudwatch-agent/cmd/amazon-cloudwatch-agent-config-wizard
-	$(WIN_BUILD)/amazon-cloudwatch-agent-config-wizard.exe github.com/aws/amazon-cloudwatch-agent/cmd/amazon-cloudwatch-agent-config-wizard
-	$(DARWIN_BUILD)/amazon-cloudwatch-agent-config-wizard github.com/aws/amazon-cloudwatch-agent/cmd/amazon-cloudwatch-agent-config-wizard
-	$(DARWIN_BUILD_ARM64)/amazon-cloudwatch-agent-config-wizard github.com/aws/amazon-cloudwatch-agent/cmd/amazon-cloudwatch-agent-config-wizard
 
-config-downloader: copy-version-file
-	@echo Building config-downloader
-	$(LINUX_AMD64_BUILD)/config-downloader github.com/aws/amazon-cloudwatch-agent/cmd/config-downloader
-	$(LINUX_ARM64_BUILD)/config-downloader github.com/aws/amazon-cloudwatch-agent/cmd/config-downloader
+amazon-cloudwatch-agent-mac: copy-version-file
```
khanhntd, line 72: Some of the plugins such as cpu is required with enable [CGO to collect CPU metrics](https://github.com/shirou/gopsutil/blob/master/cpu/cpu_darwin_cgo.go) with `CGO_ENABLED`. Therefore, building in [Linux environment](https://github.com/aws/amazon-cloudwatch-agent/blob/main/.github/workflows/build-test-linux.yml#L68) is not useful.

Makefile
```diff
@@ -136,7 +126,8 @@ lint: install-tools
 	${LINTER} run ./...
 
 test:
-	CGO_ENABLED=0 go test -timeout 15m -coverprofile coverage.txt -failfast ./awscsm/... ./cfg/... ./cmd/... ./handlers/... ./internal/... ./logger/... ./logs/... ./metric/... ./plugins/... ./profiler/... ./tool/... ./translator/...
+	CGO_ENABLED=0 go test -coverprofile coverage.txt -failfast ./awscsm/... ./cfg/... ./cmd/... ./handlers/... ./internal/... ./logger/... ./logs/... ./metric/... ./plugins/... ./profiler/... ./tool/... ./translator/...
```
khanhntd, line 129: Delete `timeout` since the default 10 is sufficient and increase timeout is not necessary since all individual unit test should within 10 or small than 5 - 10

.github/workflows/CI.yml
```diff
@@ -258,12 +269,19 @@ jobs:
...
           key: "cached_pkg_${{ github.sha }}"
-          path: go.mod
+          path: |
+            %LocalAppData%\go-build
+            ~/go/pkg/mod
 
+      - name: Unit Test 
+        if: steps.cached_pkg.outputs.cache-hit != 'true'
+        working-directory: cwa
+        run: make test
```
khanhntd, line 279: Adding `make test` for unit test Linux and MacOs in their respective environment before distributing the binaries to MacOS, Windows, Linux, Darwins

khanhntd:
> Why is this still in review after 2 months? Is this still relevant?

Yes it is and previously, the run build times improvement without cache are:
* `windows-unittest` (latest): [25m](https://github.com/aws/amazon-cloudwatch-agent/actions/runs/4507913801/jobs/7936141063)
* `windows-unittest` (2019): [31m](https://github.com/aws/amazon-cloudwatch-agent/actions/runs/4507913801/jobs/7936141063)
* `macos`:
* `linux`:

with cache are:
* `windows-unittest` (latest): [25m](https://github.com/aws/amazon-cloudwatch-agent/actions/runs/4507913801/jobs/7936141063)
* `windows-unittest` (2019): [31m](https://github.com/aws/amazon-cloudwatch-agent/actions/runs/4507913801/jobs/7936141063)
* `macos`:
* `linux`:

.github/workflows/PR-build.yml
SaxyPandaBear: This hasn't changed in 2 months, and my stance on it hasn't either. If other people want to approve this, then I will too but I don't agree with this change.

khanhntd: Yes I changed that during the past 2 months. We always run `integrationTest` when merge from main, therefore, I [added a unit test for Linux](https://github.com/aws/amazon-cloudwatch-agent/pull/665/files#diff-3ab46ee209a127470fce3c2cf106b1a1dbadbb929a4b5b13656a4bc4ce19c0b8R89-R91)  and for [Mac](https://github.com/aws/amazon-cloudwatch-agent/pull/665/files#diff-3ab46ee209a127470fce3c2cf106b1a1dbadbb929a4b5b13656a4bc4ce19c0b8R288-R291)

.github/workflows/CI.yml
```diff
@@ -61,36 +61,49 @@ jobs:
...
+      
 
       - name: Import GPG Key
+        if: steps.cached_binaries.outputs.cache-hit != 'true'
         uses: crazy-max/ghaction-import-gpg@v5
         with:
           gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY }}
           passphrase: ${{ secrets.PASSPHRASE }}
 
+      - name: Unit Test
```
sethAmazon, line 89: Unit tests are flaky, need to add a retry.

.github/workflows/PR-build.yml
```diff
@@ -0,0 +1,151 @@
...
+      - 'LICENSE'
+      - '.github/**'
+      - '!.github/workflows/PR-build'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref_name }}
+  cancel-in-progress: true
+
+jobs:
+  macos-unittest:
```
sethAmazon, line 28: Why can't this be done with os list with a single job?

Makefile
```diff
@@ -22,16 +22,8 @@ LDFLAGS +=  -X github.com/aws/amazon-cloudwatch-agent/cfg/agentinfo.BuildStr=${B
 LINUX_AMD64_BUILD = CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -trimpath -buildmode=${CWAGENT_BUILD_MODE} -ldflags="${LDFLAGS}" -o $(BUILD_SPACE)/bin/linux_amd64
 LINUX_ARM64_BUILD = CGO_ENABLED=0 GOOS=linux GOARCH=arm64 go build -trimpath -buildmode=${CWAGENT_BUILD_MODE} -ldflags="${LDFLAGS}" -o $(BUILD_SPACE)/bin/linux_arm64
 WIN_BUILD = GOOS=windows GOARCH=amd64 go build -trimpath -buildmode=${CWAGENT_BUILD_MODE} -ldflags="${LDFLAGS}" -o $(BUILD_SPACE)/bin/windows_amd64
-DARWIN_BUILD = GO111MODULE=on GOOS=darwin GOARCH=amd64 go build -trimpath -ldflags="${LDFLAGS}" -o $(BUILD_SPACE)/bin/darwin_amd64
```
sethAmazon, line 25: Will make release stop working?

khanhntd: No it won't [since](https://github.com/aws/amazon-cloudwatch-agent/pull/665/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R74) we have an `ifelse`
 when running `amazon-cloudwatch-agent-mac`
```
ifneq ($(OS),Windows_NT)
ifeq ($(shell uname -s),Darwin)	
```

.github/workflows/PR-build.yml
khanhntd: You can argue they would have similar workflows. However, there are some workflows that are specific to OS (e.g Windows when install [choco-make](https://github.com/khanhntd/amazon-cloudwatch-agent/blob/master/.github/workflows/build-test-windows.yml#L49-L50)) or having coverage-uploading that is one workflow would be enough. Instead of having multiple `if else`, I would prefer a straightforward workflow. If you have a strong urge against it, I am fine with changing it too

.github/workflows/PR-build.yml
khanhntd: If you disagree with running make test  in `integrationTest.yml`, then I'm fine with adding it back too. FYI,  I have added back the push on main branch

Issue #172, opened 2021-02-15
Custom global dimensions


okgolove:
Hi! Thank you for the software.
I'd like to ask what was the reason to forbid specifying custom global dimensions? I would like to add a global custom dimension `environment name: <env name>` for ALL metrics from cloudwatch agent for example. Yes, I can add it manually for most of agent inputs, but I can't do this for statsd and collectd, for example.

I'm talking about this part:

```
append_dimensions – Optional. Adds Amazon EC2 metric dimensions to all metrics collected by the agent. The only supported key-value pairs are shown in the following list. Any other key-value pairs are ignored.

"ImageID":"${aws:ImageId}" sets the instance's AMI ID as the value of the ImageID dimension.
"InstanceId":"${aws:InstanceId}" sets the instance's instance ID as the value of the InstanceID dimension.
"InstanceType":"${aws:InstanceType}" sets the instance's instance type as the value of the InstanceType dimension.
"AutoScalingGroupName":"${aws:AutoScalingGroupName}" sets the instance's Auto Scaling group name as the value of the AutoScalingGroupName dimension.
```

May be you are planning to add this feature or you can suggest any workaround.
Thanks!

pxaws:
There is no reason to forbid specifying custom global dimensions. Existing variables in the config like `"${aws:ImageId}"` are not environment variables. Cloudwatch agent gets their values by querying the ec2 metadata service. 

To support the feature you mentioned, we would like to know more about your use case. For example, are you running cloudwatch agent in container or not? 

For container environments, it might be easy to specify the environmental variables through docker file. When cloudwatch agent runs, it can get the values for those environmental variables. 

For non-container environments, we would like to know how do you specify the environmental variables so that cloudwatch agent can pick them up? (You know cloudwatch agent might run in a different shell from the one in which those environmental variables are set up)

okgolove:
Sorry, I must be describe my thoughts wrong.
I'm no talking about environment variables, I'm talking about environment name (I meant staging/dev/UAT/etc).
My goal is specifying global dimension instead of append dimensions per metric (**MySuperCustomDimension** is an example):

```
{
    "metrics": {
        "append_dimensions": {
            "MySuperCustomDimension": "everything_i_want",
            "ImageId": "${aws:ImageId}",
            "InstanceId": "${aws:InstanceId}",
            "InstanceType": "${aws:InstanceType}",
            "AutoScalingGroupName": "${aws:AutoScalingGroupName}"
        },
[...]

PR #673, opened 2023-01-27
Support global metric dimensions

paulbrimicombe:
# Description of the issue

Addresses https://github.com/aws/amazon-cloudwatch-agent/issues/172

- Frequently a specific EC2 deployment will have a set of Dimensions that should be associated with all published metrics
- Support this in the metrics configuration

# Description of changes

- Add a `global_dimensions` property to the metric configuration to support this use-case

  ```json
    {
      "agent": {
        "region": "us-east-1"
      },
      "metrics": {
        "append_dimensions": {
          "ImageId": "${aws:ImageId}",
        },
        "global_dimensions": {
          "Environment": "test",
          "DeploymentStack": "green"
        },
        "metrics_collected": {
          "mem": {
            "measurement": ["mem_used_percent"]
          },
          "disk": {
            "measurement": ["used_percent"],
            "resources": ["*"]
          }
        }
      }
    }
  ```

- I opted not to use the existing `append_dimensions` property because:
  - The code required is quite different for the two use-cases
  - This could be a breaking change if people have non `aws:` values in the `append_dimensions` field
- With the current implementation, any dimension values defined in `global_dimensions` override dimensions set for individual metrics. This can easily be altered.

# License

# Tests

I have added / adjusted tests for:

- JSON to TOML translation of the configuration
- Extended tests for `BuildDimensions` to include the cases where global dimensions are provided

# Requirements
2. Run `make linter`

- Are these instructions up-to-date? `make linter` fails completely and `make fmt` and `make fmt-sh` make loads of changes to files I've not touched.


plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -582,9 +583,16 @@ func BuildDimensions(mTags map[string]string) []*cloudwatch.Dimension {
 		})
 	}
 
+	for key, value := range globalDimensions {
```
SaxyPandaBear, line 586: One thing that I would like to try to optimize is how we introduce the global dimensions. I'm not sure if there is a good way to, but basically what we're doing here is taking a static set of tags and pushing them onto an empty slice for every metric we emit. I want to believe there's a more efficient approach. Haven't thought of how yet

plugins/outputs/cloudwatch/cloudwatch_test.go
```diff
@@ -59,6 +59,108 @@ func TestBuildDimensions(t *testing.T) {
 	}
 }
 
+// Test that global dimensions are added correctly
+func TestGlobalDimensions(t *testing.T) {
+	assert := assert.New(t)
```
SaxyPandaBear, line 64: I don't think we need this. I'm assuming this is copied from the other tests but I don't think it's necessary

plugins/outputs/cloudwatch/cloudwatch_test.go
```diff
@@ -59,6 +59,108 @@ func TestBuildDimensions(t *testing.T) {
...
+		i += 1
+	}
+
+	sort.Strings(tagKeys)
+
+	if len(testPoint.Tags()) >= MaxDimensions {
+		assert.Equal(MaxDimensions, len(dimensions), "Number of dimensions should be less than MaxDimensions")
+	} else {
+		assert.Equal(len(testPoint.Tags())+len(globalDimensions), len(dimensions), "Number of dimensions should be equal to number of tags plus the number of global dimensions")
+	}
```
SaxyPandaBear, line 86: Is there ambiguity in the number of dimensions? I haven't actually looked at this test code closely before. I assumed that the test metric is consistent across test executions. I'm not convinced we need to have this if-else check for asserting. We should just do a single assertion for the size. Which would warrant having another case to account for when we exceed the number of dimensions

translator/config/schema.go
```diff
@@ -112,6 +112,16 @@ var schema = `{
...
           "maxLength": 255
         },
+        "global_dimensions": {
+          "type": "object",
+          "description": "Any dimensions to be applied to all published metrics",
+          "maxProperties": 30,
+          "additionalProperties": {
+            "type": "string",
+            "minLength": 1,
+            "maxLength": 1024
```
SaxyPandaBear, line 122: 👍 was going to ask about this. Good catch here on the dimension name length

plugins/outputs/cloudwatch/cloudwatch_test.go
```diff
@@ -59,6 +59,108 @@ func TestBuildDimensions(t *testing.T) {
...
+		if i >= 10 {
+			break
+		}
+		assert.Equal(key, *dimensions[i+len(globalDimensions)].Name, "Key should be equal")
+		assert.Equal(testPoint.Tags()[key], *dimensions[i+len(globalDimensions)].Value, "Value should be equal")
+	}
+}
+
+// Test that global dimensions override metric tags correctly
+func TestGlobalDimensionsOverride(t *testing.T) {
```
SaxyPandaBear, line 111: 😄 happy to be biased on this one. Just commenting to say I agree with this decision to override.

translator/translate/metrics/global_dimensions/ruleGlobalDimensions.go
```diff
@@ -0,0 +1,41 @@
...
+
+func (ad *globalDimensions) ApplyRule(input interface{}) (returnKey string, returnVal interface{}) {
+	im := input.(map[string]interface{})
+
+	dimensions := map[string]interface{}{}
+
+	if _, ok := im[SectionKey]; !ok {
+		returnKey = ""
+		returnVal = ""
+	} else {
```
SaxyPandaBear, line 23: I _think_ that this isn't necessary, but I am a little fuzzy on the syntax here because I don't like named returns in Go. 

My understanding is you could simplify this to:
```go
if _, ok := im[SectionKey]; !ok {
    return
}

// rest of the logic, no else
```

translator/translate/metrics/global_dimensions/ruleGlobalDimensions.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	parent "github.com/aws/amazon-cloudwatch-agent/translator/translate/metrics"
+)
+
+type globalDimensions struct {
+}
+
+const SectionKey = "global_dimensions"
+
+func (ad *globalDimensions) ApplyRule(input interface{}) (returnKey string, returnVal interface{}) {
```
SaxyPandaBear, line 15: Would appreciate a unit test. It looks straightforward enough but it wouldn't hurt 😄

plugins/outputs/cloudwatch/cloudwatch_test.go
paulbrimicombe: It's not strictly needed but it is the pattern used in all of the surrounding tests and means that each assertion is _very slightly_ more terse:

```go
assert := assert.New(t)
assert.Equal(MaxDimensions, len(dimensions), "Number of dimensions should be less than MaxDimensions")
```

vs.

```go
// Need to pass in `t` as the first argument to `assert.Equal()`
assert.Equal(t, MaxDimensions, len(dimensions), "Number of dimensions should be less than MaxDimensions")
```

plugins/outputs/cloudwatch/cloudwatch_test.go
paulbrimicombe: Again, this is the pattern used in the surrounding tests which is why I've adopted it here. I think it was probably originally set up to support Property Based Testing but it doesn't look like that's actually in use. Looking at the coverage, there are no tests for the case where we're exceeding `MaxDimensions` in the pre-existing code (which definitely doesn't mean we shouldn't add the tests here!).

plugins/outputs/cloudwatch/cloudwatch_test.go
SaxyPandaBear: Huh I never noticed that. At any rate, I'm fine leaving it.

translator/translate/metrics/global_dimensions/ruleGlobalDimensions.go
paulbrimicombe: Again, this is the pattern used all over the place in the surrounding files which is why I've used it. If you'd like me to change it then I'm happy to, but it might look a bit strange compared to any of the other files in the `/translate` part of the file tree.

paulbrimicombe: Having just tested this, there is a subtle difference between assigning to empty strings vs just calling `return` in that the values are `nil` if you don't assign them. I've stuck with `""` in case that's important for some reason later on in the processing.

translator/translate/metrics/global_dimensions/ruleGlobalDimensions.go
paulbrimicombe: I've added a test for this now.

plugins/outputs/cloudwatch/cloudwatch.go
paulbrimicombe: I might look at this on another branch (based on this one) to keep this branch a bit more tidy.

plugins/outputs/cloudwatch/cloudwatch_test.go
paulbrimicombe: I've added some explicit tests around the max dimensions now and I've removed these weird checks for `MaxDimensions` in these tests.

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -571,7 +572,7 @@ func (c *CloudWatch) BuildMetricDatum(point telegraf.Metric) []*cloudwatch.Metri
 // 30 dimensions per metric so we only keep up to the first 30 alphabetically.
 // This always includes the "host" tag if it exists.
 // See https://github.com/aws/amazon-cloudwatch-agent/issues/398
-func BuildDimensions(mTags map[string]string) []*cloudwatch.Dimension {
+func BuildDimensions(mTags map[string]string, globalDimensions map[string]string) []*cloudwatch.Dimension {
```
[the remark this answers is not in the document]
paulbrimicombe: One quick question @SaxyPandaBear: do you have opinions about the priority of the global tags vs. the `host` tag that's passed in via the per-metric dimension map (I'm thinking in terms of the max dimensions permitted here)? At the moment, the `host` has highest priority, then the global tags, then the other per-metric tags.

translator/translate/metrics/global_dimensions/ruleGlobalDimensions.go
SaxyPandaBear: Sorry I keep pushing for things that are different from what's around it. I appreciate that you looked into it for me. Resolving

plugins/outputs/cloudwatch/cloudwatch.go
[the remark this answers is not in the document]
SaxyPandaBear: I'll have to trace through it, considering I didn't know that `host` had a higher priority than the other metrics. I assumed it was equal priority before. I'll take a look today.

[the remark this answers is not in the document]
paulbrimicombe: While you're looking at this, I think the `sort.Strings(keys)` business is there to keep the dimensions used consistent when the number exceeds the max permitted (map iteration order is deliberately randomised in golang). I should probably add this to the global dimensions for the same reason.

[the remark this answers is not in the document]
paulbrimicombe: (plus there should probably be tests for that behaviour as well — these are currently missing for the existing code)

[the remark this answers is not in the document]
SaxyPandaBear: The fact that the host tag is explicitly called out with this comment:
`// This is pretty ugly but we always want to include the "host" tag if it exists.` 

Makes me think this is something we need to preserve. For better or for worse. There's too much risk in changing this specific behavior. 

With respect to the key sorting, agreed. I think we'll need to sort the global dimensions to maintain consistent behavior. Good call out. I didn't catch that in the initial pass through

translator/translate/metrics/global_dimensions/globalDimensions_test.go
```diff
@@ -0,0 +1,88 @@
...
+			"global_dimensions": map[string]interface{}{
+				"Environment": "test",
+				"Dimension":   "value",
+			},
+		}
+		assert.Equal(t, expected, actualValue, "Expect values to be equal")
+		assert.Equal(t, actualKey, "outputs", "Expect keys to be equal")
+	} else {
+		panic(err)
+	}
```
SaxyPandaBear, line 36: I think rather than doing this if-else, we should just use `assert.NoError(t, err)` and unwrap it. This applies to all of the checks in this test file. In this case, I am not concerned about other test files following this pattern because this needs to be tidied up here (if you saw it elsewhere feel free to point it out. we need to clean it up)

plugins/outputs/cloudwatch/cloudwatch.go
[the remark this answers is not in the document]
SaxyPandaBear: Will wait to resolve this comment until the `sort.Strings` is added for the global dimension keys

translator/translate/metrics/global_dimensions/globalDimensions_test.go
paulbrimicombe: Good point @SaxyPandaBear. I've fixed these now.

plugins/outputs/cloudwatch/cloudwatch.go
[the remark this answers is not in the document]
khanhntd: We can divide into two append_dimensions here:
* `append_dimensions` is for [each metric plugins](https://github.com/aws/amazon-cloudwatch-agent/blob/c85501042b088014ec40b636a8b6b2ccc9739738/translator/translate/metrics/util/commonconfigutil.go#L54-L57) which is added as a telegraf tag; therefore, will be added firstly as a tag
* `append_dimensions` which is the [ec2tagger](https://github.com/aws/private-amazon-cloudwatch-agent-staging/tree/b7f7db3cd86dd8e60c2911561b2d0f5622e86be0/plugins/processors/ec2tagger). Since its a processors, its will be added secondly as a tag
* Lastly would be your global dimensions which will added in the global dimensions

However, the [global dimensions are processed first](https://github.com/aws/amazon-cloudwatch-agent/blob/13af598623255e1dfebd5088212acf863c6b7476/plugins/outputs/cloudwatch/cloudwatch.go#L586-L596) so global dimensions will be prioritized before others `append_dimensions`

khanhntd, line 575:
```suggestion
func BuildDimensions(mTags , globalDimensions map[string]string) []*cloudwatch.Dimension {
```
Or would it better to pass the mTags and globalDimensions as a map which globalDimensions will override all the key from mTags. Then `sort.Strings(keys)`

plugins/outputs/cloudwatch/cloudwatch_test.go
```diff
@@ -59,6 +59,131 @@ func TestBuildDimensions(t *testing.T) {
...
+	}
+
+	for i, key := range tagKeys {
+		assert.Equal(key, *dimensions[i+len(globalDimensions)].Name, "Key should be equal")
+		assert.Equal(testPoint.Tags()[key], *dimensions[i+len(globalDimensions)].Value, "Value should be equal")
+	}
+}
+
+// Test that global dimensions are added correctly
+func TestGlobalDimensionsExceedMaxNumber(t *testing.T) {
```
khanhntd, line 125: IMO, it would be better to have a table driven test to [test different behaviors of global dimensions here](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests) . Your thoughs?

plugins/outputs/cloudwatch/cloudwatch_test.go
```diff
@@ -59,6 +59,131 @@ func TestBuildDimensions(t *testing.T) {
...
+	tagKeys := make([]string, len(testPoint.Tags()))
+	i := 0
+	for k := range testPoint.Tags() {
+		tagKeys[i] = k
+		i += 1
+	}
+
+	sort.Strings(tagKeys)
+
+	assert.Equal(len(testPoint.Tags())+len(globalDimensions), len(dimensions), "Number of dimensions should be equal to number of tags plus the number of global dimensions")
```
khanhntd, line 103:
```suggestion
	assert.Len(dimensions,len(testPoint.Tags())+len(globalDimensions), "Number of dimensions should be equal to number of tags plus the number of global dimensions")
```

plugins/outputs/cloudwatch/cloudwatch_test.go
```diff
@@ -59,6 +59,131 @@ func TestBuildDimensions(t *testing.T) {
...
+				matchingDimension = dimension
+			}
+		}
+
+		assert.NotNil(matchingDimension, "Global dimension should exist")
+		assert.Equal(value, *matchingDimension.Value, "Global dimension value should be equal")
+	}
+
+	for i, key := range tagKeys {
+		assert.Equal(key, *dimensions[i+len(globalDimensions)].Name, "Key should be equal")
```
khanhntd, line 119: It would be better to create a const  dimension= *dimensions[i+len(globalDimensions)]

translator/config/schema.json
```diff
@@ -100,6 +100,16 @@
           "minLength": 1,
           "maxLength": 255
         },
+        "global_dimensions": {
+          "type": "object",
+          "description": "Any dimensions to be applied to all published metrics",
+          "maxProperties": 30,
```
khanhntd, line 106: I wonder if we should limit the global dimensions. Since telegraf as default for each plugin will have the [default tag (e.g cpu will have total_cpu, cpu_core)](https://github.com/influxdata/telegraf/tree/master/plugins/inputs/cpu). Therefore, for avoid the global_dimensions override those one, would it be better to limit these here?

translator/translate/metrics/global_dimensions/globalDimensions.go
```diff
@@ -0,0 +1,42 @@
...
+const SectionKey = "global_dimensions"
+
+func (ad *globalDimensions) ApplyRule(input interface{}) (returnKey string, returnVal interface{}) {
+	im := input.(map[string]interface{})
+	dimensions := map[string]interface{}{}
+
+	returnKey = ""
+	returnVal = ""
+
+	if _, ok := im[SectionKey]; ok {
```
khanhntd, line 22:
```suggestion
	if _, ok := im[SectionKey].(map[string]interface{}); ok {
```

translator/translate/metrics/global_dimensions/globalDimensions_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGlobalDimensions(t *testing.T) {
+	e := new(globalDimensions)
+	var input interface{}
+	err := json.Unmarshal([]byte(`{
+      "global_dimensions": {
+				"Environment": "test",
```
khanhntd, line 18: Another key I would imagine to be invalid is `number` or at least out of bound string.

plugins/outputs/cloudwatch/cloudwatch.go
paulbrimicombe: If this were my codebase, I'd definitely prefer explicitly typing each argument (as in the original PR). But it's not my codebase so it's up to you.

There are issues with combining the maps because:

- The `host` dimension from `mTags` needs to be added first
- The current code opts to prioritise the global tags over the metric tags when `MaxDimensions` is reached so we need to keep the two separate until we've made that check.

plugins/outputs/cloudwatch/cloudwatch_test.go
paulbrimicombe: I like your suggestion for using `assert.Len()` but unfortunately in that case it would mean switching the expectation and the actual values around which might be misleading if this test fails. I'll look out for other places I can use `assert.Len()` though!

translator/config/schema.json
paulbrimicombe: It really depends on which Telegraf metric you are talking about; some have additional tags, some don't. I suspect that 30 is _way_ more than anyone is likely to need though so limiting to something smaller (e.g. 20?) might well make sense.

translator/translate/metrics/global_dimensions/globalDimensions.go
paulbrimicombe: Nice spot — I'll make this change.

translator/translate/metrics/global_dimensions/globalDimensions_test.go
paulbrimicombe: Yes, this is true _but_ the schema should give us protection from this shouldn't it? I'm moderately happy to add checks for different JSON data types as this doesn't increase complexity very much.

By "out of bound string" do you mean too long? The other surrounding translators don't seem to have these kind of checks in place but if you really want them I could add them.

plugins/outputs/cloudwatch/cloudwatch.go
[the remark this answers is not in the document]
paulbrimicombe: I'm not sure I understand this comment @khanhntd — are you suggesting we make a change here or describing how the current implementation works?

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -582,25 +583,41 @@ func BuildDimensions(mTags map[string]string) []*cloudwatch.Dimension {
...
+		if key != "" && value != "" && globalDimensions[key] == "" {
+			keys = append(keys, key)
 		}
+	}
+
+	if len(dimensions)+len(keys) > MaxDimensions {
+		log.Printf("D! max MaxDimensions %v is less than than number of dimensions %v thus only taking the max number", MaxDimensions, len(dimensions)+len(keys))
+		sort.Strings(keys)
+		keys = keys[:MaxDimensions-len(dimensions)]
+	}
```
paulbrimicombe, line 618: This change is so that we only bother to sort the keys if we actually need to (i.e. there are too many dimensions)

plugins/outputs/cloudwatch/cloudwatch.go
paulbrimicombe: OK I've had a quick look at this. The best way I can see of doing more of the work up-front for these global dimensions is to put some of the logic into the `startRoutines()` function. In there we could do the work of constructing a slice of dimensions (`*cloudwatch.Dimension[]`) based on the `GlobalDimensions` map before we start doing any actual work.

What this would mean, though, is that the testing becomes a bit funky because either:

- You have to do a bunch of slower (but higher fidelity) tests where you actually start the routines, publish some metrics, and then check that the correct dimensions were used
- Or you have to test the two bits of dimensions processing separately (you could stick the pre-processing of the global dimensions into one function and then pass in the `*cloudwatch.Dimension[]` to the `BuildDimensions` function) and hope that the two actually get used correctly in practice.

If you have a particular preference for one of these testing options and think that this is worth pursuing, please let me know and I'll give it a go.

SaxyPandaBear: Thanks for looking into it. My initial reaction is that trying to do this pre-optimization is too disruptive. I am _mildly_ concerned with always computing the global dimensions since my understanding is that this happens to every metric. 

I think one thing that I'd like to verify is agent CPU/mem load before and after this change. We could/should test with statsd to have control over the metric throughput and then we can observe CPU and memory usage.

plugins/outputs/cloudwatch/cloudwatch.go
SaxyPandaBear: I am opposed to combining the maps into one argument. I'm indifferent if we combine the types or not

translator/config/schema.json
SaxyPandaBear: I'd rather just keep it at 30 to be consistent with API guardrails than trying to reason a number in between

plugins/outputs/cloudwatch/cloudwatch_test.go
```diff
@@ -33,30 +33,288 @@ func TestBuildDimensions(t *testing.T) {
...
 
 	for i, key := range tagKeys {
-		if i >= 10 {
-			break
+		assert.Equal(key, *dimensions[i+len(globalDimensions)].Name, "Key should be equal")
+		assert.Equal(testPoint.Tags()[key], *dimensions[i+len(globalDimensions)].Value, "Value should be equal")
+	}
+}
+
+// Test that global dimensions don't override the host metric tag
```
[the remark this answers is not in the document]
SaxyPandaBear: Thanks will take a look today

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -571,36 +572,58 @@ func (c *CloudWatch) BuildMetricDatum(point telegraf.Metric) []*cloudwatch.Metri
...
+	if len(globalDimensionKeys)+len(dimensions) > MaxDimensions {
+		log.Printf("D! max MaxDimensions %v is less than than number of global dimensions plus host dimension (if present) %v thus only taking the max number", MaxDimensions, len(globalDimensionKeys)+len(dimensions))
+		sort.Strings(globalDimensionKeys)
+		globalDimensionKeys = globalDimensionKeys[:MaxDimensions-len(dimensions)]
+	}
 
-		value := mTags[k]
-		if value == "" {
+	for _, key := range globalDimensionKeys {
+		if hasHostDimension && key == "host" {
```
SaxyPandaBear, line 604: This would be an interesting use case... Do you think we need any guard against configuring `host` in the global dimensions map?

plugins/outputs/cloudwatch/cloudwatch_test.go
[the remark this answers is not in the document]
SaxyPandaBear: Was making a comment above about what you're referring to I assume - does it make any sense for someone to add `host` to their global dimensions? Is this something that we should guard against at translation time? I personally think it doesn't make sense to override existing behavior that we have for the host tag by letting customers drop a host tag in here. There is a check above in building the dimensions map to check if a global dimension key == host.. and as I'm thinking about it, I don't think that needs to be there if we guard against it during translation. Thoughts?

SaxyPandaBear (commented):
Super appreciative of the added unit tests. Just one open question that we probably want to discuss about overriding the host tag, and someone will likely have to take on doing some performance testing to see if there's any material impact with introducing this change (doesn't have to be you 😄  )

translator/translate/metrics/global_dimensions/globalDimensions_test.go
SaxyPandaBear: I'm indifferent about adding this

plugins/outputs/cloudwatch/cloudwatch_test.go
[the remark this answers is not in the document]
paulbrimicombe: The more I think about it, the more I think that excluding the `host` dimension from the global dimensions is a good idea. If someone has a specific use-case for overriding the host in the global dimensions then we can think again but otherwise it just seems like a source of unexpected behaviour / increased complexity / bugs etc.

I'll have a go at adding some checks for that in the translation layer — do you think adding something to exclude it in the schema is a good idea or would it be better to accept the invalid config and filter it out in the translation layer (possibly logging a warning that it's being dropped)?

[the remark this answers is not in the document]
paulbrimicombe: OK, I've just pushed a change that:

- Drops (silently) any `host` dimension from `global_dimensions` in the translator
- Drops (silently) any `host` dimension from `GlobalDimensions` in the CloudWatch output plugin

Strictly I didn't need to do the second but it seems moderately sensible to take a "belt and braces" approach as the additional logic required is very small (a single `key != "host"` check). Let me know what you think @SaxyPandaBear .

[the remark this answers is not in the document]
SaxyPandaBear: Thanks. Sorry was busy yesterday will take a look today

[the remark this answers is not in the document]
SaxyPandaBear: 😓 was somehow even busier yesterday. I promise I will look by EOD Friday but not sure when I'll have time to sit down and review

SaxyPandaBear:
LGTM but looks like failed unit tests after the change. Maybe I didn't look hard enough at the unit test code. Once it's passing I'm good to approve

khanhntd:
Another interesting find I have found is that: we could leverage [global_tags](https://github.com/influxdata/telegraf/blob/53f36570fd2ecec058dfde207d5fc3fd7b8b7753/docs/CONFIGURATION.md#global-tags) from telegraf. However, it would be override by tags within each plugin. Have we considered this option before? Will try to look the PR EOD.

SaxyPandaBear:
> This could be a breaking change if people have non aws: values in the append_dimensions field

I took this at face value, but after reviewing with product today, I think I don't understand the concern here. Don't we fail config translation today if you supply a non `aws:` prefixed value in the outer append_dimensions configuration? 

If there's no real concern about breaking existing customers, I think it makes sense to consolidate the logic into the existing `append_dimensions` field, though I do hate pushing for that so late after the fact. Realistically, the existing JSON structure is already confusing enough as it is, so I think we have to be diligent with knowing what warrants adding more bloat to it.

---
Separate thing to validate with the change e2e regardless of the above - it was brought to my attention that there are some metrics configurations that do not support an inner append_dimensions field. One thing that we'll need to verify is that the global dimensions are added regardless of that fact. For example, our statsd plugin does not allow for statsd-specific append dimensions, but it's likely that if you are trying to append global dimensions, you'll want that to be present

paulbrimicombe:
> LGTM but looks like failed unit tests after the change. Maybe I didn't look hard enough at the unit test code. Once it's passing I'm good to approve

Ah yes. Sorry I didn't spot these before (some of the `TestSEH1Distribution` tests always fail on my Mac so it's sometimes hard to spot genuine failures). I've pushed a fix for the broken tests now.

paulbrimicombe:
> I took this at face value, but after reviewing with product today, I think I don't understand the concern here. Don't we fail config translation today if you supply a non `aws:` prefixed value in the outer append_dimensions configuration?
> 
> If there's no real concern about breaking existing customers, I think it makes sense to consolidate the logic into the existing `append_dimensions` field, though I do hate pushing for that so late after the fact. Realistically, the existing JSON structure is already confusing enough as it is, so I think we have to be diligent with knowing what warrants adding more bloat to it.

I totally agree about limiting the complexity of the configuration but I've tried supplying non `aws:` prefixed values and they seem to be silently dropped.  Certainly no tests fail if you e.g. add some in here: https://github.com/aws/amazon-cloudwatch-agent/blob/main/translator/config/sampleSchema/validLinuxMetrics.json#L106 and I can't find any code in the `translator/translate/metrics/append_dimensions` file tree that would reject a value that doesn't match one of the currently-supported values.

paulbrimicombe:
> Separate thing to validate with the change e2e regardless of the above - it was brought to my attention that there are some metrics configurations that do not support an inner append_dimensions field. One thing that we'll need to verify is that the global dimensions are added regardless of that fact. For example, our statsd plugin does not allow for statsd-specific append dimensions, but it's likely that if you are trying to append global dimensions, you'll want that to be present

I can't see any reason why the global dimensions _wouldn't_ be added provided they still use the CloudWatch output plugin but it would be good to check end-to-end if possible.

SaxyPandaBear:
> and I can't find any code in the `translator/translate/metrics/append_dimensions` file tree that would reject a value that doesn't match one of the currently-supported values.

Ok. I'll take a look today to try to trace through where they're getting silently dropped.

SaxyPandaBear:
```
--- FAIL: TestBackoffRetries (66.47s)
    cloudwatch_test.go:807: 
        	Error Trace:	cloudwatch_test.go:807
        	Error:      	"1s" is not greater than "1.022734529s"
        	Test:       	TestBackoffRetries
FAIL
```
looks like a weird failure. I'm assuming this is flakiness and not related to your change. Will rerun the mac tests

paulbrimicombe:
> The key problem that is causing unrecognized values to be dropped is that there isn't a matching rule. It looks kind of messy and not fit for arbitrary strings, unfortunately. Because there is no child rule that extracts the arbitrary key-value pairs, they never make it into the translated config.

@SaxyPandaBear I think this is all deliberate in that the original developers _wanted_ to drop unsupported keys and the easiest way to do that was to iterate through the supported keys. It wouldn't be particularly difficult to extract any additional keys but my worry is that people who _do_ have additional keys in their config already will suddenly have new dimensions appended to their metrics which would be a breaking change.

paulbrimicombe:
> > my worry is that people who do have additional keys in their config already will suddenly have new dimensions appended to their metrics which would be a breaking change
> 
> My gut reaction is that if someone has configured an arbitrary key-value pair in their append dimensions and hasn't complained that they don't show up as dimensions on CW, they don't care either way. 

This was what the original bug report #172 was requesting / complaining about so some people obviously do care about this (including my organisation...).

> I'm a little less worried about this being a breaking change because it's meant for static strings, at least from a cost perspective. I can get confirmation to be sure, but I don't think that customers who do this should receive a spike in CW billing. If you start with [instanceID, autoScalingGroupName], there's already cardinality there so you're paying for unique metrics per instance ID and ASG name. Adding `"foo=bar"` to the dimension set is still going to produce the same number of metrics, just with another dimensions. At least that's what my gut tells me.

I'm very confident that you're correct that this won't introduce additional costs for clients. But that's not my main worry (see commentary below).

> Back to the main point, I don't even really see an issue operationally. I just tested something out myself. I set up a config with just instance ID as the global append dimension:
> 
> **snip...**
> 
> Then I set up a dashboard widget for the mem used metric, and also set an alarm.
> 
> Afterwards, I added InstanceType as a dimension:
> 
> **snip...**
> 
> And monitored the alarm / dashboard widget. The metrics still flowed to the alarm and widget.

All you have done here is confirmed the current behaviour — that unsupported `append_dimensions` keys are silently dropped when the JSON configuration is converted to TOML.  If we suddenly start treating arbitrary static key / value pairs in `append_dimensions` as dimensions to be added to all metrics, client's alarms / metric queries / search expressions will be broken because every metric will have a different set of dimensions to those included previously.

In your second example, your `mem_used_percent` metric would have another dimension of `InstnaceType`: `${aws:InstanceType}` added to it, meaning that both the alarm configuration and the dashboard widget would be broken (dimensions have to match exactly for CloudWatch alarms — if you used a very specific type of search expression, your dashboard widget _might_ be OK, but most likely it wouldn't be).

This is why I think we should consider using `append_dimensions` as the configuration key for these global dimensions as a breaking change (some client's monitoring will be broken when they upgrade to the latest version of the CloudWatch Agent).

If any of this doesn't make sense / seems wrong @SaxyPandaBear , please let me know!

SaxyPandaBear:
🤦 I had a typo in the config so it didn't append the other dimension. That's totally my bad. Okay that clears it up. I'll talk with others internally. Don't think there's a good way to figure out who would be impacted by this if we were to introduce the breaking change

paulbrimicombe:
> > My gut reaction is that if someone has configured an arbitrary key-value pair in their append dimensions and hasn't complained that they don't show up as dimensions on CW, they don't care either way.



> I'm agree with Andrew on this with using `append_dimensions` as a replace for global_dimensions

@khanhntd to me, as a client of this application, this would _definitely_ be a breaking change. My production dashboards / alarms / monitoring etc. would be broken by a routine update to the CloudWatch Agent for the same input configuration.

paulbrimicombe:
> * The implementation shares the same mechanism with `global_tags`; however, only the dimension's order is different so I'm still not sure the reason why we are not leverage it? (e.g using global_tags with non:AWS)

This is interesting. I'd  completely forgotten about the Telegraf `global_tags` option (I'd even used it a long time ago in a different project 🤦 ).

There seems to be some half-built(?) support for use of the built-in Telegraf `global_tags` (e.g. `translator/translate/globaltags/globaltags.go`) but it's untested, undocumented and doesn't seem to actually work with the current version of the agent (e.g. if I add a `global_tags` entry to the top level of the JSON config file, it never ends up in the TOML file — the service starts and creates the TOML file but something is filtering it out before it gets there).

If we _do_ use `global_tags` instead of this PR, there are some things we need to decide / work out:

1. Where `global_tags` should appear in the JSON. Currently it looks like it's being picked up from the root of the JSON -- should it really be in `metrics` instead?
2. According to the [Telegraf docs](https://github.com/influxdata/telegraf/blob/master/docs/CONFIGURATION.md#global-tags), "Global tags are overriden [sic] by tags set by plugins." — is this the behaviour we want?
3. Do we want to have deterministic behaviour wrt. global vs local tags when there are too many tags / dimensions? I don't know whether we have access to the information we'd need in the output plugin (will we have to do anything to pass through the information about which tags are global?)
4. How would we test that this works?

khanhntd:
1. Where `global_tags` should appear in the JSON. Currently it looks like it's being picked up from the root of the JSON -- should it really be in metrics instead?
Yes we can put a translator for it. (e.g if you have global dimension in JSON, put it in global tags). Therefore, you can put the global_dimensions as part of metrics and `global_tags` are not being used based on my research (internal and external)

2. According to the [Telegraf docs](https://github.com/influxdata/telegraf/blob/master/docs/CONFIGURATION.md#global-tags), "Global tags are overriden [sic] by tags set by plugins." — is this the behaviour we want?
IMO, yes. Plugin dimension should be prioritized over global plugin.  Since if the customers are not setting at plugin level, then there is no reason why the customers would expect it to begin with.

3. Do we want to have deterministic behaviour wrt. global vs local tags when there are too many tags / dimensions? I don't know whether we have access to the information we'd need in the output plugin (will we have to do anything to pass through the information about which tags are global?)
I have the same thought of limiting the [global tags](https://github.com/aws/amazon-cloudwatch-agent/pull/673#discussion_r1094636543) to put some constraint on it. However, there are two ways to avoid that: 
1. Using aggregating dimensions which will generate more metrics but shimming down the dimensions. However, increasing the cost for customers
2. Telegraf also has a feature of `tagdrop` or `tagfilter` to filter out all the unwanted tags and can be used to filter global tags if `global tags` + `plugin tags` are over thirdty.

The difference is defining the default behavior for agent and the intention on which perspective (customer or us).

4. How would we test that this works?
By all means, unit or manual testing or exploration testing. The most and easiest way would be writing the toml config for agent (e.g 
```
[global_tags]
  user = "${USER}"

[[inputs.mem]]
```
)

khanhntd:
The reason why I would bring up `global_tags` would be:
* Firstly, it won't be specific to any output plugins
* Secondly, support dimensions as part of traces and logs in the future
* Thirdly, reduce all the functional code that we want to rely on CloudWatch 

Let's me know what are your thoughts about this too !

paulbrimicombe:
> IMO, yes. Plugin dimension should be prioritized over global plugin. Since if the customers are not setting at plugin level, then there is no reason why the customers would expect it to begin with.

Sorry but I'm not sure what you mean by "if the customers are not setting at plugin level, then there is no reason why the customers would expect it to begin with". Bear in mind that customers may well not be in control of the output of plugins in use (e.g. they are from third-party tools etc.).  The original requests for global dimensions was to allow clients to add arbitrary dimensions to *all* metrics, which also wouldn't have been set at the plugin level.

> Telegraf also has a feature of tagdrop or tagfilter to filter out all the unwanted tags and can be used to filter global tags if global tags + plugin tags are over thirdty.

How can we know at runtime if global tags + plugin tags are over thirty? As I understand it, the Telegraf config for things like tagdrop and tagfilter are static and so can't change depending on the number of tags for a specific metric etc.

> Firstly, it won't be specific to any output plugins

This I feel may be a drawback in that it might have unintended consequences for other output plugins. The feature request is for CloudWatch metrics — are we *sure* that other output plugins would want the same global tags applied (now and in future)?

paulbrimicombe:
Pragmatically I think we need to try to actually come to some decision on this because it's turning out to be taking a long time.

Do we think that using `global_dimensions` is the right thing to use? That would mean:

- The additional dimensions would *always* be applied to *all* output plugins
- We have less fine-grained control over what happens when there are too many dimensions (e.g. deterministically choosing whether we prioritise plugin or global dimensions etc.)
- I worry a bit about testing (esp. for the conditions when the dimensions overflow)

Or, do we think using something like the code in the PR is the right thing to use? This would mean:

- Additional dimensions are only added to CloudWatch metrics
- We have fine-grained control over what happens when the number of dimensions overflows
- We can unit test lots of the additional code
- There is more code / more to maintain


## Notes on testing `global_tags`

- From the code, it [_looks_ like `global_tags` is supported](https://github.com/aws/amazon-cloudwatch-agent/blob/main/translator/translate/globaltags/globaltags.go).
- I can add unit tests to (e.g. by adding it to the JSON and TOML files here https://github.com/aws/amazon-cloudwatch-agent/blob/main/translator/totomlconfig/toTomlConfig_test.go#L166-L168) that prove that the translation works correctly (there are some changes required to the tests to make sure they actually test this properly)
- On a real copy of the agent, however, any `global_tags` I add to the JSON don't get migrated to the TOML on startup for some reason. They seem to get stripped out by _something_. Any idea why that would be?

khanhntd:
Hi @paulbrimicombe , 
Will come back to the PR within today. Sorry for taking a long time to respond though

etiennechabert:
Hello @khanhntd 5 months have passed, can you please share with us what is the opinion of the CloudWatch team on this issue and the solution proposed in this PR?

I think the limitation this PR is trying to address would already help a lot for the folks monitoring Ec2 tasks, since the dimensions you are forcing your users to use, might not make sense for some (most?) use cases. 

But even more necessary for ECS or Serverless applications.... for which the current interface of the CloudWatch Agent does not allow ANY global dimensions to be set!
See: https://github.com/aws/amazon-cloudwatch-agent/issues/864

Moreover, in theory, one could add a CloudWatch agent to his on-promise stack if he wants to use CloudWatch as a monitoring solution. So the CloudWatch Agent should allow some kind of generic configuration, agnostic of the stack used, and I think this PR goes exactly in this direction.

In general, when you look at the CloudWatch Agent interface you can really feel it was designed around EC2 which make sense historically, but AWS should now try to support and facilitate way more than that because of the portfolio of products it offers 🙏

etiennechabert:
Forced to "up" so the PR does not stale...
cc @khanhntd

khanhntd:
Hi @etiennechabert , I would recommend talk with @SaxyPandaBear, @chadpatel  or @kevin-aws  since I have not been with AWS from June 2023. Therefore, I unable to support anything more or less on behalf of AWS.

etiennechabert:
Hi @khanhntd, thanks for the transparency regarding this, let's give some time to the folks you mentioned to share an update regarding this matter.

SaxyPandaBear:
@etiennechabert I'm no longer on the team either. Sorry.

etiennechabert:
Not giving up 😅

@chadpatel @kevin-aws can you please share with us if this topic and PR is still on the radar of the Cloudwatch team ?

oleksandrsv:
Is it still in progress?

etiennechabert:
@oleksandrsv pretty convinced it have never been in progress on AWS side...
And they are pretty much switching to OTEL, don't expect changes to your Cloudwatch Agent in my opinion. I personally gave up on this one

Grandmother:
Maybe @lisguo, @Paramadon or @okankoAMZ could help? We'd like to have this too.

Issue #606, opened 2022-10-06
CloudWatch agent should support SDK process credential provider


SaxyPandaBear:
Currently, the agent only supports a subset of available forms of authentication provided by the AWS SDK for Go. One that's missing is [process credentials](https://docs.aws.amazon.com/sdkref/latest/guide/feature-process-credentials.html). The agent supports static credentials and assume role (temporary) credentials, but depends directly on the SDK `credentials` file, and doesn't invoke configured external processes that are set up in the SDK config. This can lead to obscure errors such as `SharedCredsLoad: failed to get profile` and `SharedCredsAccessKey: shared credentials the_role in /something/.aws/credentials did not contain aws_access_key_id`. When using process credentials, the SDK relies on the configured process to be invoked to populate the credential values in the credentials file, and the agent is not doing that, so it fails to find the profile or the credentials in the profile.

jomojowo:
Agreed, 

Currently below error is seen: 
when agent is used with eks irsa solution for the serviceaccount of the eks. 

2022-11-25T14:27:37Z W! [outputs.cloudwatchlogs] Retried 3 time, going to sleep 1.347279726s before retrying.
2022-11-25T14:27:38Z E! [outputs.cloudwatchlogs] Aws error received when sending logs to /aws/containerinsights/(cluster-name)/prometheus/kubernetes-service-endpoints: SharedCredsLoad: failed to load shared credentials file
caused by: FailedRead: unable to open file
caused by: open /root/.aws/credentials: no such file or directory
2022-11-25T14:27:38Z W! [outputs.cloudwatchlogs] Retried 4 time, going to sleep 2.911632124s before retrying.
2022-11-25T14:27:41Z E! [outputs.cloudwatchlogs] Aws error received when sending logs to /aws/containerinsights/(cluster-name)/prometheus/kubernetes-service-endpoints: SharedCredsLoad: failed to load shared credentials file
caused by: FailedRead: unable to open file

jomojowo:
image version: amazon/cloudwatch-agent:1.247354.0b251981
though I provided security context the below was still not fixed: 

          securityContext:
            runAsUser: 0.

PR #682, opened 2023-02-02
[GH-606][FEATURE] Add config mode for ModeWithIRSA to support using d…

stewartmoreland:
[GH-606][FEATURE] Add config mode for ModeWithIRSA to support using default credential provider with RUN_WITH_IRSA environment variable as True

# Description of the issue
Config mode OnPrem injects profile & file based credential configs to `amazon-cloudwatch-agent.toml`, thereby breaking support for using the default credential provider chain when running in EKS Anywhere or other k8s on-premises environments with IRSA.

Linked Issue: [GH-606](https://github.com/aws/amazon-cloudwatch-agent/issues/606)

# Description of changes
Change adds a new config mode (`ModeWithIRSA`) to prevent from adding profile & file based credentials and allowing the SDK to assume the default credential provider chain.

# License

# Tests
Unit tests have been executed and passed.

Agent has been deployed as DaemonSet to EKS Anywhere and validated functionality with `AssumeRoleWithWebIdentity` / IRSA.

### DIFF
--- a/translator/config/defaultConfig.go
+++ b/translator/config/defaultConfig.go
@@ -195,8 +195,6 @@ func DefaultJsonConfig(os string, mode string) string {
 	default:
 		if mode == ModeEC2 {
 			return defaultLinuxEC2Config
-		} else if mode == ModeWithIRSA {
-			return defaultLinuxOnPremConfig
 		} else {
 			return defaultLinuxOnPremConfig
 		}
--- a/translator/config/defaultConfig.go
+++ b/translator/config/defaultConfig.go
@@ -195,6 +195,8 @@ func DefaultJsonConfig(os string, mode string) string {
 	default:
 		if mode == ModeEC2 {
 			return defaultLinuxEC2Config
+		} else if mode == ModeWithIRSA {
+			return defaultLinuxOnPremConfig
 		} else {
 			return defaultLinuxOnPremConfig
 		}
--- a/translator/context/context.go
+++ b/translator/context/context.go
@@ -114,7 +114,6 @@ func (ctx *Context) SetMode(mode string) {
 		ctx.mode = config.ModeOnPrem
 	case config.ModeOnPremise:
 		ctx.mode = config.ModeOnPremise
-		ctx.mode = config.ModeOnPrem
 	case config.ModeWithIRSA:
 		ctx.mode = config.ModeWithIRSA
 	default:
--- a/translator/context/context.go
+++ b/translator/context/context.go
@@ -28,7 +28,7 @@ func CurrentContext() *Context {
 	return ctx
 }
 
-//Testing only
+// Testing only
 func ResetContext() {
 	ctx = nil
 }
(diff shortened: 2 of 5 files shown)


translator/util/sdkutil.go
```diff
@@ -38,6 +39,11 @@ func DetectAgentMode(configuredMode string) string {
 		return config.ModeEC2
 	}
 
+	if runWithIrsa == config.RUN_WITH_IRSA_TRUE {
```
khanhntd, line 42: Would there be any reason why this won't be the last check? (e.g after `if DefaultECSRegion() != ""{)`)

khanhntd (commented):
Would need to think more deeply about the implementation though ! However, thanks for the contribution and we are very appreciated for it.

translator/context/context.go
```diff
@@ -114,8 +114,11 @@ func (ctx *Context) SetMode(mode string) {
 		ctx.mode = config.ModeOnPrem
 	case config.ModeOnPremise:
 		ctx.mode = config.ModeOnPremise
+		ctx.mode = config.ModeOnPrem
```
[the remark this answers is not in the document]
stewartmoreland: Correct. This was unintentional. Updated latest commit removing this line. Thank you!

translator/util/sdkutil.go
khanhntd: No. Why would this check get prioritized  over all other checks and not as a last check before return its as onPrem

khanhntd: e.g Why not putting it here
```
if DefaultEC2Region() != "" {
		fmt.Println("I! Detected the instance is EC2")
		return config.ModeEC2
	}
	if DefaultECSRegion() != "" {
		fmt.Println("I! Detected the instance is ECS")
		return config.ModeEC2
	}
        if runWithIrsa == config.RUN_WITH_IRSA_TRUE {
               return config.ModeWithIRSA
        }
	fmt.Println("I! Detected the instance is OnPremise")
	return config.ModeOnPrem
```
but the reverse?

translator/config/defaultConfig.go
```diff
@@ -195,6 +195,8 @@ func DefaultJsonConfig(os string, mode string) string {
 	default:
 		if mode == ModeEC2 {
 			return defaultLinuxEC2Config
+		} else if mode == ModeWithIRSA {
+			return defaultLinuxOnPremConfig
```
SaxyPandaBear, line 199: I think that this case is captured in the `else` block already

translator/util/sdkutil.go
SaxyPandaBear: Because the RUN_WITH_IRSA is an override flag. It should come before we try to ping IMDS and ECS

SaxyPandaBear (approved):

translator/util/sdkutil.go
khanhntd: Sounds fair to me !

khanhntd (approved):
IMO, using flag would not be recommended  since its a hack way and needs to be known for every customers. Therefore, it's not an automated mechanism. However, I'm good with this shorted term solution since its benefit for all EKS customers running `onPremise`. Since its passing the sanity and its easy to roll back, I'm good with it.

khanhntd (approved):
LGTM. Thanks for the contribution and great deep dive on it !

SaxyPandaBear (approved):
LGTM

PR #689, opened 2023-02-21
Fix Containerd ContainerFS

sethAmazon:
# Description of the issue
ContainerFS type is missing from cw logs in containerd env

# Description of changes
Create tmp fix to cadvisor to add ContainerFS types in containerd env

# License

# Tests
Manually run in my cluster output
```
{
    "AutoScalingGroupName": "eks-seth-containerd-node-group-16c33a8a-68c7-7db4-be41-07631c0c4490",
    "ClusterName": "seth-containerd-test-cluster-2",
    "EBSVolumeId": "aws://us-west-2d/vol-0458202ac43d6492b",
    "InstanceId": "i-0f5d7e57b26f05773",
    "InstanceType": "t3.large",
    "Namespace": "kube-system",
    "NodeName": "ip-172-31-63-94.us-west-2.compute.internal",
    "PodName": "kube-proxy",
    "Sources": [
        "cadvisor",
        "calculated"
    ],
    "Timestamp": "1677012635251",
    "Type": "ContainerFS",
    "Version": "0",
    "container_filesystem_available": 0,
    "container_filesystem_capacity": 53674487808,
    "container_filesystem_usage": 0,
    "container_filesystem_utilization": 0,
    "device": "/dev/nvme0n1p1",
    "fstype": "vfs",
    "kubernetes": {
        "container_name": "kube-proxy",
        "containerd": {
            "container_id": "12777945ee2d1815f9fe8bc315b4299f2c1fd8d2cc33926fda6824d5eb93521e"
        },
        "host": "ip-172-31-63-94.us-west-2.compute.internal",
        "labels": {
            "controller-revision-hash": "85987847cc",
            "k8s-app": "kube-proxy",
            "pod-template-generation": "1"
        },
        "namespace_name": "kube-system",
        "pod_id": "9e3a2c97-06a6-499c-a085-1a893ab5c560",
        "pod_name": "kube-proxy-jjgz5",
        "pod_owners": [
            {
                "owner_kind": "DaemonSet",
                "owner_name": "kube-proxy"
            }
        ]
    }
}
```

# Requirements
2. Run `make linter`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -50,6 +50,9 @@ replace github.com/prometheus/prometheus => github.com/prometheus/prometheus v0.
 // go-kit has the fix for nats-io/jwt/v2 merged but not released yet. Replacing this version for now until next release.
 replace github.com/go-kit/kit => github.com/go-kit/kit v0.12.1-0.20220808180842-62c81a0f3047
 
+// Use cadvisor main release when https://github.com/google/cadvisor/issues/2785 is fixed
+replace github.com/google/cadvisor => github.com/sethAmazon/cadvisor v0.44.1-0.20230221203935-b689186cd42f
+
 require (
 	github.com/BurntSushi/toml v0.4.1
 	github.com/Jeffail/gabs v1.4.0
@@ -62,7 +65,7 @@ require (
 	github.com/bigkevmcd/go-configparser v0.0.0-20200217161103-d137835d2579
 	github.com/go-kit/kit v0.11.0
 	github.com/gobwas/glob v0.2.3
-	github.com/google/cadvisor v0.44.0
+	github.com/google/cadvisor v0.45.0-containerd-cri
 	github.com/google/go-cmp v0.5.8
 	github.com/hashicorp/golang-lru v0.5.4
 	github.com/influxdata/telegraf v0.0.0-00010101000000-000000000000
@@ -137,7 +140,6 @@ require (
 	github.com/cilium/ebpf v0.7.0 // indirect
 	github.com/cncf/xds/go v0.0.0-20211130200136-a8f946100490 // indirect
 	github.com/containerd/console v1.0.3 // indirect
-	github.com/containerd/containerd v1.6.8 // indirect
 	github.com/containerd/ttrpc v1.1.0 // indirect
 	github.com/coreos/go-semver v0.3.0 // indirect
 	github.com/coreos/go-systemd/v22 v22.3.2 // indirect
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -62,7 +65,7 @@ require (
 	github.com/bigkevmcd/go-configparser v0.0.0-20200217161103-d137835d2579
 	github.com/go-kit/kit v0.11.0
 	github.com/gobwas/glob v0.2.3
-	github.com/google/cadvisor v0.44.0
+	github.com/google/cadvisor v0.45.0-containerd-cri
```
sethAmazon, line 68: This is the official tracked containerd support. The code to get metrics is there but not turned on.

go.mod
```diff
@@ -50,6 +50,9 @@ replace github.com/prometheus/prometheus => github.com/prometheus/prometheus v0.
 // go-kit has the fix for nats-io/jwt/v2 merged but not released yet. Replacing this version for now until next release.
 replace github.com/go-kit/kit => github.com/go-kit/kit v0.12.1-0.20220808180842-62c81a0f3047
 
+// Use cadvisor main release when https://github.com/google/cadvisor/issues/2785 is fixed
+replace github.com/google/cadvisor => github.com/sethAmazon/cadvisor v0.44.1-0.20230221203935-b689186cd42f
```
sethAmazon, line 54: This turns on containerd containerfs metrics.

khanhntd: I wonder if we can  using `docker` until 1.23 and using `containerd` later on with 1.24

sethAmazon: That is correct

khanhntd (approved):

sky333999:
As per [this](https://github.com/google/cadvisor/issues/2785#issuecomment-1204837267), due to a circular dependency, the custom branch of cadvisor will never merge back to their main. They dont explicitly say if they will maintain the custom branch with all patches, security updates etc and with how much lag.

So us introducing a fork on top of an already custom branch seems risky. I dont have a better alternative but at the same time not sure if this is the right path.

khanhntd:
> As per https://github.com/google/cadvisor/issues/2785#issuecomment-1204837267, due to a circular dependency, the custom branch of cadvisor will never merge back to their main. They dont explicitly say if they will maintain the custom branch with all patches, security updates etc and with how much lag. https://github.com/google/cadvisor/issues/2785#issuecomment-1438893896 where people are complaining about this exact issue.
> 
> So us introducing a fork on top of an already custom branch seems risky. I dont have a better alternative but at the same time not sure if this is the right path.

Its  not a one way door. Therefore, what would be the worst solution of it though? I'm fine with updating it to have the `ContainerFS`

PR #700, opened 2023-03-03
Clean host

sethAmazon:
# Description of the issue
We have internal hosts we need to terminate after 1 day.

# Description of changes
Clean hosts that need to be terminated after 1 day. 

# License

# Tests
https://github.com/aws/amazon-cloudwatch-agent/actions/runs/4326874506

# Requirements
2. Run `make linter`

### DIFF
--- a/.github/workflows/clean-aws-resources.yml
+++ b/.github/workflows/clean-aws-resources.yml
@@ -95,3 +95,42 @@ jobs:
 
       - name: Clean old dedicated host
         run: go run ./tool/clean/clean_dedicated_host/clean_dedicated_host.go --tags=clean
+
+  clean-hosts:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
+    strategy:
+      matrix:
+        role: [ internal, external]
+        include:
+          - role: internal
+            role_secret: INTERNAL_AWS_ASSUME_ROLE
+          - role: external
+            role_secret: TERRAFORM_AWS_ASSUME_ROLE
+        region: [
+          us-east-1, us-east-2,
+          us-west-1, us-west-2,
+          ap-southeast-1, ap-southeast-2,
+          ap-south-1,
+          ap-northeast-1, ap-northeast-2, ap-northeast-3,
+          ca-central-1,
+          eu-central-1,
+          eu-west-1, eu-west-2, eu-west-3,
+          eu-north-1,
+          sa-east-1
+        ]
+      fail-fast: false
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v3
+
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          role-to-assume: ${{ secrets[matrix.role_secret] }}
+          aws-region: ${{ matrix.region }}
+
+      - name: Clean old dedicated host
+        run: go run ./tool/clean/clean_host/clean_host.go --tags=clean
--- a/tool/clean/clean_host/clean_host.go
+++ b/tool/clean/clean_host/clean_host.go
@@ -1,0 +1,92 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+//go:build clean
+// +build clean
+
+package main
+
+import (
+	"context"
+	"github.com/aws/amazon-cloudwatch-agent/tool/clean"
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+	"github.com/aws/aws-sdk-go-v2/service/ec2/types"
+	"github.com/aws/aws-sdk-go/aws"
+	"log"
+	"time"
+)
+
+// Clean integration hosts if they have been open longer than 1 day
+func main() {
+	err := cleanHost()
+	if err != nil {
+		log.Fatalf("errors cleaning %v", err)
+	}
+}
+
+func cleanHost() error {
+	log.Print("Begin to clean EC2 Host")
+
+	cxt := context.Background()
+	defaultConfig, err := config.LoadDefaultConfig(cxt)
+	if err != nil {
+		return err
+	}
+	ec2client := ec2.NewFromConfig(defaultConfig)
+
+	terminateInstances(cxt, ec2client)
+	return err
+}
+
+func terminateInstances(cxt context.Context, ec2client *ec2.Client) {
+	maxResults := int32(1000)
+	nameFilter := types.Filter{Name: aws.String("tag:Name"), Values: []string{
+		"buildLinuxPackage",
+		"buildPKG",
+		"buildMSI",
+		"MSIUpgrade_*",
+		"IntegrationTestBase",
+		"CWADockerImageBuilderX86",
+		"CWADockerImageBuilderARM64",
+	}}
+
+	instanceInput := ec2.DescribeInstancesInput{
+		Filters: []types.Filter{
+			nameFilter,
+			{Name: aws.String("instance-state-name"),
+				Values: []string{"running"}}},
+		MaxResults: maxResults}
+	for {
+		instanceIds := make([]string, 0)
+		expirationDateInstance := time.Now().UTC().Add(clean.KeepDurationOneDay)
+		describeInstanceOutput, _ := ec2client.DescribeInstances(cxt, &instanceInput)
+		for _, reservation := range describeInstanceOutput.Reservations {
+			for _, instance := range reservation.Instances {
+				log.Printf("instance id %v experation date %v host creation date raw %v host state %v",
+					*instance.InstanceId, expirationDateInstance, *instance.LaunchTime, instance.State)
+				if expirationDateInstance.After(*instance.LaunchTime) {
+					log.Printf("Try to delete instance %s tags %v launch-date %s", *instance.InstanceId, instance.Tags, *instance.LaunchTime)
+					instanceIds = append(instanceIds, *instance.InstanceId)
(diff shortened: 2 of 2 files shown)


tool/clean/clean_host/clean_host.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+	"github.com/aws/aws-sdk-go-v2/service/ec2/types"
+	"github.com/aws/aws-sdk-go/aws"
+	"log"
+	"time"
+)
+
+// Clean integration hosts if they have been open longer than 1 day
+const tagName = "tag:Name"
+var tagValue = []string{
```
sethAmazon, line 22: Tags we have in the internal except soaking

.github/workflows/clean-aws-resources.yml
```diff
@@ -94,4 +94,37 @@ jobs:
           aws-region: ${{ matrix.region }}
 
       - name: Clean old dedicated host
-        run: go run ./tool/clean/clean_dedicated_host/clean_dedicated_host.go --tags=clean
\ No newline at end of file
+        run: go run ./tool/clean/clean_dedicated_host/clean_dedicated_host.go --tags=clean
+
+  clean-internal-hosts:
```
khanhntd, line 99: It would be better to have a general clean: e.g clean external hosts. We could add a matrix to change between external and internal assume role.

tool/clean/clean_host/clean_host.go
```diff
@@ -0,0 +1,93 @@
...
+	_, err = ec2client.TerminateInstances(cxt, &terminateHost)
+	return err
+}
+
+func getHost(cxt context.Context, ec2client *ec2.Client) []types.Instance {
+	// Get list of host
+	nameFilter := types.Filter{Name: aws.String(tagName), Values: tagValue}
+
+	nextToken := aws.String("")
+	hosts := make([]types.Instance, 0)
```
khanhntd, line 77: Why would we need to make a length 0 here though? Are there any restrictions on `TerminateInstances` that we can leverage to initialize the array length beforehand?

tool/clean/clean_host/clean_host.go
khanhntd: If we are using in one place, I would highly recommend not initialize it but put it inside the function that use the variables.

tool/clean/clean_host/clean_host.go
```diff
@@ -0,0 +1,93 @@
...
+	hostInput := ec2.DescribeInstancesInput{Filters: []types.Filter{nameFilter, {Name: aws.String("instance-state-name"), Values: []string{"running"}}}}
+	for {
+		describeHostOutput, _ := ec2client.DescribeInstances(cxt, &hostInput)
+		for _, reservation := range describeHostOutput.Reservations {
+			for _, host := range reservation.Instances {
```
khanhntd, line 82: It would be better to confirm the instance[ state code](https://docs.aws.amazon.com/cli/latest/reference/ec2/describe-instances.html) for avoid adding unnecessary instances host.
```
instance-state-code - The state of the instance, as a 16-bit unsigned integer. The high byte is used for internal purposes and should be ignored. The low byte is set based on the state represented. The valid values are: 0 (pending), 16 (running), 32 (shutting-down), 48 (terminated), 64 (stopping), and 80 (stopped).
```

sethAmazon: Not needed since we filter by running state.

tool/clean/clean_host/clean_host.go
sethAmazon: I needed a dynamic array. We will not know the total size ahead of time, since the for loop can run multiple times. I think you are talking about not needed efficacy, the code just looks ugly. It would be like saying new arraylist in java but go requires initial size.

.github/workflows/clean-aws-resources.yml
sethAmazon: The external does not cleaning at least not from what I have seen

khanhntd: Its the same case with ADOT. We won't know. However, we don't need to add it now but only changing  the name to `clean-EC2-host` and also add a row for internal only

tool/clean/clean_host/clean_host.go
sethAmazon: Changed

tool/clean/clean_host/clean_host.go
sethAmazon: I ran this code and we had sub 10 deletions, so I assume this is a non issue.

.github/workflows/clean-aws-resources.yml
```diff
@@ -94,4 +94,43 @@ jobs:
...
+  clean-hosts:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
+    strategy:
+      matrix:
+        role: [ internal, current]
+        include:
+          - role: internal
```
sethAmazon, line 108: Using secrets in matrix is a pain

.github/workflows/clean-aws-resources.yml
```diff
@@ -94,4 +94,43 @@ jobs:
...
+            role_secret: INTERNAL_AWS_ASSUME_ROLE
+          - role: current
```
khanhntd, line 110:
```suggestion
          - role: external
```

tool/clean/clean_host/clean_host.go
```diff
@@ -0,0 +1,98 @@
...
+		terminateInstance := ec2.TerminateInstancesInput{InstanceIds: instanceIds}
+		_, err := ec2client.TerminateInstances(cxt, &terminateInstance)
+		if err != nil {
+			log.Printf("Error %v terminating instances %v", err, instanceIds)
+		}
+		nextToken = describeInstanceOutput.NextToken
+		if describeInstanceOutput.NextToken == nil {
+			break
+		}
+		instanceInput = ec2.DescribeInstancesInput{Filters: []types.Filter{
```
khanhntd, line 90:
```suggestion
		instanceInput.NextToken = nextToken
```

tool/clean/clean_host/clean_host.go
```diff
@@ -0,0 +1,98 @@
...
+	nextToken := aws.String("")
+	instanceInput := ec2.DescribeInstancesInput{
+		Filters: []types.Filter{
+			nameFilter,
+			{Name: aws.String("instance-state-name"),
+				Values: []string{"running"}}},
+		MaxResults: maxResults}
+	for {
+		instanceIds := make([]string, 0)
+		expirationDateInstance := time.Now().UTC().Add(clean.KeepDurationOneDay)
```
khanhntd, line 63: Nice ! I like the idea of since we are waiting some time to terminate the previous patch, we can calculate again the dates to terminate the current patch.

tool/clean/clean_host/clean_host.go
```diff
@@ -0,0 +1,98 @@
...
+		if describeInstanceOutput.NextToken == nil {
+			break
+		}
+		instanceInput = ec2.DescribeInstancesInput{Filters: []types.Filter{
+				nameFilter,
+				{Name: aws.String("instance-state-name"), Values: []string{"running"}}},
+			MaxResults: maxResults,
+			NextToken: nextToken}
+		// prevent throttle https://docs.aws.amazon.com/AWSEC2/latest/APIReference/throttling.html
+		time.Sleep(time.Minute)
```
khanhntd, line 96: Throttle is within second, so its better to reduce the time here (e.g 10 * time.Second perhaps)

khanhntd (commented):
Small nitpicks. If you feel its can be ignored, I will approve it.

tool/clean/clean_host/clean_host.go
sethAmazon: The docs for 1000 are 1 minute based. Read the doc.

khanhntd:
```
100 Describe* requests in one second. so you can make up to 100 Describe* requests in one second. If you exceed 100 requests in a second, you are throttled and the remaining requests within that second fail.
```

khanhntd: Since its a non-mutating actions, its fallin within this category - only has 100 Bucket Maximum Capacity.

khanhntd (approved):
LGTM other than the throttle issues.

tool/clean/clean_host/clean_host.go
sethAmazon:
```
For example, the resource token bucket size for RunInstances is 1000 tokens, and the refill rate is two tokens per second. Therefore, you can immediately launch 1000 instances, using any number of API requests, such as one request for 1000 instances or four requests for 250 instances. After the resource token bucket is empty, you can launch up to two instances every second, using either one request for two instances or two requests for one instance.
```

khanhntd: Its rare use case for me IMO and maybe I'm focusing too much on `request rate` not `resources rate`. Deleting 1000 EC2 instances for the first call and 200 instances for every next call would be fine for me ( in term of cost, its small to neglect it). However, its not worth to block it.

khanhntd (approved):
LGTM.

tool/clean/clean_host/clean_host.go
```diff
@@ -0,0 +1,92 @@
...
+					instanceIds = append(instanceIds, *instance.InstanceId)
+				}
+			}
+		}
+		if len(instanceIds) == 0 {
+			log.Printf("No instances to terminate")
+			return
+		}
+
+		log.Printf("instances to temrinate %v", instanceIds)
```
bryce-carey, line 79: nit: typo

bryce-carey (approved):

PR #718, opened 2023-03-15
fix: no dot when converting %f to timestamp layout

manger:
# Description of the issue
In a timestamp format, %f (fraction of seconds) represents the digits, not including the decimal mark (point or comma). So when converting a format to a layout, %f should be replaced with 000.

For example, the format "%H:%M:%S.%f" should be converted to the layout "15:04:05.000". Currently, the code incorrectly converts that format to "15:04:05..000" (note the two dots).

The capturing group for %f in TimeFormatRexMap looks wrong as no other mapping uses a capturing group.

# Description of changes
%f in a format is mapped to "000", instead of ".000", in a layout.

# License

# Tests
None.

# Requirements
2. Run `make linter`

### DIFF
--- a/translator/translate/logs/logs_collected/files/collect_list/ruleTimestampFormat.go
+++ b/translator/translate/logs/logs_collected/files/collect_list/ruleTimestampFormat.go
@@ -76,7 +76,7 @@ var TimeFormatMap = map[string]string{
 	"%p":  "PM",
 	"%Z":  "MST",
 	"%z":  "-0700",
-	"%f":  ".000",
+	"%f":  "000",
 }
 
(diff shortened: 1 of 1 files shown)


manger:
> @manger can you give a reproducible example for us to validate against?

The [default timestamp format for Log4J2 %d](https://logging.apache.org/log4j/2.x/manual/layouts.html#:~:text=Example-,%25d%7BDEFAULT%7D,-2012%2D11%2D02) produces timestamps such as
`2006-01-02 22:04:05,000` or `2012-11-02 14:34:02,123`
with fraction-of-seconds to millisecond precision, and a decimal comma.

That is an example that doesn't work without a fix.

PR #723, opened 2023-03-23
Add config option to disable extraction of metrics from EMF logs for …

sky333999:
…kubernetes

# Description of the issue
Currently, by default, all metrics for Kubernetes Container Insights are published as EMF logs and are extracted into metrics behind the scenes. What distinguishes a EMF log from a regular json log is the presence of `CloudWatchMetrics` field in the json (since the header to indicate is now optional).
A subset of customers in very specific scenarios need the ability to disable the default extraction of these metrics from EMF logs.

Ideally we want this option for ECS & Prometheus as well. But these will be made available in a future commit when we contribute a similar change to `awsemfexporter` in a central place that can be leveraged for all plugins using EMF logs to push metrics (which include ECS & Prometheus).

# Description of changes
Adds a config option that allows a customer to disable the default extraction of metrics from EMF logs.
When this option is set, we skip adding the `CloudWatchMetrics` blob to the json and hence there is nothing to extract. This will leave the datapoints in the logs as-is and just not create any metrics.

Sample log event when option is not set (default):
```
{
  "CloudWatchMetrics": [
    {
      "Metrics": [
        {
          "Unit": "Count",
          "Name": "service_number_of_running_pods"
        }
      ],
      "Dimensions": [
        [
          "Service",
          "Namespace",
          "ClusterName"
        ],
        [
          "ClusterName"
        ]
      ],
      "Namespace": "ContainerInsights/Disable-1.0"
    }
  ],
  "ClusterName": "cw1",
  "Namespace": "kube-system",
  "Service": "kube-dns",
  "Sources": [
    "apiserver"
  ],
  "Timestamp": "1679370382569",
  "Type": "ClusterService",
  "Version": "0",
  "kubernetes": {
    "namespace_name": "kube-system",
    "service_name": "kube-dns"
  },
  "service_number_of_running_pods": 2
}
```

Sample log event when option is set:
```
{
  "ClusterName": "cw1",
  "Namespace": "kube-system",
  "Service": "kube-dns",
  "Sources": [
    "apiserver"
  ],
  "Timestamp": "1679371926278",
  "Type": "ClusterService",
  "Version": "0",
  "kubernetes": {
    "namespace_name": "kube-system",
    "service_name": "kube-dns"
  },
  "service_number_of_running_pods": 2
}
```

# License

# Tests
* Unit tests
* Tested changes manually in an EKS cluster

# Requirements
2. Run `make linter`

### DIFF
--- a/plugins/processors/k8sdecorator/k8sdecorator_test.go
+++ b/plugins/processors/k8sdecorator/k8sdecorator_test.go
@@ -37,7 +37,7 @@ func TestDisableMetricExtraction(t *testing.T) {
 				},
 			},
 			DimensionSets: [][]string{{"ClusterName"}},
-			Namespace:     "",
+			Namespace:     "ContainerInsights",
 		},
 	}, m1.Fields()["CloudWatchMetrics"])
 
--- a/plugins/processors/k8sdecorator/k8sdecorator_test.go
+++ b/plugins/processors/k8sdecorator/k8sdecorator_test.go
@@ -13,41 +13,82 @@ import (
 )
 
 func TestDisableMetricExtraction(t *testing.T) {
+
 	tags := map[string]string{MetricType: TypeCluster}
 	fields := map[string]interface{}{MetricName(TypeCluster, NodeCount): 10, MetricName(TypeCluster, FailedNodeCount): 1}
 
-	m1 := metric.New("testClusterMetric", tags, fields, time.Now())
-	decorator := &K8sDecorator{
-		started:                 true,
-		DisableMetricExtraction: false,
-		ClusterName:             "TestK8sCluster",
-	}
-	decorator.Apply(m1)
-	assert.Equal(t, "Sources,CloudWatchMetrics", m1.Tags()["attributesInFields"])
-	assert.Equal(t, []structuredlogscommon.MetricRule{
-		{
-			Metrics: []structuredlogscommon.MetricAttr{
+	testCases := map[string]struct {
+		k8sDecorator               *K8sDecorator
+		expectedAttributesInFields string
+		expectedCloudWatchMetrics  interface{}
+	}{
+		"WithDisableMetricExtractionDefault": {
+			k8sDecorator: &K8sDecorator{
+				started:     true,
+				ClusterName: "TestK8sCluster",
+			},
+			expectedAttributesInFields: "Sources,CloudWatchMetrics",
+			expectedCloudWatchMetrics: []structuredlogscommon.MetricRule{
 				{
-					Unit: "Count",
-					Name: "cluster_node_count",
-				},
-				{
-					Unit: "Count",
-					Name: "cluster_failed_node_count",
+					Metrics: []structuredlogscommon.MetricAttr{
+						{
+							Unit: "Count",
+							Name: "cluster_node_count",
+						},
+						{
+							Unit: "Count",
+							Name: "cluster_failed_node_count",
+						},
+					},
+					DimensionSets: [][]string{{"ClusterName"}},
+					Namespace:     "ContainerInsights",
 				},
 			},
-			DimensionSets: [][]string{{"ClusterName"}},
-			Namespace:     "ContainerInsights",
 		},
-	}, m1.Fields()["CloudWatchMetrics"])
-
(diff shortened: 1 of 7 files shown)


translator/totomlconfig/tomlConfigTemplate/tomlConfig.go
```diff
@@ -323,12 +323,13 @@ type (
...
 	k8sDecoratorConfig struct {
-		ClusterName       string `toml:"cluster_name"`
-		HostIp            string `toml:"host_ip"`
-		NodeName          string `toml:"host_name_from_env"`
-		Order             int
-		PreferFullPodName bool `toml:"prefer_full_pod_name"`
-		TagService        bool `toml:"tag_service"`
-		TagPass           map[string][]string
+		ClusterName             string `toml:"cluster_name"`
+		DisableMetricExtraction bool   `toml:"disable_metric_extraction"`
```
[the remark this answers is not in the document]
khanhntd: Would be better to add in `schema.json` (as the rest of others options) since its the source of truth for customers to configure their CWAgent json beside the public document

plugins/processors/k8sdecorator/k8sdecorator_test.go
```diff
@@ -0,0 +1,53 @@
...
+	. "github.com/aws/amazon-cloudwatch-agent/internal/containerinsightscommon"
+	"github.com/aws/amazon-cloudwatch-agent/internal/structuredlogscommon"
+	"github.com/influxdata/telegraf/metric"
+	"github.com/stretchr/testify/assert"
+	"testing"
+	"time"
+)
+
+func TestDisableMetricExtraction(t *testing.T) {
+	tags := map[string]string{MetricType: TypeCluster}
```
khanhntd, line 16: Would be better to table unit test here with other options too (e.g test default case for `DisableMetricExtraction`) since the others configuration such as (e.g PrefFullPodName, NodeName) are missing here.

plugins/processors/k8sdecorator/k8sdecorator_test.go
```diff
@@ -0,0 +1,53 @@
...
+	fields := map[string]interface{}{MetricName(TypeCluster, NodeCount): 10, MetricName(TypeCluster, FailedNodeCount): 1}
+
+	m1 := metric.New("testClusterMetric", tags, fields, time.Now())
```
khanhntd, line 19: Another thing I would consider would be using [Given - When - Then](https://en.wikipedia.org/wiki/Given-When-Then) rule for unit test: to make it easier to understand the work flow since `attributes in Fields` is a hacky way of us passing logs events through pipeline.

translator/totomlconfig/tomlConfigTemplate/tomlConfig.go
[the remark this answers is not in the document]
sky333999: Will check with our PM and if that is indeed the case, will batch that update along with other options and also once we add this for ECS & Prometheus.

plugins/processors/k8sdecorator/k8sdecorator_test.go
sky333999: Updated to add a default case when this new flag isnt set at all. Lemme know if it looks good.

plugins/processors/k8sdecorator/k8sdecorator_test.go
sky333999: Updated as suggested!

plugins/processors/k8sdecorator/k8sdecorator_test.go
```diff
@@ -0,0 +1,53 @@
...
+	m2 := metric.New("testClusterMetric", tags, fields, time.Now())
+	decorator = &K8sDecorator{
+		started:                 true,
+		DisableMetricExtraction: true,
+		ClusterName:             "TestK8sCluster",
+	}
+	decorator.Apply(m2)
+	assert.Equal(t, "Sources", m2.Tags()["attributesInFields"])
+	assert.Equal(t, nil, m2.Fields()["CloudWatchMetrics"])
+}
```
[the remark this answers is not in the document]
sky333999: Updated as suggested!

taohungyang (approved):

khanhntd (approved):

PR #740, opened 2023-04-19
CX Wants Build Hash Version Added To Release Notes

sethAmazon:
# Description of the issue
CX Wants Build Hash Version Added To Release Notes

# Description of changes
Add build hash version release notes

# License

# Tests
N/A

# Requirements
2. Run `make linter`


RELEASE_NOTES
```diff
@@ -5,6 +5,9 @@ Amazon CloudWatch Agent 1.247358.0 (2023-01-30)
 Enhancements:
 * Add log output when log files are deleted with auto_removal feature (#659)
 
+Version With Build Hash:
```
SaxyPandaBear, line 8: I don't get this change. Wouldn't we want to just put this in the title section of the release notes change for the version like 
```
========================================================================
Amazon CloudWatch Agent 1.247358.0b252413 (2023-01-30)
========================================================================
```

zhihonl (approved):

RELEASE_NOTES
```diff
@@ -1,5 +1,5 @@
 ========================================================================
-Amazon CloudWatch Agent 1.247358.0 (2023-01-30)
+Amazon CloudWatch Agent 1.247358.0b252413 (2023-01-30)
```
khanhntd, line 2: What would be the demand for this though? For easier installation of the previous agent, if that's the case, should we have a force mechanism to always install with `1.247358.0` (version without hash) instead of always having `1.247358.0b252413` (hash). 

Small nitpick would be update the hash from the [release Note too](https://github.com/aws/amazon-cloudwatch-agent/releases/tag/v1.247358.0) if the change are merged.

sethAmazon: As it will change the tag and tagging happens before the hashing, I will edit the description of the release but not the tag.

SaxyPandaBear (approved):

Issue #300, opened 2021-10-23
Question/Feature request: how can I prepend an environment name to configured log groups at runtime


danwashusen:
**Scenario**
We have implemented a 'golden image' pipeline using AWS ImageBuilder that bakes in our base cloudwatch agent configuration (core logs files and system metrics),

e.g. 
```
...
    "logs": {
    "log_stream_name": "${aws:InstanceId}",
    "force_flush_interval" : 60,
    "logs_collected": {
      "files": {
        "collect_list": [
          {
            "file_path": "/var/log/syslog",
            "log_group_name": "syslog",
            "timestamp_format": "%b %d %H:%M:%S"
          },
...
```

This ends up with a massive log group with streams of every instance we're running in the region. We want to be able to separate those out into say '/production/syslog' and '/staging/syslog'...

**Work Around**
Our current work-around is to run `sed` over the base config included in the golden image early in the boot order...

e.g. `sed -E 's/"log_group_name": "(.*)"/"log_group_name": "\/ENV_NAME\/TIER\/\1"/g' /path/to/amazon-cloudwatch-agent.json`

```
  "logs": {
    "log_stream_name": "${aws:InstanceId}",
    "force_flush_interval" : 60,
    "logs_collected": {
      "files": {
        "collect_list": [
          {
            "file_path": "/var/log/syslog",
            "log_group_name": "/ENV_NAME/TIER/syslog",
            "timestamp_format": "%b %d %H:%M:%S"
          },
          {
            "file_path": "/var/log/auth",
            "log_group_name": "/ENV_NAME/TIER/auth",
            "timestamp_format": "%b %d %H:%M:%S"
          },
          {
            "file_path": "/var/log/audit/audit.log",
            "log_group_name": "/ENV_NAME/TIER/audit"
          },
          {
            "file_path": "/var/log/aide/aide.log",
            "log_group_name": "/ENV_NAME/TIER/aide"
          }
        ]
      }
    }
```

**Desired Behaviour**
It would be great if we could just declare arbitrary environment variables that are picked up by and agent and substituted in at runtime. e.g. `"log_group_name": "/${env.ENV_NAME}/${env.TIER}/aide"`

tmederpq:
I didn't find a good solution when I hit this roadblock about a month back, and wound up sort of flipping it on it's head. Thought I'd share my own work-around. 

[...]

Issue #738, opened 2023-04-17
[Feature Request] AutoScalingGroupName variable option for log_group_name / log_stream_name in Logs section


mengmann:
**Is your feature request related to a problem? Please describe.**
Currently there is no way use the instances AutoScalingGroupName as a variable either for `log_group_name` nor for `log_stream_name` configuration.

**Describe the solution you'd like**
... to have AutoScalingGroupName as variable to use it like this:
```json
{
  "logs": {
    "logs_collected": {
      "files": {
        "collect_list": [
          {
            "file_path": "/var/log/nginx/access.log",
            "log_group_name": "{autoscaling_group_name}",
            "log_stream_name": "/var/log/nginx/access.log",
            "timestamp_format": "[%a/%b/%Y:%H:%M:%S %z]",
            "timezone": "Local",
            "retention_in_days": 30
          },
          {
            "file_path": "/var/log/nginx/error.log",
            "log_group_name": "{autoscaling_group_name}",
            "log_stream_name": "/var/log/nginx/error.log",
            "timestamp_format": "%Y/%m/%d %H:%M:%S",
            "timezone": "Local",
            "retention_in_days": 30
          }
        ]
      }
    }
  }
}
```

**Describe alternatives you've considered**
Maybe it would be a good generic approach to make use of instance tags (needs sufficient IAM permission). But so we can basically add any tag key as a aggregation key for Cloudwatch Logs.
example: {tag:key}
```
"log_group_name": "{tag:aws:autoscaling:groupName}",
```

**Additional context**
Currently the list of available variables is limited to this: {instance_id}, {hostname}, {local_hostname}, and {ip_address}
In the metrics section one can use the following variables for dimensional metric aggregation:  ${aws:ImageId},  ${aws:InstanceId}, ${aws:InstanceType}  and already ${aws:AutoScalingGroupName}

So maybe there is some room for unification / consistency improvements here?

SaxyPandaBear:
> Maybe it would be a good generic approach to make use of instance tags

This was my first thought, because doing something outside of those bounds can become kind of invasive, permission-wise. We get this information via IMDS from the running host. That doesn't directly expose the AutoScalingGroup name. It is simpler for us to just read from tags (and gives a wider benefit for other customers who want tag based labeling). 

If we tried to avoid using instance tags, we would have to do some lookup from the InstanceID or something to figure out the auto scaling group (which probably would end up looking at tags anyway in the simplest form). 

[...]

PR #744, opened 2023-05-05
Tags

kixorz:
# Description of the issue
Hello, I've had the same problem like the author of [this issue](https://github.com/aws/amazon-cloudwatch-agent/issues/300). I'd like to use an instance tag to create a log group in the agent configuration to construct and display environment name based on a tag value.

Thank you for considering this change. If there are changes needed, please provide any suggestions to improve the solution.

# Description of changes
The change fetches a list of instance tags from EC2 instance metadata, stores it locally and uses the list to translate the tag placeholders in the configuration file.

# License

# Tests
I've tested the solution on EC2 instance with Enabled instance tags and a custom tag.
Steps:
1) Add tags of your choice, for example use `TestTag` and assign `TestValue`
2) Use tags in the cloudwatch-agent `config.json`, for example use `"log_stream_name": "{instance_id}_{tag:TestTag}"`
3) Enable instance metadata tags
4) Observe the translation of `{tag:TestTag}` to `TestValue` in `amazon-cloudwatch-agent.log`

### DIFF
--- a/translator/translate/util/placeholderUtil.go
+++ b/translator/translate/util/placeholderUtil.go
@@ -4,6 +4,7 @@
 package util
 
 import (
+	"fmt"
 	"log"
 	"net"
 	"os"
@@ -19,6 +20,7 @@ type Metadata struct {
 	Hostname   string
 	PrivateIP  string
 	AccountID  string
+	Tags       map[string]string
 }
 
 type MetadataInfoProvider func() *Metadata
@@ -30,6 +32,7 @@ var Ec2MetadataInfoProvider = func() *Metadata {
 		Hostname:   ec2.Hostname,
 		PrivateIP:  ec2.PrivateIP,
 		AccountID:  ec2.AccountID,
+		Tags:       ec2.Tags,
 	}
 }
 
@@ -90,10 +93,22 @@ func GetMetadataInfo(provider MetadataInfoProvider) map[string]string {
 		accountID = unknownAccountId
 	}
 
-	return map[string]string{instanceIdPlaceholder: instanceID, hostnamePlaceholder: hostname,
-		localHostnamePlaceholder: localHostname, ipAddressPlaceholder: ipAddress, awsRegionPlaceholder: awsRegion,
-		accountIdPlaceholder: accountID,
+	metadata := map[string]string{
+		instanceIdPlaceholder:    instanceID,
+		hostnamePlaceholder:      hostname,
+		localHostnamePlaceholder: localHostname,
+		ipAddressPlaceholder:     ipAddress,
+		awsRegionPlaceholder:     awsRegion,
+		accountIdPlaceholder:     accountID,
 	}
+
+	tags := provider().Tags
+	for tagKey, tagValue := range tags {
+		tagPlaceholder := fmt.Sprintf("{tag:%s}", tagKey)
+		metadata[tagPlaceholder] = tagValue
+	}
+
+	return metadata
 }
 
 func getHostName() string {
--- a/translator/util/ec2util/ec2util.go
+++ b/translator/util/ec2util/ec2util.go
@@ -48,7 +48,7 @@ func initEC2UtilSingleton() (newInstance *ec2Util) {
 		Tags:      make(map[string]string),
 	}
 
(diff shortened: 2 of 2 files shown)


kixorz:
Hello @sethAmazon - What can I do to possibly move this forward?

sethAmazon:
Thank you for the contribution. We will discuss this with our pm.

translator/util/ec2util/ec2util.go
```diff
@@ -115,5 +121,18 @@ func (e *ec2Util) deriveEC2MetadataFromIMDS() error {
...
+	if tagKeys, err := md.GetMetadata("tags/instance"); err == nil {
+		for _, tagKey := range strings.Fields(tagKeys) {
+			if tagValue, err := md.GetMetadata("tags/instance/" + tagKey); err == nil {
+				e.Tags[tagKey] = tagValue
+			} else {
+				fmt.Printf("E! [EC2] Fetch instance tag %s from EC2 metadata fail:%v\n", tagKey, err)
+			}
+		}
+	} else {
+		fmt.Println("E! [EC2] Fetch instance tags from EC2 metadata fail:", err)
```
SaxyPandaBear, line 134: Just to be clear, my understanding is we don't want to return an error here because retrieving instance tags from IMDS requires customers to explicitly opt-in, right? Or was `return err` missed here?

translator/translate/util/placeholderUtil.go
```diff
@@ -90,10 +93,22 @@ func GetMetadataInfo(provider MetadataInfoProvider) map[string]string {
...
+		localHostnamePlaceholder: localHostname,
+		ipAddressPlaceholder:     ipAddress,
+		awsRegionPlaceholder:     awsRegion,
+		accountIdPlaceholder:     accountID,
 	}
+
+	tags := provider().Tags
+	for tagKey, tagValue := range tags {
+		tagPlaceholder := fmt.Sprintf("{tag:%s}", tagKey)
+		metadata[tagPlaceholder] = tagValue
```
SaxyPandaBear, line 108: I think we're very fortunate that instance tags are not allowed to use `{` and `}` characters in the key name when pulling from IMDS, because I was going to say that this needed to be changed to guard against that, otherwise. https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html#tag-restrictions

SaxyPandaBear (commented):
Overall looks good to me. I appreciate the contribution. I will reach out to our product team to see if getting this feature in is a current priority.

translator/util/ec2util/ec2util.go
kixorz: Yes - it's a soft error based on the opt-in. I felt like this should not be a hard error because the pattern in the file seems to be just printing the errors.

translator/translate/util/placeholderUtil.go
kixorz: Thanks for this comment! I did consider the possible requirement to do some validation - however, as you're saying, we shouldn't need it in this case.

SaxyPandaBear:
One outstanding question I have is if this also covers AWS reserved tags. It would be really awesome if this contribution also covered the use case in #738

kixorz:
I have confirmed through testing that with this change an autoscaling group propagating `aws:ec2launchtemplate:id` or other tags can be referenced in the config by `{tag:aws:ec2launchtemplate:id}` and are correctly translated to the tag value.

kixorz:
Are we good on this one?

SaxyPandaBear:
@kixorz apologies, I haven't heard from our product team yet on this. I'll try to follow up. From a technical standpoint, it looks good to me.

nickgalantowicz:
@SaxyPandaBear @kixorz can we put eyes and effort to get this through? Would be a nice feature to have.

kixorz:
Seems like the team doesn't want to move this forward. Maybe some communication clarity around this would help?

nickgalantowicz:
@SaxyPandaBear @nathalapooja @adam-mateen @mitali-salvi Can we get insights on the teams thoughts with this pr? Hoping to at least get some clarity on why or why not this is progressing.

PR #753, opened 2023-05-18
Add daily cleanup for integration test resource

zhihonl:
# Description of the issue
Currently when terraform fails during integration test, some resources will not get destroyed. Sometimes when the test fails the terraform data on instance launch is lost therefore making it difficult to enforce cleanup after test failure.

# Description of changes
Add a name regex to daily resource cleanup code that terminates instances with tag name value of `cwagent-integ-test-ec2` when the instance existed for longer than a day which is sufficient time to run the current tests. Also adds OS names to the instance names for better readabiilty.

# License

# Tests
In this [test run](https://github.com/zhihonl/amazon-cloudwatch-agent/actions/runs/5017645071/jobs/8996025587), the name regex is used to terminate all instances with name starting with cwagent-integ-test-ec2. It has been verified that instances that were created with same naming scheme but not one day old remained running.

# Requirements
2. Run `make linter`

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -498,6 +498,7 @@ jobs:
             terraform init
             if terraform apply --auto-approve \
               -var="ssh_key_value=${PRIVATE_KEY}" -var="github_test_repo=${{env.CWA_GITHUB_TEST_REPO_URL}}" \
+              -var="test_name=${{ matrix.arrays.os }}" \
               -var="cwa_github_sha=${GITHUB_SHA}" -var="install_agent=${{ matrix.arrays.installAgentCommand }}" \
               -var="github_test_repo_branch=${{env.CWA_GITHUB_TEST_REPO_BRANCH}}" \
               -var="ec2_instance_type=${{ matrix.arrays.instanceType }}" \
@@ -598,6 +599,7 @@ jobs:
             terraform init
             if terraform apply --auto-approve \
               -var="ssh_key_value=${PRIVATE_KEY}" -var="github_test_repo=${{env.CWA_GITHUB_TEST_REPO_URL}}" \
+              -var="test_name=${{ matrix.arrays.os }}" \
               -var="cwa_github_sha=${GITHUB_SHA}" -var="install_agent=${{ matrix.arrays.installAgentCommand }}" \
               -var="github_test_repo_branch=${{env.CWA_GITHUB_TEST_REPO_BRANCH}}" \
               -var="ec2_instance_type=${{ matrix.arrays.instanceType }}" \
@@ -674,6 +676,7 @@ jobs:
             if terraform apply --auto-approve \
             -var="ec2_instance_type=${{ matrix.arrays.instanceType }}" \
             -var="ssh_key_value=${PRIVATE_KEY}" -var="ssh_key_name=${KEY_NAME}"  \
+            -var="test_name=${{ matrix.arrays.os }}" \
             -var="cwa_github_sha=${GITHUB_SHA}"  \
             -var="test_dir=${{ matrix.arrays.test_dir }}" \
             -var="ami=${{ matrix.arrays.ami }}" \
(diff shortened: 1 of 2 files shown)


sethAmazon (approved):

.github/workflows/integration-test.yml
```diff
@@ -498,6 +498,7 @@ jobs:
             terraform init
             if terraform apply --auto-approve \
               -var="ssh_key_value=${PRIVATE_KEY}" -var="github_test_repo=${{env.CWA_GITHUB_TEST_REPO_URL}}" \
+              -var="test_name=${{ matrix.arrays.os }}" \
```
adam-mateen, line 501: Does `matrix.arrays.os` store the operating system?
I am confused why it is being assigned into another variable called `test_name`?

tool/clean/clean_host/clean_host.go
```diff
@@ -49,6 +49,7 @@ func terminateInstances(cxt context.Context, ec2client *ec2.Client) {
 		"IntegrationTestBase",
 		"CWADockerImageBuilderX86",
 		"CWADockerImageBuilderARM64",
+		"cwagent-integ-test-ec2*",
```
adam-mateen, line 52: When does this cleaner run?
Is it possible the cleaner runs while tests are running and therefore cause false test failures?

.github/workflows/integration-test.yml
zhihonl: `matrix.arrays.os` stores the OS that the test is running on. It is assigned to test_name because that is the variable name being used to distinguish the instance names for different OS. For example looking at `main.tf` file for mac in [amazon-cloudwatch-agent-test repo](https://github.com/aws/amazon-cloudwatch-agent-test/blob/88b1a960e56528436473416a7040b06375917e64/terraform/ec2/mac/main.tf#L70), it's the variable being used to dynamically name the instances.

We could add a variable called `test_os` instead for clarity but it accomplishes the same thing.

tool/clean/clean_host/clean_host.go
jefchien: There's a check in the cleaner code that will skip instances that are newer than a day, so shouldn't affect running tests.

.github/workflows/integration-test.yml
jefchien: In your example, we already differentiate using the prefix `cwagent-integ-test-ec2-mac-` which includes the OS it's running on. There's no harm in adding this, but I'm not sure I understand why it should be the `matrix.array.os`.

tool/clean/clean_host/clean_host.go
zhihonl: This cleaner run every day at 0:00 UTC and it's impossible for tests to be running and get deleted incorrectly because the cleaner will check if the instance has been running for more than a than day, and if so it will terminate it. I don't think we have any test besides canary and soaking tests that take that long but those should be separated from t his process.

The issue with cleanup after test calls is sometimes a failure or cancellation will completely skip the terraform destroy even if a separate job was created for it. A good example is [this run](https://github.com/zhihonl/amazon-cloudwatch-agent/actions/runs/5017163339/jobs/8995177970). The workflow was canceled but the instance still remained on my personal account even though we have a separate destroy job after `terraform apply` as an attempt to mitigate it.

.github/workflows/integration-test.yml
zhihonl: Adding `test_name` to `cwagent-integ-test-ec2-mac-` will also specify which major version of Mac we are using, since it's referring to the os in [this file](https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/generator/resources/ec2_mac_test_matrix.json).

As for why it should be `matrix.array.os`, it's the only variable that is common to all test_matrix and has clear distinctions. It's also what is being used in the internal repo so I thought that makes the most sense.

zhihonl: Technically for the main purpose of this PR, the test_name addition is not necessary, but it's just something I noticed during testing that all of our test instances end up with names such as `cwagent-integ-test-ec2---90a1bd2bdc0f9e70)` which makes it hard to pinpoint which stage of integration test we should look into since each stage is usually OS based.

jefchien: So instead it would be something like `cwagent-integ-test-ec2-mac-macOS Big Sur Amd64-[testing-id]`. Seems fine.

zhihonl: That's a good point. I don't think EC2 tag allows space. In this case we can change the `os` variable value in the matrix file or just remove `test_name` and leave it as it is for Mac. I'm fine with either option.

adam-mateen (approved):

PR #754, opened 2023-05-19
Add EKS Prometheus Integration Test

JayPolanco:
# Description of the issue
Adds a new github action for testing prometheus on EKS

# Description of changes
- Adds an eks_deployment_matrix to generate matrix for the eks prometheus test.
- Creates a EKSPrometheus github action for testing prometheus on eks. 

# License

# Tests
Tested in a branch of the main repo

# Requirements
2. Run `make linter`

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -1078,7 +1078,7 @@ jobs:
             fi
 
       - name: Terraform destroy
-        if: ${{ cancelled() && steps.eks-ec2-integration-test.outputs.cache-hit != 'true' }}
+        if: ${{ cancelled() || failure() }}
         uses: nick-fields/retry@v2
         with:
           max_attempts: 3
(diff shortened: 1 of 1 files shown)


.github/workflows/integration-test.yml
```diff
@@ -1010,6 +1013,79 @@ jobs:
...
+            terraform init
+            if terraform apply --auto-approve \
+              -var="test_dir=${{ matrix.arrays.test_dir }}"\
+              -var="cwagent_image_repo=${{ steps.login-ecr.outputs.registry }}/${{ env.ECR_INTEGRATION_TEST_REPO }}" \
+              -var="cwagent_image_tag=${{ github.sha }}" \
+              -var="k8s_version=${{ matrix.arrays.k8s_version }}"; then
+              terraform destroy -auto-approve
+            else
+              terraform destroy -auto-approve && exit 1
+            fi
```
SaxyPandaBear, line 1078: So with the changes from the ECS/EKS statsd templating, would it make more sense for us to add to that test generation, rather than duplicate the GitHub Action? @movence

JayPolanco: That was something I considered, but decided not to in order to make a clear distinction between the deployment types as those use daemon and this one uses deployment. We could change it to be one large integration test, but I'd image testing time would increase and I don't know if that trade off is worth it

movence: I think one EKS integration test `EKSIntegrationTest` makes sense here since now we can specify `terraformDir` to have tests with different infra configs. Envisioning one EKS github action with all tests under it as jobs including statsd, prometheus and etc. It feels like we don't fully utilize the benefits of having `terraformDir` and `TestSuite` when each test types get separated out to their own.

SaxyPandaBear (approved):

movence:
LGTM we can revisit the structure later if needed

movence (approved):

PR #767, opened 2023-06-16
Add Block Reading Of Next Line If Logs Agent Buffersize Is Reached

sethAmazon:
# Description of the issue

# Description of changes

# License

# Tests

# Requirements
2. Run `make linter`


translator/config/schema.go
```diff
@@ -880,6 +884,11 @@ var schema = `{
       "minLength": 4,
       "maxLength": 2048
     },
+    "maxCloudwatchLogsBufferDefinition": {
+      "type": "integer",
+      "minimum": 1000000,
```
sethAmazon, line 889: While this is technically smaller than 1mb. I think it is close enough.

translator/translate/logs/ruleMaxCloudwatchLogsBuffer.go
```diff
@@ -0,0 +1,26 @@
...
+)
+
+const key = "max_cloudwatch_logs_buffer"
+
+type MaxCloudwatchLogsBuffer struct {
+}
+
+func (m *MaxCloudwatchLogsBuffer) ApplyRule(input interface{}) (returnKey string, returnVal interface{}) {
+	returnKey = Output_Cloudwatch_Logs
+	_, val := translator.DefaultMaxCloudwatchLogsBuffer(key, float64(-1), input)
```
sethAmazon, line 17: -1 means we would rather not use this feature. If nothing is given, we will have -1 as the value.

logs/util/read_logs_blocker.go
```diff
@@ -0,0 +1,60 @@
...
+
+func (l *LogBlocker) Subtract(v int64) {
+	l.logsBufferSizeMutex.Lock()
+	defer l.logsBufferSizeMutex.Unlock()
+	l.logsBufferSize = l.logsBufferSize - v
+}
+
+// Reset this is test only code
+// do not use for application
+func (l *LogBlocker) Reset() {
```
sethAmazon, line 43: I'm not sure how to get around having this here, even though this should be for testing only

sethAmazon: I'm not certain how to deal with test side effects. Since this is a singleton instance.

translator/defaultKeyCase.go
```diff
@@ -57,6 +57,19 @@ func DefaultIntegralCase(key string, defaultVal, input interface{}) (returnKey s
 	return
 }
 
+func DefaultInteger64Case(key string, defaultVal, input interface{}) (returnKey string, returnVal interface{}) {
```
sethAmazon, line 60: I am uncertain if this is a little cargo cult programming. We do need to parse from float64, but I wonder if we need this wrapper function or should we move this logic to the calling function.

translator/config/schema.json
```diff
@@ -606,6 +606,10 @@
         "endpoint_override": {
           "description": "The override endpoint to use to access cloudwatch logs",
           "$ref": "#/definitions/endpointOverrideDefinition"
+        },
+        "max_cloudwatch_logs_buffer": {
+          "description": "The max number of bytes buffer to send to cloudwatch logs",
```
sethAmazon, line 611: Should we use bytes or another unit?

sethAmazon:
This code should be down streamed to ccwa after merge.

translator/translate/agent/ruleGlobalMaxCloudwatchLogsBuffer.go
```diff
@@ -0,0 +1,27 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+package agent
```
sethAmazon, line 4: This is so scroll will get the same value.

translator/translate/logs/ruleMaxCloudwatchLogsBuffer.go
```diff
@@ -0,0 +1,25 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+package logs
```
sethAmazon, line 4: This will need to be copied for scroll.

logs/util/read_logs_blocker.go
SaxyPandaBear: The problem is relying on a global singleton instead of using composition so we could replace it with a test log blocker.

logs/util/read_logs_blocker.go
```diff
@@ -0,0 +1,60 @@
...
+	l.logsBufferSizeMutex.Lock()
+	defer l.logsBufferSizeMutex.Unlock()
+	l.logsBufferSize = 0
+}
+
+func (l *LogBlocker) SetMaxLogBuffer(v int64) {
+	l.maxLogBufferSize = v
+}
+
+func (l *LogBlocker) Block() (block bool, bufferSize int64, maxBufferSize int64) {
```
SaxyPandaBear, line 53: It is not obvious to me what a LogBlocker does.

plugins/inputs/logfile/tail/tail.go
```diff
@@ -225,6 +226,16 @@ func (tail *Tail) readLine() (string, error) {
 	tail.lk.Lock()
 	defer tail.lk.Unlock()
 
+	// do not allow line reading if the output pusher size is greater than configured
+	block, bufferSize, maxBufferSize := util.GetLogBlocker().Block()
+	for block {
+		tail.Logger.Infof("max buffer of logs size sending to cloudwatch " +
+			"blocking reading for one second for file %s max buffer %d current buffer %d",
+			tail.Filename, maxBufferSize, bufferSize)
+		time.Sleep(time.Second)
```
SaxyPandaBear, line 235: Why do we need to wait a whole second?

plugins/inputs/logfile/tail/tail.go
```diff
@@ -225,6 +226,16 @@ func (tail *Tail) readLine() (string, error) {
 	tail.lk.Lock()
 	defer tail.lk.Unlock()
 
+	// do not allow line reading if the output pusher size is greater than configured
+	block, bufferSize, maxBufferSize := util.GetLogBlocker().Block()
+	for block {
```
SaxyPandaBear, line 231: How will we handle this in the shutdown scenario for the agent? In the scenario that we are blocking indefinitely, at what point does this tailing goroutine terminate?

plugins/outputs/cloudwatchlogs/pusher.go
```diff
@@ -100,13 +103,16 @@ func (p *pusher) AddEventNonBlocking(e logs.LogEvent) {
...
+	util.GetLogBlocker().Add(int64(len(e.Message())))
+
 	// Drain the channel until new event can be added
 	for {
 		select {
 		case p.nonBlockingEventsCh <- e:
 			return
 		default:
 			<-p.nonBlockingEventsCh
+			util.GetLogBlocker().Subtract(int64(len(e.Message())))
```
SaxyPandaBear, line 115: Instead of adding `int64(len(e.Message()))` everywhere, can't we just add a `Size()` to the `LogEvent` struct?

translator/translate/logs/logs_test.go
```diff
@@ -18,8 +18,10 @@ import (
 )
 
 func TestLogs(t *testing.T) {
+
```
SaxyPandaBear, line 21: Why do we need to add this newline here?

SaxyPandaBear (commented):
This approach doesn't make sense to me. Why are we using a global semaphore to restrict reading across the agent, when the actual mechanism for pushing the logs is separate per file configuration?

logs/util/read_logs_blocker.go
sethAmazon: I doubt that will work because the weighted semaphore will try to acquire lock on addition. We need to force acquire lock because we add the header size at and change the state size for truncating. Thus, we need to have a flexible size that allows the exporting path to add size when it needs to. So, this is not really a hard stop at 1mb (configurable) it is a soft spot that blocks the next execution of reading and opening new files. Since the exporter should constantly change line size state.

sethAmazon: Unfortunately, it is up to the reader to stop itself. Instead of fully blocking.

sethAmazon: A log event can already be added to the buffer, but before flushing, the size will change.

plugins/inputs/logfile/tail/tail.go
sethAmazon: No. This can change to a different period of time if you have any ideas.

logs/util/read_logs_blocker.go
sethAmazon: Converted to a struct that is passed down to all required places.

plugins/inputs/logfile/tail/tail.go
sethAmazon: Added a context with cancel from the background context, so it exits the chan on parent context cancel.

plugins/outputs/cloudwatchlogs/pusher.go
sethAmazon: Added Size()

translator/translate/logs/logs_test.go
sethAmazon: Not needed probs just some weird formatting from intelij.

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -137,18 +140,24 @@ func (c *CloudWatchLogs) getDest(t Target) *cwDest {
...
+
+
 func (c *CloudWatchLogs) writeMetricAsStructuredLog(m telegraf.Metric) {
 	t, err := c.getTargetFromMetric(m)
 	if err != nil {
 		c.Log.Errorf("Failed to find target: %v", err)
 	}
-	cwd := c.getDest(t)
+	// emf logs are already bound by number of metrics
+	cwd := c.getDest(t, util.DefaultLogBlocker())
```
sethAmazon, line 160: emf logs are being moved to otel anyway

sethAmazon:
Screenshot 2023-06-28 at 12 26 38 PM

sethAmazon:
Screenshot 2023-06-28 at 12 36 05 PM

PR #771, opened 2023-06-21
Correcting latency health metric units to be milliseconds

bryce-carey:
# Description of the issue
The agent stat for latency is being published in nanoseconds and it should be in milliseconds.

# Description of changes
* Changing the value of the latency to milliseconds
* Minor change to logs pusher `RecordOpData` call for `CreateLogGroup`

# License

# Tests
make

# Requirements
2. Run `make linter`

### DIFF
--- a/cfg/agentinfo/info.go
+++ b/cfg/agentinfo/info.go
@@ -61,13 +61,13 @@ type agentInfo struct {
 }
 
 type agentStats struct {
-	CpuPercent          *float64       `json:"cpu,omitempty"`
-	MemoryBytes         *uint64        `json:"mem,omitempty"`
-	FileDescriptorCount *int32         `json:"fd,omitempty"`
-	ThreadCount         *int32         `json:"th,omitempty"`
-	LatencyMillis       *time.Duration `json:"lat,omitempty"`
-	PayloadBytes        *int           `json:"load,omitempty"`
-	StatusCode          *int           `json:"code,omitempty"`
+	CpuPercent          *float64 `json:"cpu,omitempty"`
+	MemoryBytes         *uint64  `json:"mem,omitempty"`
+	FileDescriptorCount *int32   `json:"fd,omitempty"`
+	ThreadCount         *int32   `json:"th,omitempty"`
+	LatencyMillis       *int64   `json:"lat,omitempty"`
+	PayloadBytes        *int     `json:"load,omitempty"`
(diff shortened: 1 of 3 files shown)


adam-mateen (approved):

plugins/outputs/cloudwatchlogs/pusher.go
```diff
@@ -348,7 +348,7 @@ func (p *pusher) createLogGroupAndStream() error {
 		_, err = p.Service.CreateLogGroup(&cloudwatchlogs.CreateLogGroupInput{
 			LogGroupName: &p.Group,
 		})
-		p.agentInfo.RecordOpData(time.Since(opStartTime), p.bufferredSize, err)
+		p.agentInfo.RecordOpData(time.Since(opStartTime), 1, err)
```
adam-mateen, line 351: Do we include the operation name as a dimension?

Are there separate metrics for errors from CreateLogGroup vs CreateLogStream vs PutLogEvents?

bryce-carey: That's a great question. Actually I'm of the mind that these `RecordOpData` calls should be removed for all but PutMetricData and PutLogEvents for now. I realized that these will muddle the API metrics for PutLogEvents since we're sending information about the LAST request sent, which could have been CreateLogGroup or other non-PutLogEvent APIs. I'll make this change real quick.

cfg/agentinfo/info.go
```diff
@@ -61,13 +61,13 @@ type agentInfo struct {
...
-	FileDescriptorCount *int32         `json:"fd,omitempty"`
-	ThreadCount         *int32         `json:"th,omitempty"`
-	LatencyMillis       *time.Duration `json:"lat,omitempty"`
-	PayloadBytes        *int           `json:"load,omitempty"`
-	StatusCode          *int           `json:"code,omitempty"`
+	CpuPercent          *float64 `json:"cpu,omitempty"`
+	MemoryBytes         *uint64  `json:"mem,omitempty"`
+	FileDescriptorCount *int32   `json:"fd,omitempty"`
+	ThreadCount         *int32   `json:"th,omitempty"`
+	LatencyMillis       *int64   `json:"lat,omitempty"`
```
sethAmazon, line 68: Should we wrap int64 as a new struct called millis?

sethAmazon (approved):

PR #782, opened 2023-06-29
Workflow refactor and add onprem test jobs

ChenaLee:
# Description of the issue
Onprem tests have been added but not yet automatically run. 

# Description of changes
Refactored the existing workflow to share logic with EC2 test workflow since it should be almost 100% same spec-wise.

# License

# Tests
Tested with https://github.com/aws/amazon-cloudwatch-agent-test/pull/290

Need to merge after that merges. 

# Requirements
2. Run `make linter`

### DIFF
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -9,9 +9,9 @@ env:
   S3_INTEGRATION_BUCKET: ${{ secrets.S3_INTEGRATION_BUCKET }}
   KEY_NAME: ${{ secrets.KEY_NAME }}
   ECR_INTEGRATION_TEST_REPO: "cwagent-integration-test"
-  CWA_GITHUB_TEST_REPO_NAME: "aws/amazon-cloudwatch-agent-test"
-  CWA_GITHUB_TEST_REPO_URL: "https://github.com/aws/amazon-cloudwatch-agent-test.git"
-  CWA_GITHUB_TEST_REPO_BRANCH: "main"
+  CWA_GITHUB_TEST_REPO_NAME: "ChenaLee/amazon-cloudwatch-agent-test"
+  CWA_GITHUB_TEST_REPO_URL: "https://github.com/ChenaLee/amazon-cloudwatch-agent-test.git"
+  CWA_GITHUB_TEST_REPO_BRANCH: "onpremOneTest"
 
 on:
   push:
--- a/.github/workflows/integration-test.yml
+++ b/.github/workflows/integration-test.yml
@@ -9,9 +9,9 @@ env:
   S3_INTEGRATION_BUCKET: ${{ secrets.S3_INTEGRATION_BUCKET }}
   KEY_NAME: ${{ secrets.KEY_NAME }}
   ECR_INTEGRATION_TEST_REPO: "cwagent-integration-test"
-  CWA_GITHUB_TEST_REPO_NAME: "ChenaLee/amazon-cloudwatch-agent-test"
-  CWA_GITHUB_TEST_REPO_URL: "https://github.com/ChenaLee/amazon-cloudwatch-agent-test.git"
-  CWA_GITHUB_TEST_REPO_BRANCH: "onpremOneTest"
+  CWA_GITHUB_TEST_REPO_NAME: "aws/amazon-cloudwatch-agent-test"
+  CWA_GITHUB_TEST_REPO_URL: "https://github.com/aws/amazon-cloudwatch-agent-test.git"
+  CWA_GITHUB_TEST_REPO_BRANCH: "main"
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


.github/workflows/integration-test.yml
```diff
@@ -572,94 +572,64 @@ jobs:
...
+    name: 'OnpremLinux'
+    uses: ./.github/workflows/ec2-integration-test.yml
+    with:
+      github_sha: ${{github.sha}}
+      test_dir: terraform/ec2/linux_onprem
+      job_id: linux-onprem-integration-test
+      test_props: ${{needs.GenerateTestMatrix.outputs.ec2_linux_onprem_matrix}}
+      test_repo_name: ${{ needs.OutputEnvVariables.outputs.CWA_GITHUB_TEST_REPO_NAME }}
+      test_repo_url: ${{ needs.OutputEnvVariables.outputs.CWA_GITHUB_TEST_REPO_URL }}
+      test_repo_branch: ${{ needs.OutputEnvVariables.outputs.CWA_GITHUB_TEST_REPO_BRANCH }}
```
SaxyPandaBear, line 631: It would be nice to not need to add the extra step to set these, and instead to make these repository environment variables that we can override.. but idk if that's any better. https://docs.github.com/en/actions/learn-github-actions/variables#creating-configuration-variables-for-a-repository

.github/workflows/integration-test.yml
```diff
@@ -827,7 +797,7 @@ jobs:
     name: 'StopLocalStack'
     runs-on: ubuntu-latest
     if: ${{ always() }}
-    needs: [ StartLocalStack, EC2LinuxIntegrationTest ]
+    needs: [ StartLocalStack, EC2LinuxIntegrationTest, LinuxOnPremIntegrationTest ]
```
SaxyPandaBear, line 800: Does the on-prem test actually use LocalStack? This is specifically for the CA bundle tests.

SaxyPandaBear (approved):

.github/workflows/integration-test.yml
ChenaLee: will make this run all tests ec2 runs since I don't see claims anywhere they have different supported feature set.

.github/workflows/integration-test.yml
jefchien: Is there a reason we can't just use the EVs directly? Why do we need this extra step to set the outputs?

.github/workflows/integration-test.yml
```diff
@@ -572,94 +572,64 @@ jobs:
...
-        uses: aws-actions/configure-aws-credentials@v1
-        with:
-          role-to-assume: ${{ env.TERRAFORM_AWS_ASSUME_ROLE }}
-          aws-region: us-west-2
-          role-duration-seconds: ${{ env.TERRAFORM_AWS_ASSUME_ROLE_DURATION }}
-
-      - name: Cache if success
-        id: ec2-linux-integration-test
-        uses: actions/cache@v3
+      - name: Set up Go 1.x
```
jefchien, line 588: nit: Don't think we need to set up Go.

ChenaLee: why not?

.github/workflows/integration-test.yml
```diff
@@ -572,94 +572,64 @@ jobs:
...
-      - name: Echo Test Info
-        run: echo run on ec2 instance os ${{ matrix.arrays.os }} arc ${{ matrix.arrays.arc }} test dir ${{ matrix.arrays.test_dir }}
+          go-version: ~1.19.6
 
-      - name: Verify Terraform version
-        run: terraform --version
+      - name: SetOutputs
+        id: set-outputs
+        run: |
+          echo "::set-output name=CWA_GITHUB_TEST_REPO_NAME::${{ env.CWA_GITHUB_TEST_REPO_NAME }}"
```
sky333999, line 596: nit: set-output is deprecated and its preferred to use `>> $GITHUB_OUTPUT`
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

.github/workflows/integration-test.yml
sky333999: Where is go used in this GHA? I see it just outputting some vars.

.github/workflows/integration-test.yml
jefchien: Huh didn't know that. Looks like this is the only workaround. https://github.com/orgs/community/discussions/26671

jefchien (approved):

.github/workflows/integration-test.yml
jefchien: We're not using Go in this job.

ChenaLee: oh for the output var one! Didn't realize that was the job you are talking about. yeah gotcha

ChenaLee: Will come back to this in a follow up PR

PR #815, opened 2023-08-15
fix file glob matching

SaxyPandaBear:
# Description of the issue
The CloudWatch agent does not properly enforce glob patterns in its file configurations. In the scenario where a user supplies something like `[!abc]at`, the agent determines that this has a meta character `[`, but no super meta character `**`. Thus, when the agent tries to find matching files, it defaults to `filepath.Glob`, which does not interpret the `!` inversion operator for the glob pattern. 

# Description of changes
Updated the `Match` function so that it doesn't use `filepath.Glob` when there are nonstandard glob characters that are not operable with `filepath.Glob`, such as `!`. Updated the `Compile` function to change the short-circuit behavior, since there is technically a glob that needs to be compiled, even for the characters that can be interpreted in `filepath.Glob`. 

To illustrate this, I added a test for `/[something?`, which should be an invalid glob pattern because it does not close the `[` with a subsequent `]` character. The current behavior would treat this as a raw string to match against, when this _should_ be interpreted as a bad glob configuration. 

# License

# Tests
Added unit tests and ensured they failed before the change, and pass after. 

## Performance benchmarking
Screenshot 2023-08-15 at 5 55 00 PM

The blue line is with my changes, and the orange line is the baseline (what is currently released to Amazon Linux). 

Ran on t3.medium Amazon Linux 2023 instances. The log file configuration is as follows:
```json
    "logs": {
        "logs_collected": {
            "files": {
                "collect_list": [
                    {
                        "file_path": "/tmp/[a-g]oo",
                        "log_group_name": "something",
                        "log_stream_name": "somethingElse"
                    }
                ]
            }
        }
    }
```
Since I'm lazy, I wrote a script in python to write to a log file that rotates
```python
import logging
import time

handler = TimedRotatingFileHandler("/tmp/foo", when='M', interval=5, backupCount=5)
fmt = '%(asctime)s %(message)s'
handler.setFormatter(logging.Formatter(fmt))
logger = logging.getLogger('root')
logger.addHandler(handler)
logger.setLevel(logging.INFO)

while True:
    logger.info("hello there")
    time.sleep(1)
```

I also verified that the agent is publishing logs from both instances, though the glob `[a-g]oo` is valid even now. I did validate that using glob characters like `!` and `{}` work after my change, but not with the current release. 

# Requirements
2. Run `make lint`

### DIFF
--- a/plugins/inputs/logfile/globpath/globpath.go
+++ b/plugins/inputs/logfile/globpath/globpath.go
@@ -33,7 +33,7 @@ func Compile(path string) (*GlobPath, error) {
 
 	// if there are no glob meta characters in the path, don't bother compiling
 	// a glob object or finding the root directory. (see short-circuit in Match)
-	if !out.hasMeta || !out.hasSuperMeta {
+	if !out.hasMeta && !out.hasSuperMeta {
 		return &out, nil
 	}
 
@@ -59,19 +59,6 @@ func (g *GlobPath) Match() map[string]os.FileInfo {
 			out[g.path] = info
 		} else {
 			log.Printf("D! Stat file %v failed due to %v", g.path, err)
-		}
-		return out
-	}
-	if !g.hasSuperMeta {
-		out := make(map[string]os.FileInfo)
-		files, _ := filepath.Glob(g.path)
-		for _, file := range files {
-			info, err := os.Stat(file)
-			if info != nil {
-				out[file] = info
-			} else {
-				log.Printf("D! Stat file %v failed due to %v", g.path, err)
-			}
 		}
 		return out
 	}
--- a/plugins/inputs/logfile/globpath/globpath.go
+++ b/plugins/inputs/logfile/globpath/globpath.go
@@ -52,7 +52,7 @@ func Compile(path string) (*GlobPath, error) {
 }
 
 func (g *GlobPath) Match() map[string]os.FileInfo {
-	if !g.hasMeta {
+	if !g.hasMeta && !g.hasSuperMeta {
 		out := make(map[string]os.FileInfo)
 		info, err := os.Stat(g.path)
 		if info != nil {
@@ -61,8 +61,7 @@ func (g *GlobPath) Match() map[string]os.FileInfo {
 			log.Printf("D! Stat file %v failed due to %v", g.path, err)
 		}
 		return out
-	}
-	if !g.hasSuperMeta {
+	} else if !g.hasSuperMeta {
 		out := make(map[string]os.FileInfo)
 		files, _ := filepath.Glob(g.path)
 		for _, file := range files {
--- a/plugins/inputs/logfile/globpath/globpath.go
+++ b/plugins/inputs/logfile/globpath/globpath.go
@@ -62,6 +62,19 @@ func (g *GlobPath) Match() map[string]os.FileInfo {
 		}
 		return out
 	}
+	if !g.hasSuperMeta {
+		out := make(map[string]os.FileInfo)
+		files, _ := filepath.Glob(g.path)
+		for _, file := range files {
+			info, err := os.Stat(file)
+			if info != nil {
+				out[file] = info
+			} else {
+				log.Printf("D! Stat file %v failed due to %v", g.path, err)
+			}
+		}
+		return out
+	}
 	return walkFilePath(g.root, g.g)
 }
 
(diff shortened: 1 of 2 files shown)


plugins/inputs/logfile/globpath/globpath.go
```diff
@@ -62,19 +62,6 @@ func (g *GlobPath) Match() map[string]os.FileInfo {
...
-		for _, file := range files {
-			info, err := os.Stat(file)
-			if info != nil {
-				out[file] = info
-			} else {
-				log.Printf("D! Stat file %v failed due to %v", g.path, err)
-			}
-		}
-		return out
-	}
```
adam-mateen, line 77: Why is it safe to remove this code block?

Does is duplicate functionality that is also in `walkFilePath()`?

SaxyPandaBear: Because `filepath.Glob` doesn't behave how we want it to, and because the `walkFilePath` function does the same thing, but with the `glob.Glob` pattern

SaxyPandaBear: Basically, the `TestInversionGlobPattern` test that I added is the important test that illustrates the need for this code change. If I put this block of code back in, the agent will call `filepath.Glob`, which will match `log2.log`, when what we really want is `log1.log`.

jefchien: If we're removing this block, then what's the point of the `hasSuperMeta` flag? We were using the `filepath.Glob` by default because it works for most situations and is more efficient.

SaxyPandaBear: I can rip out the `hasSuperMeta` flag. I was just trying to be minimally disruptive. 

I guess now we come to a product decision. Which should we use? `filepath.Glob` is more efficient, but there is going to be confusion on why the agent doesn't match log files as customer expect with some glob patterns. 

As far as performance implications, I'll run something simple like `/tmp/*.log` and compare CPU and mem before and after the change since it would be comparing `filepath.Glob` and `Glob.Match`

jefchien: `hasSuperMeta` is basically the flag we use to tell the globpath to use Glob.Match. Right now it's only true if the path contains `**`. We can change it so it's also true if we need to handle unsupported glob characters.

SaxyPandaBear: Screenshot 2023-08-15 at 4 22 31 PM

This is me running through the repro case, which isn't super useful from a performance standpoint, but interesting to see the weird difference. 

Making some adjustments then I'll run through a simple pattern (not the repro case) to compare performance

plugins/inputs/logfile/globpath/globpath.go
```diff
@@ -33,7 +33,7 @@ func Compile(path string) (*GlobPath, error) {
 
 	// if there are no glob meta characters in the path, don't bother compiling
 	// a glob object or finding the root directory. (see short-circuit in Match)
-	if !out.hasMeta || !out.hasSuperMeta {
+	if !out.hasMeta && !out.hasSuperMeta {
```
jefchien, line 36: nit: I believe his should be an OR. We only need to compile the glob in the `hasSuperMeta` case. With an AND, we would be compiling the glob and not using it.

adam-mateen (approved):

plugins/inputs/logfile/globpath/globpath.go
SaxyPandaBear: Also if you look at the diff for https://github.com/aws/amazon-cloudwatch-agent/pull/815/commits/d81957b82fef78ccd2686e2b542110b3afb4c599, I think this also illustrates it. If I flip it to an OR condition here, then the valid glob `/nested1/nested2/{nested, foo}.txt` does not return the expected result. It matches zero elements when it should match `nested1/nested2/nested.txt`

jefchien: You're right. I was thinking we were implementing the `hasMeta` differently. In a different implementation, we used
```go
// isMeta determines whether the rune is a glob character.
func isMeta(char rune) bool {
	return syntax.Special(byte(char))
}
```
from "github.com/gobwas/glob/syntax"

jefchien (approved):

PR #822, opened 2023-08-23
Container insights ssl

bhanuba:
# Description of the issue
In ccwa emf uses an otel exporter instead of a cwa exporter. We need to test this still works with ca bundle or not. 

# Description of changes
Here we are testing the container insights to check whether it works with the ca_bundle or not. 
- Automated container insights test using ca bundle.
- Fix the ca bundle issue with CCWA in container insights.

# License

# Tests
Unit tests passed 
# Requirements
2. Run `make lint`


translator/tocwconfig/tocwconfig_test.go
```diff
@@ -50,7 +51,13 @@ func TestBaseContainerInsightsConfig(t *testing.T) {
 	context.CurrentContext().SetRunInContainer(true)
 	t.Setenv(config.HOST_NAME, "host_name_from_env")
 	t.Setenv(config.HOST_IP, "127.0.0.1")
-	expectedEnvVars := map[string]string{}
+	t.Setenv(envconfig.AWS_CA_BUNDLE, "sampleConfig/public_amazon_cert.pem")
+	defer func() {
```
sky333999, line 55: Can we just call `os.Clearenv()` in `resetContext()` so we cleanup all env vars in a common place. Not sure if any tests will break due to this but worth a try.

go.mod
```diff
@@ -153,6 +153,8 @@ require (
 	k8s.io/klog/v2 v2.90.1
 )
 
+require github.com/open-telemetry/opentelemetry-collector-contrib/internal/aws v0.24.0
```
sky333999, line 156: Dont think this needs a separate require block

sky333999: Where is this being used? I dont see any new imports from this in this PR.

translator/translate/otel/exporter/awsemf/translator.go
```diff
@@ -78,6 +81,11 @@ func (t *translator) Translate(c *confmap.Conf) (component.Config, error) {
 		}
 	}
 	cfg.AWSSessionSettings.Region = agent.Global_Config.Region
+	if c.IsSet(endpointOverrideKey) {
+		cfg.AWSSessionSettings.Endpoint, _ = common.GetString(c, endpointOverrideKey)
+	}
+	cfg.AWSSessionSettings.CertificateFilePath = os.Getenv(envconfig.AWS_CA_BUNDLE)
```
sky333999, line 87: Do we always want to set this even if the env var is not set?

sky333999:
Mind updating the title and the description of the PR to be provide some more details. `sll` seems like a typo. And Im not very sure what `Fixed Ca_Bundle issue` is.

Also, how did we verify that this fix works?

translator/tocwconfig/tocwconfig_test.go
SaxyPandaBear: `t.Setenv()` should be unique to the test executed, which I think is better than trying to set with `os.Setenv` and clearing that.

translator/translate/otel/exporter/awsemf/translator.go
SaxyPandaBear: I think it looks messy, but is effectively the same. `Getenv` returns an empty string, and the default value for the CertificateFilePath would be the empty string already.

translator/tocwconfig/sampleConfig/emf_and_kubernetes_config.yaml
```diff
@@ -34,7 +34,7 @@ exporters:
     disable_metric_extraction: true
     eks_fargate_container_insights_enabled: false
     certificate_file_path: ""
-    endpoint: ""
+    endpoint: "https://fake_endpoint"
```
SaxyPandaBear, line 37: Just to confirm, were these supposed to be overridden before, but we weren't doing translation correctly in the EMF exporter for it to be reflected? Or is this new behavior?

bhanuba: Now we are using the yaml translator. Previously we used toml translator. I think thats the reason it might be the erroring out as mismatch.

go.mod
bhanuba: removed form the code.

translator/tocwconfig/tocwconfig_test.go
```diff
@@ -50,7 +51,13 @@ func TestBaseContainerInsightsConfig(t *testing.T) {
 	context.CurrentContext().SetRunInContainer(true)
 	t.Setenv(config.HOST_NAME, "host_name_from_env")
 	t.Setenv(config.HOST_IP, "127.0.0.1")
-	expectedEnvVars := map[string]string{}
+	t.Setenv(envconfig.AWS_CA_BUNDLE, "/etc/test/ca_bundle.pem")
+	defer func() {
+		t.Setenv(envconfig.AWS_CA_BUNDLE, "")
+	}()
```
SaxyPandaBear, line 57: Why do we need this block? As far as I know, setting environment variables with `t.Setenv` keeps it local to the individual unit test.

bhanuba: We can remove the differ function if it keep local for unit test. Removed!!

SaxyPandaBear:
Firstly, can you please update the PR title? It doesn't properly convey what this code change is for.

The PR looks good but I think it's lacking some end-to-end testing. Unit testing isn't enough here. Can you build the agent with this code change and validate it yourself?

SaxyPandaBear:
There are merge conflicts that you should resolve in your PR before it can be merged

SaxyPandaBear (commented):
Please add some evidence of manual testing to verify that this change works

sethAmazon (approved):

sethAmazon:
We ran integration tests that are failing in some os and passing in others. The failures are for reasons apart from SSL connection.

amazon-cloudwatch-agent
SaxyPandaBear, line 1: We should not have this. This looks like a Git submodule. Please remove it

SaxyPandaBear (changes requested):

amazon-cloudwatch-agent
bhanuba: removed

SaxyPandaBear (approved):

sethAmazon (approved):

PR #825, opened 2023-08-24
deprecate metric_granularity in favor of enhanced_container_insights …

chadpatel:
# Description of the issue
There are no longer tiers for enhanced container insights, it is all or nothing

# Description of changes
I didn't remove metric_granularity but instead I updated the translator to first check a new flag, enhanced_container_insights.  If that is not set to true, it will check metric_granularity.  If metric_granularity is >1 then it will enabled enhanced container insights.  So it kind of bridges us by adding a new flag and also maintaining backwards compatibility for our beta testers

# License

# Tests
My cluster is set to metric_granularity 2, now I have FullPodName metrics

Screen Shot 2023-08-24 at 1 37 21 PM

# Requirements
2. Run `make lint`


SaxyPandaBear (approved):
WRT backwards-compatibility, it would be nice to have a JSON translation test that still validates that setting `metric_granularity` works as expected.. unless you didn't update all of the CI test JSONs and there is one that still uses the old config.

sethAmazon (approved):

sethAmazon:
I'm assuming this is only for the enhanced feature and not the current container insihgts

chadpatel:
> I'm assuming this is only for the enhanced feature and not the current container insihgts

correct, this enables enhanced container insights.  If one of the two flags is not set, customers will get regular container insights

chadpatel:
> WRT backwards-compatibility, it would be nice to have a JSON translation test that still validates that setting `metric_granularity` works as expected.. unless you didn't update all of the CI test JSONs and there is one that still uses the old config.

That is covered by WithLevel1Granularity/WithLevel2Granularity/WithLevel3Granularity

I decided to not put it in the end-to-end json translation tests so we can minimize people propagating this deprecated flag.

PR #838, opened 2023-08-29
Get Stats For IMDSV2 Fallback Succeed

sethAmazon:
# Description of the issue

# Description of changes

# License

# Tests

# Requirements
2. Run `make lint`

### DIFF
--- a/handlers/agentinfo/info.go
+++ b/handlers/agentinfo/info.go
@@ -49,6 +49,7 @@ var (
 	fullVersion             = getFullVersion(version)
 	id                      = uuid.NewString()
 	sharedConfigFallback    atomic.Bool
+	imdsFallbackSucceed     atomic.Bool
 )
 
 var isRunningAsRoot = defaultIsRunningAsRoot
@@ -75,6 +76,7 @@ type agentStats struct {
 	PayloadBytes         *int     `json:"load,omitempty"`
 	StatusCode           *int     `json:"code,omitempty"`
 	SharedConfigFallback *int     `json:"scfb,omitempty"`
+	ImdsFallbackSucceed  *int     `json:"ifs,omitempty"`
 }
 
 func New(groupName string) AgentInfo {
@@ -123,6 +125,7 @@ func (ai *agentInfo) RecordOpData(latency time.Duration, payloadBytes int, err e
 		stats.FileDescriptorCount = ai.fileDescriptorCount()
 		stats.ThreadCount = ai.threadCount()
 		stats.SharedConfigFallback = getSharedConfigFallback()
+		stats.ImdsFallbackSucceed = succeedImdsFallback()
 		ai.nextUpdate = now.Add(updateInterval)
 	}
 
@@ -150,6 +153,15 @@ func (ai *agentInfo) memoryBytes() *uint64 {
 func (ai *agentInfo) fileDescriptorCount() *int32 {
 	if fdCount, err := ai.proc.NumFDs(); err == nil {
 		return aws.Int32(fdCount)
+	}
+	return nil
+}
+
+// we only need to know if value is 1
+// thus return nil if not set
+func succeedImdsFallback() *int {
+	if imdsFallbackSucceed.Load() {
+		return aws.Int(1)
 	}
 	return nil
 }
@@ -286,6 +298,8 @@ func readVersionFile() string {
 	return strings.Trim(string(byteArray), " \n\r\t")
 }
 
+// this returns true for true or invalid
+// examples of invalid are not set env var, "", "invalid"
 func getUsageDataEnabled() bool {
 	ok, err := strconv.ParseBool(os.Getenv(envconfig.CWAGENT_USAGE_DATA))
 	return ok || err != nil
@@ -311,4 +325,8 @@ func getSharedConfigFallback() *int {
 		return aws.Int(1)
 	}
 	return nil
+}
+
+func SetImdsFallbackSucceed() {
+	imdsFallbackSucceed.Store(true)
 }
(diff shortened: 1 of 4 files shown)


handlers/agentinfo/info.go
```diff
@@ -282,6 +293,8 @@ func readVersionFile() string {
 	return strings.Trim(string(byteArray), " \n\r\t")
 }
 
+// this returns true for true or invalid
+// examples of invalid are not set env var, "", "invalid"
 func getUsageDataEnabled() bool {
 	ok, err := strconv.ParseBool(os.Getenv(envconfig.CWAGENT_USAGE_DATA))
 	return ok || err != nil
```
sethAmazon, line 300: !ok && err == nil

maybe this is a little cleaner?

handlers/agentinfo/info.go
```diff
@@ -122,6 +125,8 @@ func (ai *agentInfo) RecordOpData(latency time.Duration, payloadBytes int, err e
 		ai.nextUpdate = now.Add(updateInterval)
 	}
 
+	stats.ImdsFallbackSucceed = ai.imdsFallbackSucceed()
```
bryce-carey, line 128: The metadata provider `InstanceID` and `Hostname` functions are not called anywhere outside of unit tests. The provider `Get` function is only called once upon the processor `Start` function, which is called before any `RecordOpData` call. Could we use a `sync.Once` here to only set this field only one time?

sethAmazon: Do you mean as apart from the enabled once code you have when creating a new agent info? That happens in the start func of cw. I think that will have a  thread timing issue.

sethAmazon: Talked with @bryce-carey on this over slack. Changed to only publish this metric 1 time per minute. No need for single sync.

SaxyPandaBear (commented):
Didn't realize I had old comments here. publishing them now

handlers/agentinfo/info.go
```diff
@@ -297,3 +310,9 @@ func isUsageDataEnabled() bool {
 func defaultIsRunningAsRoot() bool {
 	return os.Getuid() == 0
 }
+
+func SetImdsFallbackSucceed() {
```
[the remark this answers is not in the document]
sethAmazon: That is correct, yes. As long as the agent continues to run, this value will be true. Spoke with @bryce-carey about this.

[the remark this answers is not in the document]
sethAmazon: This is because we would rather not handle how to reset this value during runtime.

.github/workflows/PR-build.yml
```diff
@@ -54,6 +54,13 @@ jobs:
         if: needs.changes.outputs.lint == 'true'
         uses: actions/checkout@v3
 
+      - name: Cache if success
+        id: cache_if_success
+        uses: actions/cache@v3
+        with:
+          path: go.mod
+          key: v1-go-pkg-mod-${{ hashFiles('**/go.sum') }}
```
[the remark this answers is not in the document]
sethAmazon: This caches to build. So as long as the go.sum does not change it will keep the downloaded mods. I looked into the set up go and it does this by default. This can be removed.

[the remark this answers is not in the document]
SaxyPandaBear: This is running cache on the lint step though so I don't think I get it

handlers/agentinfo/info.go
```diff
@@ -47,6 +47,8 @@ var (
 	version                 = readVersionFile()
 	fullVersion             = getFullVersion(version)
 	id                      = uuid.NewString()
+	imdsFallbackSucceed     bool
+	imdsFallbackSucceedLock sync.Mutex
```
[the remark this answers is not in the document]
sethAmazon: Changed

.github/workflows/PR-build.yml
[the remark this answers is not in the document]
sethAmazon: Removed

handlers/agentinfo/info.go
```diff
@@ -47,6 +48,7 @@ var (
 	version                 = readVersionFile()
 	fullVersion             = getFullVersion(version)
 	id                      = uuid.NewString()
+	imdsFallbackSucceed     atomic.Bool
```
SaxyPandaBear, line 51: Just to be sure, do we know that this defaults with a `false` value loaded? What is the zero value for atomic.Bool? Is it nil?

sethAmazon: It is false. 
```
func TestSetImdsFallbackSucceed(t *testing.T) {
	defer imdsFallbackSucceed.Store(false)
	assert.False(t, imdsFallbackSucceed.Load())
	SetImdsFallbackSucceed()
	assert.True(t, imdsFallbackSucceed.Load())
}
```

sethAmazon: https://pkg.go.dev/sync/atomic

```
Types [¶](https://pkg.go.dev/sync/atomic#pkg-types)
type [Bool](https://cs.opensource.google/go/go/+/go1.21.0:src/sync/atomic/type.go;l=11) [¶](https://pkg.go.dev/sync/atomic#Bool)
added in go1.19
type Bool struct {
	// contains filtered or unexported fields
}
A Bool is an atomic boolean value. The zero value is false.
```

sethAmazon: I had the same assumption that it would be nil until I looked into the docs and added a unit test for it.

handlers/agentinfo/info.go
```diff
@@ -72,6 +74,7 @@ type agentStats struct {
 	LatencyMillis       *int64   `json:"lat,omitempty"`
 	PayloadBytes        *int     `json:"load,omitempty"`
 	StatusCode          *int     `json:"code,omitempty"`
+	ImdsFallbackSucceed *bool    `json:"ifs,omitempty"`
```
jefchien, line 77: Should this be an int instead? It'll need to be converted into an int at some point.

sethAmazon: Changed json output to int

jefchien (approved):

jefchien (approved):

okankoAMZ (approved):

PR #839, opened 2023-08-30
Persistent queue

yifanFengg:
# Description of the issue
The persistentQueue can writes logEvents in memory to the disk when the retryErrors happens during the process that Pusher publish logEvent to destination, and read logEvent from disk. 

# Description of changes
When a customer generates a large number of log files, the Agent reads lines from these files, parses them into logEvents, and then publishes them to the destination. However, if retryErrors persistently occur during the publish process, the pusher can be blocked for up to 14 days, During this period, the pusher will continue reading lines from log files and parsing them into logEvent, this gone cause memory usage to steadily increase. As a result, the agent will retains a significant amount of memory in the system.

# License

# Tests

•Current CloudWatch-Agent version : 1.300025.0   
•Prototype which agent implemented with persistentQueue
•EC2 instance type :  t3.2xlarge (32G ram, 512G ssd)
•100 log file with different log stream name
•Each log file has 100 log lines
•Each line of log file has 250,000 characters, so size of one log line would be 250KB
•MockServer : receive publish request from agent, return aws error 

**Below is graph that compare current agent with prototype**
image

**memory usage of EC2 instance on which prototype runs**  
image
**memory usage of EC2 instance on which agent runs**
image

**Below image list data file that contain compressed logEvent that persisted by prototype** 
Screenshot 2023-08-30 at 4 09 55 PM

# Requirements
2. Run `make lint`

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -155,7 +155,6 @@ require (
 	github.com/nsqio/go-diskqueue v1.1.0
 )
 
-
 require (
 	cloud.google.com/go/compute v1.19.0 // indirect
 	cloud.google.com/go/compute/metadata v0.2.3 // indirect
--- a/translator/cmdutil/translatorutil.go
+++ b/translator/cmdutil/translatorutil.go
@@ -208,12 +208,10 @@ func TranslateJsonMapToTomlConfig(jsonConfigValue interface{}) (interface{}, err
 	if !translator.IsTranslateSuccess() {
 		return nil, fmt.Errorf("%v", translator.ErrorMessages)
 	}
-
 	// Translation is valid, log info messages and continue to convert/write to toml
 	for _, infoMessage := range translator.InfoMessages {
 		log.Println(infoMessage)
 	}
-
 	return val, nil
 }
 
--- a/packaging/linux/amazon-cloudwatch-agent.spec
+++ b/packaging/linux/amazon-cloudwatch-agent.spec
@@ -45,7 +45,6 @@ ln -f -s /opt/aws/amazon-cloudwatch-agent/var ${RPM_BUILD_ROOT}/var/run/amazon/a
 %dir /opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.d
 %dir /opt/aws/amazon-cloudwatch-agent/logs
 %dir /opt/aws/amazon-cloudwatch-agent/var
-
 /opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent
 /opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent-ctl
 /opt/aws/amazon-cloudwatch-agent/bin/CWAGENT_VERSION
--- a/translator/translate/translate.go
+++ b/translator/translate/translate.go
@@ -66,7 +66,6 @@ func (t *Translator) ApplyRule(input interface{}) (returnKey string, returnVal i
 
 	// sort rule here so that we could get the output plugin instance in a stable order
 	sortedRuleKey := make([]string, 0, len(targetRuleMap))
-
 	for k := range targetRuleMap {
 		sortedRuleKey = append(sortedRuleKey, k)
 	}


plugins/outputs/cloudwatchlogs/pusher.go
```diff
@@ -55,29 +68,56 @@ type pusher struct {
...
+
 	initNonBlockingChOnce sync.Once
 	startNonBlockCh       chan struct{}
 	wg                    *sync.WaitGroup
 	agentInfo             agentinfo.AgentInfo
 }
 
-func NewPusher(target Target, service CloudWatchLogsService, flushTimeout time.Duration, retryDuration time.Duration, logger telegraf.Logger, stop <-chan struct{}, wg *sync.WaitGroup, agentInfo agentinfo.AgentInfo) *pusher {
+func NewPusher(target Target, service CloudWatchLogsService, flushTimeout time.Duration, retryDuration time.Duration, logger telegraf.Logger, stop <-chan struct{}, wg *sync.WaitGroup, agentInfo agentinfo.AgentInfo, flagPersistentQueue bool) *pusher {
+	marshaler := func(i interface{}) ([]byte, error) { return json.Marshal(i) }
```
SaxyPandaBear, line 83: I don't think we need to wrap this in another function. I think that you could just refer directly to `json.Marshal` for the queue's marshaller function

plugins/outputs/cloudwatchlogs/pusher.go
```diff
@@ -55,29 +68,56 @@ type pusher struct {
...
+		Target:        target,
+		Service:       service,
+		FlushTimeout:  flushTimeout,
+		RetryDuration: retryDuration,
+		Log:           logger,
+		events:        make([]*cloudwatchlogs.InputLogEvent, 0, 10),
+		eventsCh:      make(chan logs.LogEvent, 100),
+		flushTimer:    time.NewTimer(flushTimeout),
+		stop:          stop,
+		backlogQueue: diskqueue.NewPersistentQueue(
```
SaxyPandaBear, line 100: So we always create a persistent queue, even if the feature flag is off?

plugins/outputs/cloudwatchlogs/pusher.go
```diff
@@ -147,56 +187,142 @@ func (p *pusher) start() {
...
 
-	for {
-		select {
-		case e := <-ec:
-			// Start timer when first event of the batch is added (happens after a flush timer timeout)
-			if len(p.events) == 0 {
-				p.resetFlushTimer()
-			}
+	if p.flagPersistentQueue {
+		for {
```
SaxyPandaBear, line 191: IMO these new `for` blocks should be pulled out into separate functions to simplify the `start` function code

plugins/outputs/cloudwatchlogs/pusher.go
```diff
@@ -147,56 +187,142 @@ func (p *pusher) start() {
...
+			select {
+			case <-p.ticker.C:
+				//p.Log.Debugf("start dequeue from disk")
```
SaxyPandaBear, line 194: If we don't need these debug logs, we can just remove them from the code. No need to comment them out

plugins/outputs/cloudwatchlogs/pusher.go
```diff
@@ -212,6 +338,92 @@ func (p *pusher) reset() {
...
+			if info.ExpiredLogEventEndIndex != nil {
+				p.Log.Warnf("%d log events for log '%s/%s' are expired", *info.ExpiredLogEventEndIndex, p.Group, p.Stream)
+			}
+		}
+		p.Log.Debugf("in the persistentQueue, Pusher published %v log events to group: %v stream: %v with size %v KB in %v.", len(input.LogEvents), *input.LogGroupName, *input.LogStreamName, temp.BufferredSize/1024, time.Since(startTime))
+		p.addStats("rawSize", float64(temp.BufferredSize))
+		p.lastSentTime = time.Now()
+		return
+	}
+	p.Log.Debugf("errors not nil, we need tell exactly what's error")
```
SaxyPandaBear, line 381: I think this log is unnecessary

tool/gzip.go
```diff
@@ -0,0 +1,39 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+package tool
```
SaxyPandaBear, line 4: Why are we putting the persistent queue code in the `tool` package?

tool/persistentqueue/diskqueue/diskqueue.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/aws/amazon-cloudwatch-agent/tool"
+	"github.com/aws/amazon-cloudwatch-agent/tool/persistentqueue"
+)
+
+type diskQueue struct {
+	marshal   persistentqueue.Marshaler
+	unmarshal persistentqueue.Unmarshaler
+	queue     diskqueue.Interface
+	size      int64
+	logger    telegraf.Logger
```
SaxyPandaBear, line 22: nit - I think we should transition from the telegraf logger to zap

translator/cmdutil/translatorutil.go
```diff
@@ -93,7 +93,7 @@ func checkSchema(inputJsonMap map[string]interface{}) {
 		log.Panicf("E! Failed to run schema validation because of %v", err)
 	}
 	if result.Valid() {
-		log.Print("I! Valid Json input schema.")
+		log.Print("I! tool.")
```
SaxyPandaBear, line 96: What is this change for?

translator/tocwconfig/tocwconfig_unix_test.go
```diff
@@ -15,11 +17,13 @@ func TestCompleteConfigUnix(t *testing.T) {
...
-
-	// The translation needs to use the runtime.GOOS value in order to generate the proper configuration YAML,
-	// so this is separate
+	//The translation needs to use the runtime.GOOS value in order to generate the proper configuration YAML,
+	//so this is separate
 	checkTranslation(t, "complete_linux_config", "linux", expectedEnvVars, "")
 	checkTranslation(t, "complete_darwin_config", "darwin", nil, "")
+
+	//checkTranslation(t, "complete_linux_config", "linux", nil, "")
+
```
SaxyPandaBear, line 26: What is this change for?

translator/tocwconfig/totomlconfig/tomlConfigTemplate/tomlConfig.go
```diff
@@ -241,6 +241,7 @@ type (
 	// Output plugins
 
 	cloudWatchOutputConfig struct {
+		PersistentQueue     bool   `toml:"persistentQueue"`
```
SaxyPandaBear, line 244: I thought this change was for logs. Why are we including this in the CloudWatch metrics config too?

.github/workflows/integration-test.yml
```diff
@@ -54,6 +54,7 @@ jobs:
   GenerateTestMatrix:
     name: 'GenerateTestMatrix'
     runs-on: ubuntu-latest
+    if: false
```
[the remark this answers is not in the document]
yifanFengg: no, I wanted to run integration test before, so i added it, but now it's useless, so I will delete it now.
good morning panda, thanks for your review, appreciate it.

go.mod
```diff
@@ -154,6 +154,8 @@ require (
 	k8s.io/klog/v2 v2.90.1
 )
 
+require github.com/nsqio/go-diskqueue v1.1.0
```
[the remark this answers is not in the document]
yifanFengg: yeah, I will change it now.

plugins/outputs/cloudwatchlogs/pusher.go
yifanFengg: yeah, you are right, panda, I don't need to wrap this in another function. I will change it right now.

tool/gzip.go
yifanFengg: yeah , I will change it to util folder, do you think this is properly?

tool/persistentqueue/diskqueue/diskqueue.go
yifanFengg: sure, I will change it right now

translator/tocwconfig/totomlconfig/tomlConfigTemplate/tomlConfig.go
yifanFengg: My bad, panda, I will change it right now.

plugins/outputs/cloudwatchlogs/pusher.go
yifanFengg: but if i refer directly to`json.Marshal` without wrap in `marshaler` , the `ummarshaler` function left there will  seems like not paired , right ? how do you think, panda, do I still need to change it? I am not sure now.

tool/persistentqueue/diskqueue/diskqueue.go
yifanFengg: but this logger parameter is come from `func NewPusher(target Target, service CloudWatchLogsService, flushTimeout time.Duration, retryDuration time.Duration, logger telegraf.Logger, stop <-chan struct{}, wg *sync.WaitGroup, agentInfo agentinfo.AgentInfo)`, and the `telegraf.Logger ` cannot convert to `zap.Logger`,
so i have to use `telegraf.Logger`, right ? or I can just not pass this `telegraf.Logger` parameter to diskQueue and directly create a logger using zap.Logger?

how do you think

yifanFengg: hi, andrew, how do you think , do I need to directly create a new `zap.Logger` in `diskQueue` without pass it as parameter from `NewPusher`

plugins/outputs/cloudwatchlogs/pusher.go
SaxyPandaBear: Not sure what you mean? I don't think they need to be mirrored. If the marshaller function can be simplified to just call `json.Marshal` directly, we should do that. The reason the other function needs to be wrapped is because the function doesn't exactly match what we want.

tool/persistentqueue/diskqueue/diskqueue.go
SaxyPandaBear: Ok if there's no zap logger to pass in, using the telegraf one is fine I guess

PR #842, opened 2023-09-06
Add shared config fallback stat to agent info.

jefchien:
# Description of the issue
We want insight into the potential impact that removing the fallback override logic would have.

# Description of changes
Adds a shared config fallback stat to the agent info. Similar to https://github.com/aws/amazon-cloudwatch-agent/pull/838.

# License

# Tests
Added unit tests.

# Requirements
2. Run `make lint`

### DIFF
--- a/handlers/agentinfo/info.go
+++ b/handlers/agentinfo/info.go
@@ -16,6 +16,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -47,6 +48,7 @@ var (
 	version                 = readVersionFile()
 	fullVersion             = getFullVersion(version)
 	id                      = uuid.NewString()
+	sharedConfigFallback    atomic.Bool
 )
 
 var isRunningAsRoot = defaultIsRunningAsRoot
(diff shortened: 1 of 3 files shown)


sethAmazon (approved):

handlers/agentinfo/info_test.go
```diff
@@ -233,3 +236,13 @@ func TestIsUsageDataEnabled(t *testing.T) {
...
 }
+
+func TestSharedConfigFallback(t *testing.T) {
+	defer sharedConfigFallback.Store(false)
+	assert.Nil(t, getSharedConfigFallback())
+	RecordSharedConfigFallback()
+	assert.Equal(t, 1, *(getSharedConfigFallback()))
+	RecordSharedConfigFallback()
+	RecordSharedConfigFallback()
+	assert.Equal(t, 1, *(getSharedConfigFallback()))
```
zhihonl, line 247: I'm not sure how we are trying to record the fallback counts. Is this supposed to be 2?

handlers/agentinfo/info.go
```diff
@@ -65,13 +67,14 @@ type agentInfo struct {
...
-	PayloadBytes        *int     `json:"load,omitempty"`
-	StatusCode          *int     `json:"code,omitempty"`
+	CpuPercent           *float64 `json:"cpu,omitempty"`
+	MemoryBytes          *uint64  `json:"mem,omitempty"`
+	FileDescriptorCount  *int32   `json:"fd,omitempty"`
+	ThreadCount          *int32   `json:"th,omitempty"`
+	LatencyMillis        *int64   `json:"lat,omitempty"`
+	PayloadBytes         *int     `json:"load,omitempty"`
+	StatusCode           *int     `json:"code,omitempty"`
+	SharedConfigFallback *int     `json:"scfb,omitempty"`
```
zhihonl, line 77: nit: `scfb` isn't intuitive to understand. Maybe something like `cred_fallback`? 

Don't feel too strong on this since it looks like only the developers will use this.

handlers/agentinfo/info_test.go
jefchien: It's a boolean, so no matter how many times RecordSharedConfigFallback is called, it'll always be 1. We convert it into an int because it's intended to be a metric.

handlers/agentinfo/info.go
jefchien: We have a limit on the header size, so we're trying to keep the names brief. It's an acronym (kinda).

zhihonl (approved):

PR #847, opened 2023-09-08
Drop unsupported NaN, Inf, and out of range values.

jefchien:
# Description of the issue
The CloudWatch Agent doesn't currently account for NaN or Inf values. When we have aggregation enabled, we will add these metrics to the distribution. When the regular distribution is [converted](https://github.com/aws/amazon-cloudwatch-agent/blob/main/plugins/outputs/cloudwatch/util.go#L87) into an SEH one, we're unable to retrieve the counts for the NaN value due to the way GoLang handles NaN lookups. This leads to an empty distribution list and the CloudWatch Agent creates the individual metric datum instead of the aggregate. CloudWatch does not support NaN values.

# Description of changes
- Drop NaN/Inf/Out of range values during OTEL conversion.
- Rejects NaN/Inf/Out of range values when they reach the distributions.
- Drop individual MetricDatums with NaN/Inf/Out of range values.

Matches https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_MetricDatum.html values restrictions.

# License

# Tests
Added unit tests.

# Requirements
2. Run `make lint`

### DIFF
--- a/internal/util/type_conversion.go
+++ b/internal/util/type_conversion.go
@@ -35,7 +35,7 @@ func ToOtelValue(value interface{}) interface{} {
 		return float64(v)
 	case float64:
 		if math.IsNaN(v) || math.IsInf(v, 0) {
-			return float64(0)
+			return nil
 		}
 		return v
 	case bool:
--- a/internal/util/type_conversion.go
+++ b/internal/util/type_conversion.go
@@ -3,7 +3,11 @@
 
 package util
 
-import "github.com/aws/amazon-cloudwatch-agent/metric/distribution"
+import (
+	"math"
+
+	"github.com/aws/amazon-cloudwatch-agent/metric/distribution"
+)
 
 func ToOtelValue(value interface{}) interface{} {
 	switch v := value.(type) {
@@ -30,6 +34,9 @@ func ToOtelValue(value interface{}) interface{} {
 	case float32:
 		return float64(v)
 	case float64:
+		if math.IsNaN(v) || math.IsInf(v, 0) {
+			return float64(0)
+		}
 		return v
 	case bool:
 		if v {
--- a/internal/util/type_conversion.go
+++ b/internal/util/type_conversion.go
@@ -4,49 +4,50 @@
 package util
 
 import (
+	"fmt"
 	"math"
 
 	"github.com/aws/amazon-cloudwatch-agent/metric/distribution"
 )
 
-func ToOtelValue(value interface{}) interface{} {
+func ToOtelValue(value interface{}) (interface{}, error) {
 	switch v := value.(type) {
 	case int:
-		return int64(v)
+		return int64(v), nil
 	case int8:
-		return int64(v)
+		return int64(v), nil
 	case int16:
-		return int64(v)
+		return int64(v), nil
 	case int32:
-		return int64(v)
+		return int64(v), nil
 	case int64:
-		return v
+		return v, nil
 	case uint:
-		return int64(v)
+		return int64(v), nil
 	case uint8:
-		return int64(v)
+		return int64(v), nil
 	case uint16:
-		return int64(v)
+		return int64(v), nil
 	case uint32:
-		return int64(v)
+		return int64(v), nil
 	case uint64:
-		return int64(v)
+		return int64(v), nil
 	case float32:
-		return float64(v)
+		return float64(v), nil
 	case float64:
 		if math.IsNaN(v) || math.IsInf(v, 0) {
-			return nil
+			return nil, fmt.Errorf("unsupported value: %v", v)
 		}
-		return v
+		return v, nil
 	case bool:
 		if v {
-			return int64(1)
+			return int64(1), nil
 		} else {
-			return int64(0)
+			return int64(0), nil
 		}
 	case distribution.Distribution:
-		return v
+		return v, nil
 	default:
-		return nil
+		return nil, fmt.Errorf("unsupported type: %T", v)
 	}
 }
--- a/internal/util/type_conversion_test.go
+++ b/internal/util/type_conversion_test.go
@@ -15,9 +15,9 @@ func TestToOtelValue(t *testing.T) {
 		input interface{}
 		want  interface{}
 	}{
-		{input: math.NaN(), want: float64(0)},
-		{input: math.Inf(1), want: float64(0)},
-		{input: math.Inf(-1), want: float64(0)},
+		{input: math.NaN(), want: nil},
+		{input: math.Inf(1), want: nil},
+		{input: math.Inf(-1), want: nil},
 		{input: "test", want: nil},
 		{input: int32(3), want: int64(3)},
 		{input: 5.5, want: 5.5},
--- a/internal/util/type_conversion_test.go
+++ b/internal/util/type_conversion_test.go
@@ -8,19 +8,42 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+
+	"github.com/aws/amazon-cloudwatch-agent/metric/distribution/regular"
 )
 
 func TestToOtelValue(t *testing.T) {
+	distribution := regular.NewRegularDistribution()
 	testCases := []struct {
 		input interface{}
 		want  interface{}
 	}{
+		// ints
+		{input: 5, want: int64(5)},
+		{input: int8(5), want: int64(5)},
+		{input: int16(5), want: int64(5)},
+		{input: int32(5), want: int64(5)},
+		{input: int64(5), want: int64(5)},
+		// uints
+		{input: uint(5), want: int64(5)},
+		{input: uint8(5), want: int64(5)},
+		{input: uint16(5), want: int64(5)},
+		{input: uint32(5), want: int64(5)},
+		{input: uint64(5), want: int64(5)},
+		// floats
+		{input: float32(5.5), want: 5.5},
+		{input: 5.5, want: 5.5},
+		// bool
+		{input: false, want: int64(0)},
+		{input: true, want: int64(1)},
+		// distribution
+		{input: distribution, want: distribution},
+		// unsupported floats
 		{input: math.NaN(), want: nil},
 		{input: math.Inf(1), want: nil},
 		{input: math.Inf(-1), want: nil},
+		// unsupported types
 		{input: "test", want: nil},
-		{input: int32(3), want: int64(3)},
-		{input: 5.5, want: 5.5},
 	}
 	for _, testCase := range testCases {
 		assert.Equal(t, testCase.want, ToOtelValue(testCase.input))
--- a/internal/util/type_conversion_test.go
+++ b/internal/util/type_conversion_test.go
@@ -4,6 +4,7 @@
 package util
 
 import (
+	"errors"
 	"math"
 	"testing"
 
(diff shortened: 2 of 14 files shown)


metric/distribution/regular/regular_distribution.go
```diff
@@ -69,6 +70,9 @@ func (regularDist *RegularDistribution) Size() int {
 
 // weight is 1/samplingRate
 func (regularDist *RegularDistribution) AddEntryWithUnit(value float64, weight float64, unit string) error {
+	if math.IsNaN(value) || math.IsInf(value, 0) {
```
adam-mateen, line 73: Could refactor this a bit and remove a level of indentation.
func ... {
    if weight <= 0 || math.IsNaN(value) || math.IsInf(value, 0) || value < 0 {
        return fmt.Errof("E! unsupported weight or value, w %v, v %v", weight, value)
    }
    .
    .
    .
}
```

**Review by sethAmazon — DISMISSED** [review]

File: internal/util/type_conversion.go [hunk 2]
```diff
@@ -30,6 +34,9 @@ func ToOtelValue(value interface{}) interface{} {
 	case float32:
 		return float64(v)
 	case float64:
+		if math.IsNaN(v) || math.IsInf(v, 0) {
+			return nil
+		}
```
adam-mateen, line 39: I am fine with this, but just wanted to bring it up for discussion.
If OTEL is okay with NaN and INF than we shouldn't drop it as this point.

The thing that doesn't support NaN and INF is the backend, so the exporter should be the place that decides to drop it.

Functionally its the same though.

adam-mateen (dismissed):

metric/distribution/regular/regular_distribution.go
```diff
@@ -69,34 +69,33 @@ func (regularDist *RegularDistribution) Size() int {
...
-		//min
-		if value < regularDist.minimum {
-			regularDist.minimum = value
-		}
-		//max
-		if value > regularDist.maximum {
-			regularDist.maximum = value
-		}
+	if weight <= 0 {
+		return fmt.Errorf("weight must be larger than 0: %v", weight)
```
straussb, line 73: You sure this should change from debug to error (or I guess warning where it's logged at the caller)?  Is there any case where it could be too much spam?

adam-mateen (approved):

internal/util/type_conversion.go
jefchien: Sure I can remove this check. It should get dropped at the distribution step, but that'll lead to sending unnecessary metrics through our pipeline.

adam-mateen (dismissed):

sethAmazon (approved):

internal/util/type_conversion_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+import (
+	"math"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestToOtelValue(t *testing.T) {
+	testCases := []struct {
```
straussb, line 14: Do you have time to add test cases for all the other types, while you're here...

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -398,6 +399,10 @@ func (c *CloudWatch) BuildMetricDatum(metric *aggregationDatum) []*cloudwatch.Me
 			continue
 		}
 		if len(distList) == 0 {
+			if math.IsNaN(*metric.Value) || math.IsInf(*metric.Value, 0) {
```
straussb, line 402: Do you want to check for being "in the range of -2^360 to 2^360" too?  Telegraf seems to have [something](https://github.com/influxdata/telegraf/blob/67f9a0ea8276e06e3c3d4028cbbd113dfe9bbc6f/plugins/outputs/cloudwatch/cloudwatch.go#L404-L405) around that...

straussb (commented):
I assume you ran the test program that sends `NaN` values over collectd, to sanity check the fix?

metric/distribution/regular/regular_distribution.go
jefchien: That's fair. I think it's more useful to return it as an error because then we can log the metric that caused it as well. Can change that to a debug log.

jefchien: Or have it so it logs the weight one as a debug and the value one as a warning.

plugins/outputs/cloudwatch/cloudwatch.go
jefchien: I'm not sure what telegraf has, but I can add those bounds. I'm guessing they're inclusive.

jefchien: Having trouble getting the check to validate for values close to the bounds.

straussb (dismissed):

metric/distribution/regular/regular_distribution.go
```diff
@@ -72,7 +72,7 @@ func (regularDist *RegularDistribution) AddEntryWithUnit(value float64, weight f
 	if weight <= 0 {
 		return fmt.Errorf("unsupported weight %v: %w", weight, distribution.ErrUnsupportedWeight)
 	}
-	if value < 0 || math.IsNaN(value) || math.IsInf(value, 0) {
+	if value < 0 || math.IsNaN(value) || math.IsInf(value, 0) || !distribution.IsValueInRange(value) {
```
adam-mateen, line 75: in range check is half redundant with the value < 0 check

adam-mateen: IsValueInRange could just take in the upper and lower bound, 0, x).
Or it could hard code the boundaries if we don't expect them to change.

Another option is to move all the conditions into a single function that checks value.

jefchien: That's fair. I wanted to keep it consistent, but I can change it.

metric/distribution/distribution.go
```diff
@@ -5,13 +5,20 @@ package distribution
...
+const (
+	epsilon = 0.001
+)
+
 var (
 	ErrUnsupportedWeight = errors.New("weight must be larger than 0")
-	ErrUnsupportedValue  = errors.New("value cannot be negative, NaN, or Inf")
+	ErrUnsupportedValue  = errors.New("value cannot be negative, NaN, Inf, or greater than 2^360")
+	MinValue             = -math.Pow(2, 360)
+	MaxValue             = math.Pow(2, 360)
```
adam-mateen, line 21: Could this be a `const` instead of `var`?

jefchien: No. `math.Pow` is a function, so it needs to be a var.

metric/distribution/regular/regular_distribution.go
jefchien: Moved it into a single `IsAcceptedValue` function.

jefchien:
> I assume you ran the test program that sends `NaN` values over collectd, to sanity check the fix?

@sethAmazon was able to verify that it drops the metric. 

```
2023-09-08T21:55:25.619Z	warn	accumulator/accumulator.go:121	Conversion of metric values failed	{"kind": "receiver", "name": "telegraf_socket_listener", "data_type": "metrics", "name": "collectd_seth_test_gauge_1_value", "tags": {"aws:AggregationInterval":"10s","host":"<redacted>","type":"gauge"}, "fields": {}, "type": 3, "error": "empty metrics after converting fields: [value]"}
2023-09-08T21:55:25.619Z	warn	accumulator/accumulator.go:121	Conversion of metric values failed	{"kind": "receiver", "name": "telegraf_socket_listener", "data_type": "metrics", "name": "collectd_seth_test_gauge_1_value", "tags": {"aws:AggregationInterval":"10s","host":"<redacted>","type":"gauge"}, "fields": {}, "type": 3, "error": "empty metrics after converting fields: [value]"}
```

plugins/outputs/cloudwatch/aggregator.go
```diff
@@ -144,7 +145,11 @@ func (durationAgg *durationAggregator) aggregating() {
 					m.distribution = distribution.NewDistribution()
 					err := m.distribution.AddEntryWithUnit(*m.Value, 1, *m.Unit)
 					if err != nil {
-						log.Printf("W! err %s, metric %s", err, *m.MetricName)
+						if errors.Is(err, distribution.ErrUnsupportedValue) {
+							log.Printf("W! err %s, metric %s", err, *m.MetricName)
+						} else if errors.Is(err, distribution.ErrUnsupportedWeight) {
+							log.Printf("D! err %s, metric %s", err, *m.MetricName)
+						}
```
straussb, line 152: Super minor - if `UnsupportedWeight` is the only one we want at debug, then just make it `if`/`else` (instead of `else if`), just in case we add more errors later and forget to update this caller.

receiver/adapter/accumulator/accumulator.go
```diff
@@ -200,7 +201,7 @@ func (o *otelAccumulator) modifyMetricAndConvertToOtelValue(m telegraf.Metric) (
 	}
 
 	if len(mMetric.Fields()) == 0 {
-		return nil, errEmptyAfterConvert
+		return nil, fmt.Errorf("empty metrics after converting fields: %w", errs)
```
straussb, line 204: What does an example of a full error string look like in this case?

receiver/adapter/accumulator/accumulator.go
```diff
@@ -188,9 +185,13 @@ func (o *otelAccumulator) modifyMetricAndConvertToOtelValue(m telegraf.Metric) (
 	// Otel only supports numeric data. Therefore, filter unsupported data type and convert metrics value to corresponding value before
 	// converting the data model
 	// https://github.com/open-telemetry/opentelemetry-collector/blob/bdc3e22d28006b6c9496568bd8d8bcf0aa1e4950/pdata/pmetric/metrics.go#L106-L113
+	var errs error
 	for field, value := range mMetric.Fields() {
 		// Convert all int,uint to int64 and float to float64 and bool to int.
-		otelValue := util.ToOtelValue(value)
+		otelValue, err := util.ToOtelValue(value)
+		if err != nil {
+			errs = multierr.Append(errs, fmt.Errorf("field (%s): %w", field, err))
```
adam-mateen, line 193: It looks like if at least 1 field/value is successfully converted then no error is returned.
So errors are ignored unless every single field/value fails to get converted.
Does this match previous behavior?

receiver/adapter/accumulator/accumulator.go
jefchien: As mentioned above, it'd be something like

```
empty metrics after converting fields: field (client): unsupported type: string; field (value): unsupported value: NaN
```

receiver/adapter/accumulator/accumulator.go
jefchien: Yes, we only return the error if all the fields were dropped.

adam-mateen (approved):

receiver/adapter/accumulator/accumulator.go
jefchien: Changed it to quote the fields, so it'd look like 

```
empty metrics after converting fields: field ("client"): unsupported type: string; field ("value"): unsupported value: NaN
```

Hopefully that makes it clearer that those are the fields.

plugins/outputs/cloudwatch/aggregator.go
jefchien: Changed it so all non-value errors are logged as debug.

adam-mateen (approved):

straussb (approved):

PR #851, opened 2023-09-13
Fix windows event logs to only starts once

okankoAMZ:
# Description of the issue
Currently, when the agent start in windows both log agent and telegraf start the windows log event plugin. As a result, the the plugins generates duplicate logs.

# Description of changes
Changed the start function of the plugin to be exclusively ran once to avoid any duplicate plugin calls to be started.

# License

# Tests
Created and ran a units test this can be seen in  [TestWindowsDuplicateStart](https://github.com/aws/amazon-cloudwatch-agent/blob/95e3f82cafd164ddae91c9e30246187af8bb8484/plugins/inputs/windows_event_log/windows_event_log_test.go#L106)
Also ran integration test:  [this is the instance where there is only logs](https://github.com/aws/amazon-cloudwatch-agent-test/pull/346)

# Requirements
2. Run `make lint`

### DIFF
--- a/plugins/inputs/windows_event_log/windows_event_log.go
+++ b/plugins/inputs/windows_event_log/windows_event_log.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/influxdata/telegraf"
@@ -24,6 +25,8 @@ import (
 const (
 	forcePullInterval = 250 * time.Millisecond
 )
+
+var startOnlyOnce sync.Once
 
 type EventConfig struct {
 	Name          string   `toml:"event_name"`
@@ -77,6 +80,13 @@ func (s *Plugin) FindLogSrc() []logs.LogSrc {
  * We can do any initialization in this method.
  */
 func (s *Plugin) Start(acc telegraf.Accumulator) error {
+	alreadyRan := true
+	startOnlyOnce.Do(func() {
+		alreadyRan = false
+	})
+	if alreadyRan {
+		return nil
+	}
 	for _, eventConfig := range s.Events {
 		// Assume no 2 EventConfigs have the same combination of:
 		// LogGroupName, LogStreamName, Name.
--- a/plugins/inputs/windows_event_log/windows_event_log_test.go
+++ b/plugins/inputs/windows_event_log/windows_event_log_test.go
@@ -10,6 +10,8 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+
+	"github.com/stretchr/testify/require"
 )
 
 // TestGetStateFilePathGood tests getStateFilePath with good input.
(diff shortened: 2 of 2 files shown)


plugins/inputs/windows_event_log/windows_event_log.go
```diff
@@ -77,6 +81,15 @@ func (s *Plugin) FindLogSrc() []logs.LogSrc {
  * We can do any initialization in this method.
  */
 func (s *Plugin) Start(acc telegraf.Accumulator) error {
+	alreadyRan := true
+	startOnlyOnce.Do(func() {
+		fmt.Println("Warning! This function can only be executed once, this is the first and last time it will be ran.")
```
jefchien, line 86: Do not use `fmt`. Use `log` for log lines. I don't think we want to have these logged anyways. It's understood that this will be called multiple times in some scenarios. We don't want to mislead customers into thinking this is an issue.

okankoAMZ: should we just ignore if they run it twice?

plugins/inputs/windows_event_log/windows_event_log_test.go
```diff
@@ -101,3 +103,21 @@ func TestGetStateFilePathSpecialChars(t *testing.T) {
...
+	}
+	ec := EventConfig{
+		LogGroupName:  "My  Group/:::",
+		LogStreamName: "My::Stream//  ",
+		Name:          "System  Event//Log::",
+	}
+	plugin.Events = append(plugin.Events, ec)
+	plugin.Start(nil)
+	plugin.Start(nil)
+	require.Equal(t, 1, len(plugin.newEvents), "Start should be ran only once so there should be only 1 new event")
```
jefchien, line 122: Validate the state before and after each call to `Start`.

plugins/inputs/windows_event_log/windows_event_log_test.go
```diff
@@ -101,3 +103,21 @@ func TestGetStateFilePathSpecialChars(t *testing.T) {
 		t.Errorf("expected non-nil")
 	}
 }
+
+func TestWindowsDuplicateStart(t *testing.T) {
+	fileStateFolder := filepath.Join(os.TempDir(), "CloudWatchAgentTest")
+	// cleanup
+	defer os.RemoveAll(fileStateFolder)
```
jefchien, line 110: `testing.T` has a `TempDir` function that will do the clean up for you.

https://pkg.go.dev/testing#T.TempDir

plugins/inputs/windows_event_log/windows_event_log.go
```diff
@@ -25,6 +27,8 @@ const (
 	forcePullInterval = 250 * time.Millisecond
 )
 
+var startOnlyOnce sync.Once
```
jefchien, line 30: nit: I think we might be able to move this into the `Plugin` struct since the plugin should only be initialized in the `init` function and should be the same instance wherever it's used.

bhanuba (approved):

plugins/inputs/windows_event_log/windows_event_log_test.go
```diff
@@ -101,3 +103,21 @@ func TestGetStateFilePathSpecialChars(t *testing.T) {
...
+	plugin := Plugin{
+		FileStateFolder: fileStateFolder,
+	}
+	ec := EventConfig{
+		LogGroupName:  "My  Group/:::",
+		LogStreamName: "My::Stream//  ",
+		Name:          "System  Event//Log::",
+	}
+	plugin.Events = append(plugin.Events, ec)
+	require.Equal(t, 0, len(plugin.newEvents), "Start should be ran only once so there should be only 1 new event")
```
jefchien, line 118: nit: The error log would be misleading.

jefchien (approved):

jefchien:
Please update the title to better reflect the change. Maybe something like "Fix windows event logs to only starts once"

bhanuba (approved):

Issue #672, opened 2023-01-20
CWA reports duplicate metric observation with same dimension list.


adam-mateen:
**Describe the bug**
If CWA is reporting a metric "foo" with dimensions A, B, C, D.
And you have:
```
"aggregation_dimensions": [["A", "B", "C", "D"]]
```
Then CWA detects that aggregation is the same dimension list as the original with this code:
```
if i == len(targetDimensions) && !reflect.DeepEqual(rawDimension, extraDimensions) {
	fullDimensionsList = append(fullDimensionsList, extraDimensions)
}
```
And the desired behavior occurs - which is each metric value/observation is reported just once.

But if we change to:
```
"aggregation_dimensions": [["D", "B", "C", "A"]]
```
Then the metric gets reported twice with:
1. A, B, C, D (original)
2. D, B, C, A (aggregation)
Even though the order really shouldn’t matter once it is in CloudWatch and you are graphing it.

**Steps to reproduce**
Modify this existing `TestProcessRollup` code:
```
	cw.config.RollupDimensions = [][]string{{"d1", "d2", "d3"}}
	rawDimension = []*cloudwatch.Dimension{
		{
			Name:  aws.String("d1"),
			Value: aws.String("v1"),
		},
		{
			Name:  aws.String("d2"),
			Value: aws.String("v2"),
		},
		{
			Name:  aws.String("d3"),
			Value: aws.String("v3"),
		},
	}

	actualDimensionList = cw.ProcessRollup(rawDimension)
	expectedDimensionList = [][]*cloudwatch.Dimension{
		{
			{
				Name:  aws.String("d1"),
				Value: aws.String("v1"),
			},
			{
				Name:  aws.String("d2"),
				Value: aws.String("v2"),
			},
			{
				Name:  aws.String("d3"),
				Value: aws.String("v3"),
			},
		},
	}
	assert.EqualValues(t, expectedDimensionList, actualDimensionList,
		"Unexpected dimension roll up list with duplicate roll up")
```
By adding these lines:
```
	cw.config.RollupDimensions = [][]string{{"d1", "d2", "d3"}}
	actualDimensionList = cw.ProcessRollup(rawDimension)
	assert.EqualValues(t, expectedDimensionList, actualDimensionList,
		"Unexpected dimension roll up list with duplicate roll up")
```

**What did you expect to see?**
I expect that the order of `aggregation_dimensions` does not matter.
Whether it is:
```
[...]

PR #856, opened 2023-09-20
Prevent aggregate dimension duplicates.

jefchien:
# Description of the issue
Closes https://github.com/aws/amazon-cloudwatch-agent/issues/672

When we create the aggregate dimensions in `ProcessRollup`, we have a check to prevent duplicate dimension sets from occurring. The [reflect.DeepEqual](https://pkg.go.dev/reflect#DeepEqual) comparison requires the order to be the same.
> Slice values are deeply equal when all of the following are true: they are both nil or both non-nil, they have the same length, and either they point to the same initial entry of the same underlying array (that is, &x[0] == &y[0]) or their corresponding elements (up to length) are deeply equal. Note that a non-nil empty slice and a nil slice (for example, []byte{} and []byte(nil)) are not deeply equal. 

This means that if the metric has dimensions `[a, b, c]` and the configuration is `[[a, c, b]]`, then we'll get `[[a, b, c], [a, c, b]]` as the resulting aggregate dimensions (original and configured). This results in duplicate metrics/dimension combinations being sent to CloudWatch.

# Description of changes
Skip dimension aggregations that have the same number of dimensions as the original metric. There are two scenarios in this case: either the same dimensions have been specified in the aggregate dimensions or a dimension that does not exist for the metric is being specified. Either way, that combination can be dropped from the aggregation.

Update `GetUniqueRollupList` to also account for duplicates and ignore order within the dimension sets, so `[[a, a, b]]` and `[[b, a]]` is equivalent to `[[a, b]]`.

# License

# Tests
Added tests for different ordered aggregate dimensions configuration.

# Requirements
2. Run `make lint`

### DIFF
--- a/plugins/outputs/cloudwatch/cloudwatch.go
+++ b/plugins/outputs/cloudwatch/cloudwatch.go
@@ -22,6 +22,7 @@ import (
 	"go.opentelemetry.io/collector/consumer"
 	"go.opentelemetry.io/collector/exporter"
 	"go.opentelemetry.io/collector/pdata/pmetric"
+	"golang.org/x/exp/maps"
 
 	configaws "github.com/aws/amazon-cloudwatch-agent/cfg/aws"
 	"github.com/aws/amazon-cloudwatch-agent/handlers"
@@ -502,6 +503,7 @@ func BuildDimensions(tagMap map[string]string) []*cloudwatch.Dimension {
 	return dimensions
 }
 
+// ProcessRollup creates the dimension sets based on the dimensions available in the original metric.
 func (c *CloudWatch) ProcessRollup(rawDimensions []*cloudwatch.Dimension) [][]*cloudwatch.Dimension {
 	rawDimensionMap := map[string]string{}
 	for _, v := range rawDimensions {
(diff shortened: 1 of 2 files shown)


bryce-carey (approved):

adam-mateen:
I think this is the same as https://github.com/aws/amazon-cloudwatch-agent/issues/672

adam-mateen (dismissed):

bryce-carey (approved):

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -523,29 +525,35 @@ func (c *CloudWatch) ProcessRollup(rawDimension []*cloudwatch.Dimension) [][]*cl
 			}
 			i += 1
 		}
-		if i == len(targetDimensions) && !reflect.DeepEqual(rawDimension, extraDimensions) {
+		// skip if same number of dimensions as the original metric
+		if i == len(targetDimensions) && len(targetDimensions) != len(rawDimensions) {
```
adam-mateen, line 529: We are checking `len(targetDimensions) != len(rawDimensions)` after the inner loop.
We could have checked it before the inner loop and saved some time.

The inner loop is small and probably not that expensive so I don't care either way.

jefchien: You're right. Will make the change.

adam-mateen (approved):

bryce-carey (approved):

PR #870, opened 2023-09-28
Default timestamp_format to be compatible with zero padding and non zero padding

ymtaye:
# Description of the issue
There is a regression between the old Python-based log agent and the new Go-based agent in how strict the timestamp format is. In the old Python logs agent, %d would match both zero and non-zero padded. There wasn't another option.
Although the documentation calls out %d as zero padded, the reality is that strptime will accept both.

This caused users to experience the following scenario:

When config is set with a `timestamp_format` that is inconsistent with the log's timestamp, the CloudWatch Agent would revert to last valid timestamp and push that as the log's events timestamp. This scenario involves where config could be set as
```bash
"timestamp_format": "%b %d %H:%M:%S"
```
while log file would have log timestamp formats
`Apr  5 02:09:45`

# Description of changes
Add defaulting mechanism, so when config is set to zero padding timestamp regex and log timestamp has a non zero padded format the CloudWatch Agent will default and use a non zero padding time stamp. This was modified for the following timestamp format field options. See [doc](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Agent-Configuration-File-Details.html#CloudWatch-Agent-Configuration-File-Logssection)
```
%m
Month as a zero-padded decimal number

%-m
Month as a decimal number (not zero-padded)

%d
Day of the month as a zero-padded decimal number

%-d
Day of the month as a decimal number (not zero-padded)
```
### Summary of changes

`%d` and `%-d` in json `timestamp_format` both translate to `"\\s{0,1}\\d{1,2}"` in `timestamp_regex` for toml
`%m` and `%-m`   in json `timestamp_format` both translate to `"\\s{0,1}\\d{1,2}"`  in `timestamp_regex` for toml

Since the [ParseInLocation](https://github.com/aws/amazon-cloudwatch-agent/blob/a656bb931efbd4caa096abcba764c8def8493b4a/plugins/inputs/logfile/fileconfig.go#L182) function requires a `timestamp_layout` and go only supports `_2` for %d and not %m, `timestamp_layout` has been modified to be an array of layouts with both options

`%d` and `%-d` both translate to _2. For example 
`timestamp_format: "%b %-d %Y %H:%M:%S"` and  `timestamp_format: "%b %d %Y %H:%M:%S"`
both map to
```
timestamp_regex : "(\\w{3} \\s{0,1}\\d{1,2} \\d{4} \\d{2}:\\d{2}:\\d{2})"
```
`%m` and `%-m`   in json `timestamp_format` translate to the following arrays. Setting the month format in json config as the first option in `timestamp_layout` array the alternative approach being the 2nd layout in array
timestamp_format: `"%m %-d  %H:%M:%S"` maps to the following
```
timestamp_layout : ["1 _2 15:04:05", "01 _2 15:04:05"]
```

# License

# Tests
## Unit tests
```bash
--- PASS: TestTimestampRegexRule (0.00s)
    --- PASS: TestTimestampRegexRule/WithInvalidTimestampFormat (0.00s)
    --- PASS: TestTimestampRegexRule/WithNonZeroPaddedOptions (0.00s)
    --- PASS: TestTimestampRegexRule/WithZeroPaddedOptions (0.00s)
    --- PASS: TestTimestampRegexRule/WithZeroPaddedMonthWord (0.00s)
    --- PASS: TestTimestampRegexRule/WithNonZeroPaddedMonthWord (0.00s)
    --- PASS: TestTimestampRegexRule/WithYearAsTwoDigits (0.00s)
    --- PASS: TestTimestampRegexRule/WithYearAsFourDigits (0.00s)
    --- PASS: TestTimestampRegexRule/WithNoTimestampFormat (0.00s)
=== RUN   TestTimestampLayoutxRule
=== RUN   TestTimestampLayoutxRule/WithYearAsFourDigits
=== RUN   TestTimestampLayoutxRule/WithNoTimestampFormat
=== RUN   TestTimestampLayoutxRule/WithInvalidTimestampFormat
=== RUN   TestTimestampLayoutxRule/WithNonZeroPaddedOptions
=== RUN   TestTimestampLayoutxRule/WithZeroPaddedOptions
=== RUN   TestTimestampLayoutxRule/WithZeroPaddedMonthWord
=== RUN   TestTimestampLayoutxRule/WithNonZeroPaddedMonthWord
=== RUN   TestTimestampLayoutxRule/WithYearAsTwoDigits
--- PASS: TestTimestampLayoutxRule (0.00s)
    --- PASS: TestTimestampLayoutxRule/WithYearAsFourDigits (0.00s)
    --- PASS: TestTimestampLayoutxRule/WithNoTimestampFormat (0.00s)
    --- PASS: TestTimestampLayoutxRule/WithInvalidTimestampFormat (0.00s)
    --- PASS: TestTimestampLayoutxRule/WithNonZeroPaddedOptions (0.00s)
    --- PASS: TestTimestampLayoutxRule/WithZeroPaddedOptions (0.00s)
    --- PASS: TestTimestampLayoutxRule/WithZeroPaddedMonthWord (0.00s)
    --- PASS: TestTimestampLayoutxRule/WithNonZeroPaddedMonthWord (0.00s)
    --- PASS: TestTimestampLayoutxRule/WithYearAsTwoDigits (0.00s)
PASS
ok  	github.com/aws/amazon-cloudwatch-agent/translator/translate/logs/logs_collected/files/collect_list	0.019s

```

## Manual test checks for timestamp translation
json
```
 "logs": {
        "logs_collected": {
            "files": {
                "collect_list": [
                    {
                        "log_group_name": "amazon-cloudwatch-agent.log"
                    },
                    {
                        "file_path": "/opt/aws/amazon-cloudwatch-agent/logs/testing.log",
                        "log_group_name": "testing.log",
                        "log_stream_name": "testing.log",
                        "timestamp_format": "%m %d %H:%M:%S"
                    }
                ]
            }
        }
    }
```

toml
```
[inputs]

  [[inputs.cpu]]
    fieldpass = ["usage_idle", "usage_iowait", "usage_user", "usage_system"]
    interval = "60s"
    percpu = true
    totalcpu = false

  [[inputs.logfile]]
    destination = "cloudwatchlogs"
    file_state_folder = "/opt/aws/amazon-cloudwatch-agent/logs/state"

    [[inputs.logfile.file_config]]
      file_path = "/opt/aws/amazon-cloudwatch-agent/logs/amazon-cloudwatch-agent.log"
      from_beginning = true
      log_group_name = "amazon-cloudwatch-agent.log"
      pipe = false
      retention_in_days = -1

    [[inputs.logfile.file_config]]
      file_path = "/opt/aws/amazon-cloudwatch-agent/logs/testing.log"
      from_beginning = true
      log_group_name = "testing_ip-172-31-7-172"
      log_stream_name = "testing2.log"
      pipe = false
      retention_in_days = -1
      timestamp_layout = ["01 _2 15:04:05", "1 _2 15:04:05"]
      timestamp_regex = "(\\d{1,2} \\s{0,1}\\d{1,2} \\d{2}:\\d{2}:\\d{2})"
```
## Manual test for error Messaging Added

```
[ec2-user@ip-172-31-7-172 ~]$ cat /opt/aws/amazon-cloudwatch-agent/logs/amazon-cloudwatch-agent.log | grep -m 10 "Unable to parse valid timestamp testing.log"
2023-09-29T16:06:16Z D! [outputs.cloudwatchlogs] Unable to parse valid timestamp testing.log:
2023-09-29T16:06:16Z D! [outputs.cloudwatchlogs] Unable to parse valid timestamp testing.log:
2023-09-29T16:06:16Z D! [outputs.cloudwatchlogs] Unable to parse valid timestamp testing.log:
2023-09-29T16:06:16Z D! [outputs.cloudwatchlogs] Unable to parse valid timestamp testing.log:
2023-09-29T16:06:16Z D! [outputs.cloudwatchlogs] Unable to parse valid timestamp testing.log:
2023-09-29T16:06:16Z D! [outputs.cloudwatchlogs] Unable to parse valid timestamp testing.log:
2023-09-29T16:06:16Z D! [outputs.cloudwatchlogs] Unable to parse valid timestamp testing.log:
2023-09-29T16:06:16Z D! [outputs.cloudwatchlogs] Unable to parse valid timestamp testing.log:
2023-09-29T16:06:16Z D! [outputs.cloudwatchlogs] Unable to parse valid timestamp testing.log:
2023-09-29T16:06:16Z D! [outputs.cloudwatchlogs] Unable to parse valid timestamp testing.log:
```

```
[ec2-user@ip-172-31-7-172 ~]$ cat /opt/aws/amazon-cloudwatch-agent/logs/testing.log | grep "Oct 3 15:15:12" -A 5 -B 5
 ip-10-4-213-132 sudo: vtayyare : TTY=pts/0 ; PWD=/home/vtayyare ; USER=root ; COMMAND=/bin/cat /var/log/secure
10 3 15:15:11  instance-id: i-02fce21a425a2efb3: Testing for %-d and %-m
Oct 3 15:15:11  ip-10-4-213-132 sudo: vtayyare : TTY=pts/0 ; PWD=/home/vtayyare ; USER=root ; COMMAND=/bin/cat /var/log/secure
 ip-10-4-213-132 sudo: vtayyare : TTY=pts/0 ; PWD=/home/vtayyare ; USER=root ; COMMAND=/bin/cat /var/log/secure
10 3 15:15:12  instance-id: i-02fce21a425a2efb3: Testing for %-d and %-m
Oct 3 15:15:12  ip-10-4-213-132 sudo: vtayyare : TTY=pts/0 ; PWD=/home/vtayyare ; USER=root ; COMMAND=/bin/cat /var/log/secure
 ip-10-4-213-132 sudo: vtayyare : TTY=pts/0 ; PWD=/home/vtayyare ; USER=root ; COMMAND=/bin/cat /var/log/secure
10 3 15:15:14  instance-id: i-02fce21a425a2efb3: Testing for %-d and %-m
Oct 3 15:15:14  ip-10-4-213-132 sudo: vtayyare : TTY=pts/0 ; PWD=/home/vtayyare ; USER=root ; COMMAND=/bin/cat /var/log/secure
 ip-10-4-213-132 sudo: vtayyare : TTY=pts/0 ; PWD=/home/vtayyare ; USER=root ; COMMAND=/bin/cat /var/log/secure
10 3 15:15:15  instance-id: i-02fce21a425a2efb3: Testing for %-d and %-m
```
Screenshot 2023-10-03 at 11 13 34 AM

# Requirements
2. Run `make lint`


translator/translate/logs/logs_collected/files/collect_list/ruleTimestampFormat.go
```diff
@@ -82,11 +82,11 @@ var TimeFormatRexMap = map[string]string{
 	"%B":  "\\w{7}",
 	"%b":  "\\w{3}",
 	"%-m": "\\s{0,1}\\d{1,2}",
-	"%m":  "\\d{2}",
+	"%m":  "\\s{0,1}\\d{1,2}",
 	"%A":  "\\w{6,9}",
 	"%a":  "\\w{3}",
 	"%-d": "\\s{0,1}\\d{1,2}",
-	"%d":  "\\d{2}",
+	"%d":  "\\s{0,1}\\d{1,2}",
```
bryce-carey, line 89: Based on the timestamp format documentation, the agent is doing the correct thing. So are we making this change because there was a behavior change between agent versions 1.2X and 1.3X?

sky333999: +1. I dont fully understand why we are doing this either. There already is a documented way to configure non-zero-padded days with `%-d` so if customers want that behavior, they have an option to do so.
And this is fairly standard for all timestamp specifiers from what I can tell - they have an option for zero padded and an option for non-zero-padded.

ymtaye: It was between logs agent and cwa, when agent added support for `%-d` and `%-m` https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Agent-common-scenarios.html#CloudWatch-Agent-logs-timestamp-differences

jefchien: I think the `Description of the Issue` is focusing on the wrong part. The issue isn't that configuring the timestamp wrong could lead to incorrect timestamps. It's that there is a regression between the old Python-based log agent and the new Go-based agent in how strict the timestamp format is. In the old Python logs agent, `%d` would match both zero and non-zero padded. There wasn't another option.

https://docs.python.org/3/library/datetime.html#strftime-strptime-behavior

Although the documentation calls out `%d` as zero padded, the reality is that `strptime` will accept both.

translator/translate/logs/logs_collected/files/collect_list/collect_list_test.go
```diff
@@ -113,13 +113,13 @@ func TestTimestampFormatAll(t *testing.T) {
 			input: `{
 					"collect_list":[
 						{
-							"file_path":"path1",
+							"file_path":"path2",
```
jefchien, line 116: nit: Why are we changing these?

plugins/inputs/logfile/fileconfig_test.go
```diff
@@ -133,6 +133,29 @@ func TestTimestampParserWithPadding(t *testing.T) {
 	assert.Equal(t, 10, timestamp.Minute(), fmt.Sprintf("Timestamp does not match: %v, act: %v", "10", timestamp.Minute()))
 }
 
+func TestTimestampParserDefault(t *testing.T) {
+	// Check when timestamp_format is "%b %d %H:%M:%S"
+	// %d and %-d are both treated as s{0,1}\\d{1,2}
+	timestampRegex := "(\\w{3} \\s{0,1}\\d{1,2} \\d{2}:\\d{2}:\\d{2})"
+	timestampLayout := "Jan  9 15:04:05"
```
jefchien, line 140: This is an invalid timestamp layout. Go will interpret `9` differently. Day of the month must be some variant of 2. See https://pkg.go.dev/time

translator/translate/logs/logs_collected/files/collect_list/collect_list_test.go
```diff
@@ -149,13 +149,49 @@ func TestTimestampFormatAll(t *testing.T) {
...
+                            "timestamp_format": "%b %d %H:%M:%S"
+						}
+					]
+				}`,
+			expected: []interface{}{map[string]interface{}{
+				"file_path":         "path4",
+				"from_beginning":    true,
+				"pipe":              false,
+				"retention_in_days": -1,
+				"timestamp_layout":  "Jan 02 15:04:05",
```
jefchien, line 162: Does this work for zero padded and non-zero padded days? `02` indicates that it's zero padded only.

plugins/outputs/cloudwatchlogs/pusher.go
```diff
@@ -413,8 +413,10 @@ func (p *pusher) convertEvent(e logs.LogEvent) *cloudwatchlogs.InputLogEvent {
 			// a valid timestamp and use the last valid timestamp for new entries that does
 			// not have a timestamp.
 			t = p.lastValidTime
+			p.Log.Debugf("Unable to parse valid timestamp, using last valid timestamp %v: for log group %v: ", p.lastValidTime, p.Group)
```
jefchien, line 416: We don't want to log this every time. There are valid scenarios where this path would be accessed. This log line would be misleading.

translator/translate/logs/logs_collected/files/collect_list/collect_list_test.go
ymtaye: I can revert them back

plugins/outputs/cloudwatchlogs/pusher.go
ymtaye: Yeah I will add a interval check

plugins/outputs/cloudwatchlogs/pusher.go
```diff
@@ -413,6 +413,9 @@ func (p *pusher) convertEvent(e logs.LogEvent) *cloudwatchlogs.InputLogEvent {
 			// a valid timestamp and use the last valid timestamp for new entries that does
 			// not have a timestamp.
 			t = p.lastValidTime
+			if ((time.Now().UnixNano() / 1000000000) - (t / 1000)) > 432000 {
```
jefchien, line 416: What are these magic numbers?

plugins/outputs/cloudwatchlogs/pusher.go
```diff
@@ -413,6 +413,9 @@ func (p *pusher) convertEvent(e logs.LogEvent) *cloudwatchlogs.InputLogEvent {
...
+				p.Log.Debugf("Unable to parse timestamp, using last valid timestamp %v: which is older than 5 days for log group %v: ", p.lastValidTime, p.Group)
```
jefchien, line 417: This seems like it should be a warning message. Also this will get logged for every log line we try to send. Could we throttle the amount of these?

translator/translate/logs/logs_collected/files/collect_list/ruleTimestampFormat_test.go
```diff
@@ -0,0 +1,133 @@
...
+	var input interface{}
+	e := json.Unmarshal([]byte(`{
+			"timestamp_format": "%b %d %H:%M:%S"
+	}`), &input)
+	if e == nil {
+		actualReturnKey, retVal := r.ApplyRule(input)
+		assert.Equal(t, "timestamp_regex", actualReturnKey)
+		assert.NotNil(t, retVal)
+	} else {
+		panic(e)
```
jefchien, line 24: Don't panic in the test. Use `require.NoError(t, e)` instead so it'll fail fast.

translator/translate/logs/logs_collected/files/collect_list/ruleTimestampFormat_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestApplyTimestampFormatRule(t *testing.T) {
+	r := new(TimestampRegax)
```
jefchien, line 14: nit: misspelling
```suggestion
	r := new(TimestampRegex)
```

translator/translate/logs/logs_collected/files/collect_list/ruleTimestampFormat_test.go
```diff
@@ -0,0 +1,133 @@
...
+	var input interface{}
+	e := json.Unmarshal([]byte(`{
```
jefchien, line 16: Why are we unmarshaling JSON? Couldn't we just use a `map[string]interface{}`?

translator/translate/logs/logs_collected/files/collect_list/ruleTimestampFormat_test.go
```diff
@@ -0,0 +1,133 @@
...
+package collect_list
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestApplyTimestampFormatRule(t *testing.T) {
```
jefchien, line 13: nit: I think these tests can be consolidated into two tests (TestTimestampFormat and TestTimestampLayout) with multiple test cases.

translator/translate/logs/logs_collected/files/collect_list/ruleTimestampFormat_test.go
ymtaye: That's already defined here, I was using the existing interface https://github.com/aws/amazon-cloudwatch-agent/blob/70c071bb9093bd4f63a7630abb002f8660ea8341/translator/translate/logs/logs_collected/files/collect_list/ruleTimestampFormat.go#L138

translator/translate/logs/logs_collected/files/collect_list/ruleTimestampFormat_test.go
ymtaye: I can group them for conciseness, thanks for the callout

translator/translate/logs/logs_collected/files/collect_list/ruleTimestampFormat_test.go
ymtaye: I was following other translation rule tests for consistency.
https://github.com/aws/amazon-cloudwatch-agent/blob/70c071bb9093bd4f63a7630abb002f8660ea8341/translator/translate/logs/logs_collected/files/collect_list/ruleLogFilters_test.go#L19
https://github.com/aws/amazon-cloudwatch-agent/blob/70c071bb9093bd4f63a7630abb002f8660ea8341/translator/translate/logs/logs_collected/files/collect_list/ruleRetentionInDays_test.go#L32
https://github.com/aws/amazon-cloudwatch-agent/blob/70c071bb9093bd4f63a7630abb002f8660ea8341/translator/translate/logs/logs_collected/files/collect_list/collect_list_test.go#L24

plugins/outputs/cloudwatchlogs/pusher.go
ymtaye: For calculating the days, I convert the nanosecond value and divide by 1000000000 to convert into seconds. `t`is going to be in miliseconds so divide that by 1000

`432000` is the second value of 5 days so I'm trying to log the message when `lastValidTime` is 5 days from now

plugins/inputs/logfile/fileconfig_test.go
```diff
@@ -133,6 +133,45 @@ func TestTimestampParserWithPadding(t *testing.T) {
...
+		TimestampRegex:  timestampRegex,
+		TimestampRegexP: timestampRegexP,
+		TimestampLayout: timestampLayout,
+		Timezone:        timezone,
+		TimezoneLoc:     timezoneLoc}
+
+	// make sure layout is compatible for "Sep 9", "Sep  9" , "Sep 09", "Sep  09" options
+	logEntry := fmt.Sprintf("Sep 9 02:00:43  ip-10-4-213-132 sudo: vtayyare : TTY=pts/0 ; PWD=/home/vtayyare ; USER=root ; COMMAND=/bin/cat /var/log/secure\n")
+	timestamp := fileConfig.timestampFromLogLine(logEntry)
+	assert.Equal(t, 02, timestamp.Hour(), fmt.Sprintf("Timestamp does not match: %v, act: %v", "2", timestamp.Hour()))
```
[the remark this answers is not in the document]
ymtaye: Ok I will fix this

translator/translate/logs/logs_collected/files/collect_list/ruleTimestampFormat_test.go
ymtaye: Changed test grouping

translator/translate/logs/logs_collected/files/collect_list/ruleTimestampFormat_test.go
ymtaye: I have changed the `ruleTimestampFormat_test` to follow this pattern. Note: there is still a large chunck of unit tests using unmarshalling JSON, refactoring those is out of scope for this PR

translator/translate/logs/logs_collected/files/collect_list/ruleTimestampFormat_test.go
ymtaye: Refractored the regex interfacce

translator/translate/logs/logs_collected/files/collect_list/ruleTimestampFormat_test.go
ymtaye: Addressed this

plugins/inputs/logfile/fileconfig.go
```diff
@@ -43,7 +43,7 @@ type FileConfig struct {
 	//The regex of the timestampFromLogLine presents in the log entry
 	TimestampRegex string `toml:"timestamp_regex"`
 	//The timestampFromLogLine layout used in GoLang to parse the timestampFromLogLine.
-	TimestampLayout string `toml:"timestamp_layout"`
+	TimestampLayout []string `toml:"timestamp_layout"`
```
sethAmazon, line 46: @jefchien I thought you said not to do this

translator/translate/logs/logs_collected/files/collect_list/ruleTimestampFormat.go
```diff
@@ -181,7 +181,20 @@ func (t *TimestampLayout) ApplyRule(input interface{}) (returnKey string, return
 		res := checkAndReplace(val.(string), TimeFormatMap)
 		//If user provide with the specific timestamp_format, use the one that user provide
 		returnKey = "timestamp_layout"
-		returnVal = res
+		input := val.(string)
```
jefchien, line 184: nit: This shadows the `input` variable that's passed in. Choose another name.

translator/translate/logs/logs_collected/files/collect_list/ruleTimestampFormat.go
```diff
@@ -181,7 +181,20 @@ func (t *TimestampLayout) ApplyRule(input interface{}) (returnKey string, return
...
+		// Go doesn't support _2 option for month in day as a result need to set
+		// timestamp_layout with 2 strings which support %m and %-m
+		if strings.Contains(input, "%m") {
+			input := strings.Replace(input, "%m", "%-m", -1)
+			defaultOption := checkAndReplace(input, TimeFormatMap)
```
jefchien, line 189: nit: `defaultOption` doesn't really fit what this is doing. Consider renaming.

plugins/inputs/logfile/fileconfig.go
ymtaye: We can't satisfy the requirement to default for `%m` and `%-m` without setting both options prior to getting the log timestamp. For `%d` and `%-d` we are able to achieve this by using `_2` but that's not an option for month. So this is the only way to get `timestamp_layout` to be compatible.

translator/tocwconfig/sampleConfig/logs_and_kubernetes_config.conf
```diff
@@ -28,7 +28,7 @@
       multi_line_start_pattern = "{timestamp_regex}"
       pipe = false
       retention_in_days = -1
-      timestamp_layout = "02 Jan 2006 15:04:05"
+      timestamp_layout = "_2 Jan 2006 15:04:05"
```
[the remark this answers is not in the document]
ymtaye: I'll fix this now, hmm the unit test didn't catch it seems. I fixed the ones that were failing

plugins/inputs/logfile/README.md
```diff
@@ -63,7 +63,7 @@ The plugin expects messages in one of the
       log_group_name = "varlog"
       log_stream_name = "<log_stream_name>"
       timestamp_regex = "^(\\d{2} \\w{3} \\d{4} \\d{2}:\\d{2}:\\d{2}).*$"
-      timestamp_layout = "02 Jan 2006 15:04:05"
+      timestamp_layout = "_2 Jan 2006 15:04:05"
```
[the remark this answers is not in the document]
ymtaye: I changed these based on unit test results, I'll modify all instances of `timestamp_layout`

translator/translate/logs/logs_collected/files/collect_list/ruleTimestampFormat.go
ymtaye: changed to `timestampInput`

translator/translate/logs/logs_collected/files/collect_list/ruleTimestampFormat.go
ymtaye: renamed to `alternativeLayout`

translator/translate/logs/logs_collected/files/collect_list/collect_list_test.go
```diff
@@ -223,8 +277,10 @@ func TestTimestampFormat_NonZeroPadding(t *testing.T) {
 		match := regex.FindStringSubmatch(sampleLogEntry)
 		assert.NotNil(t, match)
 		assert.Equal(t, 2, len(match))
-
-		parsedTime, err := time.ParseInLocation(expectedLayout, match[1], time.UTC)
+		parsedTime, err := time.ParseInLocation(expectedLayout[0], match[1], time.UTC)
+		if err != nil && len(expectedLayout) > 1 {
+			parsedTime, err = time.ParseInLocation(expectedLayout[1], match[1], time.UTC)
```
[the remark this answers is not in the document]
ymtaye: Changed the test to use only the compatible `timestamp_layout`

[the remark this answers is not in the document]
ymtaye: We are testing both cases so there will need to be an if check
```
sampleLogEntries := []string{
		"1:2:3 18 3 8 - Log Content",
		"1:2:3 18  3  8 - Log Content",
	}
```

plugins/inputs/logfile/fileconfig.go
jefchien: @sethAmazon I thought we didn't need to do this, but since Go's time layout doesn't support zero padded/non-padded months with a single token, we have to have multiple layouts.

plugins/inputs/logfile/README.md
```diff
@@ -45,7 +45,7 @@ The plugin expects messages in one of the
       log_group_name = "logfile.log"
       log_stream_name = "<log_stream_name>"
       timestamp_regex = "^(\\d{2} \\w{3} \\d{4} \\d{2}:\\d{2}:\\d{2}).*$"
-      timestamp_layout = "02 Jan 2006 15:04:05"
+      timestamp_layout = "_2 Jan 2006 15:04:05"
```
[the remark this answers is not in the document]
ymtaye: Well, going forward the toml cfg will always have `_2` instead of `02` for `timestamp_layout` thought I should keep consistency

plugins/outputs/cloudwatchlogs/pusher.go
```diff
@@ -413,12 +448,21 @@ func (p *pusher) convertEvent(e logs.LogEvent) *cloudwatchlogs.InputLogEvent {
 			// a valid timestamp and use the last valid timestamp for new entries that does
 			// not have a timestamp.
 			t = p.lastValidTime
+            if p.lastUpdateTime != 0 {
+              // Check when timestamp has an interval of 5 days.
+              // 432000 seconds = 5 days
+              // p.lastUpdateTime -> is set in miliseconds
+                if ((time.Now().UnixNano() / 1000000000) - (p.lastUpdateTime / 1000)) > 432000 {
```
okankoAMZ, line 455: >100000000 
is this to convert from Nano seconds to seconds? can we not just use Unix() instead of UnixNano(), is there any particular reason we are using UnixNano ?

SaxyPandaBear: Agreed. The log timestamps that CWL accepts do not accept nanosecond precision. There's no need for us to use that level of precision.

plugins/outputs/cloudwatchlogs/pusher.go
```diff
@@ -82,6 +90,32 @@ func NewPusher(target Target, service CloudWatchLogsService, flushTimeout time.D
...
+    // where N = sc.Thereafter, M = sc.Initial, S = sc.Tick.
+    samplingCore := zapcore.NewSamplerWithOptions(
+        jsonOutCore,
+        time.Second, // interval
+        3, // log first 3 entries
+        0, // thereafter log zero entires within the interval
+    )
+    p.SampledLog = zap.New(samplingCore)
+    return
+}
```
SaxyPandaBear, line 117: A couple things. 
1. I think that the code for creating the Zap logger here should not be tied to the pusher code. You should pull it out to `/logger`. 
2. It's bad code smell to have this create the logger, configure it and also set the pusher's logger to it. If we're pulling the creation code out of this file, I think realistically all you need to do is add the logger to the constructor function for the pusher. 
3. This is creating a separate logger for each pusher. Is that acceptable?

plugins/outputs/cloudwatchlogs/pusher.go
jefchien: Better yet, store the `lastUpdateTime` as a `time.Time` so you can do the check without all this conversion.

```go
time.Since(p.lastUpdateTime) > [name_of_threshold_variable]
```

where [name_of_threshold_variable] is like `5 * 24 * time.Hour`

plugins/outputs/cloudwatchlogs/pusher.go
jefchien: I don't think we should create a new logger just for one log line. Even if we wanted to use the `zap.Logger`, we should be passing it into the pusher and reusing the one created in the agent.

ymtaye: I removed the Zap sampler since we are using the Telegraf.Logger for cloudwatch logs.

plugins/outputs/cloudwatchlogs/pusher.go
ymtaye: Fixed this, to follow format as Jeffrey suggested
```
			t = p.lastValidTime
			if !p.lastUpdateTime.IsZero() {
				// Check when timestamp has an interval of 5 days.
				if time.Since(p.lastUpdateTime) > elaspedPeriod {
					p.Log.Warnf("Unable to parse timestamp, using last valid timestamp %v: which is older than 5 days for log group %v: ", p.lastValidTime, p.Group)
				}
			}
```

translator/translate/logs/logs_collected/files/collect_list/collect_list_test.go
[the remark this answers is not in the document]
ymtaye: Fixed this

plugins/outputs/cloudwatchlogs/pusher.go
ymtaye: Cleaned up check on time to use time.Hour for comparison and go builtin time.Since()

plugins/outputs/cloudwatchlogs/pusher.go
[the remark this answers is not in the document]
ymtaye: Changed to time.Hour as time.Days isn't supported

plugins/inputs/logfile/fileconfig_test.go
```diff
@@ -133,9 +133,48 @@ func TestTimestampParserWithPadding(t *testing.T) {
...
+	require.NoError(t, err, fmt.Sprintf("Failed to compile regex %s", timestampRegex))
+	fileConfig := &FileConfig{
+		TimestampRegex:  timestampRegex,
+		TimestampRegexP: timestampRegexP,
+		TimestampLayout: timestampLayout,
+		Timezone:        timezone,
+		TimezoneLoc:     timezoneLoc}
+
+	// make sure layout is compatible for "Sep 9", "Sep  9" , "Sep 09", "Sep  09" options
+	logEntry := fmt.Sprintf("Sep 9 02:00:43  ip-10-4-213-132 sudo: vtayyare : TTY=pts/0 ; PWD=/home/vtayyare ; USER=root ; COMMAND=/bin/cat /var/log/secure\n")
```
jefchien, line 153: nit: Could make this a set of testcases. Especially since they're all expecting the same hour/minute value. Also, probably can be a simpler log line.

plugins/outputs/cloudwatchlogs/pusher.go
```diff
@@ -22,6 +22,7 @@ import (
 const (
 	reqSizeLimit   = 1024 * 1024
 	reqEventsLimit = 10000
+	elaspedPeriod  = 5 * 24 * time.Hour
```
jefchien, line 25: nit: misspelled
```suggestion
	elapsedPeriod  = 5 * 24 * time.Hour
```
 but also this isn't descriptive. What is this used for? Change the name to reflect that.

jefchien: 5 days is pretty long to wait before emitting the log since we're using the update timestamp now. How did we decide on 5 days?

translator/translate/logs/logs_collected/files/collect_list/collect_list_test.go
```diff
@@ -182,8 +236,8 @@ func applyRule1(t *testing.T, buf string) interface{} {
 	return val
 }
 
-// stdNumMonth     // "1"         //%-m
-// stdDay          // "2"         //%-d
+// stdNumMonth     // "_1"         //%-m
```
jefchien, line 239: This should be reverted.

ymtaye: Correct must have forgot to remove it in the revision where I changed `timestamp_layout`

plugins/inputs/logfile/fileconfig_test.go
ymtaye: I will shorten the log lines and address this. Keeping test case as is for consistency in other `fileconfig_test.go`

plugins/outputs/cloudwatchlogs/pusher.go
bryce-carey: Would you all be open to a 1-day elapsed time for this log warning message? This is a hard one and I see Yared is trying to strike a balance between raising awareness and avoiding noise. IMO 1 day would be better since it is closer in time to when the impact started.

plugins/outputs/cloudwatchlogs/pusher.go
```diff
@@ -20,8 +20,9 @@ import (
 )
 
 const (
-	reqSizeLimit   = 1024 * 1024
-	reqEventsLimit = 10000
+	reqSizeLimit    = 1024 * 1024
+	reqEventsLimit  = 10000
+	logWarnInterval = 1 * 24 * time.Hour
```
jefchien, line 25: nit: This should be more specific. `logWarnInterval` is too generic. Is this interval for all `log.Warn` messages? It's also not an interval. An interval determines how often something happens. This is more like a threshold.

plugins/outputs/cloudwatchlogs/pusher.go
```diff
@@ -413,12 +415,19 @@ func (p *pusher) convertEvent(e logs.LogEvent) *cloudwatchlogs.InputLogEvent {
 			// a valid timestamp and use the last valid timestamp for new entries that does
 			// not have a timestamp.
 			t = p.lastValidTime
+			if !p.lastUpdateTime.IsZero() {
+				// Check when timestamp has an interval of 5 days.
+				if time.Since(p.lastUpdateTime) > logWarnInterval {
+					p.Log.Warnf("Unable to parse timestamp, using last valid timestamp %v: which is at least older by 1 day for log group %v: ", p.lastValidTime, p.Group)
```
jefchien, line 421:
```suggestion
					p.Log.Warnf("Unable to parse timestamp, using last valid timestamp found in the logs %v: which is at least older than 1 day for log group %v: ", p.lastValidTime, p.Group)
```

jefchien (approved):

plugins/outputs/cloudwatchlogs/pusher.go
```diff
@@ -20,9 +20,9 @@ import (
 )
 
 const (
-	reqSizeLimit    = 1024 * 1024
-	reqEventsLimit  = 10000
-	logWarnInterval = 1 * 24 * time.Hour
+	reqSizeLimit     = 1024 * 1024
+	reqEventsLimit   = 10000
+	warnOldTimeStamp = 1 * 24 * time.Hour
```
jefchien, line 25: minor nit: Still think threshold should be somewhere in the name, but it isn't blocking.

jefchien (approved):

sethAmazon (approved):

PR #889, opened 2023-10-04
Use GZIP 6 RPM Compression To Fix Install On AL2 Without Kernel Patch

sethAmazon:
# Description of the issue
Can't install rpm on the kernel patchless version of al2 

# Description of changes
Use Gzip 6 RPM Compression for rpm build spec
Current binary size 76mb new binary size 96mb 
Current install time 11 seconds new install time 5 seconds 

# License

# Tests
Manually build on al2 with kernel patch.
Install on al2 without kernel patch. 

# Requirements
2. Run `make lint`

### DIFF
--- a/packaging/linux/amazon-cloudwatch-agent.spec
+++ b/packaging/linux/amazon-cloudwatch-agent.spec
@@ -10,6 +10,8 @@ Source:     amazon-cloudwatch-agent.tar.gz
 
 %define _enable_debug_packages 0
 %define debug_package %{nil}
+%define _source_payload w6.gzdio
+%define _binary_payload w6.gzdio
 
 %prep
 %setup -c %{name}-%{version}


okankoAMZ:
What happens if the OS doesn't have Gzip0 ?

okankoAMZ:
nit: Do you know the installation times for al2? Can you check with `time` command

sethAmazon:
It installed almost instantly

okankoAMZ (dismissed):

okankoAMZ (dismissed):

packaging/linux/amazon-cloudwatch-agent.spec
```diff
@@ -10,6 +10,8 @@ Source:     amazon-cloudwatch-agent.tar.gz
 
 %define _enable_debug_packages 0
 %define debug_package %{nil}
+%define _source_payload w9.xzdio
```
sethAmazon, line 13: For testing, set this to 0, so build is faster bc compression takes a long time when set to 9

adam-mateen (approved):

packaging/linux/amazon-cloudwatch-agent.spec
```diff
@@ -10,6 +10,8 @@ Source:     amazon-cloudwatch-agent.tar.gz
 
 %define _enable_debug_packages 0
 %define debug_package %{nil}
+%define _source_payload w6.gzdio
+%define _binary_payload w6.gzdio
```
adam-mateen, line 14: overview still says gzip 0.

After evaluating xz compression and gzip compression with different compression levels we are choosing level 6 (which is the default)

Current binary size is X compared to before it was Y.
Current install time is X compared to before it was Y.

sethAmazon: Done

zhihonl (approved):

PR #915, opened 2023-10-20
EMF Exporter does not forward metrics if IRSA is enabled

dchappa:
# Description of the issue
If a customer is using EMF along with IRSA, customer metrics would not be forwarded to the configured account. 

# Description of changes
The EMF exporter translator was missing a statement to pass the RoleARN component from the agent configuration.

# License

# Tests
Tested on personal cluster, sucessfully forwarded logs/metrics the configured secondary account.

# Requirements
2. Run `make lint`


SaxyPandaBear (approved):

translator/translate/otel/exporter/awsemf/translator.go
```diff
@@ -93,6 +93,7 @@ func (t *translator) Translate(c *confmap.Conf) (component.Config, error) {
 	if credentialsFileKey, ok := agent.Global_Config.Credentials[agent.CredentialsFile_Key]; ok {
 		cfg.AWSSessionSettings.SharedCredentialsFile = []string{fmt.Sprintf("%v", credentialsFileKey)}
 	}
+	cfg.AWSSessionSettings.RoleARN = agent.Global_Config.Role_arn
```
adam-mateen, line 96: Can you add a unit or integration test for this?

SaxyPandaBear: I debated if it was worthwhile to add anything, but I think that adding a translation test here is pretty intentional and self documenting so I'm for adding a unit test. I don't know how we mechanize this so that we don't miss this in other exporters though.

dchappa: Added a simple unit test that checks for the roleARN and region, similar to other exporters. I think it is definitely worthwhile to add an integration test for this use case as well

adam-mateen (dismissed):

plugins/inputs/logfile/logfile_test.go
```diff
@@ -434,7 +434,7 @@ func createWriteRead(t *testing.T, prefix string, logFile *LogFile, done chan bo
 	defer close(evts)
 	// Choose a large enough number of lines so that even high-spec hosts will not
 	// complete receiving logEvents before the 2nd createWriteRead() goroutine begins.
-	const numLines int = 1000000
+	const numLines int = 10000
```
dchappa, line 437: Changed this to resolve macOS build issue

SaxyPandaBear: I think before, this was 100,000. Did you need to lower it down 2 orders of magnitude for the test to pass? @adam-mateen how long would it take to write an adequate integration test that validates the auto removal behavior?

SaxyPandaBear (approved):

plugins/inputs/logfile/logfile_test.go
```diff
@@ -453,7 +453,7 @@ func createWriteRead(t *testing.T, prefix string, logFile *LogFile, done chan bo
 		select {
 		case <-done2:
 			t.Log("Child completed before timeout (as expected)")
-		case <-time.After(time.Second * 10):
+		case <-time.After(time.Second * 20):
```
dchappa, line 456: MacOS build was failing, increased the time to timeout

SaxyPandaBear: @adam-mateen @dchappa can someone explain what increasing the timeout here means for the test? From context clues, we expect the goroutine to finish in < X seconds. How bad is a regression in this timing test? What is the likelihood that we have to come back to this value and update it? I think that testing that something completed in a unit test in X time is an addition of flakiness in testing.

SaxyPandaBear (approved):

plugins/inputs/logfile/logfile_test.go
adam-mateen: >// TestLogsFileAutoRemoval verifies when a new file matching the configured
// FilePath is discovered, the old file will be automatically deleted ONLY after
// being read to the end-of-file. Also verifies the new log file is discovered
// before finishing the old file.

In other words, the intent is to have a certain order of events:
1. create first file
2. write lots of data to first file
3. while "the agent" is reading first file, create a 2nd file. it is *important** that the agent is not done reading the first file.
4. verify the agent reports everything from the first file and then deletes it.

It is unlikely that we need to increase the timeout further. Computers generally get faster not slower. If anything we may get to the point there the computer is reading all the data from the first file before we create the 2nd file and write to it.

I am definitely open to rewrites here if you've got a better idea.

AFAIK this test was passing (not flaky) until my most recent change to it. S

adam-mateen (approved):

plugins/inputs/logfile/logfile_test.go
SaxyPandaBear: I don't think we can make it cleaner without rewriting the actual mechanism for tailing.. so it is what it is.

PR #919, opened 2023-10-20
Add agenthealth extension with user agent handler.

jefchien:
# Description of the issue
The agenthealth extension implements the `awsmiddleware.Extension` interface and provides the request/response handlers to configure on the AWS Clients. This allows the agent's user agent to be prepended on AWS requests.

# Description of changes
Adds the agenthealth extension scaffolding with a user agent handler. There's a singleton user agent that gets updated by its `Set*` functions. On each change, the handler is updated and on each requests, the user agent string is added.

# License
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

# Tests
Adds unit tests.

# Requirements
_Before commit the code, please do the following steps._
1. Run `make fmt` and `make fmt-sh`
2. Run `make lint`

PR #920, opened 2023-10-23
Add agent stats handler.

jefchien:
# Description of the issue
Depends on https://github.com/aws/amazon-cloudwatch-agent/pull/919.

In an effort to move towards using the agenthealth extension, we need to support the agent stats that the current `agentinfo` provides in a more generic manner.

# Description of changes
This is done in 3 stats providers:
- `stats/client`: The client stats handler measures per operation (e.g. ListBuckets) latency, payload size, and status codes.
- `stats/provider/process`: Provides metrics about the agent process on a 1-minute interval. It is a singleton. Only one thread can access it during that given interval. All others will not receive it.
- `stats/provider/flags`: Allows flags to be set in the stats header. This is on a 5-minute interval and also a singleton.

On each request that gets sent, the stats handler will send whatever header it has available for that operation. On each response, it will queue up a refresh of the header where it will grab the stats from each of the 3 providers and merge them together.

# License

# Tests
Added unit testing.

# Requirements
2. Run `make lint`


bryce-carey (approved):

extension/agenthealth/handler/stats/client/client.go
```diff
@@ -0,0 +1,117 @@
...
+	recorder, ok := csh.operationRecorders[operation]
+	if !ok {
+		return
+	}
+	recorder.stats = agent.Stats{
+		PayloadBytes: aws.Int(int(recorder.payloadBytes)),
+		StatusCode:   aws.Int(r.StatusCode),
+	}
+	if recorder.lastRequestID == csh.getRequestID(ctx) {
+		latency := time.Since(recorder.start).Milliseconds()
```
sky333999, line 108: Do we need to think about matching responses to the corresponding requests only? Trying to think what would happen if multiple requests from diff places called the Handler above.

extension/agenthealth/handler/stats/handler.go
```diff
@@ -0,0 +1,87 @@
...
+
+func (sh *statsHandler) HandleRequest(ctx context.Context, r *http.Request) {
+	header := sh.Header(awsmiddleware.GetOperationName(ctx))
+	if header != "" {
+		r.Header.Set(headerKeyAgentStats, header)
+	}
+}
+
+func (sh *statsHandler) HandleResponse(ctx context.Context, _ *http.Response) {
+	go sh.refreshHeader(awsmiddleware.GetOperationName(ctx))
```
sky333999, line 66: Do we call refresh on handling responses mainly for calculating latency?
Am I reading this write that we first need to have a response, which will populate the headers, that will then get set on the request handler (for subsequent request??) ?

extension/agenthealth/handler/stats/client/client.go
jefchien: If we don't match them to the request IDs, then if we have multiple requests inflight at the same time, the latency measurement could be off. Maybe that's not that big of a deal.

You have a good point though. It's possible that the request ID and start time could be replaced every time before the response comes back, so it might be better to have the request ID/start time pairs in a map or something. The stats would just get whichever is the latest to when the next request comes in.

extension/agenthealth/handler/stats/client/client.go
```diff
@@ -0,0 +1,117 @@
...
+	return handlerID
+}
+
+func (csh *clientStatsHandler) Position() awsmiddleware.HandlerPosition {
+	return awsmiddleware.Before
+}
+
+func (csh *clientStatsHandler) HandleRequest(ctx context.Context, r *http.Request) {
+	operation := csh.getOperationName(ctx)
+	if !csh.allowAllOperations && !csh.allowedOperations.Contains(operation) {
```
lisguo, line 77: is there a case where we don't allow all operations?

extension/agenthealth/handler/stats/client/client.go
```diff
@@ -0,0 +1,117 @@
...
+		return
+	}
+	csh.mu.Lock()
+	defer csh.mu.Unlock()
+	recorder, ok := csh.operationRecorders[operation]
```
lisguo, line 82: is it possible to have a race condition of two identical operations that overwrite each other?

extension/agenthealth/handler/stats/client/client.go
jefchien: Yeah one use case is in the awsemf exporter. We don't want or need to include CreateLogGroup.

extension/agenthealth/handler/stats/client/client.go
jefchien: The lock should prevent that, but it is possible. We don't necessarily need the stats of every single request, so if it gets overwritten by the next request, I think that's fine.

SaxyPandaBear (approved):

lisguo (approved):

extension/agenthealth/handler/stats/client/client.go
jefchien: Added a TTL cache.

lisguo (approved):

sky333999 (approved):

PR #929, opened 2023-10-25
Implement AppSignals on EKS and native EC2

lisguo:
# Description of the issue
This is the implementation for app signals within the cloudwatch agent. Includes config translation of opentelemetry components to enable collection of Latency, Error, and Fault metrics (through emf) and traces.

# Description of changes
- New awsappsignals processor
- Translation of `app_signals` under `logs.metrics_collected` and `traces.traces_collected`
- Creates otel pipeline with the otlp receiver ->  resourcedetection, awsappsignals processor -> emfexporter / xrayexporter

# License

# Tests
Created a dev build and tested on EKS and EC2.

# Requirements
2. Run `make lint`


mxiamxia (dismissed):
LGTM! Thanks!

plugins/processors/awsappsignals/customconfiguration/replacer.go
```diff
@@ -0,0 +1,54 @@
...
+			_, isExist := attributes.Get(targetDimensionKey)
+			if !isExist {
+				continue
+			}
+			// every replacement in one specific dimension only will be performed once
+			_, ok := finalRules[targetDimensionKey]
+			if ok {
+				continue
+			}
+			finalRules[targetDimensionKey] = replacement.Value
```
jefchien, line 46: nit: Why not just do the `attributes.PutStr` here? Instead of having the `finalRules` be a `map[string]string`, you can have it be a `collections.Set[string]` so it serves the same purpose.

plugins/processors/awsappsignals/factory.go
```diff
@@ -0,0 +1,96 @@
...
+		typeStr,
+		createDefaultConfig,
+		processor.WithTraces(createTracesProcessor, stability),
+		processor.WithMetrics(createMetricsProcessor, stability),
+	)
+}
+
+func createDefaultConfig() component.Config {
+	return &Config{
+		Resolvers: []string{"eks"}, // we need to change the default config later
```
jefchien, line 37: When is later? Should this be a `TODO`?

plugins/processors/awsappsignals/internal/normalizer/attributesnormalizer.go
```diff
@@ -0,0 +1,107 @@
...
+		return
+	}
+	for k, v := range copyMapForMetric {
+		if resourceAttrValue, ok := resourceAttributes.Get(k); ok {
+			// print some debug info when an attribute value is overwritten
+			if originalAttrValue, ok := attributes.Get(k); ok {
+				n.logger.Debug("attribute value is overwritten", zap.String("attribute", k), zap.String("original", originalAttrValue.AsString()), zap.String("new", resourceAttrValue.AsString()))
+			}
+			attributes.PutStr(v, resourceAttrValue.AsString())
+			if k == "k8s.pod.name" {
```
jefchien, line 83: nit: Can we move these strings into consts? They're used all over the place.

plugins/processors/awsappsignals/internal/resolver/attributesresolver.go
```diff
@@ -0,0 +1,86 @@
...
+		if err := subResolver.Process(attributes, resourceAttributes); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (r *attributesResolver) Stop(ctx context.Context) error {
+	for _, subResolver := range r.subResolvers {
+		if err := subResolver.Stop(ctx); err != nil {
```
jefchien, line 53: Doesn't this mean that if the first subResolver fails to stop, none of the others will be stopped? Can we use an `errors.Join` to just combine all the errors we get from trying to stop the subResolvers and then return that?

plugins/processors/awsappsignals/internal/resolver/eks.go
```diff
@@ -0,0 +1,739 @@
...
+			serviceAndNamespaceToSelectors: serviceWatcher.serviceAndNamespaceToSelectors,
+			ipToPod:                        podWatcher.ipToPod,
+			podToWorkloadAndNamespace:      podWatcher.podToWorkloadAndNamespace,
+			workloadAndNamespaceToLabels:   podWatcher.workloadAndNamespaceToLabels,
+			serviceToWorkload:              serviceToWorkload,
+			workloadPodCount:               podWatcher.workloadPodCount,
+			safeStopCh:                     safeStopCh,
+		}
+
+		go instance.debugPrint()
```
jefchien, line 550: Should this be enabled by default?

plugins/processors/awsappsignals/internal/resolver/eks.go
```diff
@@ -0,0 +1,739 @@
...
+			logger.Fatal("Failed to create config", zap.Error(err))
+		}
+
+		clientset, err := kubernetes.NewForConfig(config)
+		if err != nil {
+			logger.Fatal("Failed to create eks client", zap.Error(err))
+		}
+
+		// add a time jitter of 10 seconds
+		jitterSleep(10)
```
jefchien, line 515: What is the jitter for?

plugins/processors/awsappsignals/processor.go
```diff
@@ -0,0 +1,268 @@
...
+					ap.logger.Debug(failedToProcessAttribute, zap.Error(err))
+				}
+			}
+		}
+		dps.RemoveIf(func(d pmetric.NumberDataPoint) bool {
+			for _, Mutator := range ap.customAllowlistMutators {
+				shouldBeDropped, err := Mutator.ShouldBeDropped(d.Attributes(), resourceAttribes)
+				if err != nil {
+					ap.logger.Debug(failedToProcessAttributeWithCustomRule, zap.Error(err))
+					return true
```
jefchien, line 140: So if a rule fails, we don't run the rest of the mutators?

plugins/processors/awsappsignals/processor.go
```diff
@@ -0,0 +1,268 @@
...
+			for _, Mutator := range ap.metricMutators {
+				err := Mutator.Process(dps.At(i).Attributes(), resourceAttribes, false)
+				if err != nil {
+					ap.logger.Debug(failedToProcessAttribute, zap.Error(err))
+				}
+			}
+		}
+		dps.RemoveIf(func(d pmetric.NumberDataPoint) bool {
+			for _, Mutator := range ap.customAllowlistMutators {
+				shouldBeDropped, err := Mutator.ShouldBeDropped(d.Attributes(), resourceAttribes)
```
jefchien, line 137: nit: misspelling of `resourceAttributes`

plugins/processors/awsappsignals/processor.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+	return md, nil
+}
+
+// Attributes are provided for each log and trace, but not at the metric level
+// Need to process attributes for every data point within a metric.
+func (ap *awsappsignalsprocessor) processMetricAttributes(ctx context.Context, m pmetric.Metric, resourceAttribes pcommon.Map) {
+
+	// This is a lot of repeated code, but since there is no single parent superclass
+	// between metric data types, we can't use polymorphism.
```
jefchien, line 123: There isn't a single parent superclass, but you can still reduce the repeated code.

The functions within each `RemoveIf` are exactly the same. The metricsMutators for loop is exactly the same for each too.

plugins/processors/awsappsignals/processor.go
```diff
@@ -0,0 +1,268 @@
...
+			return false
+		})
+		for i := 0; i < dps.Len(); i++ {
+			err := ap.customReplacer.Process(dps.At(i).Attributes(), resourceAttribes, false)
+			if err != nil {
+				ap.logger.Debug(failedToProcessAttribute, zap.Error(err))
+			}
+		}
+	default:
+		ap.logger.Info("Ignore unknown metric type", zap.String("type", m.Type().String()))
```
jefchien, line 266: nit: Should this be an info level log? It seems more like either debug or warn depending on if you want to surface it.

translator/translate/otel/pipeline/appsignals/translator.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	if conf == nil || !conf.IsSet(configKey) {
+		return nil, &common.MissingKeyError{ID: t.ID(), JsonKey: configKey}
+	}
+
+	translators := &common.ComponentTranslators{
+		Receivers:  common.NewTranslatorMap(otlp.NewTranslatorWithName(common.AppSignals, otlp.WithDataType(t.dataType))),
+		Processors: common.NewTranslatorMap(resourcedetection.NewTranslator(resourcedetection.WithDataType(t.dataType)), awsappsignals.NewTranslator(awsappsignals.WithDataType(t.dataType))),
+		Exporters:  common.NewTranslatorMap[component.Config](),
+		Extensions: common.NewTranslatorMap[component.Config](),
```
jefchien, line 50: Should add the agenthealth extension based on the dataType.

plugins/processors/awsappsignals/processor.go
lisguo: Yes, because we will drop the metric entirely if we fail to process the metric attribute. There is no point in reviewing other mutators if the attribute cannot be read.

translator/translate/otel/pipeline/appsignals/translator.go
lisguo: ack

plugins/processors/awsappsignals/processor.go
lisguo: ack. changing to debug

plugins/processors/awsappsignals/processor.go
lisguo: fixed

plugins/processors/awsappsignals/processor.go
```diff
@@ -0,0 +1,268 @@
...
+	stoppers                []stopper
+}
+
+func (ap *awsappsignalsprocessor) Start(_ context.Context, _ component.Host) error {
+	attributesResolver := resolver.NewAttributesResolver(ap.config.Resolvers, ap.logger)
+	ap.stoppers = append(ap.stoppers, attributesResolver)
+	ap.metricMutators = append(ap.metricMutators, attributesResolver)
+
+	attributesNormalizer := normalizer.NewAttributesNormalizer(ap.logger)
+	ap.metricMutators = append(ap.metricMutators, attributesNormalizer)
```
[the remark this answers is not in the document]
lisguo: fixed

plugins/processors/awsappsignals/processor.go
```diff
@@ -0,0 +1,268 @@
...
+	"go.uber.org/zap"
+
+	"github.com/aws/amazon-cloudwatch-agent/plugins/processors/awsappsignals/customconfiguration"
+	"github.com/aws/amazon-cloudwatch-agent/plugins/processors/awsappsignals/internal/normalizer"
+	"github.com/aws/amazon-cloudwatch-agent/plugins/processors/awsappsignals/internal/resolver"
+)
+
+const (
+	failedToProcessAttribute               = "failed to Process attributes"
+	failedToProcessAttributeWithCustomRule = "failed to process attributes with custom rule, will drop it"
```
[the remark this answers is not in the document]
lisguo: drop the metric. added to log

plugins/processors/awsappsignals/internal/resolver/eks.go
lisguo: not sure, @thpierce / @nanzhenAWS might know?

plugins/processors/awsappsignals/internal/resolver/eks.go
lisguo: probably not. @thpierce / @nanzhenAWS can we remove this and implement debug lines as part of the logger instead?

plugins/processors/awsappsignals/internal/resolver/attributesresolver.go
lisguo: changed to the following

```
func (r *attributesResolver) Stop(ctx context.Context) error {
	var errs error
	for _, subResolver := range r.subResolvers {
		if err := subResolver.Stop(ctx); err != nil {
			errs = errors.Join(errs, err)
		}
	}
	return errs
}
```

plugins/processors/awsappsignals/processor.go
jefchien: It looks like `isSelected` is the only place where we can get an error and will only return an error if the specified attribute doesn't exist. So, if I configured the agent to drop `attribute-I-don't-want` and that attribute wasn't present on a metric, it would still drop it?

plugins/processors/awsappsignals/config.go
```diff
@@ -0,0 +1,18 @@
...
+	"github.com/aws/amazon-cloudwatch-agent/plugins/processors/awsappsignals/customconfiguration"
+)
+
+type Config struct {
+	Resolvers []string                   `mapstructure:"resolvers"`
+	Rules     []customconfiguration.Rule `mapstructure:"rules"`
+}
+
+func (cfg *Config) Validate() error {
+	// TODO: validate those mandatory fields (if exist) in the config
```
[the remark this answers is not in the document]
lisguo: yes it's still a todo. @thpierce, will this be added later?

plugins/processors/awsappsignals/README.md
```diff
@@ -0,0 +1,130 @@
+# AWS AppSignals Processor for Amazon Cloudwatch Agent
+
+The AWS AppSignals processor is used to reduce the cardinality of telemtry metrics and traces before exporting them to CloudWatch Logs via [EMF](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/awsemfexporter) and [X-Ray](github.com/open-telemetry/opentelemetry-collector-contrib/exporter/awsxrayexporter) respectively.
+It reduces the cardinality of metrics/traces via 3 types of actions, `keep`, `drop` and `replace`, which are configured by users. CloudWatch Agent(CWA) customers will configure these rules with their CWA configurations, which will get translated to yaml.
```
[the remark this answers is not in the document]
lisguo: removed translation part

plugins/processors/awsappsignals/processor.go
lisguo: You have a point. I'm looking at the code as well and it appears if `isSelected` returns an error. The intent is that we will keep the metric. But in the caller we return true for `shouldBeDropped`. Should we just be returning `shouldBeDropped`?

```
		isMatched, err := isSelected(attributes, element.SelectorMatchers, false)
		if isMatched {
			// The datapoint will be dropped if one of keep rule matched
			return true, nil
		}
		if err != nil {
			// The datapoint will be kept if error is occurred in match process
			return false, err
		}
```

plugins/processors/awsappsignals/config.go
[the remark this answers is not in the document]
pxaws: I don't think we have any mandatory fields. So this TODO can be removed.

plugins/processors/awsappsignals/internal/resolver/eks.go
pxaws: Yes

plugins/processors/awsappsignals/internal/resolver/eks.go
pxaws: Just to add a jitter to avoid the case while all instances of cloudwatch agent will call the kubernetes control plane at the same time. The time here is just an arbitrary value. We can probably add an TODO for future optimization. But for now, I don't think we want to do extensive testing to figure out a optimal value.

plugins/processors/awsappsignals/config.go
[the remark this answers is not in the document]
bjrara: Validation doesn't only check if any mandatory field is missing, it's also about whether the values customer set there are legal/valid. Is there any validation we want to put for the later?

plugins/processors/awsappsignals/customconfiguration/common.go
```diff
@@ -0,0 +1,116 @@
...
+		selectorMatcherItem := SelectorMatcherItem{
+			selector.Dimension,
+			glob.MustCompile(selector.Match),
+		}
+		selectorMatchers = append(selectorMatchers, selectorMatcherItem)
+	}
+	return selectorMatchers
+}
+
+func generateTestAttributes(service string, operation string, remoteService string, remoteOperation string,
```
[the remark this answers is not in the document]
pxaws: agree

plugins/processors/awsappsignals/internal/resolver/eks.go
pxaws: This debug print is to give a full picture about the various mappings for debug purposes. This should be disabled by default.

plugins/processors/awsappsignals/config.go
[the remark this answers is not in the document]
jefchien: Like mentioned above, could we validate the globs so if they're invalid they are discovered at translation time?

plugins/processors/awsappsignals/customconfiguration/common.go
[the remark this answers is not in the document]
bjrara: moved it to common_test.go

plugins/processors/awsappsignals/customconfiguration/dropper.go
```diff
@@ -0,0 +1,35 @@
...
+
+func NewCustomDropper(rules []Rule) *DropActions {
+	return &DropActions{
+		Actions: generateActionDetails(rules, "drop"),
+	}
+}
+
+func (d *DropActions) ShouldBeDropped(attributes, _ pcommon.Map) (bool, error) {
+	// nothing will be dropped if no rule is defined
+	if d.Actions == nil || len(d.Actions) == 0 {
```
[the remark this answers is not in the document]
bjrara: updated

plugins/processors/awsappsignals/customconfiguration/dropper.go
```diff
@@ -0,0 +1,35 @@
...
+		return false, nil
+	}
+	for _, element := range d.Actions {
+		isMatched, err := isSelected(attributes, element.SelectorMatchers, false)
+		if isMatched {
+			// The datapoint will be dropped if one of keep rule matched
```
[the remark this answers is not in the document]
bjrara: updated the comment

plugins/processors/awsappsignals/customconfiguration/keeper.go
```diff
@@ -0,0 +1,35 @@
...
+
+package customconfiguration
+
+import "go.opentelemetry.io/collector/pdata/pcommon"
+
+type KeepActions struct {
+	Actions []ActionItem
+}
+
+func NewCustomKeeper(rules []Rule) *KeepActions {
```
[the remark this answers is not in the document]
bjrara: renamed.

plugins/processors/awsappsignals/processor.go
```diff
@@ -0,0 +1,268 @@
...
+	for i := 0; i < rss.Len(); i++ {
+		rs := rss.At(i)
+		ilss := rs.ScopeSpans()
+		resourceAttributes := rs.Resource().Attributes()
+		for j := 0; j < ilss.Len(); j++ {
+			ils := ilss.At(j)
+			spans := ils.Spans()
+			for k := 0; k < spans.Len(); k++ {
+				span := spans.At(k)
+				for _, Mutator := range ap.traceMutators {
```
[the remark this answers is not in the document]
bjrara: updated

plugins/processors/awsappsignals/customconfiguration/common.go
```diff
@@ -0,0 +1,116 @@
...
+
+type Replacement struct {
+	TargetDimension string `mapstructure:"target_dimension"`
+	Value           string `mapstructure:"value"`
+}
+
+type Rule struct {
+	Selectors    []Selector    `mapstructure:"selectors"`
+	Replacements []Replacement `mapstructure:"replacements,omitempty"`
+	Action       string        `mapstructure:"action"`
```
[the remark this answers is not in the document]
bjrara: done.

plugins/processors/awsappsignals/customconfiguration/replacer.go
bjrara: To me, it's more of a personal coding style choice. I actually prefer the original version cuz it would make the actual writing to original `attributes` happen in only one place.

Unless there's a strong reason to change it, I'll keep what it is. :)

plugins/processors/awsappsignals/processor.go
bjrara: Good catch. I changed it to log the error if exists, and respect the original decision of `shouldBeDropped`.

plugins/processors/awsappsignals/processor.go
bjrara: They **look** the same, but they don't. The datapoints are of different types.

plugins/processors/awsappsignals/README.md
jefchien, line 1: Move `plugins/processors/awsappsignals` to `processor/awsappsignals`. `plugins` is a legacy directory for telegraf. Going forward, all new components should be top level like `receiver/adapter` and `extension/agenthealth`

plugins/processors/awsappsignals/internal/resolver/attributesresolver.go
jefchien: nit: Could do 

```go
func (r *attributesResolver) Stop(ctx context.Context) error {
	var errs error
	for _, subResolver := range r.subResolvers {
        errs = errors.Join(errs, subResolver.Stop(ctx))
	}
	return errs
}
```

translator/config/schema.json
```diff
@@ -812,6 +888,11 @@
         "traces_collected": {
           "type": "object",
           "properties": {
+            "app_signals": {
+              "type": "object",
+              "properties": {},
+              "additionalProperties": true
```
jefchien, line 894: nit: If this should always be empty, then consider setting `additionalProperties` to false, so the schema validation will fail. This way there's less confusion about the configuration.

translator/translate/otel/exporter/awsxray/translator.go
```diff
@@ -4,6 +4,7 @@
 package awsxray
 
 import (
+	_ "embed"
```
jefchien, line 7: This is unused.

jefchien (approved):

JayPolanco (approved):

plugins/processors/awsappsignals/README.md
lisguo: I see, I put this in plugins since I saw the other processors in there. If we do this then we should add the /processor directory to be unit tested

plugins/processors/awsappsignals/processor.go
jefchien: You can use interfaces/generics to get them to behave the same.

```go
type DataPoints[T HasAttributes] interface {
    At(i int) T
    Len() int
}

type HasAttributes interface {
    Attributes() pcommon.Map
}

func getAttributes[T HasAttributes](dps DataPoints[T]) []pcommon.Map {
    var attributes []pcommon.Map
    for i := 0; i < dps.Len(); i++ {
       attributes = append(attributes, dps.At(i).Attributes())
    }
    return attributes
}

func Test(m pmetric.Metric) {
   switch m.Type() {
      case pmetric.MetricTypeGauge:
           getAttributes(m.Gauge().DataPoints())
   }
}
```

PR #945, opened 2023-11-03
Fix empty measurement field causing agent crash

zhihonl:
# Description of the issue
CloudWatch agent team released major version v1.3x which migrates to OpenTelemetry. Old agent with major version 1.2x silently passes when measurement field is empty because the plugins with empty measurement field will be removed during translation, therefore not emitting anything for that metric.

New version will fail the agent during startup because the OTEL translators will still include the plugins with empty measurement field in the YAML config but the component will not be recognized.

# Description of changes
To achieve backward compatibility and avoid breaking existing customers config that may be using the empty measurement field, we will allow empty measurement but print out a warning message to remind the customers.

# License

# Tests
Unit tests passed. Also manually tested on EC2 instance using the following config
```
{
    "agent": {
        "debug": true,
        "aws_sdk_log_level": "LogDebugWithHTTPBody"
    },
    "metrics": {
        "append_dimensions": {
        },
        "metrics_collected": {
            "cpu": {
                "measurement": ["cpu_usage_idle", "cpu_usage_iowait", "cpu_usage_user", "cpu_usage_system"],
                "metrics_collection_interval": 1,
                "resources": ["*"],
                "totalcpu": true
            },
            "diskio": {
                "measurement": [],
                "resources": ["*"],
                "append_dimensions": {}
            }
        }
    }
}
```
Screenshot 2023-11-03 at 2 01 17 PM

The agent was able to start and the CloudWatch console only receives the CPU metrics. 
**Note**: The gap between the metric was when I uninstalled the agent for another test build

# Requirements
2. Run `make lint`

### DIFF
--- a/translator/translate/otel/receiver/adapter/translators.go
+++ b/translator/translate/otel/receiver/adapter/translators.go
@@ -152,6 +152,9 @@ func fromInputs(conf *confmap.Conf, validInputs map[string]bool, baseKey string)
 			if skipInputSet.Contains(inputName) {
 				// logs agent is separate from otel agent
 				continue
+			} else if measurement := common.GetArray[any](conf, common.ConfigKey(cfgKey, common.MeasurementKey)); measurement != nil && len(measurement) == 0 {
+				log.Printf("W! Agent will not emit any metrics for %s due to empty measurement field ", inputName)
+				continue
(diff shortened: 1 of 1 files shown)


sethAmazon (approved):

ymtaye (approved):
LGTM!

PR #962, opened 2023-11-16
Upgrade Go SDK To Add Feature To Turn Off IMDSV1 Via ENV VAR

sethAmazon:
# Description of the issue
Could not turn on imdsv1 without code configuration of the agent. 

# Description of changes
Allow user to disable imdsv1 via an env var in the sdk. Upgrade the go sdk to a version that supports this feature. 

# License

# Tests
N/A

# Requirements
2. Run `make lint`


okankoAMZ (approved):

ZhenyuTan-amz:
Can we add a test to ensure v1 is disabled?

ZhenyuTan-amz (approved):
LGTM

zhihonl:
Did we test using the environment variable that is supposed to disable IMDS V1? On Linux, the agent only read from systemd environment, so I assumed just doing `export` would not work and the customer has to manually modify the systemd setting.

sethAmazon:
So I did test this against a preview version. You need to change the /etc/environment file

PR #965, opened 2023-11-22
Add test case for awsappsignals processor

bjrara:
# Description of the issue
Add unit test for awsappsignals processor

# Description of changes
To avoid bug issue like https://github.com/aws/amazon-cloudwatch-agent/pull/946.

# License

# Tests

# Requirements
2. Run `make lint`

### DIFF
--- a/plugins/processors/awsappsignals/processor_test.go
+++ b/plugins/processors/awsappsignals/processor_test.go
@@ -1,0 +1,239 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+package awsappsignals
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"go.opentelemetry.io/collector/pdata/pmetric"
+	"go.opentelemetry.io/collector/pdata/ptrace"
+	"go.uber.org/zap"
+
+	"github.com/aws/amazon-cloudwatch-agent/plugins/processors/awsappsignals/rules"
+)
+
+var testRules = []rules.Rule{
+	{
+		Selectors: []rules.Selector{
+			{
+				Dimension: "dim_action",
+				Match:     "reserved",
+			},
+			{
+				Dimension: "dim_val",
+				Match:     "test1",
+			},
+		},
+		Replacements: []rules.Replacement{
+			{
+				TargetDimension: "dim_val",
+				Value:           "test2",
+			},
+		},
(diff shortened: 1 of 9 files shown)


bjrara:
@mxiamxia @lisguo

mxiamxia (approved):
LGTM! Thanks

pxaws (approved):

plugins/processors/awsappsignals/processor_test.go
```diff
@@ -0,0 +1,239 @@
...
+		if agreedValue != newVal {
+			t.Errorf("inconsistent dimension value, agreed value is %s, new %s\n", agreedValue, newVal)
+		}
+	}
+	return agreedValue
+}
+
+func isMetricNil(m pmetric.Metrics) bool {
+	dps := m.ResourceMetrics().At(0).ScopeMetrics().At(0).Metrics().At(0).Gauge().DataPoints()
+	if dps.Len() > 0 {
```
lisguo, line 215: wont dps.Len() be the same for all of the if conditions since we are never re-assigning `dps`?

plugins/processors/awsappsignals/processor_test.go
```diff
@@ -0,0 +1,239 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"go.opentelemetry.io/collector/pdata/pmetric"
+	"go.opentelemetry.io/collector/pdata/ptrace"
+	"go.uber.org/zap"
+
+	"github.com/aws/amazon-cloudwatch-agent/plugins/processors/awsappsignals/rules"
+)
+
+var testRules = []rules.Rule{
```
lisguo, line 18: is it worth having multiple test cases where we test different sets of rules configurations?

ie. Having keep but no replace, drop. Having replace, but not keep, drop...etc.

plugins/processors/awsappsignals/processor_test.go
bjrara: Thanks for pointing it out. Updated.

plugins/processors/awsappsignals/processor_test.go
bjrara: In the tests of `keeper`, `dropper` and `replacer`, we have cases to verify the correct rule extraction given the component a set of mixed rules.

What's missing from those tests is the verification of whether the rules are correctly passed down the 3 components, which is covered in the processor. Let me know if there are other scenarios we want to cover by different sets of rules, and I can add more.

movence (approved):

PR #988, opened 2023-12-07
Removes LogGroupClass parameter in CreateLogGroup request when customer config does not define parameter

dchappa:
# Description of the issue
In regions where the LogGroupClass parameter is not yet deployed, customers are unable to create LogGroups, even with valid configurations.

# Description of changes
When a customer defines a config without a `log_group_class`, the agent will no longer send a CreateLogGroup request with a `STANDARD` log group class. Instead, the CreateLogGroup request will be made without the `LogGroupClass` parameter, which the CWL Service also accepts as a `STANDARD` Log Group Class 

# License

# Tests
On us-gov-west-1, I deployed the following config
```
{
  "agent": {
    "run_as_user": "root",
    "debug": true
  },
  "logs": {
    "logs_collected": {
      "files": {
        "collect_list": [
          {
            "file_path": "/var/gazpacho-standard.log",
            "log_group_name": "{instance_id}-standard",
            "log_stream_name": "{instance_id}",
            "timezone": "UTC"
          }
        ]
      }
    }
  }
}
```
Previously, it would fail with the error, `2023-12-07T16:49:03Z D! [outputs.cloudwatchlogs] creating group fail due to : InvalidParameterException: Only Standard log class is supported.`

Now it succeeds in log group creation
image

Setting the `log_group_class` parameter will continue to fail in these regions until this CW Logs feature is available in these regions.

I also verified that the fix continues to work in commercial regions
image

# Requirements
2. Run `make lint`

### DIFF
--- a/translator/defaultKeyCase.go
+++ b/translator/defaultKeyCase.go
@@ -93,9 +93,9 @@ func DefaultLogGroupClassCase(key string, defaultVal, input interface{}) (return
 		//CreateLogGroup API only accepts values STANDARD or INFREQUENT_ACCESS
 		returnVal = strings.ToUpper(classVal)
 	} else {
-		AddErrorMessages(
+		AddInfoMessages(
 			fmt.Sprintf("LogGroupClass key: %s", key),
-			fmt.Sprintf("%s value (%v) in is not a valid Log Group Class field. Defaulting to standard.", key, returnVal))
+			fmt.Sprintf("%s value (%v) in is not a valid Log Group Class field. Agent will not set the LogGroupClass parameter.", key, returnVal))
 		returnVal = ""
 	}
 	return
--- a/translator/defaultKeyCase.go
+++ b/translator/defaultKeyCase.go
@@ -6,8 +6,6 @@ package translator
 import (
 	"fmt"
 	"strings"
-
-	"github.com/aws/amazon-cloudwatch-agent/tool/util"
 )
 
 // DefaultCase check if current input overrides the default value for the given config entry key.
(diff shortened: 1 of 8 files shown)


translator/defaultKeyCase.go
```diff
@@ -95,10 +93,10 @@ func DefaultLogGroupClassCase(key string, defaultVal, input interface{}) (return
 		//CreateLogGroup API only accepts values STANDARD or INFREQUENT_ACCESS
 		returnVal = strings.ToUpper(classVal)
 	} else {
-		returnVal = util.StandardLogGroupClass
 		AddErrorMessages(
 			fmt.Sprintf("LogGroupClass key: %s", key),
 			fmt.Sprintf("%s value (%v) in is not a valid Log Group Class field. Defaulting to standard.", key, returnVal))
```
nathalapooja, line 98: We should change this log which says defaulting to standard and change it to Defaulting to empty

translator/isValid.go
```diff
@@ -94,5 +94,5 @@ func IsValidRetentionDays(days int) bool {
 }
 
 func IsValidLogGroupClass(class string) bool {
-	return slices.Contains(ValidLogGroupClasses, class)
+	return slices.Contains(ValidLogGroupClasses, class) || class == ""
```
nathalapooja, line 97: We might not need this change as this returns false for class =="" which is handled in else section setting the returnVal to ""

translator/defaultKeyCase.go
dchappa: Yes so the agent wouldn't be sending a request with `LogGroupClass` parameter in this instance, but thinking from a customer perspective, the log group would be created with the default value, which is Standard. If a customer were to go through logs, it might cause more confusion if they see that it's empty

translator/isValid.go
dchappa: After discussion, determined that it should not return errors, will change log level

sky333999 (approved):

movence (approved):

PR #1000, opened 2023-12-26
Add metrics cardinality control for Application Signals

bjrara:
# Description of the issue
Due to the nature of metrics generated from auto-instrumentation, high cardinality metrics may occur if attributes which are intended to have low cardinality, are assigned values with high cardinality.

# Description of changes
A new component `metric_limiter` is introduced into `awsappsignalsprocessor` to enhance cardinality control. Customers can set `drop_threshold` to indicate the anticipated number of low cardinality metrics for each `service`. The local metrics limiter will prioritize exporting the most crucial metrics (e.g. most frequently used, manually retained) up to that threshold. Metrics identified as high cardinality will be rolled up, and only the number of observations will be reported.

# License

# Tests
1. Unit tests.
2. E2E tests.
3. Performance tests.

# Requirements
4. Run `make lint`

### DIFF
--- a/plugins/processors/awsappsignals/internal/cardinalitycontrol/metrics_limiter.go
+++ b/plugins/processors/awsappsignals/internal/cardinalitycontrol/metrics_limiter.go
@@ -1,0 +1,414 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+package cardinalitycontrol
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"strings"
+	"sync"
+	"time"
+
+	"go.opentelemetry.io/collector/pdata/pcommon"
+	"go.uber.org/zap"
+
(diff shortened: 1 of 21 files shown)


plugins/processors/awsappsignals/internal/cardinalitycontrol/metrics_limiter.go
```diff
@@ -0,0 +1,414 @@
...
+		s.secondaryCMS.Insert(md)
+		updatedFrequency := s.secondaryCMS.Get(md)
+		updatedMd := copyMetricDataWithUpdatedFrequency(md, updatedFrequency)
+		s.secondaryTopK.Push(md, updatedMd)
+	}
+}
+
+// MetricData represents a key-value pair.
+type MetricData struct {
+	labels    map[string]string
```
mxiamxia, line 214: not blocker: we can remove `labels` field in MetricData but just keep `hashKey` to reduce the memory usage. As MetricData will be stored as value for every service TopKMetrics map.

PR #1004, opened 2024-01-09
implement eksdetector

JayPolanco:
# Description of the issue
Currently, the CWA is unable to distinguish between EKS and native Kubernetes. 

# Description of changes
This change implements an EKS detector which can determine if the agent is on an EKS cluster or native K8s. It does this by checking for the aws-auth configmap in the kube-system namespace, which is only present if it's an EKS cluster.

# License

# Tests
Unit  tests.
Not being used anywhere currently. Pending confirmation of k8s configurations

# Requirements
2. Run `make lint`

### DIFF
--- a/translator/translate/otel/common/eksdetector.go
+++ b/translator/translate/otel/common/eksdetector.go
@@ -63,7 +63,7 @@ func IsEKS() (bool, error) {
 	// Make HTTP GET request
 	awsAuth, err := eksDetector.getConfigMap(authConfigNamespace, authConfigConfigMap)
 	if err != nil {
-		return false, err
+		return false, nil
 	}
 
 	return awsAuth != nil, nil
--- a/translator/translate/otel/common/eksdetector.go
+++ b/translator/translate/otel/common/eksdetector.go
@@ -1,0 +1,91 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+package common
+
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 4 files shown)


translator/translate/otel/common/eksdetector.go
```diff
@@ -0,0 +1,91 @@
...
+var (
+	detectorSingleton Detector
+	once              sync.Once
+)
+
+var (
+	getInClusterConfig  = func() (*rest.Config, error) { return rest.InClusterConfig() }
+	getKubernetesClient = func(confs *rest.Config) (kubernetes.Interface, error) { return kubernetes.NewForConfig(confs) }
+	// NewDetector creates a new singleton detector for EKS
+	NewDetector = func() (Detector, error) {
```
JayPolanco, line 40: This was done to make unit testing the singleton easier and it's exported for unit testing in the translators

translator/translate/otel/common/eksdetector.go
```diff
@@ -0,0 +1,91 @@
...
+			}
+		})
+
+		return detectorSingleton, errors
+	}
+)
+
+// IsEKS checks if the agent is running on EKS. This is done by using the kubernetes API to determine if the aws-auth
+// configmap exists in the kube-system namespace
+func IsEKS(eksDetector Detector) bool {
```
lisguo, line 56: Is it possible to have this just be `IsEKS()` so that we do not need to initialize a detector when it is just going to be a singleton anyways? 

I assume the first call of `IsEKS()` will initialize the singleton, and then future calls with re-use it

JayPolanco: Yup, was following the pattern of the resourcedetection processor, but this works better

lisguo (approved):
you might have to rerun the failed checks

mitali-salvi (approved):

PR #1014, opened 2024-01-19
Add metrics cardinality control for Application Signals

mxiamxia:
This PR is an updated version from https://github.com/aws/amazon-cloudwatch-agent/pull/1000, Since @bjrara is not available atm so I create this new PR with a few changes below. 
1. Removed Metrics labels referenece in `MetricData` structure
2. Disabled Metric Limiter by default until AppSignals console team completed the related changes for supporting E2E CX.
# Description of the issue
Due to the nature of metrics generated from auto-instrumentation, high cardinality metrics may occur if attributes which are intended to have low cardinality, are assigned values with high cardinality.

# Description of changes
A new component `metric_limiter` is introduced into `awsappsignalsprocessor` to enhance cardinality control. Customers can set `drop_threshold` to indicate the anticipated number of low cardinality metrics for each `service`. The local metrics limiter will prioritize exporting the most crucial metrics (e.g. most frequently used, manually retained) up to that threshold. Metrics identified as high cardinality will be rolled up, and only the number of observations will be reported.

# License

# Tests
1. Unit tests.
3. E2E tests.
4. Performance tests.

# Requirements
5. Run `make lint`

### DIFF
--- a/plugins/processors/awsappsignals/common/types.go
+++ b/plugins/processors/awsappsignals/common/types.go
@@ -1,0 +1,27 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+package common
+
+const (
+	AttributeRemoteService       = "aws.remote.service"
+	AttributeHostedInEnvironment = "aws.hostedin.environment"
+)
+
+const (
+	MetricAttributeRemoteNamespace = "K8s.RemoteNamespace"
+	MetricAttributeLocalService    = "Service"
+	MetricAttributeLocalOperation  = "Operation"
+	MetricAttributeRemoteService   = "RemoteService"
+	MetricAttributeRemoteOperation = "RemoteOperation"
+	MetricAttributeRemoteTarget    = "RemoteTarget"
+)
+const (
+	HostedInAttributeClusterName  = "HostedIn.EKS.Cluster"
+	HostedInAttributeK8SNamespace = "HostedIn.K8s.Namespace"
+	HostedInAttributeEnvironment  = "HostedIn.Environment"
+)
+
+const (
+	AttributeTmpReserved = "aws.tmp.reserved"
+)
--- a/plugins/processors/awsappsignals/config/config.go
+++ b/plugins/processors/awsappsignals/config/config.go
@@ -4,14 +4,47 @@
 package config
 
 import (
+	"context"
 	"errors"
+	"time"
 
 	"github.com/aws/amazon-cloudwatch-agent/plugins/processors/awsappsignals/rules"
 )
 
 type Config struct {
-	Resolvers []Resolver   `mapstructure:"resolvers"`
-	Rules     []rules.Rule `mapstructure:"rules"`
+	Resolvers []Resolver     `mapstructure:"resolvers"`
+	Rules     []rules.Rule   `mapstructure:"rules"`
+	Limiter   *LimiterConfig `mapstructure:"limiter"`
+}
+
+type LimiterConfig struct {
+	Threshold                 int             `mapstructure:"drop_threshold"`
+	Disabled                  bool            `mapstructure:"disabled"`
+	LogDroppedMetrics         bool            `mapstructure:"log_dropped_metrics"`
+	RotationInterval          time.Duration   `mapstructure:"rotation_interval"`
+	GarbageCollectionInterval time.Duration   `mapstructure:"-"`
+	ParentContext             context.Context `mapstructure:"-"`
+}
+
+const (
+	DefaultThreshold        = 500
+	DefaultRotationInterval = 1 * time.Hour
+	DefaultGCInterval       = 10 * time.Minute
+)
+
+func NewDefaultLimiterConfig() *LimiterConfig {
+	return &LimiterConfig{
+		Threshold:         DefaultThreshold,
+		Disabled:          true,
+		LogDroppedMetrics: false,
+		RotationInterval:  DefaultRotationInterval,
+	}
+}
+
+func (lc *LimiterConfig) Validate() {
+	if lc.GarbageCollectionInterval == 0 {
+		lc.GarbageCollectionInterval = DefaultGCInterval
+	}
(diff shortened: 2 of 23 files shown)


plugins/processors/awsappsignals/rules/keeper.go
```diff
@@ -3,15 +3,21 @@
...
+import (
+	"go.opentelemetry.io/collector/pdata/pcommon"
+
+	"github.com/aws/amazon-cloudwatch-agent/plugins/processors/awsappsignals/common"
+)
 
 type KeepActions struct {
-	Actions []ActionItem
+	Actions                 []ActionItem
+	markDataPointAsReserved bool
```
lisguo, line 14: Can you remind me what `markDataPointAsReserved` does? 

I see that we pass in `!limiterConfig.Disabled` to this value... not sure I completely understand

plugins/processors/awsappsignals/internal/cardinalitycontrol/count_min_sketch.go
```diff
@@ -0,0 +1,98 @@
...
+	for i := 0; i < cms.depth; i++ {
+		hashFunc := cms.hashFuncs[i]
+		hashValue := hashFunc(obj.HashKey())
+		pos := int(hashValue % int64(cms.width))
+
+		cms.matrix[i][pos] += obj.Frequency()
+	}
+}
+
+func NewCountMinSketch(depth, width int, hashFuncs ...CountMinSketchHashFunc) *CountMinSketch {
```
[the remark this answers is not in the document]
mxiamxia: It is being used under `MetircLimiter.go::newService()` function. so we need to create a new CMS probability structure for a new individual micro-service. :)

plugins/processors/awsappsignals/rules/keeper.go
mxiamxia: Yeah, it is to make sure the metrics that `Replaced` or `Kept` by the custom rules will not be aggregated by MetricLimiter which happens later.  

So if MetricLimiter is enabled, we will inject an placeholder `markDataPointAsReserved` attribute to Metrics, and the MetricLimiter will not process such metrics later. If MetricLimiter is not enabled in CWAgent, we just don't do anything.

plugins/processors/awsappsignals/config/config.go
```diff
@@ -4,14 +4,47 @@
...
+	Limiter   *LimiterConfig `mapstructure:"limiter"`
+}
+
+type LimiterConfig struct {
+	Threshold                 int             `mapstructure:"drop_threshold"`
+	Disabled                  bool            `mapstructure:"disabled"`
+	LogDroppedMetrics         bool            `mapstructure:"log_dropped_metrics"`
+	RotationInterval          time.Duration   `mapstructure:"rotation_interval"`
+	GarbageCollectionInterval time.Duration   `mapstructure:"-"`
+	ParentContext             context.Context `mapstructure:"-"`
```
lisguo, line 26: Can you describe a use case where we will need to use the parent context?

plugins/processors/awsappsignals/internal/cardinalitycontrol/metrics_limiter.go
```diff
@@ -0,0 +1,415 @@
...
+			svc = newService(serviceName, m.DropThreshold, m.RotationInterval, m.ctx, m.logger)
+			m.services[serviceName] = svc
+		}
+		m.mapLock.Unlock()
+	}
+
+	metricData := newMetricData(serviceName, metricName, labels)
+
+	reserved, _ := attributes.Get(common.AttributeTmpReserved)
+	if reserved.Bool() {
```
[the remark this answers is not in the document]
mxiamxia: Yes, in the case of MetricLimiter is enabled, but no metric being matched from Keeper or Replacer rules (or no rule being defined). These metrics will have `reserved.Bool()` is false.

plugins/processors/awsappsignals/config/config.go
mxiamxia: Yes, we inherit the `ParentContext` to make sure the following new GoRoutine can be properly returned. 

```
func NewMetricsLimiter(config *config.LimiterConfig, logger *zap.Logger) Limiter {
	logger.Info("creating metrics limiter with config", zap.Any("config", config))

	ctx := config.ParentContext
	if ctx == nil {
		ctx = context.TODO()
	}

	limiter := &MetricsLimiter{
		DropThreshold:     config.Threshold,
		LogDroppedMetrics: config.LogDroppedMetrics,
		RotationInterval:  config.RotationInterval,

		logger:   logger,
		ctx:      ctx,
		services: map[string]*service{},
	}

	go func() {
		for {
			select {
			case <-ctx.Done():
				return
			default:
				limiter.removeStaleServices()
				time.Sleep(config.GarbageCollectionInterval)
			}
		}
	}()
```

mxiamxia:
Rebased!

pxaws (approved):

plugins/processors/awsappsignals/common/types.go
```diff
@@ -0,0 +1,27 @@
...
+	MetricAttributeLocalOperation  = "Operation"
+	MetricAttributeRemoteService   = "RemoteService"
+	MetricAttributeRemoteOperation = "RemoteOperation"
+	MetricAttributeRemoteTarget    = "RemoteTarget"
+)
+const (
+	HostedInAttributeClusterName  = "HostedIn.EKS.Cluster"
+	HostedInAttributeK8SNamespace = "HostedIn.K8s.Namespace"
+	HostedInAttributeEnvironment  = "HostedIn.Environment"
+)
```
majanjua-amzn, line 23: Do we not also want `HostedIn.K8s.Cluster` now?

majanjua-amzn (approved):

plugins/processors/awsappsignals/common/types.go
majanjua-amzn: Followed up offline, team will take a followup action to add any attributes needed for new platform support

PR #1018, opened 2024-01-26
add application log group info for AppSignals

pxaws:
# Description of changes
* set resource attribute `aws.log.group.names` to be the application log group name for Container Insights use case so that later we can correlate application logs with traces

# License

# Tests
`make test`

# Requirements
2. Run `make lint`

### DIFF
--- a/plugins/processors/awsappsignals/internal/resolver/kubernetes.go
+++ b/plugins/processors/awsappsignals/internal/resolver/kubernetes.go
@@ -685,6 +685,12 @@ func (h *kubernetesHostedInAttributeResolver) Process(attributes, resourceAttrib
 		attributes.PutStr(attr.HostedInClusterNameK8s, h.clusterName)
 	}
 
+	//The application log group in Container Insights is a fixed pattern:
+	// "/aws/containerinsights/{Cluster_Name}/application"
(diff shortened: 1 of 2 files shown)


plugins/processors/awsappsignals/internal/resolver/kubernetes.go
```diff
@@ -685,6 +685,11 @@ func (h *kubernetesHostedInAttributeResolver) Process(attributes, resourceAttrib
 		attributes.PutStr(attr.HostedInClusterNameK8s, h.clusterName)
 	}
 
+	//The application log group in Container Insights is a fixed pattern:
+	// "/aws/containerinsights/{Cluster_Name}/application"
```
straussb, line 689: Since we don't have this as a shared constant in code anywhere, could you add a link to where it's defined in the operator?  I suppose that would be https://github.com/aws/amazon-cloudwatch-agent-operator/blob/fe144bb02d7b1930715aa3ea32e57a5ff13406aa/helm/templates/fluent-bit-configmap.yaml#L82 .

straussb (approved):

plugins/processors/awsappsignals/internal/resolver/kubernetes.go
pxaws: Sure,  I will add it.

straussb (approved):

mxiamxia (approved):

PR #1026, opened 2024-02-01
[AppSignals][Bug Fix] Ensure resource detection processor is translated for EC2

JayPolanco:
# Description of the issue
E2E EC2 integration tests were failing for appsignals. This was caused by the translation not including the resourcedetection processor for EC2 route.

# Description of changes
- Adds logic to include the resourcedetection translation if on EKS or EC2 (Not included if on native K8s)
- Fixes Unit Testing that missed the bug
- Modified xray exporter translation to include the correct attributes depending if on EKS, K8s or EC2

# License

# Tests
unit tests, passes integ tests

# Requirements
2. Run `make lint`

### DIFF
--- a/translator/translate/otel/exporter/awsxray/translator.go
+++ b/translator/translate/otel/exporter/awsxray/translator.go
@@ -68,11 +68,15 @@ func (t *translator) Translate(conf *confmap.Conf) (component.Config, error) {
 	cfg := t.factory.CreateDefaultConfig().(*awsxrayexporter.Config)
 
 	if isAppSignals(conf) {
-		isEks := common.IsEKS()
-		if isEks.Value {
-			cfg.IndexedAttributes = indexedAttributesEKS
+		if common.IsAppSignalsKubernetes() {
+			isEks := common.IsEKS()
+			if isEks.Value {
+				cfg.IndexedAttributes = indexedAttributesEKS
+			} else {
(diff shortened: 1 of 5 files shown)


translator/translate/otel/exporter/awsxray/translator.go
```diff
@@ -68,11 +68,15 @@ func (t *translator) Translate(conf *confmap.Conf) (component.Config, error) {
...
+		if common.IsAppSignalsKubernetes() {
+			isEks := common.IsEKS()
+			if isEks.Value {
+				cfg.IndexedAttributes = indexedAttributesEKS
+			} else {
+				cfg.IndexedAttributes = indexedAttributesK8s
+			}
 		} else {
-			cfg.IndexedAttributes = indexedAttributesK8s
+			cfg.IndexedAttributes = indexedAttributesEKS
```
lisguo, line 79: prob this will need to be changed to EC2 in the future?

lisguo (approved):

okankoAMZ (approved):
nit: can you link the integ-test that passed in testing so we can refer back to it if we need to

translator/translate/otel/exporter/awsxray/translator.go
JayPolanco: I would assume so, this just keeps the current behavior

PR #1028, opened 2024-02-06
[Application Signals] [TLS] Translate CWAgent Receiver JSON Config to take in TLS settings

okankoAMZ:
# Description of the issue
Currently CWAgent Receiver  does not support TLS settings. We want to allow customers to provide TLS settings by default to promote better cybersecurity. 

# Description of changes
* Modified the receiver to parse JSON and add TLS settings.
* Modified schema.json to accept TLS Settings
* Modified json and yaml test configs to test the new TLS Settings 
* Added new units tests to receiver translator

>  [!IMPORTANT]
>  Modified  `translator/cmdutil/translatorutil.go`  to remove key-value pairs labeled as `REDACTED`. This change is essential because OTLP generates these pairs regardless of whether they are defined in JSON. As a result, they are always present in YAML, and when OTLP attempts to utilize these `REDACTED` keys (as they are defined in YAML), it leads to a crash. To prevent this, I've adjusted the `translatorutil` to parse and delete any key-value pairs with a `REDACTED` value. However, this adjustment implies that we _now **exclusively** accept certificates via a path and not embedded within JSON._

> [!IMPORTANT]
> Added unique name for OTLP receiver in traces and metric setting. This is needed because currently both traces and metric OTLP receiver has the same name, and in the case that there are one trace and one metric OTLP receiver, the traces OTLP receiver will take precedence and overwrite the metric OTLP receiver. by [zhihonl](https://github.com/zhihonl)

# License

# Tests
Ran `make test` and all tests passed. 
# Requirements
2. Run `make lint`


translator/cmdutil/translatorutil.go
```diff
@@ -224,7 +225,32 @@ func TranslateJsonMapToYamlConfig(jsonConfigValue interface{}) (interface{}, err
 	if err = conf.Marshal(cfg); err != nil {
 		return nil, err
 	}
-	return conf.ToStringMap(), nil
+	strMap := conf.ToStringMap()
+	RemoveTLSRedacted(strMap)
+	return strMap, nil
+}
+
+// TODO: Switch to a custom translator approach at some point
```
lisguo, line 233: nit: not sure how helpful this is -- would it be better to have a short comment on what we are doing instead?

translator/cmdutil/translatorutil.go
```diff
@@ -224,7 +225,32 @@ func TranslateJsonMapToYamlConfig(jsonConfigValue interface{}) (interface{}, err
...
+func RemoveTLSRedacted(stringMap map[string]interface{}) {
```
lisguo, line 234: Curious -- how did we handle this for the metrics forwarding tls configurations previously? Did it not crash then?

translator/translate/otel/receiver/otlp/translator.go
```diff
@@ -16,13 +18,16 @@ import (
 )
 
 const (
-	defaultGrpcEndpoint = "127.0.0.1:4317"
-	defaultHttpEndpoint = "127.0.0.1:4318"
+	defaultGrpcEndpoint          = "127.0.0.1:4317"
+	defaultHttpEndpoint          = "127.0.0.1:4318"
+	defaultAppSignalGrpcEndpoint = "0.0.0.0:4315"
```
lisguo, line 23: Super nit: AppSignals should be plural

translator/translate/otel/receiver/otlp/translator.go
```diff
@@ -56,18 +62,20 @@ func WithDataType(dataType component.DataType) Option {
...
-	for _, opt := range opts {
-		opt.apply(t)
-	}
-	return t
+	return NewTranslatorWithName("", opts...)
 }
 
 func NewTranslatorWithName(name string, opts ...Option) common.Translator[component.Config] {
-	t := &translator{name: name, factory: otlpreceiver.NewFactory()}
+	t := &translator{name: name, index: -1, factory: otlpreceiver.NewFactory()}
```
lisguo, line 69: what does the index do here?

translator/translate/otel/receiver/otlp/translator.go
```diff
@@ -77,26 +85,47 @@ func (t *translator) ID() component.ID {
 
 func (t *translator) Translate(conf *confmap.Conf) (component.Config, error) {
 	cfg := t.factory.CreateDefaultConfig().(*otlpreceiver.Config)
-
-	// TODO: Should follow pattern done in awsemf and awsexray exporter translations (i.e should be integrated with standard otlp translation)
+	// init default configuration
```
lisguo, line 88: thank you, glad to see this TODO completed

lisguo (commented):
Thank you for the detailed description

```
However, this adjustment implies that we now exclusively accept certificates via a path and not embedded within JSON.
```
I think this is okay. We shouldn't be putting secrets embedded in the JSON anyways.

translator/translate/otel/receiver/otlp/translator.go
okankoAMZ: The index allows us to distinguish between different otlp receivers. In this case, if you specify a fixed name to your receiver there is no need to track it with a index.

lisguo: oh so it's really just used for otlp/1, otlp/2..etc..? Should we just have it be instanceName or something? seems misleading

okankoAMZ: You can have multiple otlp running on the same instance. I'd rather have a hash or something instead of InstanceName

lisguo: yeah it just seems confusing. I'd rather have it called `instanceNum` or something to refer it to the instance name.

lisguo (dismissed):

translator/translate/otel/receiver/otlp/translator.go
```diff
@@ -77,26 +85,48 @@ func (t *translator) ID() component.ID {
...
-	if endpoint, ok := common.GetString(conf, common.ConfigKey(configKey, "grpc_endpoint")); ok {
-		cfg.GRPC.NetAddr.Endpoint = endpoint
+
+	var otlpKeyMap map[string]interface{}
+	if otlpSlice := common.GetArray[any](conf, configKey); t.index != -1 && len(otlpSlice) > t.index {
+		otlpKeyMap = otlpSlice[t.index].(map[string]interface{})
+	} else {
+		otlpKeyMap = conf.Get(configKey).(map[string]interface{})
+	}
+	if t.name == common.AppSignals {
```
sky333999, line 112: There is some conditional logic for AppSignals further up already. Can we keep it all in one place?

translator/translate/otel/receiver/otlp/translator.go
```diff
@@ -77,26 +85,48 @@ func (t *translator) ID() component.ID {
...
 	cfg := t.factory.CreateDefaultConfig().(*otlpreceiver.Config)
-
-	// TODO: Should follow pattern done in awsemf and awsexray exporter translations (i.e should be integrated with standard otlp translation)
+	// init default configuration
+	configKeyExt := common.OtlpKey
+	cfg.GRPC.NetAddr.Endpoint = defaultGrpcEndpoint
+	cfg.HTTP.Endpoint = defaultHttpEndpoint
 	if t.name == common.AppSignals {
-		return common.GetYamlFileToYamlConfig(cfg, appSignalsConfig)
+		configKeyExt = common.AppSignals
```
sky333999, line 93: Is `appSignalsConfig` and `appsignals_config.yaml` no longer being used?

translator/cmdutil/translatorutil.go
```diff
@@ -224,7 +225,31 @@ func TranslateJsonMapToYamlConfig(jsonConfigValue interface{}) (interface{}, err
...
+}
+
+func RemoveTLSRedacted(stringMap map[string]interface{}) {
+	type Node struct {
+		isTLSParent bool
+		data        map[string]interface{}
+	}
+	root := Node{isTLSParent: false, data: stringMap}
+	queue := []Node{root}
+	// Using BFS search through string Map and find sub settings of TLS
```
sky333999, line 240: Nice!

sky333999: Mind adding a quick unit test for this func?

translator/config/schema.json
```diff
@@ -987,6 +1010,9 @@
             "http_endpoint": {
               "description": "HTTP endpoint to use to listen for OTLP JSON traces",
               "$ref": "#/definitions/endpointOverrideDefinition"
+            },
+            "tls": {
```
sky333999, line 1014: So this is under `otlpDefinition` which is for `otlp` as per [this](https://github.com/aws/amazon-cloudwatch-agent/pull/1028/files#diff-b316c7f1304a4a8f2e1707d12dce4eebd0d9cf1c9c3add74cf31b01bfcdd5299R939). How does it work for your when its under `app_signals` thats [here](https://github.com/aws/amazon-cloudwatch-agent/pull/1028/files#diff-b316c7f1304a4a8f2e1707d12dce4eebd0d9cf1c9c3add74cf31b01bfcdd5299R930)?

translator/cmdutil/translatorutil.go
okankoAMZ: The other unit test actually cover this case because, the config test will fail if there is a `REDACTED`

translator/config/schema.json
okankoAMZ: Doesn't app_signals inherit otlp definitions? I may have misunderstood that, if thats not the case I can move the tls reference to under app_signals

translator/translate/otel/receiver/otlp/translator.go
okankoAMZ: yes, I'll remove it

translator/config/schema.json
sky333999: Can you check why none of your testing or unit tests failed with this bug?
And can you add a test to prevent that from happening?

translator/translate/otel/receiver/otlp/translator.go
```diff
@@ -56,18 +59,20 @@ func WithDataType(dataType component.DataType) Option {
...
 
 func NewTranslatorWithName(name string, opts ...Option) common.Translator[component.Config] {
-	t := &translator{name: name, factory: otlpreceiver.NewFactory()}
+	t := &translator{name: name, instanceNum: -1, factory: otlpreceiver.NewFactory()}
 	for _, opt := range opts {
 		opt.apply(t)
 	}
+	if name == "" && t.dataType != "" {
+		t.name = string(t.dataType)
+		if t.instanceNum != -1 {
```
sky333999, line 72: I see instanceNum being set to -1 above but when is it not that?
Do we have a test to ensure having multiple otlp sections in the same config works as you are expecting it to here?

okankoAMZ: Added the unit test

translator/translate/otel/receiver/otlp/translator_test.go
```diff
@@ -52,6 +52,25 @@ func TestTracesTranslator(t *testing.T) {
...
+						"cert_file": "path/to/cert.crt",
+						"key_file":  "path/to/key.key",
+					},
+				}},
+			wantErr: &common.MissingKeyError{
+				ID:      tt.ID(),
+				JsonKey: common.ConfigKey(common.TracesKey, common.TracesCollectedKey, common.OtlpKey),
+			},
+		},
 		"WithCompleteConfig": {
```
ymtaye, line 74: Nit: Do you think it would be better to add the tls changes to the testdata configs?

lisguo (approved):
Approved - but as a sanity check can we at least manually test that we can translate appsignals with and without the tls config correctly before merging? That would give more confidence in the change.

ymtaye (approved):
LGTM!

translator/translate/otel/pipeline/host/translator.go
```diff
@@ -42,7 +43,23 @@ func (t translator) ID() component.ID {
 func (t translator) Translate(conf *confmap.Conf) (*common.ComponentTranslators, error) {
 	if conf == nil || !conf.IsSet(common.MetricsKey) {
 		return nil, &common.MissingKeyError{ID: t.ID(), JsonKey: common.MetricsKey}
-	} else if t.receivers.Len() == 0 {
+	}
+
+	hostReceivers := t.receivers
+	if common.PipelineNameHost == t.name {
+		switch v := conf.Get(common.ConfigKey(common.MetricsKey, common.MetricsCollectedKey, common.OtlpKey)).(type) {
```
sky333999, line 50: What is all of this logic for?

translator/translate/otel/receiver/otlp/translator.go
```diff
@@ -16,23 +18,24 @@ import (
...
+	defaultHttpEndpoint           = "127.0.0.1:4318"
+	defaultAppSignalsGrpcEndpoint = "0.0.0.0:4315"
+	defaultAppSignalsHttpEndpoint = "0.0.0.0:4316"
 )
 
 var (
 	configKeys = map[component.DataType]string{
-		component.DataTypeTraces: common.ConfigKey(common.TracesKey, common.TracesCollectedKey, common.OtlpKey),
+		component.DataTypeTraces:  common.ConfigKey(common.TracesKey, common.TracesCollectedKey),
+		component.DataTypeMetrics: common.ConfigKey(common.LogsKey, common.MetricsCollectedKey),
```
sky333999, line 30: Im confused why we introduced `logs -> metrics_collected -> otlp` support in this PR.

PR #1052, opened 2024-02-21
Add support for new nvidia driver output format/schema (v12)

adam-mateen:
# Description of the issue
Latest nvidia driver versions, e.g. `535.129.03` output the `power_draw` metric in a different format and the agent's input plugin fails to parse it.

# Description of changes
Copy the latest input plugin (nvidia_smi) code from the influxdata/telegraf package to amazon-cloudwatch-agent.
Also copy some pieces of the influxtelegraf/telegraf/internal package.

# License

# Tests
- manual testing on `g5.xlarge` instance type with nvidia driver version `535.154.05`:
- (ignore gaps when the agent was stopped, just take note that the latest timestamps show metric values)
Screenshot 2024-02-21 at 4 45 14 PM

- manual testing on `g5.2xlarge` instance type with nvidia driver version `510.47.03`:
(New agent binary started at `Thu Feb 22 13:45:33 UTC 2024`)
Screenshot 2024-02-22 at 7 54 12 AM

# Requirements
2. Run `make lint`


internal/exec_windows.go
ZhenyuTan-amz, line 1: Double checking that we won't introduce Windows support by adding this here? Windows support is still in Appsec review.

plugins/inputs/nvidia_smi/README.md
```diff
@@ -0,0 +1,155 @@
...
+
+  ## Optional: timeout for GPU polling
+  # timeout = "5s"
+```
+
+### Linux
+
+On Linux, `nvidia-smi` is generally located at `/usr/bin/nvidia-smi`
+
+### Windows
```
ZhenyuTan-amz, line 40: Same as the other comment, probably should remove windows related comments.

adam-mateen: I prefer to minimize differences with upstream. Also consider customers do not enable this plugin directly, they have to configure CWAgent using the JSON config. I think the JSON config translate into TOML and YAML and will ignore this input plugin for Windows.

internal/exec_windows.go
adam-mateen: this is just a generic function that can be used to exec any other program. It is not actually used on Windows.

plugins/inputs/nvidia_smi/README.md
adam-mateen: I recommend keeping the README.md file the same as upstream.

zhihonl (approved):

ZhenyuTan-amz (approved):

PR #1058, opened 2024-02-27
Add kubernetes modes and use eksdetector on startup

lisguo:
# Description of the issue
Adds new kubernetes modes that is set on startup

```
        ModeEKS       = "EKS"
	ModeK8sEC2    = "K8sEC2"
	ModeK8sOnPrem = "K8sOnPrem"
```

Create kubernetesMode variable that is accessible via the context to be used in translation. The shortMode also gets overridden if we are on kubernetes.

If customers are on EKS:
kubernetesMode = "EKS"
mode = "EC2"

if customers are on k8s on prem:
kubernetesMode = K8sOnPrem
mode = "onPrem"

# Description of changes
- move the eksdetector from common to it's own package in translator -> util
- Uses the eksdetector on startup to determine if we are on eks, native k8s on ec2, or native k8s on prem.
- Create KubernetesMode() to access the kubernetes mode which is separate from the original mode
- Refactored some translation logic to use the new mods instead of the eksdetector

# License

# Tests
added unit tests

# Requirements
2. Run `make lint`


translator/context/context.go
```diff
@@ -127,6 +127,15 @@ func (ctx *Context) SetMode(mode string) {
...
+		ctx.mode = config.ModeEKS
+		ctx.shortMode = config.ShortModeEKS
+	case config.ModeK8sEC2:
+		ctx.mode = config.ModeK8sEC2
+		ctx.shortMode = config.ShortModeK8sEC2
+	case config.ModeK8sOnPrem:
+		ctx.mode = config.ModeK8sOnPrem
+		ctx.shortMode = config.ShortModeK8sOnPrem
 	default:
 		log.Panicf("Invalid mode %s. Valid mode values are %s, %s, %s and %s.", mode, config.ModeEC2, config.ModeOnPrem, config.ModeOnPremise, config.ModeWithIRSA)
```
lisguo, line 140: forgot to add the new modes to the log...will add in revision

translator/config/mode.go
```diff
@@ -8,10 +8,16 @@ const (
 	ModeOnPrem    = "onPrem"
 	ModeOnPremise = "onPremise"
 	ModeWithIRSA  = "withIRSA"
+	ModeEKS       = "EKS"
+	ModeK8sEC2    = "K8sEC2"
+	ModeK8sOnPrem = "K8sOnPrem"
```
jefchien, line 13: We have a few places in the code where we depend entirely on the mode being `onPrem`/`onPremise` and I'm concerned that this will cause a regression.

One option is to have a function that checks if the mode is an `onPrem` mode and include the `K8sOnPrem` as part of that and replace the existing checks.

translator/context/context.go
```diff
@@ -127,8 +127,17 @@ func (ctx *Context) SetMode(mode string) {
...
+		ctx.shortMode = config.ShortModeEKS
+	case config.ModeK8sEC2:
+		ctx.mode = config.ModeK8sEC2
+		ctx.shortMode = config.ShortModeK8sEC2
+	case config.ModeK8sOnPrem:
+		ctx.mode = config.ModeK8sOnPrem
+		ctx.shortMode = config.ShortModeK8sOnPrem
 	default:
-		log.Panicf("Invalid mode %s. Valid mode values are %s, %s, %s and %s.", mode, config.ModeEC2, config.ModeOnPrem, config.ModeOnPremise, config.ModeWithIRSA)
+		log.Panicf("Invalid mode %s. Valid mode values are %s, %s, %s, %s, %s, %s, and %s.", mode, config.ModeEC2, config.ModeOnPrem, config.ModeOnPremise, config.ModeWithIRSA, config.ModeEKS, config.ModeK8sEC2, config.ModeK8sOnPrem)
```
jefchien, line 140: I don't think we should be listing out all the modes. Don't think this can be reached, but if it were logged, it would give the wrong impression to that these new modes can be set with `-m`.

translator/context/context.go
```diff
@@ -128,7 +133,23 @@ func (ctx *Context) SetMode(mode string) {
...
-		log.Panicf("Invalid mode %s. Valid mode values are %s, %s, %s and %s.", mode, config.ModeEC2, config.ModeOnPrem, config.ModeOnPremise, config.ModeWithIRSA)
+		log.Panicf("Invalid mode %s. Valid mode values are %s, %s, %s, and %s.", mode, config.ModeEC2, config.ModeOnPrem, config.ModeOnPremise, config.ModeWithIRSA)
+	}
+}
+
+func (ctx *Context) SetKubernetesMode(mode string) {
+	switch mode {
+	case config.ModeEKS:
+		ctx.kubernetesMode = config.ModeEKS
+		ctx.shortMode = config.ShortModeEKS
```
sky333999, line 144: So in this case, the shortMode will be `EKS` but the mode would be `EC2` - would that lead to any issues?

lisguo: Yeah, I spoke to Jeffrey and he mentioned we only use shortMode for our AgentHealth metrics

sky333999: Got it. Not a blocker but perhaps we should just rename this to userAgentMode or something.
mode & shortMode sets an implicit expectation that they will be in sync and we might add future usages of this shortMode without realizing it gets overwritten here.

jefchien (dismissed):

translator/util/sdkutil.go
```diff
@@ -27,6 +28,7 @@ var DetectRegion = detectRegion
 var DetectCredentialsPath = detectCredentialsPath
 var DefaultEC2Region = defaultEC2Region
 var DefaultECSRegion = defaultECSRegion
+var IsEKS = isEKS
```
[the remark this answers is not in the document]
lisguo: yeah I did this so that I can override the function for our unit tests to return different detectors

translator/context/context.go
sky333999: Alternatively, can we set both shortMode & kubernetesMode on the header so we dont do this override here?
Not sure if thats going too far to prevent this.

sky333999 (dismissed):

jefchien (approved):

translator/context/context.go
jefchien: That would require changes to the frontend.

sky333999 (approved):

jefchien (approved):

okankoAMZ (approved):
Looks good to me!

Paramadon (approved):
lgtm

PR #1060, opened 2024-02-28
[AppSignals] Update components translators to set missing fields

sky333999:
# Description of the issue
We are currently not setting all the required fields for our components during translation that are required to work on envs when IMDS isnt available.

# Description of changes
* Consistently set fields on our components.
* Remove deviation in awsemfexporter translator for APM to make it consistent with remaining plugins.

# License

# Tests
* Unit tests

# Requirements
2. Run `make lint`

### DIFF
--- a/translator/translate/otel/exporter/awsemf/translator.go
+++ b/translator/translate/otel/exporter/awsemf/translator.go
@@ -78,22 +78,10 @@ func (t *translator) Translate(c *confmap.Conf) (component.Config, error) {
 	cfg := t.factory.CreateDefaultConfig().(*awsemfexporter.Config)
 	cfg.MiddlewareID = &agenthealth.LogsID
 
-	if common.IsAppSignalsKubernetes() && t.name == common.AppSignals {
-		isEks := common.IsEKS()
-		if isEks.Value {
-			return common.GetYamlFileToYamlConfig(cfg, appSignalsConfigEks)
-		}
-		return common.GetYamlFileToYamlConfig(cfg, appSignalsConfigK8s)
-	} else if t.name == common.AppSignals {
-		ctx := context.CurrentContext()
-		if ctx.Mode() == config.ModeEC2 {
-			return common.GetYamlFileToYamlConfig(cfg, appSignalsConfigEC2)
-		}
-		return common.GetYamlFileToYamlConfig(cfg, appSignalsConfigGeneric)
-	}
-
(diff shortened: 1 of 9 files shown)


lisguo (dismissed):

translator/translate/otel/exporter/awsemf/translator.go
```diff
@@ -157,6 +171,10 @@ func isPrometheus(conf *confmap.Conf) bool {
 	return conf.IsSet(prometheusBasePathKey)
 }
 
+func setAppSignalsFields(_ *confmap.Conf, _ *awsemfexporter.Config) error {
```
jefchien, line 174: Why do we need this function?

translator/translate/otel/exporter/awsemf/translator_test.go
```diff
@@ -727,7 +792,61 @@ func TestTranslateAppSignals(t *testing.T) {
 						"app_signals": map[string]any{},
 					},
 				}},
-			want:           testutil.GetConf(t, filepath.Join("appsignals_config_k8s.yaml")),
+			want: map[string]any{
```
jefchien, line 795: nit: Instead of redefining all the fields in the file, we should be able to use `conf.Merge` to add the dynamic fields.

jefchien (approved):

translator/translate/otel/exporter/awsemf/translator.go
sky333999: I originally had `setDisableMetricExtraction(ecsBasePathKey, conf, cfg)` and then removed it as I wasnt sure if it disabling metrics extraction from EMF logs should be an option for APM.

Thought Id just leave this for anything we might add in the future.

More importantly, in the place this is being called, I need to short circuit that one way or another. The ordering there kinda matters and it shouldnt hit isKubernetes when translating the emfexporter for appsignals - since its common for both app_signals and kubernetes to exist in the same config. And the k8s one will overwrite with unintended changes for the appsignals emfexporter section.

If someone does change that logic, our unit tests will complain.

translator/translate/otel/exporter/awsemf/translator_test.go
sky333999: Ah yeah :(
Lemme see if I can use that.

jefchien (approved):

PR #1062, opened 2024-02-28
Do not update RemoteService dimension on the to be aggregated dependency metrics

mxiamxia:
# Description of changes
Most of  high cardinality metrics case in AppSignals. the high cardinality values will come from Local and Remote Operation dimensions. and we do want to keep some of the key dimensions if it will not be in the high cardinality situation. So in this changes, we will keep the original dimension value of RemoteService that was previously overrode by MetricLimiter, but change RemoteOperation dimension value to `AllOtherRemoteOperations`

The new MetricLimiter overall aggregation rule, we don't aggregate the dimension of `LocalService`, `RemoteService` and `HostedIn_*` dimensions . We will replace `LocalOperation` to `AllOtherOperations` and `RemoteOperation` dimension into `AllOtherRemoteOperations` . The rest of other dimension values  into "-"

# License

# Tests
unit test

# Requirements
2. Run `make lint`

### DIFF
--- a/plugins/processors/awsappsignals/internal/cardinalitycontrol/metrics_limiter.go
+++ b/plugins/processors/awsappsignals/internal/cardinalitycontrol/metrics_limiter.go
@@ -19,8 +19,8 @@ import (
 )
 
 const (
-	UnprocessedServiceOperationValue = "AllOtherOperations"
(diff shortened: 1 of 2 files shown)


majanjua-amzn (approved):
LGTM

jefchien (approved):

PR #1066, opened 2024-02-29
[AppSignals] Add hostname, ec2 instance id, ec2 ASG attributes to emf log for kubernetes

lisguo:
# Description of the issue
AppSignals on EKS and Native Kubernetes on EC2 were missing the following attributes:
```
host.name
EC2.InstanceId
EC2.AutoScalingGroupName
```

AppSignals on Native Kubernetes on OnPrem instances were missing just `host.name`

# Description of changes
- Modified awsappsignals processor to read the attributes accordingly from the resource detection processor in the kubernetes resolver
- Modified emfexporter to pass in the attributes

# License

# Tests
unit tests

# Requirements
2. Run `make lint`

### DIFF
--- a/translator/tocwconfig/sampleConfig/appsignals_and_eks_config.yaml
+++ b/translator/tocwconfig/sampleConfig/appsignals_and_eks_config.yaml
@@ -28,12 +28,6 @@ exporters:
                     - aws.span.kind
                   regex: ^(SERVER|LOCAL_ROOT)$
                   separator: ;
-                - label_names:
-                    - EC2.InstanceId
-                    - EC2.AutoScalingGroupName
-                    - host.name
-                  regex: .*
-                  separator: ;
               metric_name_selectors:
                 - Latency
                 - Fault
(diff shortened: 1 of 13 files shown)


translator/tocwconfig/sampleConfig/appsignals_and_eks_config.yaml
```diff
@@ -28,6 +28,12 @@ exporters:
                     - aws.span.kind
                   regex: ^(SERVER|LOCAL_ROOT)$
                   separator: ;
+                - label_names:
+                    - EC2.InstanceId
+                    - EC2.AutoScalingGroupName
+                    - host.name
+                  regex: .*
```
mxiamxia, line 35: I trying to recall why do we need to add these labels here? it is not being used as metric dimensions which should be auto-populated as EMF regular attributes in my mind. 
In this change, we have `regex: .*`, what if otel raw metric data doesn't have   `EC2.AutoScalingGroupName` existed.

mxiamxia: I am pretty sure now that we should not have these rules under `label_mathers`, these rules are used to for extracting certain metrics from the matched label values. We don't actually need it for our cases. :)

```
- label_names:
    - EC2.InstanceId
    - EC2.AutoScalingGroupName
    - host.name
   regex: .*
```

lisguo: got it, yeah I was thinking about this too. We also have k8s attributes that didn't have to be added to the emf config. Probably just the processor change was enough. Will remove

movence (approved):

PR #1093, opened 2024-03-15
[AppSignals] Explicitly set the GCCollectionInterval for MetricsLimiter

sky333999:
# Description of the issue
When not explicitly set, the current validate logic doesnt seem to get invoked before the start of the component, so it doesnt respect the default GC collection interval causing an infinite loop leading to CPU maxing out.

# Description of changes
Explicitly set the GC collection interval when initializing the config.

# License

# Tests
Manually tested on an EKS cluster that CPU spike goes away.

# Requirements
2. Run `make lint`

### DIFF
--- a/plugins/processors/awsappsignals/config/config.go
+++ b/plugins/processors/awsappsignals/config/config.go
@@ -22,7 +22,7 @@ type LimiterConfig struct {
 	Disabled                  bool            `mapstructure:"disabled"`
 	LogDroppedMetrics         bool            `mapstructure:"log_dropped_metrics"`
 	RotationInterval          time.Duration   `mapstructure:"rotation_interval"`
-	GarbageCollectionInterval time.Duration   `mapstructure:"-"`
+	GarbageCollectionInterval time.Duration   `mapstructure:"garbage_collection_interval"`
 	ParentContext             context.Context `mapstructure:"-"`
 }
 
--- a/plugins/processors/awsappsignals/config/config.go
+++ b/plugins/processors/awsappsignals/config/config.go
@@ -34,10 +34,11 @@ const (
 
 func NewDefaultLimiterConfig() *LimiterConfig {
 	return &LimiterConfig{
-		Threshold:         DefaultThreshold,
-		Disabled:          false,
-		LogDroppedMetrics: false,
-		RotationInterval:  DefaultRotationInterval,
+		Threshold:                 DefaultThreshold,
+		Disabled:                  false,
+		LogDroppedMetrics:         false,
+		RotationInterval:          DefaultRotationInterval,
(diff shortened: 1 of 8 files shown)


jefchien (dismissed):

lisguo (dismissed):

plugins/processors/awsappsignals/config/config.go
```diff
@@ -34,10 +34,11 @@ const (
...
 	return &LimiterConfig{
-		Threshold:         DefaultThreshold,
-		Disabled:          false,
-		LogDroppedMetrics: false,
-		RotationInterval:  DefaultRotationInterval,
+		Threshold:                 DefaultThreshold,
+		Disabled:                  false,
+		LogDroppedMetrics:         false,
+		RotationInterval:          DefaultRotationInterval,
+		GarbageCollectionInterval: DefaultGCInterval,
```
bjrara, line 41: Hmmm, interesting, I thought`GarbageCollectionInterval` will be assigned with `DefaultGCInterval` in the `Validate()` if no custom value is provided.

Does the test shows that after assigning a default value, the CPU usage drops?

mxiamxia (approved):

translator/translate/otel/processor/awsappsignals/testdata/invalidRulesConfig.json
```diff
@@ -6,7 +6,8 @@
         "limiter": {
           "drop_threshold": 20,
           "log_dropped_metrics": true,
-          "rotation_interval": "10m"
+          "rotation_interval": "10m",
+          "garbage_collection_interval": "10m"
```
jefchien, line 10: Was this added to the schema.json?

lisguo: yeah, also not sure if this was meant to be configurable by customers

sky333999: Ppl can use it even if its not in our schema afaik depending on additonalPros..
None of the metric limiter props are in our schema.json - we should revisit.
For now, I just added to all the places `rotation_interval` is present.

plugins/processors/awsappsignals/config/config.go
jefchien: Do you mean the config `Validate()` function? That's called by the OTEL collector on start up (https://github.com/open-telemetry/opentelemetry-collector/blob/8990be35990b9833ad98d9bc254667560b7a9951/otelcol/collector.go#L182). We really shouldn't be relying on the Validate function to set defaults in the config.

jefchien (approved):

lisguo (approved):

mxiamxia (approved):

PR #1105, opened 2024-03-29
Add unit test to ensure log timestamp defaults to local timezone

zhihonl:
# Description of the issue
We experienced timezone issues with other products and want to ensure CloudWatch Agent is also in a good state. After code audit, we are missing edge case unit test where the timezone is set to a valid value from the host timezone options but not allowlisted in the agent.

# Description of changes
Add unit test to check for timezone that is not allowlisted by the agent(We only use timezone when it's specified to UTC, if not, default to local)

# License

# Tests
Unit test

# Requirements
2. Run `make lint`

### DIFF
--- a/plugins/inputs/logfile/fileconfig_test.go
+++ b/plugins/inputs/logfile/fileconfig_test.go
@@ -233,6 +233,17 @@ func TestTimestampParserWithFracSeconds(t *testing.T) {
 		fmt.Sprintf("The timestampFromLogLine value %v is not the same as expected %v.", timestamp, expectedTimestamp))
 }
 
+func TestNonAllowlistedTimezone(t *testing.T) {
+	fileConfig := &FileConfig{
+		Timezone: "EST",
+	}
+
+	err := fileConfig.init()
+	assert.NoError(t, err)
+
+	assert.Equal(t, time.Local, fileConfig.TimezoneLoc, "The timezone location should be in local timezone.")
+}
+
 func TestMultiLineStartPattern(t *testing.T) {
 	multiLineStartPattern := "---"
 	fileConfig := &FileConfig{


plugins/inputs/logfile/fileconfig_test.go
```diff
@@ -233,6 +233,36 @@ func TestTimestampParserWithFracSeconds(t *testing.T) {
 		fmt.Sprintf("The timestampFromLogLine value %v is not the same as expected %v.", timestamp, expectedTimestamp))
 }
 
+func TestNonAllowlistedTimezone(t *testing.T) {
+	fileConfig := &FileConfig{
+		FilePath:              "/tmp/logfile.log",
+		LogGroupName:          "logfile.log",
+		TimestampRegex:        "(\\d{2} \\w{3} \\d{4} \\d{2}:\\d{2}:\\d{2})",
+		TimestampLayout:       []string{"02 Jan 2006 15:04:05"},
+		Timezone:              "EST",
```
zhihonl, line 242: This line and line 250 are the main purpose of this CR. We just want to ensure when Timezone is not "UTC", we default to "Local"

plugins/inputs/logfile/fileconfig_test.go
```diff
@@ -233,6 +233,17 @@ func TestTimestampParserWithFracSeconds(t *testing.T) {
...
 
+func TestNonAllowlistedTimezone(t *testing.T) {
+	fileConfig := &FileConfig{
+		Timezone: "EST",
+	}
+
+	err := fileConfig.init()
+	assert.NoError(t, err)
+
+	assert.Equal(t, time.Local, fileConfig.TimezoneLoc, "The timezone location should be in local timezone.")
```
zhihonl, line 244: If Timezone is not "UTC", we default everything else to be local

Paramadon (approved):
Looks good!

plugins/inputs/logfile/fileconfig_test.go
```diff
@@ -233,6 +233,17 @@ func TestTimestampParserWithFracSeconds(t *testing.T) {
 		fmt.Sprintf("The timestampFromLogLine value %v is not the same as expected %v.", timestamp, expectedTimestamp))
 }
 
+func TestNonAllowlistedTimezone(t *testing.T) {
+	fileConfig := &FileConfig{
+		Timezone: "EST",
```
dchappa, line 238: may be a dumb question - Will this fail if the test is run outside of the EST?

zhihonl: You can input any value and it will always pass since we always default to Local when timezone is not specified as "UTC"

dchappa (approved):

PR #1114, opened 2024-04-02
[Logfile] Clean up state folder on init

lisguo:
# Description of the issue
Currently, customers using the logs plugin have a state folder tracking log files on the host that are being watched. This folder gets cleaned up periodically every hour. If the agent dies or crashes before the 1 hour, the agent could end up creating more state files causing excessive inodes to fill up on the host.

# Description of changes
We can clean up the state folder on initialization of the logfile plugin, as well as periodically.

# License

# Tests
pass unit tests

# Requirements
2. Run `make lint`

### DIFF
--- a/plugins/inputs/logfile/logfile.go
+++ b/plugins/inputs/logfile/logfile.go
@@ -107,8 +107,9 @@ func (t *LogFile) Start(acc telegraf.Accumulator) error {
 		return fmt.Errorf("failed to create state file directory %s: %v", t.FileStateFolder, err)
 	}
 
-	// Clean state file regularly
+	// Clean state file on init and regularly
 	go func() {
+		t.cleanupStateFolder()
 		ticker := time.NewTicker(1 * time.Hour)
 		defer ticker.Stop()
 		for {


plugins/inputs/logfile/logfile.go
```diff
@@ -107,7 +107,8 @@ func (t *LogFile) Start(acc telegraf.Accumulator) error {
 		return fmt.Errorf("failed to create state file directory %s: %v", t.FileStateFolder, err)
 	}
 
-	// Clean state file regularly
+	// Clean state file on init and regularly
+	t.cleanupStateFolder()
```
zhihonl, line 111: The agent reads the state to track which part of the log file it left off reading. Does this execute before or after the state file has been read?

jefchien: nit: Move this into the goroutine. I don't think we need to block on the clean up.

jefchien (commented):
Separately, are we sure this is run when the agent is started with the OTEL collector? I'm not sure if the `Start` function in the `logfile` plugin is called.

jefchien (approved):

jefchien (approved):

zhihonl (approved):

plugins/inputs/logfile/logfile.go
```diff
@@ -107,8 +107,9 @@ func (t *LogFile) Start(acc telegraf.Accumulator) error {
 		return fmt.Errorf("failed to create state file directory %s: %v", t.FileStateFolder, err)
 	}
 
-	// Clean state file regularly
+	// Clean state file on init and regularly
 	go func() {
+		t.cleanupStateFolder()
```
adam-mateen, line 112: Can you add a unit test or integ test which creates some files monitored by the agent, write 1 MiB of logs to them, verify the state files point to the EOF, then kill the agent, then write more data, then start the agent back up and make sure they report logs events from the 1 MiB point onwards?

Maybe we already have a test like this, if so great.

My concern is that you are deleting the state files before the agent has a chance to start up and read them. (I think Zhihong has the same concern).

PR #1126, opened 2024-04-08
Drop container/pod level Nvidia GPU metrics when there is no active workload.

movence:
# Description of the issue
The bug is that we always emit pod_gpu_ metrics at the Cluster level for GPU Devices which are not assigned to any pod. This is wrong - there should be no metrics at the pod level because no pod owns the GPU.

# Description of changes
GPUAttribute processor will drop container/pod level nvidia GPU metrics when there is no active workload. It determines by checking `PodName` attribute on a DCGM metric.

# License

# Tests
Unit test and with a test cluster.

EMF log validations:
- Without an active workload
Screenshot 2024-04-08 at 9 19 31 AM

---
- **WITH an ACTIVE WORKLOAD**
Screenshot 2024-04-08 at 9 22 02 AM
Screenshot 2024-04-08 at 9 23 20 AM

# Requirements
2. Run `make lint`

### DIFF
--- a/plugins/processors/gpuattributes/processor.go
+++ b/plugins/processors/gpuattributes/processor.go
@@ -134,6 +134,8 @@ func (d *gpuAttributesProcessor) processMetrics(_ context.Context, md pmetric.Me
 			ils := ilms.At(j)
 			metrics := ils.Metrics()
 
+			d.filterGpuMetricsWithoutPodName(metrics)
+
 			metricsLength := metrics.Len()
 			for k := 0; k < metricsLength; k++ {
 				m := metrics.At(k)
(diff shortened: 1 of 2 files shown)


straussb:
Is it not possible to detect this at the receiver layer?

movence:
Yeah, I had a similar concern of processing metrics over and over in the gpu attribute processor. However, original DCGM metrics get duplicated by [`metrictransformer`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/metricstransformprocessor/README.md) processor which doesn't support options we need. So there is not much we can do in the receiver unless we change the flow of metrics.

straussb:
The decorator consumer could do the duplication logic instead of metricstransformprocessor... it's a bigger change but maybe makes sense as we're having to add more logic.

I'll leave it up to you.  Approving the change as is.

straussb (approved):

movence:
As discussed, we will follow up with another PR to:
- Move metric duplication logic to decorator consumer
- Refactor dcgm and neuron to use the local decorator that EFA already uses for common tasks like labeling and attribute decoration

PR #1137, opened 2024-04-12
upgrade protobuf version - GO-2024-2611, CVE-2024-24786

chadpatel:
**Description:** 
CloudWatch Agent was flagged for CGO-2024-2611, CVE-2024-24786
https://pkg.go.dev/vuln/GO-2024-2611

We are not impacted by this CVE as we don't have any instances where we are unmarshaling into a message which contains a google.protobuf.Any value, or when the UnmarshalOptions.DiscardUnknown option is set.

That being said, the best thing we can do is to update to the latest and greatest protobuf so we know with certainty we aren't impacted and it will also satisfy scanners

**Link to tracking Issue:** N/A
**Testing:** N/A

**Documentation:** N/A

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -101,7 +101,7 @@ require (
 	github.com/gobwas/glob v0.2.3
 	github.com/google/go-cmp v0.6.0
 	github.com/google/uuid v1.4.0
-	github.com/grafana/regexp v0.0.0-20221122212121-6b5c0a4cb7fd
+	github.com/grafana/regexp v0.0.0-20221122212121-6b5c0a4cb7fd // indirect
 	github.com/hashicorp/golang-lru v1.0.2
 	github.com/influxdata/telegraf v0.0.0-00010101000000-000000000000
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -101,7 +101,7 @@ require (
 	github.com/gobwas/glob v0.2.3
 	github.com/google/go-cmp v0.6.0
 	github.com/google/uuid v1.4.0
-	github.com/grafana/regexp v0.0.0-20221122212121-6b5c0a4cb7fd
+	github.com/grafana/regexp v0.0.0-20221122212121-6b5c0a4cb7fd // indirect
```
movence, line 104: nit: should this be moved to indirect group?

chadpatel: idk go mod tidy did this.  It would surprise me if we were using grafana's regex library directly

mitali-salvi (approved):

movence (approved):

PR #1141, opened 2024-04-19
launch canary test once a day

adam-mateen:
# Description of the issue
We previously had a canary test against a private S3 bucket.
We stopped updating that private S3 bucket.
We should have a canary test that exercises the same download/install mechanisms that users use.
(aka post release testing)

# Description of changes
Related change - https://github.com/aws/amazon-cloudwatch-agent-test/pull/404

The github workflow runs on a CRON schedule - once a day at 15:00 UTC.
The workflow just launches a long lived EC2 instance which has a local CRON job that runs a Go test.
The Go test will download, install, configure, and start the agent.
The test reports a few custom metrics.
And the agent obviously reports some metrics based on the simple config it is given.

After this change is merged I will add alarms on these various metrics in our test account.

# License

# Tests
Refer to the linked PR.

# Requirements
2. Run `make lint`

### DIFF
--- a/.github/workflows/deploy-canary.yml
+++ b/.github/workflows/deploy-canary.yml
@@ -1,0 +1,88 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT
+
+name: Deploy Canary
+env:
+  TERRAFORM_AWS_ASSUME_ROLE: ${{ secrets.TERRAFORM_AWS_ASSUME_ROLE }}
+  S3_INTEGRATION_BUCKET: ${{ secrets.S3_INTEGRATION_BUCKET }}
+  KEY_NAME: ${{ secrets.KEY_NAME }}
+  PRIVATE_KEY: ${{ secrets.AWS_PRIVATE_KEY }}
+  CWA_GITHUB_TEST_REPO_NAME: "aws/amazon-cloudwatch-agent-test"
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-canary.yml
```diff
@@ -0,0 +1,88 @@
...
+  S3_INTEGRATION_BUCKET: ${{ secrets.S3_INTEGRATION_BUCKET }}
+  KEY_NAME: ${{ secrets.KEY_NAME }}
+  PRIVATE_KEY: ${{ secrets.AWS_PRIVATE_KEY }}
+  CWA_GITHUB_TEST_REPO_NAME: "aws/amazon-cloudwatch-agent-test"
+  CWA_GITHUB_TEST_REPO_URL: "https://github.com/aws/amazon-cloudwatch-agent-test.git"
+  CWA_GITHUB_TEST_REPO_BRANCH: "main"
+
+on:
+  schedule:
+    - cron: "0 15 * * *" # Run daily at 15:00 UTC
```
adam-mateen, line 16: We could just run the workflow once and the test will repeatedly run on its own thanks to a local CRON job. Running the workflow once a day just helps in case someone accidentally deletes the host. It also helps with patching, although that probably isn't needed if SSM is auto updating.

nathalapooja (approved):

lisguo (approved):

zhihonl:
Maybe as a separate PR but do we also need to bring over the soak test? I see that both soak test and canary are removed from private repo https://github.com/aws/private-amazon-cloudwatch-agent-staging/pull/696/files

zhihonl (approved):

PR #1148, opened 2024-04-24
Added LoggingExporter for AppSignals during translation time if debugging is enabled.

musa-asad:
# Description of the issue
There is no LoggingExporter for AppSignals during translation time.

When debugging is enabled, we want to add the LoggingExporter with detailed verbosity to the translated `yaml`.

# Description of changes
During translation:
- Adds LoggingExporter as an exporter.
- Adds LoggingExporter to AppSignals pipelines when debugging is enabled.

# License

# Tests
- Ran unit tests: `common_test.go`, `logging/translator_test.go`, and `appsignals/translator_test.go`.
  - `common_test.go` was just run to ensure that unit test still works for the new changes in `common.go`.
  - `logging/translator_test.go` checks to see if the `logging.go` is set up correctly and sees if the agent key missing throws an error and if enabling debugging provides the correct yaml configuration for the loggingexporter.
  - `appsignals/translator_test.go` includes a new check to see if the loggingexporter is appropriately added to the exporters when both appsignals and debugging are enabled.

# Requirements
2. Run `make lint`


extension/agenthealth/handler/stats/provider/interval_test.go
```diff
@@ -14,6 +14,7 @@ import (
 )
 
 func TestIntervalStats(t *testing.T) {
+	t.Skip()
```
jefchien, line 17: If it's only flaky in windows, only skip the windows run of the test.

```suggestion
	if runtime.GOOS == windowsOS {
		t.Skip("Interval stats tests have been flaky on Windows.")
	}
```

translator/translate/otel/common/common.go
```diff
@@ -18,6 +18,7 @@ import (
 
 const (
 	AgentKey                           = "agent"
+	Debug                              = "debug"
```
jefchien, line 21: nit: For consistency.
```suggestion
	DebugKey                              = "debug"
```

translator/translate/otel/common/common.go
```diff
@@ -174,6 +177,17 @@ func (e *MissingKeyError) Error() string {
 	return fmt.Sprintf("%q missing key in JSON: %q", e.ID, e.JsonKey)
 }
 
+// A DebugLoggingEnabledError occurs when app signals monitoring is not enabled for
+// an app signals translator.
+type DebugLoggingEnabledError struct {
```
jefchien, line 182: This is very specific to the debug exporter. Move it out of `common`. I don't think you need a custom error for this either.

translator/translate/otel/pipeline/appsignals/translator.go
```diff
@@ -54,6 +55,12 @@ func (t *translator) Translate(conf *confmap.Conf) (*common.ComponentTranslators
 	translators.Processors.Set(resourcedetection.NewTranslator(resourcedetection.WithDataType(t.dataType)))
 	translators.Processors.Set(awsappsignals.NewTranslator(awsappsignals.WithDataType(t.dataType)))
 
+	if conf.IsSet(common.DebugLogging) {
+		if enabled, _ := common.GetBool(conf, common.DebugLogging); enabled {
+			translators.Exporters.Set(logging.NewTranslator())
+		}
+	}
```
jefchien, line 62: nit: Don't need to check if it's set.
```suggestion
		if enabled, ok := common.GetBool(conf, common.DebugLogging); ok && enabled {
			translators.Exporters.Set(logging.NewTranslator())
		}
```

translator/translate/otel/exporter/logging/translator.go
```diff
@@ -0,0 +1,53 @@
...
+
+	cfg := t.factory.CreateDefaultConfig().(*loggingexporter.Config)
+
+	c := confmap.NewFromStringMap(map[string]interface{}{
+		"verbosity": configtelemetry.LevelDetailed,
+	})
+
+	if err := c.Unmarshal(&cfg); err != nil {
+		return nil, fmt.Errorf("unable to unmarshal into loggingexporter config: %w", err)
+	}
```
jefchien, line 50: This unmarshaling is unnecessary. `Verbosity` is a public field that you can set directly (https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/debugexporter/config.go#L26).

translator/translate/otel/exporter/logging/translator.go
```diff
@@ -0,0 +1,53 @@
...
+	return component.NewIDWithName(t.factory.Type(), "")
+}
+
+func (t *translator) Translate(conf *confmap.Conf) (component.Config, error) {
+	if conf == nil || !conf.IsSet(common.DebugLogging) {
+		return nil, &common.MissingKeyError{ID: t.ID(), JsonKey: common.DebugLogging}
+	}
+	if enabled, _ := common.GetBool(conf, common.DebugLogging); !enabled {
+		return nil, &common.DebugLoggingEnabledError{ID: t.ID(), JsonKey: common.DebugLogging}
+	}
```
jefchien, line 40: In the case of this exporter, we already check this before we add it to the pipeline, so it's not necessary to check it again.

translator/translate/otel/common/common.go
```diff
@@ -174,6 +177,17 @@ func (e *MissingKeyError) Error() string {
...
+	ID      component.ID
+	JsonKey string
```
jefchien, line 184: These fields don't make sense for this type of error.

translator/translate/otel/pipeline/appsignals/translator.go
```diff
@@ -54,6 +55,12 @@ func (t *translator) Translate(conf *confmap.Conf) (*common.ComponentTranslators
 	translators.Processors.Set(resourcedetection.NewTranslator(resourcedetection.WithDataType(t.dataType)))
 	translators.Processors.Set(awsappsignals.NewTranslator(awsappsignals.WithDataType(t.dataType)))
 
+	if conf.IsSet(common.DebugLogging) {
+		if enabled, _ := common.GetBool(conf, common.DebugLogging); enabled {
+			translators.Exporters.Set(logging.NewTranslator())
```
jefchien, line 60: The logging exporter should use the same pipeline name. If we want to add logging exporters for other pipelines in the future, this will help to differentiate the logs from each.

zhihonl:
In the test section, can you add info of how you verify this works in a real usage?

translator/translate/otel/pipeline/appsignals/translator.go
musa-asad: Could you elaborate on what you mean by 'same pipeline name'?

jefchien: An OTEL component ID can have a type and a name, which is used to differentiate itself from other components. The pipeline that this logging exporter is part of is the `application_signals` pipeline, so to differentiate itself from other logging exporters, it should include `application_signals` as its name.

musa-asad: Ah, got it. Understood and fixed.

lisguo (approved):

translator/translate/otel/pipeline/appsignals/translator.go
```diff
@@ -54,6 +55,10 @@ func (t *translator) Translate(conf *confmap.Conf) (*common.ComponentTranslators
 	translators.Processors.Set(resourcedetection.NewTranslator(resourcedetection.WithDataType(t.dataType)))
 	translators.Processors.Set(awsappsignals.NewTranslator(awsappsignals.WithDataType(t.dataType)))
 
+	if enabled, _ := common.GetBool(conf, common.DebugLogging); enabled {
```
jefchien, line 58: Still think there's risk to enabling this level of debug logging using the existing `agent/debug` field. We can determine if we need to change this before we release it.

jefchien (approved):

PR #1166, opened 2024-05-07
Revert disk volume

sethAmazon:
# Description of the issue
Dimensions were wrong for disk metrics without volume-id attached

# Description of changes
Revert change to add volume-id to disk metrics

# License

# Tests
Unit

# Requirements
2. Run `make lint`


lisguo (approved):

musa-asad (approved):

sethAmazon:
False alarm

mda590:
@sethAmazon @lisguo It looks like the changes for Volume ID were removed and then never added back? Is that intentional? Is there any plan to bring back this functionality? We are trying to decide if we need to write our own script to map EBS volume IDs to metrics or if this is something that can be added to the agent? Thanks!

PR #1191, opened 2024-05-28
Optimize the memory utilization of Kubernetes resolver in Application Signals

bjrara:
# Description of the issue
The memory utilization of CloudWatch Agent is high after enabling Application Signals. In an EKS cluster of ~7k Pods, the memory usage went up to 90% for leader instance, 70% for follower instances.

From heap profiling, it was confirmed that the issue was caused by `applicationsignalsprocessor` which list/watch Pods and Services cluster-wise for IP/Service translation.

# Description of changes
This PR does the following changes to `applicationsignalsprocessor`:
1. Reduce the memory usage of list/watching.
  It is achieved by customizing transformer in informers to empty fields that might contain big objects. By applying the changes, the memory usage went down from 90% to 50% for leader, 70% to 40% for followers.
1. Remove logics of setting RemoteService to UnknownRemoteService due to IP translation failure.
  Forcing low cardinality `RemoteService` is no longer needed as we have metrics limiting in place.
1. Fix bug in Pod IP cache with `hostNetwork` as true.
  In the current implementation, if a Pod is enabled with `hostNetwork`, its IP cache will never be cleared even after the Pod is deleted.
6. Code refactoring Kubernetes resolver.

# Result
## Baseline (w/o Application Signals)
image (1)

## Before
image (2)

## After
image

# License

# Tests

# Requirements
8. Run `make lint`

### DIFF
--- a/plugins/processors/awsapplicationsignals/internal/resolver/kubernetes.go
+++ b/plugins/processors/awsapplicationsignals/internal/resolver/kubernetes.go
@@ -8,10 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"math/rand"
-	"net"
-	"regexp"
-	"strconv"
-	"strings"
+	"reflect"
 	"sync"
 	"time"
 
@@ -20,6 +17,7 @@ import (
 	semconv "go.opentelemetry.io/collector/semconv/v1.22.0"
 	"go.uber.org/zap"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
@@ -31,10 +29,6 @@ import (
 )
 
 const (
-	// kubeAllowedStringAlphaNums holds the characters allowed in replicaset names from as parent deployment
-	// https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/rand/rand.go#L121
-	kubeAllowedStringAlphaNums = "bcdfghjklmnpqrstvwxz2456789"
-
 	// Deletion delay adjustment:
 	// Previously, EKS resolver would instantly remove the IP to Service mapping when a pod was destroyed.
 	// This posed a problem because:
@@ -47,22 +41,6 @@ const (
 	deletionDelay = 2 * time.Minute
 
 	jitterKubernetesAPISeconds = 10
-)
-
-var (
-	// ReplicaSet name = Deployment name + "-" + up to 10 alphanumeric characters string, if the ReplicaSet was created through a deployment
-	// The suffix string of the ReplicaSet name is an int32 number (0 to 4,294,967,295) that is cast to a string and then
-	// mapped to an alphanumeric value with only the following characters allowed: "bcdfghjklmnpqrstvwxz2456789".
-	// The suffix string length is therefore nondeterministic. The regex accepts a suffix of length 6-10 to account for
-	// ReplicaSets not managed by deployments that may have similar names.
-	// Suffix Generation: https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/controller_utils.go#L1201
-	// Alphanumeric Mapping: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/util/rand/rand.go#L121)
-	replicaSetWithDeploymentNamePattern = fmt.Sprintf(`^(.+)-[%s]{6,10}$`, kubeAllowedStringAlphaNums)
-	deploymentFromReplicaSetPattern     = regexp.MustCompile(replicaSetWithDeploymentNamePattern)
-	// if a pod is launched directly by a replicaSet (with a given name by users), its name has the following pattern:
-	// Pod name = ReplicaSet name + 5 alphanumeric characters long string
-	podWithReplicaSetNamePattern = fmt.Sprintf(`^(.+)-[%s]{5}$`, kubeAllowedStringAlphaNums)
-	replicaSetFromPodPattern     = regexp.MustCompile(podWithReplicaSetNamePattern)
 )
 
 type kubernetesResolver struct {
@@ -108,69 +86,6 @@ func jitterSleep(seconds int) {
 	time.Sleep(jitter)
 }
 
-func attachNamespace(resourceName, namespace string) string {
-	// character "@" is not allowed in kubernetes resource names: https://unofficial-kubernetes.readthedocs.io/en/latest/concepts/overview/working-with-objects/names/
-	return resourceName + "@" + namespace
-}
-
-func getServiceAndNamespace(service *corev1.Service) string {
-	return attachNamespace(service.Name, service.Namespace)
-}
-
-func extractResourceAndNamespace(serviceOrWorkloadAndNamespace string) (string, string) {
-	// extract service name and namespace from serviceAndNamespace
-	parts := strings.Split(serviceOrWorkloadAndNamespace, "@")
-	if len(parts) != 2 {
-		return "", ""
-	}
-	return parts[0], parts[1]
-}
-
-func extractWorkloadNameFromRS(replicaSetName string) (string, error) {
-	match := deploymentFromReplicaSetPattern.FindStringSubmatch(replicaSetName)
-	if match != nil {
-		return match[1], nil
-	}
-
-	return "", errors.New("failed to extract workload name from replicatSet name: " + replicaSetName)
-}
-
-func extractWorkloadNameFromPodName(podName string) (string, error) {
-	match := replicaSetFromPodPattern.FindStringSubmatch(podName)
-	if match != nil {
-		return match[1], nil
-	}
-
-	return "", errors.New("failed to extract workload name from pod name: " + podName)
-}
-
-func getWorkloadAndNamespace(pod *corev1.Pod) string {
-	var workloadAndNamespace string
-	if pod.ObjectMeta.OwnerReferences != nil {
-		for _, ownerRef := range pod.ObjectMeta.OwnerReferences {
-			if workloadAndNamespace != "" {
-				break
-			}
-
-			if ownerRef.Kind == "ReplicaSet" {
-				if workloadName, err := extractWorkloadNameFromRS(ownerRef.Name); err == nil {
-					// when the replicaSet is created by a deployment, use deployment name
-					workloadAndNamespace = attachNamespace(workloadName, pod.Namespace)
-				} else if workloadName, err := extractWorkloadNameFromPodName(pod.Name); err == nil {
-					// when the replicaSet is not created by a deployment, use replicaSet name directly
-					workloadAndNamespace = attachNamespace(workloadName, pod.Namespace)
-				}
-			} else if ownerRef.Kind == "StatefulSet" {
-				workloadAndNamespace = attachNamespace(ownerRef.Name, pod.Namespace)
-			} else if ownerRef.Kind == "DaemonSet" {
-				workloadAndNamespace = attachNamespace(ownerRef.Name, pod.Namespace)
-			}
-		}
-	}
-
-	return workloadAndNamespace
-}
-
 // Deleter represents a type that can delete a key from a map after a certain delay.
 type Deleter interface {
 	DeleteWithDelay(m *sync.Map, key interface{})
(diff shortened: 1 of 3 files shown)


plugins/processors/awsapplicationsignals/internal/resolver/kubernetes.go
```diff
@@ -242,110 +157,107 @@ func updateHostNetworkRecords(newPod *corev1.Pod, oldPod *corev1.Pod, ipToPod *s
...
 	}
 }
 
-func onAddOrUpdatePod(newObj, oldObj interface{}, ipToPod, podToWorkloadAndNamespace, workloadAndNamespaceToLabels *sync.Map, workloadPodCount map[string]int, isAdd bool, logger *zap.Logger, deleter Deleter) {
-	pod := newObj.(*corev1.Pod)
-
+func (p *podWatcher) onAddOrUpdatePod(pod, oldPod *corev1.Pod, isAdd bool) {
 	if isAdd {
-		handlePodAdd(pod, ipToPod)
+		p.handlePodAdd(pod)
```
mxiamxia, line 216: nice refactoring works on these verbose functions!

plugins/processors/awsapplicationsignals/internal/resolver/kubernetes.go
```diff
@@ -523,38 +481,48 @@ func getKubernetesResolver(platformCode, clusterName string, logger *zap.Logger)
 		// jitter calls to the kubernetes api
 		jitterSleep(jitterKubernetesAPISeconds)
 
-		sharedInformerFactory := informers.NewSharedInformerFactory(clientset, 0)
+		sharedInformerFactory := informers.NewSharedInformerFactoryWithOptions(clientset, 0, informers.WithTweakListOptions(func(options *metav1.ListOptions) {
+			options.ResourceVersion = ""
```
mxiamxia, line 485: Q: is `WithTweakListOptions` with `options.ResourceVersion = ""` necessary?

plugins/processors/awsapplicationsignals/internal/resolver/kubernetes.go
```diff
@@ -523,38 +481,48 @@ func getKubernetesResolver(platformCode, clusterName string, logger *zap.Logger)
...
+		}))
 		podInformer := sharedInformerFactory.Core().V1().Pods().Informer()
+		err = podInformer.SetTransform(filterPodIPFields)
+		if err != nil {
+			logger.Error("failed to set custom transform into pod store", zap.Error(err))
+		}
```
mxiamxia, line 491: will Informer fallback to the default caching behavior with the full Pod metadata stored if `SetTransform` failed?

mxiamxia (commented):
LGTM! Left a few comments/questions but non-blocker.

plugins/processors/awsapplicationsignals/internal/resolver/kubernetes.go
bjrara: This is mainly to reduce the memory spike on start.

plugins/processors/awsapplicationsignals/internal/resolver/kubernetes.go
bjrara: Yes.

plugins/processors/awsapplicationsignals/internal/resolver/kubernetes.go
bjrara: I reverted this part of the logics as it causes inconsistency in our stateful mappings.

plugins/processors/awsapplicationsignals/internal/resolver/kubernetes.go
```diff
@@ -501,13 +390,67 @@ func (m *ServiceToWorkloadMapper) Start(stopCh chan struct{}) {
...
+		pod.Spec.EphemeralContainers = nil
+		pod.Spec.ImagePullSecrets = nil
+		pod.Spec.HostAliases = nil
+		pod.Spec.SchedulingGates = nil
+		pod.Spec.ResourceClaims = nil
+		pod.Spec.Tolerations = nil
+		pod.Spec.Affinity = nil
+
+		pod.Status.InitContainerStatuses = nil
+		pod.Status.ContainerStatuses = nil
```
mxiamxia, line 417: please add a comment to indicate which fields will not be reset that will be used by the processor

mxiamxia (commented):
🚢

plugins/processors/awsapplicationsignals/internal/resolver/kubernetes.go
bjrara: added.

mxiamxia (dismissed):

plugins/processors/awsapplicationsignals/internal/resolver/kubernetes.go
```diff
@@ -525,36 +477,44 @@ func getKubernetesResolver(platformCode, clusterName string, logger *zap.Logger)
 
 		sharedInformerFactory := informers.NewSharedInformerFactory(clientset, 0)
 		podInformer := sharedInformerFactory.Core().V1().Pods().Informer()
+		err = podInformer.SetTransform(filterPodIPFields)
+		if err != nil {
+			logger.Error("failed to set custom transform into pod store", zap.Error(err))
```
lisguo, line 482: nit: change error message to something more meaningful
```suggestion
			logger.Error("failed to minimize pod object", zap.Error(err))
```

plugins/processors/awsapplicationsignals/internal/resolver/kubernetes.go
```diff
@@ -525,36 +477,44 @@ func getKubernetesResolver(platformCode, clusterName string, logger *zap.Logger)
...
+		}
 		serviceInformer := sharedInformerFactory.Core().V1().Services().Informer()
+		err = serviceInformer.SetTransform(filterServiceIPFields)
+		if err != nil {
+			logger.Error("failed to set custom transform into service store", zap.Error(err))
```
lisguo, line 487: similar to above
```suggestion
			logger.Error("failed to minimize service object", zap.Error(err))
```

plugins/processors/awsapplicationsignals/internal/resolver/kubernetes.go
```diff
@@ -501,13 +390,76 @@ func (m *ServiceToWorkloadMapper) Start(stopCh chan struct{}) {
...
 		}
 	}()
 }
 
+// filterPodIPFields removes fields that could contain large objects, and retain essential
+// fields needed for IP/name translation. The following fields must be kept:
+// - ObjectMeta: Namespace, Name, Labels, OwnerReference
+// - Spec: HostNetwork, ContainerPorts
+// - Status: PodIP/s, HostIP/s
+func filterPodIPFields(obj interface{}) (interface{}, error) {
```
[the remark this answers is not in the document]
bjrara: It was firstly done so, but turned out creating new objects increased the memory consumption to a new level.

> If k8s api adds a new pod spec for example, then we have to update this every time.
It's true, but we don't need to update every time, only when large object is introduced.

plugins/processors/awsapplicationsignals/internal/resolver/kubernetes.go
bjrara: Updated.

plugins/processors/awsapplicationsignals/internal/resolver/kubernetes.go
bjrara: Updated.

plugins/processors/awsapplicationsignals/internal/resolver/kubernetes.go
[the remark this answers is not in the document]
bjrara: Updated to `minimizePod`

plugins/processors/awsapplicationsignals/internal/resolver/kubernetes.go
```diff
@@ -501,13 +390,76 @@ func (m *ServiceToWorkloadMapper) Start(stopCh chan struct{}) {
...
+		}
+	}
+	return obj, nil
+}
+
+// filterServiceIPFields removes fields that could contain large objects, and retain essential
+// fields needed for IP/name translation. The following fields must be kept:
+// - ObjectMeta: Namespace, Name
+// - Spec: Selectors, ClusterIP
+func filterServiceIPFields(obj interface{}) (interface{}, error) {
```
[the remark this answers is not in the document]
bjrara: Updated to `minimizeService`.

plugins/processors/awsapplicationsignals/internal/resolver/kubernetes.go
```diff
@@ -202,150 +114,120 @@ func onAddOrUpdateService(obj interface{}, ipToServiceAndNamespace, serviceAndNa
...
 	pod := obj.(*corev1.Pod)
-	if pod.Status.PodIP != "" {
-		deleter.DeleteWithDelay(ipToPod, pod.Status.PodIP)
-	} else if pod.Status.HostIP != "" {
+	if pod.Spec.HostNetwork && pod.Status.HostIP != "" {
+		p.logger.Debug("deleting host ip from cache", zap.String("hostNetwork", pod.Status.HostIP))
 		for _, port := range getHostNetworkPorts(pod) {
-			deleter.DeleteWithDelay(ipToPod, pod.Status.HostIP+":"+port)
+			p.deleter.DeleteWithDelay(p.ipToPod, pod.Status.HostIP+":"+port)
 		}
```
jefchien, line 210: Same here. Can reuse `p.removeHostNetworkRecords`

plugins/processors/awsapplicationsignals/internal/resolver/kubernetes.go
```diff
@@ -202,150 +114,120 @@ func onAddOrUpdateService(obj interface{}, ipToServiceAndNamespace, serviceAndNa
...
-			ipToPod.Store(newPod.Status.PodIP, newPod.Name)
+			p.ipToPod.Store(newPod.Status.PodIP, newPod.Name)
 		}
 	}
 }
 
-func onAddOrUpdatePod(newObj, oldObj interface{}, ipToPod, podToWorkloadAndNamespace, workloadAndNamespaceToLabels *sync.Map, workloadPodCount map[string]int, isAdd bool, logger *zap.Logger, deleter Deleter) {
-	pod := newObj.(*corev1.Pod)
-
+func (p *podWatcher) onAddOrUpdatePod(pod, oldPod *corev1.Pod, isAdd bool) {
```
jefchien, line 179: Couldn't we determine if it's an add based on the presence of `oldPod`?

plugins/processors/awsapplicationsignals/internal/resolver/kubernetes_utils.go
```diff
@@ -0,0 +1,142 @@
...
+
+func getWorkloadAndNamespace(pod *corev1.Pod) string {
+	var workloadAndNamespace string
+	if pod.ObjectMeta.OwnerReferences != nil {
+		for _, ownerRef := range pod.ObjectMeta.OwnerReferences {
+			if workloadAndNamespace != "" {
+				break
+			}
+
+			if ownerRef.Kind == "ReplicaSet" {
```
jefchien, line 83: nit: The kind consts are defined in `containerinsightscommon` and can be reused.

jefchien (approved):

plugins/processors/awsapplicationsignals/internal/resolver/kubernetes.go
```diff
@@ -202,150 +114,120 @@ func onAddOrUpdateService(obj interface{}, ipToServiceAndNamespace, serviceAndNa
...
-		if oldPod.Status.PodIP != "" {
-			deleter.DeleteWithDelay(ipToPod, oldPod.Status.PodIP)
+func (p *podWatcher) handlePodUpdate(newPod *corev1.Pod, oldPod *corev1.Pod) {
+	// HostNetwork is an immutable field
+	if newPod.Spec.HostNetwork && oldPod.Status.HostIP != newPod.Status.HostIP {
+		if oldPod.Status.HostIP != "" {
+			p.logger.Debug("deleting host ip from cache", zap.String("hostNetwork", oldPod.Status.HostIP))
+			for _, port := range getHostNetworkPorts(oldPod) {
+				p.deleter.DeleteWithDelay(p.ipToPod, oldPod.Status.HostIP+":"+port)
+			}
```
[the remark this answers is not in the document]
bjrara: Thanks! Updated.

plugins/processors/awsapplicationsignals/internal/resolver/kubernetes.go
bjrara: Updated.

plugins/processors/awsapplicationsignals/internal/resolver/kubernetes.go
bjrara: We can. Updated.

plugins/processors/awsapplicationsignals/internal/resolver/kubernetes_utils.go
bjrara: IMHO, including `containerinsightscommon` package just for removing these one-time hardcodes alone doesn't bring a lot benefits. I can add them as consts in our package if any hardcoded string should be avoided. Rn, I'm keeping them.

lisguo (approved):

jefchien (approved):

PR #1266, opened 2024-08-01
Fix potential out of bounds panic

dricross:
# Description of the issue
The `TestTranslator/WithKubernetes/WithoutClusterName` unit test is failing locally due to a panic in `backoffSleep` on this line: https://github.com/aws/amazon-cloudwatch-agent/blob/main/translator/translate/logs/util/get_eks_cluster_name.go#L117. The call to ec2.DescribeTags to attempt to find the Kubernetes cluster name is failing (due to an environment issue), and it retries 6 times. On the sixth attempt, it attempts to access sleeps[5] which is invalid causing a panic.

This panic behavior would happen outside unit testing if the ec2.DescribeTags call fails more than 5 times.

# Description of changes
Maintain behavior of retrying 6 times, but only pull from pre-defined sleep durations on attempts 1-5. On 6th attempt, use default duration.

# License

# Tests
Create new unit test for function that was panicing. Before changes, the unit test panics regardless of environment configuration. After changes, the unit test does not panic.

# Requirements
2. Run `make lint`


translator/translate/logs/util/get_eks_cluster_name_test.go
```diff
@@ -0,0 +1,16 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+package util
+
+import (
+	"testing"
+)
+
+func TestBackoffSleep(t *testing.T) {
```
jefchien, line 10: I don't think we want a test that actually sleeps for a minute. Consider creating another function like `getBackoffDuration` and just verifying the values it gets.

```
github.com/aws/amazon-cloudwatch-agent/translator/translate/logs/util	66.231s	coverage: 1.3% of statements
```

musa-asad (approved):

lisguo (approved):

PR #1269, opened 2024-08-02
Separate build jobs for artifacts and trigger tests with workflow_run

jefchien:
# Description of the issue
The `integration-test.yml` workflow has exceeded the reusable workflow limit of 20 (https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/reusing-workflows#limitations) due to changes in the nested reusable workflows in the `application-signals-e2e-test.yml` workflow and no longer runs (https://github.com/aws/amazon-cloudwatch-agent/actions/runs/10221982238).

>Invalid workflow file: .github/workflows/integration-test.yml#L1 too many workflows are referenced, total: 25, limit: 20

There isn't currently a way to increase the limit.

# Description of changes
One way around this limitation is to move the application signals E2E workflow out of the integration test workflow. The build jobs that were previously in the integration test workflow have been separated out into their own workflow since both test workflows rely on the build artifacts for the commit. The test workflows are now triggered on the build workflow completion and success.

# License

# Tests
Triggered the build with a PR event (https://github.com/aws/amazon-cloudwatch-agent/actions/runs/10222942184?pr=1269), but am unable to test the workflow_run

> Note: This event will only trigger a workflow run if the workflow file is on the default branch.

https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_run

# Requirements
2. Run `make lint`

### DIFF
--- a/.github/workflows/application-signals-e2e-test.yml
+++ b/.github/workflows/application-signals-e2e-test.yml
@@ -4,9 +4,12 @@
 # This is a reusable workflow for running the E2E test for Application Signals.
 # It is meant to be called from another workflow.
 # Read more about reusable workflows: https://docs.github.com/en/actions/using-workflows/reusing-workflows#overview
-name: E2E Testing
+name: Application Signals E2E Test
 on:
-  workflow_call:
+  workflow_run:
+    workflows: [ Build Test Artifacts ]
+    types:
+      - completed
 
 permissions:
   id-token: write
@@ -18,7 +21,14 @@ concurrency:
 
 
 jobs:
+  CheckBuildTestArtifacts:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    steps:
+      - run: echo 'The triggering build workflow succeeded'
+
   java-eks-e2e-test:
+    needs: CheckBuildTestArtifacts
     uses: aws-observability/aws-application-signals-test-framework/.github/workflows/java-eks-e2e-test.yml@main
     secrets: inherit
     with:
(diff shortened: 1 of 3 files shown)


.github/workflows/application-signals-e2e-test.yml
```diff
@@ -18,7 +21,14 @@ concurrency:
 
 
 jobs:
+  CheckBuildTestArtifacts:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    steps:
+      - run: echo 'The triggering build workflow succeeded'
```
okankoAMZ, line 28: So this job only echos ?

.github/workflows/application-signals-e2e-test.yml
```diff
@@ -27,21 +37,23 @@ jobs:
...
   java-ec2-default-e2e-test:
+    needs: CheckBuildTestArtifacts
     uses: aws-observability/aws-application-signals-test-framework/.github/workflows/java-ec2-default-e2e-test.yml@main
     secrets: inherit
     with:
       aws-region: us-east-1
       caller-workflow-name: 'main-build'
 
   java-ec2-asg-e2e-test:
+    needs: [ CheckBuildTestArtifacts ]
```
okankoAMZ, line 48: nit: why is it a list if there is only one dependency

okankoAMZ (approved):

.github/workflows/application-signals-e2e-test.yml
jefchien: They did that for their other jobs, so I just followed the pattern.

.github/workflows/application-signals-e2e-test.yml
jefchien: The main purpose of this step is to act as a blocker. If the build fails, then the tests shouldn't run. Think a job requires there to be a step.

sky333999 (approved):

.github/workflows/application-signals-e2e-test.yml
okankoAMZ: Makes sense

PR #1302, opened 2024-08-15
Reduce the size of the start-amazon-cloudwatch-agent binary

jefchien:
# Description of the issue
The `start-amazon-cloudwatch-agent`, which is meant to just call the translator and start the agent, has grown to 95MB-112MB.
```
> du -h build/bin/linux_amd64/start-amazon-cloudwatch-agent 
95M    build/bin/linux_amd64/start-amazon-cloudwatch-agent
```

From looking at a build with https://github.com/Zxilly/go-size-analyzer, this is due to all the dependencies it brings in from the translator. The main source of this is the `run_as_user` logic.

image

# Description of changes
Moves the `run_as_user` logic out of the translator into an `internal` package. Replaces the JSON merge to find the `run_as_user` in the `start-amazon-cloudwatch-agent` with a lookup in the TOML generated by the translator and a change in Telegraf to allow for unsupported fields (https://github.com/aws/telegraf/pull/154).

image

Reduces the size of the binary to 2MB.
```
> du -h build/bin/linux_amd64/start-amazon-cloudwatch-agent 
2.1M    build/bin/linux_amd64/start-amazon-cloudwatch-agent
```

# License

# Tests
Built and analyzed the binaries locally. Ran the integration tests against them.
Build: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/10406066610
Integration Tests: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/10406485196

# Requirements
2. Run `make lint`


cmd/amazon-cloudwatch-agent/amazon-cloudwatch-agent.go
```diff
@@ -241,6 +241,8 @@ func runAgent(ctx context.Context,
 	c := config.NewConfig()
 	c.OutputFilters = outputFilters
 	c.InputFilters = inputFilters
+	c.AllowUnusedFields = true
```
chadpatel, line 244: what is this change for?

jefchien: `start-amazon-cloudwatch-agent` is responsible for a few things: running the `config-translator` binary to generate the TOML/YAML, processing the `run_as_user` to potentially switch users, and starting the `amazon-cloudwatch-agent` binary.

To determine the `run_as_user`, `start-amazon-cloudwatch-agent` currently merges the available JSONs and then parses the JSON to read the field. This is work that is already done by the `config-translator` before it writes the TOML/YAML. This change adds `run_as_user` under the `agent` section of the Telegraf configuration so the `start-amazon-cloudwatch-agent` no longer needs to merge the JSONs to read the file, which avoids repeated work and unnecessary dependencies.

Telegraf normally validates that all fields in the TOML configuration are known fields and used by their components.
`run_as_user` is not a standard field. Setting `AllowUnusedFields` will tell the Telegraf configuration unmarshaler to not fail based on the presence of this unknown field.

Makefile
```diff
@@ -109,7 +109,6 @@ build-for-docker-amd64:
 	$(LINUX_AMD64_BUILD)/amazon-cloudwatch-agent github.com/aws/amazon-cloudwatch-agent/cmd/amazon-cloudwatch-agent
 	$(LINUX_AMD64_BUILD)/start-amazon-cloudwatch-agent github.com/aws/amazon-cloudwatch-agent/cmd/start-amazon-cloudwatch-agent
 	$(LINUX_AMD64_BUILD)/config-translator github.com/aws/amazon-cloudwatch-agent/cmd/config-translator
-	cp $(BASE_SPACE)/packaging/opentelemetry-jmx-metrics.jar $(BUILD_SPACE)/bin/linux_amd64/opentelemetry-jmx-metrics.jar
```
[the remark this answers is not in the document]
jefchien: It's not and shouldn't be in this PR, but should be removed in another one. The JAR can't be used by the image, so it shouldn't be included in the build.

chadpatel (approved):

cmd/amazon-cloudwatch-agent/amazon-cloudwatch-agent.go
sky333999: This is effectively turning off the check entirely, so we do run the risk of adding new unused fields in the future and not getting flagged - although Im not sure why that is so bad. Id imagine any unused fields wont cause any issues, just that they are present for no reason.

Alternative might be to just add `run_as_user` to some allow list like https://github.com/aws/telegraf/blob/483bd77d26a2900938ad835d286a719c5efe954c/config/config.go#L1838?

sky333999 (approved):

PR #1312, opened 2024-08-21
Increase retry delay on PLE for certain exceptions

dricross:
# Description of the issue
The CloudWatchAgent has a delayed retry for failed PutLogEvents requests which starts out small (200ms, jittered) and increases up to 1 minute based on the number of consecutive failed requests, and then all subsequent retries take 1 minute (jittered). If the back end servers are experiencing an outage, this retry strategy can overwhelm the already struggling servers, exacerbating the problem.

# Description of changes
Use a "long" retry strategy if the CloudWatchAgent receives response codes that indicate a backend outage may be occurring.

For only 500, 503, Connection Refused, Connection Reset by Peer, Connection Timeout, Throttling errors for PutLogEvents:
 * Set base retry interval to 2 seconds
 * Reduce retries to 2 retries
 * Keep long-retry to 1 minute
 
This will not increase the delay generally since the agent does rely on exception handling and retries for some core behavior (i.e. creating the log group/stream if they don't exist).

This implementation borrows some error checking logic from the AWS Go SDK (v1):
* [Checking for throttle errors](https://github.com/aws/aws-sdk-go/blob/main/aws/request/retryer.go#L230)
* [Checking for connection refused errors](https://github.com/aws/aws-sdk-go/blob/main/aws/request/retryer.go#L192)
* Checking for timeout errors (several places in [aws/request/retryer](https://github.com/aws/aws-sdk-go/blob/main/aws/request/retryer.go))
* [Checking for connection reset errors](https://github.com/aws/aws-sdk-go/blob/main/aws/request/connection_reset_error.go#L7)

# License

# Tests
Unit tests

# Requirements
2. Run `make lint`


musa-asad (dismissed):

plugins/outputs/cloudwatchlogs/pusher.go
```diff
@@ -297,38 +308,103 @@ func (p *pusher) send() {
...
+//   - Connection Reset by Peer
+//   - Connection Timeout
+//   - Throttling
+func chooseRetryWaitStrategy(err error) retryWaitStrategy {
+	if netErr, ok := err.(net.Error); ok && netErr.Timeout() {
+		return retryLong
+	}
+
+	// Connection reset/refused are actually syscalls
+	if errors.Is(err, syscall.ECONNRESET) || errors.Is(err, syscall.ECONNREFUSED) {
```
jefchien, line 386: Will these fall under non-AWS errors? There's this section on line 277, 
```go
awsErr, ok := err.(awserr.Error)
		if !ok {
			p.Log.Errorf("Non aws error received when sending logs to %v/%v: %v. CloudWatch agent will not retry and logs will be missing!", p.Group, p.Stream, err)
			// Messages will be discarded but done callbacks not called
			p.reset()
			return
}
```
Will this error cause the logs to get dropped?

plugins/outputs/cloudwatchlogs/pusher.go
```diff
@@ -297,38 +308,103 @@ func (p *pusher) send() {
...
+		d = base * time.Duration(1<<int64(retryCount))
+	}
+	return time.Duration(seededRand.Int63n(int64(d/2)) + int64(d/2))
+}
+
+// retryWaitLong returns a duration to wait before retrying a request.
+// This implements an exponential backoff strategy with jitter.
+// The base wait time is 2 seconds, and the maximum wait time is 1 minute (jittered).
+// The wait time is capped at 1 minute if the long-retry count exceeds 2.
+func retryWaitLong(retryCount int) time.Duration {
```
jefchien, line 359: nit: Both functions are identical with just different base wait times and retry counts. Couldn't we just pass those in?

plugins/outputs/cloudwatchlogs/pusher_test.go
```diff
@@ -767,6 +769,184 @@ func TestResendWouldStopAfterExhaustedRetries(t *testing.T) {
...
+				assert.LessOrEqual(t, duration, tt.maxDuration, "retryWait(%v) should be less than or equal to %v", tt.n, tt.maxDuration)
+			}
+		})
+	}
+}
+
+func TestRetryWaitLong(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		n           int
```
jefchien, line 893: nit: Don't need n since `range` will give you the index.
```go
for n, tt := range tests
```

plugins/outputs/cloudwatchlogs/pusher.go
dricross: The way I interpret the [SDK docs](https://docs.aws.amazon.com/sdk-for-go/api/aws/awserr/#Error) is that SDK API calls should always wrap lower level errors in an awserr.Error before returning to the caller. If that's the case, then these syscall errors would be wrapped by awserr.Error, and it would be rather unexpected to ever get a non-AWS error. If that's not how it works, then yeah I believe it would cause the logs to get dropped.

Perhaps there's a way to verify that behavior (unit tests mocking the return types obviously won't work)...

plugins/outputs/cloudwatchlogs/pusher_test.go
dricross: That's true, but I think it's more clear when the test case specifies the test conditions rather than relying on the index to form the test case inputs.

plugins/outputs/cloudwatchlogs/pusher.go
dricross: Yeah, we could parametrize it. I felt the extra layer of abstraction wasn't worth it for these two functions (and I don't anticipate we'd be growing the number of strategies) so I left them separate.

plugins/outputs/cloudwatchlogs/pusher.go
dricross: I created a unit test which replaces the underyling HttpClient used by the cloudwatchlogs client with a mock client that returns errors to see what the PutLogEvents API call returns.

```
customHTTPClient := &errorRoundTripper{
	rt:     http.DefaultTransport,
	errMsg: "connection refused",
}
			
sess, err := session.NewSession(&aws.Config{
	HTTPClient: &http.Client{
		Transport: customHTTPClient,
	},
	Region: aws.String("us-east-1"),
})
require.NoError(t, err)

cwl := cloudwatchlogs.New(sess)
params := &cloudwatchlogs.PutLogEventsInput{
	LogEvents: []*cloudwatchlogs.InputLogEvent{
		{
			Message:   aws.String("test message"),
			Timestamp: aws.Int64(time.Now().UnixNano() / int64(time.Millisecond)),
		},
	},
	LogGroupName:  aws.String("hello"),
	LogStreamName: aws.String("world"),
}
_, err = cwl.PutLogEvents(params)
fmt.Println(reflect.TypeOf(err))
require.True(t, strings.Contains(err.Error(), tt.errMsg))
_, ok := err.(awserr.Error)
require.True(t, ok, "error returned is not an awserr")
```
The unit test verifies that an error returned by the underlying HTTPClient would be wrapped by an `awserr.Error` before being returned to the caller. So, underlying errors will not get caught by the section on line 277.

However, I did find an example in aws-sdk-go where the [SDK itself checks for connection reset errors](https://github.com/aws/aws-sdk-go/blob/02c1f723b528251a45068001bee8b56d904d7484/aws/request/connection_reset_error.go) by looking at the error message instead of the error type:
```
func isErrConnectionReset(err error) bool {
	if strings.Contains(err.Error(), "read: connection reset") {
		return false
	}

	if strings.Contains(err.Error(), "use of closed network connection") ||
		strings.Contains(err.Error(), "connection reset") ||
		strings.Contains(err.Error(), "broken pipe") {
		return true
	}

	return false
}
```

So checking for the syscall errors may not be what we want to do in the agent either. It may be better to borrow the logic from the SDK since there seems to be a series of issues that lead to this particular implementation:
* https://github.com/aws/aws-sdk-go/commit/135dd4cb18a208322feca0a3f46519920beae53a
* https://github.com/aws/aws-sdk-go/commit/c3d27102088451890a9ba5be0dc1f2e0425f0fc3
* https://github.com/aws/aws-sdk-go/commit/2c957a88ad3e912c48e72b9691eba01912a0218d

musa-asad (dismissed):

plugins/outputs/cloudwatchlogs/pusher_test.go
jefchien: Yeah you're right since they're separate test cases.

jefchien (dismissed):

jefchien (dismissed):

musa-asad (dismissed):
Nit appropriately addressed, reapproved

musa-asad (approved):

jefchien (approved):

PR #1314, opened 2024-08-23
Adhere to SOURCE_DATE_EPOCH specification to fix reproducibility

ohnatiuk1:
# Description of the issue
Currently Makefile uses current date and time as an argument to the linker which leads to the timestamp being embedded in the binary causing reproducibility issues: we will get different binaries each time we build we build the project.
Reproducible builds are ones where the same source code input produces the same binary output. This helps with caching, authentication of binaries, etc.

# Description of changes
My change adds support for a standardised variable. When set, this variable indicates that the build environment wishes to have the value of this variable used as timestamps instead of "current time".

See:
https://reproducible-builds.org/specs/source-date-epoch/
https://reproducible-builds.org/docs/source-date-epoch/

# License
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your current license (MIT).

# Tests
I am able to build the project after applying this patch.
Furthermore, examining the resulting binaries shows that timestamps are set as expected.


ohnatiuk1:
Hi. When will this be reviewed and merged?

Issue #1328, opened 2024-09-03
v1.300045.0 doesn't work on ECS Fargate


rchanbg:
**Describe the bug**
Latest version v1.300045.0 doesn't work on ECS Fargate. The following errors are presented and then the container is stopped:
`Error: failed to build pipelines: failed to create "resourcedetection" processor, in pipeline "metrics/application_signals": failed creating detector type "eks": failed to create config: unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined`

`E! [telegraf] Error running agent: failed to build pipelines: failed to create "resourcedetection" processor, in pipeline "traces/application_signals": failed creating detector type "eks": failed to create config: unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined`

**Steps to reproduce**
Running the cloudwatch agent as a container in ECS Fargate

**What did you expect to see?**
The container should be in running state

**What did you see instead?**
The container fails to start

**What version did you use?**
v1.300045.0 

**What config did you use?**
```
- Name: cloudwatch-agent
    Image: public.ecr.aws/cloudwatch-agent/cloudwatch-agent:1.300044.0b793-arm64
    Essential: false
    Cpu: 32
    MemoryReservation: 100
    Memory: 200
    Secrets:
      - Name: CW_CONFIG_CONTENT
        ValueFrom: /config/all/common/ecs-cwagent/configuration
    LogConfiguration:
      LogDriver: awslogs
      Options:
        awslogs-group: !Ref LogGroup
        awslogs-region: !Ref AWS::Region
        awslogs-stream-prefix: cloudwatch-agent
```

**Environment**
ECS Fargate

**Additional context**
Previous version v1.300044.0 works fine but on Application Signals Console you get the following warning: 
The services in your account are using a version of the CloudWatch agent or EKS add-on that will no longer be supported by Application Signals after August 9, 2024. To continue using Application Signals please upgrade to the latest version. After August 9, 2024 Application Signals will not discover and display services instrumented with previous agent or EKS add-on versions. 

And also it shows this link https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Application-Signals-Agent-Versions.html that says versions older than 1.300040.0 would not longer be supported eventhough I'm running v1.300044.0

jefchien:
[...]

PR #1336, opened 2024-09-04
Add full support for ECS in Application Signals

bjrara:
# Description of changes
Currently, traffic from ECS is recognized as from generic platform. This PR adds ECS as a supported platform, and includes more ECS specific attributes in the Application Signals EMF log.

In addition, it removes resource detection for Task related attributes, and relies on client side to provide the accurate values. This is mainly to support deploying CloudWatch agent as Daemon or a standalone service on an ECS Cluster.

# License

# Tests
Added unit tests.

Integration tests
https://github.com/aws-observability/aws-application-signals-test-framework/actions/workflows/python-ecs-canary.yml
https://github.com/aws-observability/aws-application-signals-test-framework/actions/workflows/java-ecs-canary.yml

# Requirements
2. Run `make lint`

### DIFF
--- a/plugins/processors/awsapplicationsignals/common/types.go
+++ b/plugins/processors/awsapplicationsignals/common/types.go
@@ -28,6 +28,14 @@ const (
 	AttributeHost                = "Host"
 )
 
+// Platform attribute used as CloudWatch EMF log field.
+const (
+	MetricAttributeECSCluster                = "ECS.Cluster"
+	MetricAttributeECSTaskId                 = "ECS.TaskId"
+	MetricAttributeECSTaskDefinitionFamily   = "ECS.TaskDefinitionFamily"
+	MetricAttributeECSTaskDefinitionRevision = "ECS.TaskDefinitionRevision"
+)
+
 // Telemetry attributes used as CloudWatch EMF log fields.
 const (
 	MetricAttributeTelemetrySDK    = "Telemetry.SDK"
--- a/plugins/processors/awsapplicationsignals/internal/normalizer/attributesnormalizer.go
+++ b/plugins/processors/awsapplicationsignals/internal/normalizer/attributesnormalizer.go
@@ -42,10 +42,12 @@ var attributesRenamingForMetric = map[string]string{
 	attr.AWSRemoteResourceType:                 common.CWMetricAttributeRemoteResourceType,
 	attr.AWSRemoteDbUser:                       common.MetricAttributeRemoteDbUser,
 	attr.AWSRemoteResourceCfnPrimaryIdentifier: common.MetricAttributeRemoteResourceCfnPrimaryIdentifier,
+	attr.AWSECSClusterName:                     common.MetricAttributeECSCluster,
+	attr.AWSECSTaskID:                          common.MetricAttributeECSTaskId,
 }
 
 var resourceAttributesRenamingForTrace = map[string]string{
-	// these kubernetes resource attributes are set by the openTelemetry operator
+	// these kubernetes resource attributes are set by the OpenTelemetry operator
 	// see the code references from upstream:
 	// * https://github.com/open-telemetry/opentelemetry-operator/blob/0e39ee77693146e0924da3ca474a0fe14dc30b3a/pkg/instrumentation/sdk.go#L245
 	// * https://github.com/open-telemetry/opentelemetry-operator/blob/0e39ee77693146e0924da3ca474a0fe14dc30b3a/pkg/instrumentation/sdk.go#L305C43-L305C43
@@ -61,9 +63,9 @@ var attributesRenamingForTrace = map[string]string{
 	attr.AWSRemoteTarget: attr.AWSRemoteResourceIdentifier,
 }
 
-var copyMapForMetric = map[string]string{
-	// these kubernetes resource attributes are set by the openTelemtry operator
-	// see the code referecnes from upstream:
+var resourceToMetricAttributes = map[string]string{
+	// these kubernetes resource attributes are set by the OpenTelemetry operator
+	// see the code references from upstream:
 	// * https://github.com/open-telemetry/opentelemetry-operator/blob/0e39ee77693146e0924da3ca474a0fe14dc30b3a/pkg/instrumentation/sdk.go#L245
 	// * https://github.com/open-telemetry/opentelemetry-operator/blob/0e39ee77693146e0924da3ca474a0fe14dc30b3a/pkg/instrumentation/sdk.go#L305C43-L305C43
 	semconv.AttributeK8SDeploymentName:  common.AttributeK8SWorkload,
@@ -73,6 +75,8 @@ var copyMapForMetric = map[string]string{
 	semconv.AttributeK8SCronJobName:     common.AttributeK8SWorkload,
 	semconv.AttributeK8SPodName:         common.AttributeK8SPod,
 	semconv.AttributeAWSLogGroupNames:   "aws.log.group.names",
+	semconv.AttributeAWSECSTaskRevision: common.MetricAttributeECSTaskDefinitionRevision,
+	semconv.AttributeAWSECSTaskFamily:   common.MetricAttributeECSTaskDefinitionFamily,
 }
 
 const (
@@ -107,7 +111,7 @@ func (n *attributesNormalizer) copyResourceAttributesToAttributes(attributes, re
 	if isTrace {
 		return
 	}
-	for k, v := range copyMapForMetric {
+	for k, v := range resourceToMetricAttributes {
 		if resourceAttrValue, ok := resourceAttributes.Get(k); ok {
 			// print some debug info when an attribute value is overwritten
 			if originalAttrValue, ok := attributes.Get(k); ok {
--- a/plugins/processors/awsapplicationsignals/internal/resolver/attributesresolver.go
+++ b/plugins/processors/awsapplicationsignals/internal/resolver/attributesresolver.go
@@ -7,7 +7,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strings"
 
 	"go.opentelemetry.io/collector/pdata/pcommon"
 	semconv "go.opentelemetry.io/collector/semconv/v1.22.0"
@@ -16,7 +15,6 @@ import (
 	"github.com/aws/amazon-cloudwatch-agent/plugins/processors/awsapplicationsignals/common"
 	appsignalsconfig "github.com/aws/amazon-cloudwatch-agent/plugins/processors/awsapplicationsignals/config"
 	attr "github.com/aws/amazon-cloudwatch-agent/plugins/processors/awsapplicationsignals/internal/attributes"
-	"github.com/aws/amazon-cloudwatch-agent/translator/util/ecsutil"
 )
 
 const (
@@ -25,6 +23,7 @@ const (
 	AttributePlatformGeneric = "Generic"
 	AttributePlatformEC2     = "AWS::EC2"
 	AttributePlatformEKS     = "AWS::EKS"
+	AttributePlatformECS     = "AWS::ECS"
 	AttributePlatformK8S     = "K8s"
 )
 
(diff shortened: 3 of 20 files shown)


plugins/processors/awsapplicationsignals/internal/normalizer/attributesnormalizer.go
```diff
@@ -60,9 +62,9 @@ var attributesRenamingForTrace = map[string]string{
 	attr.AWSRemoteTarget: attr.AWSRemoteResourceIdentifier,
 }
 
-var copyMapForMetric = map[string]string{
+var resourceAttributesToMetric = map[string]string{
 	// these kubernetes resource attributes are set by the openTelemtry operator
```
jefchien, line 66: nit: Since you're fixing another typo from the line below.
```suggestion
	// these kubernetes resource attributes are set by the opentelemetry operator
```

plugins/processors/awsapplicationsignals/internal/normalizer/attributesnormalizer.go
```diff
@@ -60,9 +62,9 @@ var attributesRenamingForTrace = map[string]string{
 	attr.AWSRemoteTarget: attr.AWSRemoteResourceIdentifier,
 }
 
-var copyMapForMetric = map[string]string{
+var resourceAttributesToMetric = map[string]string{
```
jefchien, line 65: nit: The current name makes it sound like the resource attributes are being converted into metrics whereas I believe it's a copy and rename.
```suggestion
var resourceToMetricAttributes = map[string]string{
```

plugins/processors/awsapplicationsignals/internal/resolver/ecs.go
```diff
@@ -0,0 +1,94 @@
...
+			platformType:     AttributePlatformECS,
+			attributeMap:     DefaultInheritedAttributes,
+		},
+		clusterName: clusterName,
+	}
+}
+
+func getECSResourcesFromResourceAttributes(resourceAttributes pcommon.Map) (clusterName, taskId string) {
+	if clusterAttr, ok := resourceAttributes.Get(semconv.AttributeAWSECSClusterARN); ok {
+		parts := strings.Split(clusterAttr.Str(), "/")
```
jefchien, line 74: Do we need to validate the contents of this attribute or can we assume that it's in a valid ARN format?

translator/tocwconfig/sampleConfig/appsignals_and_eks_config.yaml
```diff
@@ -353,6 +353,7 @@ processors:
         detectors:
             - eks
             - env
+            - ecs
```
jefchien, line 356: Why is this getting added to the EKS configuration?

translator/translate/otel/processor/resourcedetection/configs/config.yaml
```diff
@@ -1,6 +1,38 @@
...
 override: true
 timeout: 2s
+ecs:
+  resource_attributes:
+    aws.ecs.cluster.arn:
+      enabled: true
+    aws.ecs.launchtype:
+      enabled: true
+    aws.ecs.task.arn:
+      enabled: false
```
jefchien, line 11: Why isn't this enabled? Isn't this used in the resolver?

plugins/processors/awsapplicationsignals/internal/resolver/ecs.go
```diff
@@ -0,0 +1,94 @@
...
+		resourceAttributesResolver: resourceAttributesResolver{
+			defaultEnvPrefix: defaultEnvPrefix,
+			platformType:     AttributePlatformECS,
+			attributeMap:     DefaultInheritedAttributes,
+		},
+		clusterName: clusterName,
+	}
+}
+
+func getECSResourcesFromResourceAttributes(resourceAttributes pcommon.Map) (clusterName, taskId string) {
```
jefchien, line 72: Why not get the task ID from the `aws.ecs.task.id` attribute?

plugins/processors/awsapplicationsignals/internal/resolver/ecs.go
```diff
@@ -0,0 +1,94 @@
...
+	if clusterAttr, ok := resourceAttributes.Get(semconv.AttributeAWSECSClusterARN); ok {
+		parts := strings.Split(clusterAttr.Str(), "/")
+		clusterName = parts[len(parts)-1]
+	}
+	if taskAttr, ok := resourceAttributes.Get(semconv.AttributeAWSECSTaskARN); ok {
```
jefchien, line 77: If we already got the cluser name from the Cluster ARN, why are we trying to parse this attribute again?

plugins/processors/awsapplicationsignals/internal/resolver/ecs.go
bjrara: `aws.ecs.task.id` is added to semantic convention `v1.25.0`, see [CHANGELOG](https://github.com/open-telemetry/semantic-conventions/blob/26e3779ba8367c7f4c1d5c6e23000cdb51455d86/CHANGELOG.md#v1250).

We choose to extract from `aws.ecs.task.arn` because it's more reliable.

plugins/processors/awsapplicationsignals/internal/resolver/ecs.go
bjrara: The cluster name is only set through `aws.ecs.task.arn` if it's empty, meaning the `aws.ecs.cluster.arn` doesn't exist. I don't think we would ever hit the logic, but I think it will be nice to have it as a protection. `aws.ecs.task.arn` is mainly used to extract task id.

plugins/processors/awsapplicationsignals/internal/normalizer/attributesnormalizer.go
bjrara: Updated

translator/tocwconfig/sampleConfig/appsignals_and_eks_config.yaml
bjrara: Added a new resourcedetection template for ecs.

plugins/processors/awsapplicationsignals/internal/resolver/ecs.go
bjrara: It should be valid, because the cluster name is set by resourcedetection processor.

plugins/processors/awsapplicationsignals/internal/resolver/attributesresolver.go
```diff
@@ -121,40 +118,15 @@ func getLocalEnvironment(attributes, resourceAttributes pcommon.Map, defaultEnvP
...
 	}
-	if defaultEnvPrefix == appsignalsconfig.PlatformECS {
-		if clusterName, _ := getECSClusterName(resourceAttributes); clusterName != "" {
-			return getDefaultEnvironment(defaultEnvPrefix, clusterName)
-		}
-		if clusterName := ecsutil.GetECSUtilSingleton().Cluster; clusterName != "" {
-			return getDefaultEnvironment(defaultEnvPrefix, clusterName)
-		}
-	} else if defaultEnvPrefix == appsignalsconfig.PlatformEC2 {
+	if defaultEnvPrefix == appsignalsconfig.PlatformEC2 {
```
sky333999, line 121: Would removing this old logic end up being a breaking change if someone updates their agent version but not the instr agent version?
I guess we can document valid version tuples that ppl have to stick to.

plugins/processors/awsapplicationsignals/common/types.go
```diff
@@ -28,6 +28,14 @@ const (
 	AttributeHost                = "Host"
 )
 
+// Platform attribute used as CloudWatch EMF log field.
+const (
+	MetricAttributeECSCluster                = "ECS.Cluster"
+	MetricAttributeECSTaskId                 = "ECS.TaskId"
+	MetricAttributeECSTaskDefinitionFamily   = "ECS.TaskDefinitionFamily"
+	MetricAttributeECSTaskDefinitionRevision = "ECS.TaskDefinitionRevision"
```
sky333999, line 36: Is it intentional to have nothing about the ECS.Service?

sky333999:
I was expecting to see `awsemf` translator changes to introduce `appsignals_config_ecs.yaml` defining a schema for ECS. Is that not required?

translator/translate/otel/processor/resourcedetection/configs/ecs_config.yaml
```diff
@@ -0,0 +1,38 @@
+detectors: [env, ecs, ec2]
```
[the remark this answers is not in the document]
bjrara: May I know what existing detectors you are referring to?

plugins/processors/awsapplicationsignals/common/types.go
bjrara: Yes, today, ECS.Service is not available in the otel attributes.

bjrara:
> I was expecting to see `awsemf` translator changes to introduce `appsignals_config_ecs.yaml` defining a schema for ECS. Is that not required?

EMF template is shared across non EKS/K8s platforms, it reuses the generic one.

plugins/processors/awsapplicationsignals/internal/resolver/ecs.go
```diff
@@ -0,0 +1,94 @@
...
+
+func (e *ecsResourceAttributesResolver) Process(attributes, resourceAttributes pcommon.Map) error {
+	for attrKey, mappingKey := range e.attributeMap {
+		if val, ok := resourceAttributes.Get(attrKey); ok {
+			attributes.PutStr(mappingKey, val.Str())
+		}
+	}
+
+	clusterName := e.clusterName
+	clusterNameOverwrite, taskId := getECSResourcesFromResourceAttributes(resourceAttributes)
```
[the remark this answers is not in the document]
bjrara: That's a good question. I agree this PR should better handle the priority of different host in customization.

In the next update, the priority of setting the environment is
1. aws.local.environment (which comes from user specified deployment.environment)
2. user specified aws.hostedin.environment (should be deprecated soon)
3. user specified hosted_in
4. auto detected aws.ecs.cluster.arn
5. auto detected aws.ecs.task.arn
6. auto detected cluster name from CWA
7. hardcoded `default`

plugins/processors/awsapplicationsignals/internal/resolver/attributesresolver.go
bjrara: No it won't. It's fully controlled by the agent itself. Perhaps you can explain more about where the concern comes from?

plugins/processors/awsapplicationsignals/internal/resolver/ecs.go
jefchien: Can we use it if it exists and then fallback on extracting from the `aws.ecs.task.arn`?

plugins/processors/awsapplicationsignals/internal/resolver/ecs.go
jefchien: The resourcedetection processor will only set it if the attribute doesn't already exist on the metric. So if I set it to "myclustername" on the SDK for example or if I sent a custom metric to the endpoint, we'd end up using "myclustername" as the attribute. Can we check the `parts` length at least?

translator/translate/otel/processor/resourcedetection/configs/ecs_config.yaml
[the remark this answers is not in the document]
sky333999: I meant the config.yaml with eks, env, ec2. But I see you had it that way originally and changed it later per Jeffrey's suggestion.
Im fine either way. This is probably cleaner so we dont even attempt ECS on K8s.

plugins/processors/awsapplicationsignals/internal/resolver/attributesresolver.go
sky333999: I mean currently the agent is extracting the ECS attributes and setting those as a generic platform. But going forward, the agent wont be doing that and instead rely on the instr agent to send these.
So if the agent version is updated but the instr agent is not, would it appear like we now no longer send these attributes.

translator/translate/otel/processor/resourcedetection/configs/ecs_config.yaml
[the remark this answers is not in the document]
bjrara: Yes. I think Jeffrey had that concern because of an issue reported in the last release. https://github.com/aws/amazon-cloudwatch-agent/issues/1328

plugins/processors/awsapplicationsignals/internal/resolver/attributesresolver.go
bjrara: Thanks for the explanation.

Today, the CloudWatch agent gets the `ClusterName` if it's unavailable from the instr agent, and sets it in `Environment`. There's no additional attributes the agent adds to spans or metrics because `ecs` isn't registered as a `resourcedetection` processor.

If the CloudWatch agent is upgraded, but the instr agent stays old, customers won't benefit from the new versions, but the existing behavior should not break (except that we will upgrade `PlatformType` from `Generic` to `ECS` which will be handled to be backward compatible from backend side).

After the agent upgrade, `ecs` will be registered as `resourcedetection` processor so it adds/overwrites attributes in the `Resource`. Rather than worrying some attribute may be missing, I would call out risks in `overwrites`. For example, the cluster arn will be overwritten by the `resourcedetection` in the agent, which means if the application and the agent lives in different clusters, the cluster name will be populated wrong. However, since we don't support such cases today in CloudWatch agent (I think), it should be fine?

jefchien (approved):

sky333999 (approved):

PR #1347, opened 2024-09-11
Specifying region in EC2 and STS clients in entity store

bryce-carey:
# Description of the issue
Resolving issue where the region is not specified when creating the STS and EC2 metadata clients.

# Description of changes
Adds region to the AWS config when instantiating the clients.

# License

# Tests
Manual in us-east-1

# Requirements
2. Run `make lint`

### DIFF
--- a/extension/entitystore/extension.go
+++ b/extension/entitystore/extension.go
@@ -5,9 +5,12 @@ package entitystore
 
 import (
 	"context"
+	"net/http"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/client"
+	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
 	"github.com/aws/aws-sdk-go/service/sts"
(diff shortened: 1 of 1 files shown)


extension/entitystore/extension.go
```diff
@@ -225,6 +226,7 @@ func getEC2Provider(region string, ec2CredentialConfig *configaws.CredentialConf
 	return ec2.New(
 		ec2CredentialConfig.Credentials(),
 		&aws.Config{
+			Region:   aws.String(region),
```
nathalapooja, line 229: Can we add unit tests ?

extension/entitystore/extension.go
```diff
@@ -203,6 +203,7 @@ func (e *EntityStore) shouldReturnEntity() bool {
 		e.stsClient = sts.New(
 			e.nativeCredential,
 			&aws.Config{
+				Region:   aws.String(e.config.Region),
```
zhihonl, line 206: this may not even be an issue but just thought I would mention. 

`e.config.region` comes from default EC2 region. But if customers configures the log agent credential to use a different account in a different region(i think this is possible?), there will be a mismatch in the region used for `e.nativeCredential` and the region used in this line.

bryce-carey: Is there a way to get the customer configured region here? So when we do the config translation, we don't feed in the customer configuration to the entity store?

nathalapooja (approved):

zhihonl (approved):

PR #1349, opened 2024-09-12
Caches result of shouldReturnEntity() function to reduce number of STS GetCallerIdentity requests.

dchappa:
# Description of the issue
In the current implementation, we currently make a request to the `sts:GetCallerIdentity` so that we can verify that the account ID for the instance matches the account ID that is assuming the role for the current credential. 

However, this check was being done at every time a PutLogEvent call was being made, causing impact to the STS service.

# Description of changes
The logic of checking instance account vs. the account assuming the role remains untouched. This PR instead caches that result, and now only periodically checks the match on an interval. 

With the current implementation, a successful match will set the interval to 24 hours, and a failed match will set the interval to 5 minutes. 

# License

# Tests
Unit testing has shown that the agent is not making calls to STS as aggressively.

Manual testing: I added a debug line right after a successful GetCallerIdentity call is made. Agent Logs show that while we still continue to successfully publish logs to CWL, we are clearly reducing the number of `GetCallerIdentity` calls.
```
[ec2-user@ip-172-31-75-235 amazon-cloudwatch-agent]$ sudo less /opt/aws/amazon-cloudwatch-agent/logs/amazon-cloudwatch-agent.log | grep GetCallerIdentity
2024-09-13T14:40:10Z I! {"caller":"entitystore/extension.go:244","msg":"Made call to STS GetCallerIdentity","kind":"extension","name":"entitystore"}

[ec2-user@ip-172-31-75-235 amazon-cloudwatch-agent]$ sudo less /opt/aws/amazon-cloudwatch-agent/logs/amazon-cloudwatch-agent.log | grep 'Pusher published' |tail -n 5
2024-09-13T14:44:59Z D! [outputs.cloudwatchlogs] Pusher published 1 log events to group: i-03b6ab25131f0f8a0 stream: i-03b6ab25131f0f8a0 with size 0 KB in 23.811657ms.
2024-09-13T14:45:04Z D! [outputs.cloudwatchlogs] Pusher published 2 log events to group: i-03b6ab25131f0f8a0 stream: i-03b6ab25131f0f8a0 with size 0 KB in 10.225546ms.
2024-09-13T14:45:14Z D! [outputs.cloudwatchlogs] Pusher published 1 log events to group: i-03b6ab25131f0f8a0 stream: i-03b6ab25131f0f8a0 with size 0 KB in 31.33703ms.
2024-09-13T14:45:19Z D! [outputs.cloudwatchlogs] Pusher published 2 log events to group: i-03b6ab25131f0f8a0 stream: i-03b6ab25131f0f8a0 with size 0 KB in 13.017723ms.
2024-09-13T14:45:29Z D! [outputs.cloudwatchlogs] Pusher published 2 log events to group: i-03b6ab25131f0f8a0 stream: i-03b6ab25131f0f8a0 with size 0 KB in 31.463429ms.


[ec2-user@ip-172-31-75-235 amazon-cloudwatch-agent]$ sudo less /opt/aws/amazon-cloudwatch-agent/logs/amazon-cloudwatch-agent.log | grep GetCallerIdentity
2024-09-13T14:40:10Z I! {"caller":"entitystore/extension.go:244","msg":"Made call to STS GetCallerIdentity","kind":"extension","name":"entitystore"}
```

# Requirements
2. Run `make lint`

### DIFF
--- a/extension/entitystore/extension.go
+++ b/extension/entitystore/extension.go
@@ -36,6 +36,8 @@ const (
 	Type                 = "Type"
 	Name                 = "Name"
 	Environment          = "Environment"
+	successfulRefresh    = time.Hour * 24
+	failureRefresh       = time.Minute * 5
 )
 
 type ec2ProviderType func(string, *configaws.CredentialConfig) ec2iface.EC2API
@@ -77,6 +79,14 @@ type EntityStore struct {
 	metadataprovider ec2metadataprovider.MetadataProvider
 
 	stsClient stsiface.STSAPI
+
+	// nextRefreshTime is the time when the next refresh should happen
+	// determining whether we should return the EntityStore
+	nextRefreshTime time.Time
+
+	// currentTime will always return the current time. This parameter exists mainly for testing
+	// purposes, so we can mock the time
+	currentTime func() time.Time
 }
 
 var _ extension.Extension = (*EntityStore)(nil)
@@ -130,7 +140,9 @@ func (e *EntityStore) NativeCredentialExists() bool {
 
 // CreateLogFileEntity creates the entity for log events that are being uploaded from a log file in the environment.
 func (e *EntityStore) CreateLogFileEntity(logFileGlob LogFileGlob, logGroupName LogGroupName) *cloudwatchlogs.Entity {
-	if !e.shouldReturnEntity() {
+	// To prevent unnecessary calls to the sts:GetCallerIdentity API, we verify whether we should return the entity every 24 hours on success and
+	// every 5 minutes on failure.
+	if !e.needToRefreshInterval() {
 		return nil
 	}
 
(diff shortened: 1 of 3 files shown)


extension/entitystore/extension.go
```diff
@@ -36,6 +36,8 @@ const (
 	Type                 = "Type"
 	Name                 = "Name"
 	Environment          = "Environment"
+	successfulRefresh    = time.Hour * 24
+	failureRefresh       = time.Minute * 5
```
dchappa, line 40: I'm definitely not married to these intervals, particularly the failure one. Not sure what a good interval is for this, but one request per 5 minutes should still be much better than X amount of PLE calls.

extension/entitystore/extension_test.go
```diff
@@ -332,12 +344,115 @@ func TestEntityStore_shouldReturnRID(t *testing.T) {
...
+				metadataprovider: tt.fields.metadataprovider,
+				stsClient:        tt.fields.stsClient,
+				nativeCredential: tt.fields.nativeCredential,
+				currentTime: func() time.Time {
+					currentTime = currentTime.Add(time.Duration(tt.mockTimeSkipInterval) + 1)
+					return currentTime
+				},
+			}
+			for i := 0; i < tt.numberOfFunctionCalls; i++ {
+				e.CreateLogFileEntity(glob, group)
```
dchappa, line 442: This is the function called by the logs agent when making the PLE call, so thought it was best to test by calling this function.

extension/entitystore/extension_test.go
Paramadon, line 1: Can we consider adding more tests that simulate the conditions where the IMDS or STS calls fail? Just to ensure behavior is correct in that case.

PR #1370, opened 2024-10-04
Update AMI cleaner script to properly handle macOS images

varunch77:
# Description of the issue
We run a daily GitHub workflow ([AWS Daily Resources Cleaner](https://github.com/aws/amazon-cloudwatch-agent/actions/workflows/clean-aws-resources.yml)) to delete AMIs older than 60 days. This works for non-macOS AMIs because we use ImageBuilder to handle building them. However, macOS AMIs are not currently supported by ImageBuilder. As a result, our macOS AMIs are deleted by the workflow when they cross the 60 day threshold and are not automatically replaced, leaving a gap in our integration testing. 

# Description of changes
I modified the logic of the script to specifically handle macOS AMIs. If there is only one image, it won't be deleted regardless of age. If there are multiple images and they are all past the expiration date, the script will keep the youngest and delete the rest. Otherwise, it'll keep the images within the limit and delete those that are past it. The script also avoids the edge case where multiple AMIs are created on the same day (and set to be deleted together). It ensures that they don't get deleted at the same time, leaving a gap in our macOS images. 

To accomplish this, I use a map that has each macOS version and architecture as a key and a list of corresponding machine images as its value. 

# License

# Tests
 I ran the script with a lowered expiration threshold (60 days -> 7 days) to check that AMIs identified to be deleted were all older than 7 days—this was a quick verification that original functionality with non macOS AMIs was unchanged.

I also created multiple AMIs for each macOS version and architecture and ran the script with the following behavior:
  - 1 AMI - under the age limit (didn’t delete)
  - 1 AMI - over the age limit (didn’t delete)
  - 2 AMIs - both under the age limit (didn’t delete any)
  - 2 AMIs - one under the age limit, one over the age limit (deleted the one over, kept the one under)
  - 2 AMIs - both over the age limit (deleted the older one, kept the younger one)
  - 3 AMIs - all under the age limit (didn't delete any)
  - 3 AMIs - one over the age limit (deleted the older one, kept the two under the limit)
  - 3 AMIs - two over the age limit (deleted the two over, kept the one under)
  - 3 AMIs - all over the age limit (kept the youngest, deleted the older two)

This matched intended behavior. We can be confident that a situation with more AMIs would reduce to an example listed above.

### DIFF
--- a/tool/clean/clean_ami/clean_ami.go
+++ b/tool/clean/clean_ami/clean_ami.go
@@ -59,13 +59,13 @@ func deregisterAMIs(ctx context.Context, ec2client *ec2.Client, images []types.I
     for _, image := range images {
 		if image.Name != nil && image.ImageId != nil && image.CreationDate != nil {
 			log.Printf("Try to delete ami %v tags %v image id %v image creation date raw %v", *image.Name, image.Tags, *image.ImageId, *image.CreationDate)
-			// deregisterImageInput := &ec2.DeregisterImageInput{ImageId: image.ImageId}
-			// _, err := ec2client.DeregisterImage(ctx, deregisterImageInput)
+			deregisterImageInput := &ec2.DeregisterImageInput{ImageId: image.ImageId}
+			_, err := ec2client.DeregisterImage(ctx, deregisterImageInput)
  
-			// if err != nil && errList != nil {
-			// 	log.Printf("Error while deregistering ami %v", *image.Name)
-			// 	*errList = append(*errList, err)
-			// }
+			if err != nil && errList != nil {
+				log.Printf("Error while deregistering ami %v", *image.Name)
+				*errList = append(*errList, err)
+			}
 		}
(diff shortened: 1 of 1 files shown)


zhihonl (approved):

dchappa:
Nice job for manually testing all the scenarios, and glad that we shouldn't have any issues with `macos` testing anymore.

Not blocking - but given that there are some new nuances to the way we handle the cleanup images (just based off of your PR description), I think we could stand benefit to add some unit tests for this. That way in the future, we can also see what sort of behavior we expected when we made these changes. 

I see that for a lot of the scripts in `tool/` we don't have unit tests, so maybe we can save it for a future improvement.

varunch77:
@dchappa Agree, I think more comprehensive unit tests—while not super critical for a cleanup script—would definitely make it easier to ensure everything is working as intended as we make more modifications down the line

PR #1377, opened 2024-10-07
Fix config detection when path is invalid.

jefchien:
# Description of the issue
On ECS, the `/etc/cwaconfig` directory may not be available. This currently causes a panic because the `DirEntry` is nil.

# Description of changes
Adds a check to make sure this doesn't change the current behavior.

# License

# Tests
Updated unit tests.

# Requirements
2. Run `make lint`


okankoAMZ (approved):
LGTM!

mitali-salvi (approved):

PR #1378, opened 2024-10-07
Update metrics handling in applicationsignalsprocessor

bjrara:
# Description of the issue
This PR is to prepare for the runtime metrics release in Application Signals.

# Description of changes
1. Drop metrics if metric doesn't contain `Telemetry.Source` attribute.
2. Try to copy `aws.local.service` from resource attributes if the attribute is not found on metrics. This is only used for runtime metrics.
3. Respect custom metric names.

# License

# Tests

# Requirements
4. Run `make lint`


mxiamxia (approved):

plugins/processors/awsapplicationsignals/internal/prune/metric_pruner.go
```diff
@@ -21,6 +22,9 @@ func (p *MetricPruner) ShouldBeDropped(attributes pcommon.Map) (bool, error) {
 				return true, errors.New("Metric attribute " + attributeKey + " must contain only ASCII characters.")
 			}
 		}
+		if _, ok := attributes.Get(common.MetricAttributeTelemetrySource); !ok {
```
lisguo, line 25: are there any risks in dropping metrics without Telemetry.Source? Are there any edge cases where a customer might emit application signals metrics without this attribute?

plugins/processors/awsapplicationsignals/processor.go
```diff
@@ -137,7 +138,10 @@ func (ap *awsapplicationsignalsprocessor) processMetrics(ctx context.Context, md
 			metrics := ils.Metrics()
 			for k := 0; k < metrics.Len(); k++ {
 				m := metrics.At(k)
-				m.SetName(metricCaser.String(m.Name())) // Ensure metric name is in sentence case
+				// Check if the first letter of the metric name is not capitalized
+				if len(m.Name()) > 0 && !unicode.IsUpper(rune(m.Name()[0])) {
```
lisguo, line 142: will there be documentation for custom metrics to ensure the first letter is not capitalized?

plugins/processors/awsapplicationsignals/internal/prune/metric_pruner.go
bjrara: In the case where customer is using preview SDK, all the metrics will be dropped. Those SDKs are deprecated though. Also even the traffic is sent to logs, Application Signals has stopped processing those logs emitted from preview SDK while using the new CloudWatch agent.

plugins/processors/awsapplicationsignals/processor.go
bjrara: It's a good question. I personally prefer respecting the original metric name...

I don't know if there will be documentation for naming convention, perhaps we should revisit this logic whether it makes sense to change the letter case.

sky333999 (approved):

PR #1392, opened 2024-10-15
Fix logs-only check in agent and container environment merge

jefchien:
# Description of the issue
If the YAML file passed in is not the `paths.YamlConfigPath`, then it will not check to see if the file exists before trying to start up the OTEL collector part of the agent. For logs-only users, the agent does not need a YAML file and should not use one. If using the control script or start-amazon-cloudwatch-agent, the `paths.YamlConfigPath` will be correct, so the agent is unimpacted.

# Description of changes
Changed the check so it only confirms that a single YAML file is provided. If multiple YAML files are provided, then assume that the intention is to use them.

Fixes the container environment merge. For container environments, the `CW_OTEL_CONFIG_CONTENT` environment variable is available if only one YAML is provided. Matches the behavior of the translator and the JSON `CW_CONFIG_CONTENT`.

# License

# Tests
Updated unit tests.

# Requirements
2. Run `make lint`


cmd/amazon-cloudwatch-agent/amazon-cloudwatch-agent.go
```diff
@@ -310,18 +310,16 @@ func runAgent(ctx context.Context,
...
 		go logAgent.Run(ctx)
 
-		// If only the default CWAgent config yaml is provided and does not exist, then ASSUME
+		// If only a single YAML is provided and does not exist, then ASSUME the agent is
 		// just monitoring logs since this is the default when no OTEL config flag is provided.
 		// So just start Telegraf.
-		if isOnlyDefaultConfigPath {
+		if len(fOtelConfigs) == 1 {
 			_, err = os.Stat(fOtelConfigs[0])
 			if errors.Is(err, os.ErrNotExist) {
```
movence, line 324: can we add a warning message here? or it won't be necessary? I'm just thinking about a case where there is a typo in a given path.

movence (approved):

sky333999 (approved):

PR #1413, opened 2024-11-05
Add null checks to service provider in entitystore extension

nathalapooja:
# Description of the issue
With LogsAgent plugin, Agent has a race condition that can cause some customers to lose logs as the agent panics.

# Description of changes
The panic is due to LogsAgent calling the CreateLogFileEntity method in entity store extension before the extension is initialized/started. Added a null check for serviceProvider to return nil when it is not initialized.

Also added null checks in other methods of entitystore extension to avoid panics.
# License

# Tests
Unit Tests

# Requirements
2. Run `make lint`

### DIFF
--- a/extension/entitystore/extension.go
+++ b/extension/entitystore/extension.go
@@ -139,6 +139,9 @@ func (e *EntityStore) NativeCredentialExists() bool {
 
 // CreateLogFileEntity creates the entity for log events that are being uploaded from a log file in the environment.
 func (e *EntityStore) CreateLogFileEntity(logFileGlob LogFileGlob, logGroupName LogGroupName) *cloudwatchlogs.Entity {
+	if e.serviceprovider == nil {
+		return nil
+	}
 	serviceAttr := e.serviceprovider.logFileServiceAttribute(logFileGlob, logGroupName)
 
 	keyAttributes := e.createServiceKeyAttributes(serviceAttr)
@@ -153,6 +156,9 @@ func (e *EntityStore) CreateLogFileEntity(logFileGlob LogFileGlob, logGroupName 
 
 // GetMetricServiceNameAndSource gets the service name source for service metrics if not customer provided
 func (e *EntityStore) GetMetricServiceNameAndSource() (string, string) {
+	if e.serviceprovider == nil {
+		return "", ""
+	}
 	return e.serviceprovider.getServiceNameAndSource()
 }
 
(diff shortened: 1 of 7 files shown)


zhihonl:
Per discussion with @jefchien, we should add a new struct variable [here](https://github.com/aws/amazon-cloudwatch-agent/blob/7ea621c8700610a613041291a102ae32fa7ed8f2/extension/entitystore/extension.go#L48) named `ready`, then at the [end of the extension start function](https://github.com/aws/amazon-cloudwatch-agent/blob/7ea621c8700610a613041291a102ae32fa7ed8f2/extension/entitystore/extension.go#L103), we set ready to true.

Then in [GetEntityStore](https://github.com/aws/amazon-cloudwatch-agent/blob/b67a45d67464c19ac6025bccd37bd7228aebaab2/extension/entitystore/factory.go#L18) function implementation, we should return nil if `entitystore.ready` is false. This help us bulletproof against race conditions

zhihonl:
Per discussion with @jefchien , for thread safety when accessing the entitystore extension, we should have a RWMutex initialized [here](https://github.com/aws/amazon-cloudwatch-agent/blob/b67a45d67464c19ac6025bccd37bd7228aebaab2/extension/entitystore/factory.go#L15). Then have a read lock for [GetEntityStore function](https://github.com/aws/amazon-cloudwatch-agent/blob/b67a45d67464c19ac6025bccd37bd7228aebaab2/extension/entitystore/factory.go#L18) and write lock for [createExtension function](https://github.com/aws/amazon-cloudwatch-agent/blob/b67a45d67464c19ac6025bccd37bd7228aebaab2/extension/entitystore/factory.go#L36)

zhihonl (approved):

extension/entitystore/factory.go
```diff
@@ -13,10 +14,16 @@ import (
...
 	TypeStr, _  = component.NewType("entitystore")
 	entityStore *EntityStore
+	mutex       sync.RWMutex
 )
 
 func GetEntityStore() *EntityStore {
-	return entityStore
+	mutex.RLock()
+	defer mutex.RUnlock()
+	if entityStore != nil && entityStore.ready {
```
jefchien, line 23: The lock only protects the `entityStore`. It doesn't protect multiple threads from reading/writing to the `ready` bool.

extension/entitystore/extension.go
```diff
@@ -48,6 +48,7 @@ type EntityStore struct {
 	logger *zap.Logger
 	config *Config
 	done   chan struct{}
+	ready  bool
```
jefchien, line 51: If you don't want to use a lock, you could use an atomic.Bool (https://pkg.go.dev/sync/atomic#Bool)

extension/entitystore/factory.go
nathalapooja: We are writing ready bool only once in entitystore extension Start method. Do we still need to have lock on it?

jefchien (approved):

zhihonl (approved):

PR #1423, opened 2024-11-12
Adding Statuscode Handler to Agent Health Extension 

Paramadon:
---

### **Description of the Issue**

The CloudWatch Agent lacked an `agenthealth` configuration to monitor API health by tracking HTTP status codes for specific responses. Monitoring the status codes (`200`, `400`, `408`, `419`, and `429`) across all APIs is critical for diagnosing issues and ensuring comprehensive observability of API behaviors. Without this configuration, users were unable to quickly identify trends in API health metrics or correlate specific status codes with performance issues.

---

### **Changes Made**

- **Added `agenthealth` Configuration**:
  - Configured the CloudWatch Agent to track the following status codes across all APIs:
    - `200`: Success  
    - `400`: Bad Request  
    - `408`: Request Timeout  
    - `419`: Authentication Timeout  
    - `429`: Too Many Requests  
  - This ensures that health metrics for these common response codes are captured and reported.

- **Updated CloudWatch Agent Configuration JSON**:
  - Introduced a new section under the metrics collection configuration to specify `agenthealth` metrics.
  - Ensured compatibility with existing metrics and logs collection by integrating the new configuration seamlessly.

- **Validated the Configuration**:
  - Tested the updated configuration with the CloudWatch Agent to ensure the new metrics are collected and reported correctly.
  - Confirmed that the metrics appear as expected in CloudWatch for all specified APIs.

---

### **Impact**

- Provides real-time visibility into API status codes, helping identify anomalies or patterns that indicate issues.
- Enhances observability for critical services by adding detailed health metrics.

---

### **Testing**

- Deployed the updated CloudWatch Agent health configuraion
- Verified that metrics for status codes `200`, `400`, `408`, `419`, and `429` are captured in the header as you can see from the image below:
Screenshot 2024-11-15 at 4 36 21 PM

---


extension/agenthealth/handler/stats/provider/interval.go
```diff
@@ -39,8 +40,17 @@ func (p *intervalStats) Stats(string) agent.Stats {
...
-		stats = value.(agent.Stats)
+		if s, ok := value.(agent.Stats); ok {
+			stats = s
+		} else {
+			log.Println("Error: Loaded value is not of type agent.Stats")
+		}
+	}
+
+	if stats.StatusCodes == nil {
+		stats.StatusCodes = make(map[string][5]int)
```
mitali-salvi, line 53: Does this map need to be created with a int array of len 5 ? if we add more status code in the future it will be cumbersome to update all the references

mitali-salvi (commented):
Could you fix the linting issues ?

tool/clean/clean_host/clean_host.go
```diff
@@ -66,6 +67,7 @@ func terminateInstances(cxt context.Context, ec2client *ec2.Client) {
 	for {
 		instanceIds := make([]string, 0)
 		expirationDateInstance := time.Now().UTC().Add(clean.KeepDurationOneDay)
+		log.Println("yooooo we in above describe instances for host")
```
musa-asad, line 70: testing logs still present

extension/agenthealth/handler/stats/agent/agent.go
```diff
@@ -118,3 +159,39 @@ type StatsConfig struct {
...
+func NewStatusCodeOperationsFilter() OperationsFilter {
+	allowed := collections.NewSet[string](StatusCodeOperations...)
+
+	return OperationsFilter{
+		operations: allowed,
+		allowAll:   allowed.Contains(AllowAllOperations),
+	}
+}
+
+func NewStatusCodeAndOtherOperationsFilter(operations []string) OperationsFilter {
```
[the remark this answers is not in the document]
dricross: Perhaps NewStatusCodeAndOtherOperationsFilter simply calls NewStatusCodeOperationsFilter

```suggestion
func NewStatusCodeAndOtherOperationsFilter(operations []string) OperationsFilter {
   filter := NewStatusCodeOperationsFilter()
   for _, operation := range operations {
	filter.allowed.Add(operation)
   }
   return filter
```

cmd/amazon-cloudwatch-agent/amazon-cloudwatch-agent.go
jefchien, line 1: Why are we changing this file?

extension/agenthealth/handler/stats/agent/agent.go
```diff
@@ -80,6 +82,45 @@ func (s *Stats) Merge(other Stats) {
...
+					existing[1] + value[1], // 400
+					existing[2] + value[2], // 408
+					existing[3] + value[3], // 413
+					existing[4] + value[4], // 429
+				}
+
+				log.Printf(
+					"Updated key %s: 200=%d, 400=%d, 408=%d, 413=%d, 429=%d",
+					key, s.StatusCodes[key][0], s.StatusCodes[key][1], s.StatusCodes[key][2], s.StatusCodes[key][3], s.StatusCodes[key][4],
+				)
```
dricross, line 115: Are these logs going to flood the log file?

extension/agenthealth/handler/stats/agent/agent_test.go
```diff
@@ -10,105 +10,72 @@ import (
...
-	assert.EqualValues(t, 1, *stats.EnhancedContainerInsights)
-	assert.EqualValues(t, 0, *stats.RunningInContainer)
-	assert.EqualValues(t, "RegionType", *stats.RegionType)
-	assert.EqualValues(t, "Mode", *stats.Mode)
+
+	assert.Equal(t, [5]int{3, 5, 7, 9, 11}, stats.StatusCodes["operation1"])
+	assert.Equal(t, [5]int{0, 1, 2, 3, 4}, stats.StatusCodes["operation2"])
 }
 
-func TestMarshal(t *testing.T) {
```
dricross, line 55: We should keep the existing marshal tests

extension/agenthealth/handler/stats/provider/interval.go
```diff
@@ -39,8 +40,17 @@ func (p *intervalStats) Stats(string) agent.Stats {
...
 func (p *intervalStats) getStats() agent.Stats {
 	var stats agent.Stats
+
 	if value := p.stats.Load(); value != nil {
-		stats = value.(agent.Stats)
+		if s, ok := value.(agent.Stats); ok {
+			stats = s
+		} else {
+			log.Println("Error: Loaded value is not of type agent.Stats")
+		}
```
jefchien, line 49: Why do we need this now?

extension/agenthealth/handler/stats/provider/interval.go
jefchien: Why are we initializing it to an empty map here? What is it used for?

extension/agenthealth/handler/stats/agent/agent.go
```diff
@@ -80,6 +82,45 @@ func (s *Stats) Merge(other Stats) {
 	if other.EntityRejected != nil {
 		s.EntityRejected = other.EntityRejected
 	}
+	if other.StatusCodes != nil {
+		log.Println("Merging status codes from another source.")
```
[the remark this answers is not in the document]
jefchien: The merge function should not have logs.

extension/agenthealth/handler/stats/provider/process_test.go
```diff
@@ -77,6 +75,7 @@ func TestProcessStats(t *testing.T) {
 	mock.mu.Unlock()
 	provider.refresh()
 	assert.Eventually(t, func() bool {
-		return provider.getStats() == agent.Stats{}
+		stats := provider.getStats()
+		return len(stats.StatusCodes) == 0 //map isn't comparable so we can just do this
```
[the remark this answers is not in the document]
jefchien: Why did we change this to only look at status codes?

extension/agenthealth/handler/stats/handler.go
```diff
@@ -21,12 +21,25 @@ const (
 	headerKeyAgentStats = "X-Amz-Agent-Stats"
 )
 
-func NewHandlers(logger *zap.Logger, cfg agent.StatsConfig) ([]awsmiddleware.RequestHandler, []awsmiddleware.ResponseHandler) {
-	filter := agent.NewOperationsFilter(cfg.Operations...)
+func NewHandlers(logger *zap.Logger, cfg agent.StatsConfig, statuscodeonly bool) ([]awsmiddleware.RequestHandler, []awsmiddleware.ResponseHandler) {
+	statusCodeFilter := agent.NewStatusCodeOperationsFilter()
+	if statuscodeonly {
+		statusCodeStats := provider.GetStatusCodeStats(statusCodeFilter)
```
dricross, line 27: In both branches (statuscodeonly true or false), you are executing this line so you can move it outside the if block

extension/agenthealth/handler/stats/agent/agent.go
```diff
@@ -118,3 +159,39 @@ type StatsConfig struct {
...
+	"DescribeServices",
+	"DescribeTaskDefinition",
+	"ListServices",
+	"ListTasks",
+	"CreateLogGroup",
+	"CreateLogStream",
+}
+
+// NewStatusCodeOperationsFilter creates a new filter for allowed operations and status codes.
+func NewStatusCodeOperationsFilter() OperationsFilter {
```
dricross, line 177: Keep these `OperationsFilter ` functions close to the type definition (move `StatsConfig` down)

extension/agenthealth/handler/stats/provider/statuscode.go
```diff
@@ -0,0 +1,180 @@
...
+	case "ListTasks":
+		return "lt"
+	case "CreateLogGroup":
+		return "clg"
+	case "CreateLogStream":
+		return "cls"
+	case "AssumeRole":
+		return "sts"
+	default:
+		return operation
```
jefchien, line 149: So if we don't recognize the operation, then we'll still use the full name in the map?

extension/agenthealth/config.go
```diff
@@ -12,6 +12,7 @@ import (
 type Config struct {
 	IsUsageDataEnabled bool              `mapstructure:"is_usage_data_enabled"`
 	Stats              agent.StatsConfig `mapstructure:"stats"`
+	StatusCodeOnly     *bool             `mapstructure:"is_status_code_only,omitempty"`
```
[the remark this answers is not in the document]
dricross: Agreed with mitali. Instead of a pointer, can we treat the default value as false?

Additionally, can you explain a bit more what this field is used for? It seems like a way to set the agent health extension such that it turns the other things off. I think that may be confusing because you could set `IsUsageDataEnabled` to true and also set `StatusCodeOnly` to true, so it's not clear what it would actually do (without looking at the implementation).

plugins/inputs/prometheus/prometheus.go
jefchien, line 1: Seems incomplete. Remove from this PR.

extension/agenthealth/handler/stats/handler.go
```diff
@@ -21,12 +21,25 @@ const (
...
+	stats := newStatsHandler(logger, filter, []agent.StatsProvider{
+		clientStats,
+		provider.GetProcessStats(),
+		provider.GetFlagsStats(),
+		statusCodeStats,
+	})
 	agent.UsageFlags().SetValues(cfg.UsageFlags)
-	return []awsmiddleware.RequestHandler{stats, clientStats}, []awsmiddleware.ResponseHandler{clientStats}
+
+	return []awsmiddleware.RequestHandler{stats, clientStats, statusCodeStats}, []awsmiddleware.ResponseHandler{statusCodeStats}
```
dricross, line 42: Should there be two response handlers? One for clientStats and one for statusCodeStats?

extension/agenthealth/handler/stats/handler_test.go
```diff
@@ -37,9 +37,12 @@ func TestStatsHandler(t *testing.T) {
 		ThreadCount:          aws.Int32(789),
 		LatencyMillis:        aws.Int64(1234),
 		PayloadBytes:         aws.Int(5678),
-		StatusCode:           aws.Int(200),
```
dricross, line 40: This wasn't removed from the structure, so I wouldn't expect to see it removed from the test.

extension/agenthealth/handler/stats/handler_test.go
```diff
@@ -59,15 +62,9 @@ func TestStatsHandler(t *testing.T) {
...
 	handler.HandleRequest(ctx, req)
-	assert.Equal(t, `"cpu":1.2,"mem":123,"fd":456,"th":789,"lat":1234,"load":5678,"code":200,"scfb":1,"ifs":1`, req.Header.Get(headerKeyAgentStats))
+	assert.Equal(t, `"cpu":1.2,"mem":123,"fd":456,"th":789,"lat":1234,"load":5678,"scfb":1,"ifs":1,"codes":{"di":[0,1,0,0,0],"pmd":[1,0,0,0,0]}`, req.Header.Get(headerKeyAgentStats))
 	stats.StatusCode = aws.Int(404)
 	stats.LatencyMillis = nil
 	handler.HandleRequest(ctx, req)
-	assert.Equal(t, `"cpu":1.2,"mem":123,"fd":456,"th":789,"load":5678,"code":404,"scfb":1,"ifs":1`, req.Header.Get(headerKeyAgentStats))
-}
-
-func TestNewHandlers(t *testing.T) {
```
dricross, line 69: What happened to this test?

extension/agenthealth/handler/stats/agent/agent.go
```diff
@@ -118,3 +159,39 @@ type StatsConfig struct {
...
+	allowed := collections.NewSet[string](StatusCodeOperations...)
```
jefchien, line 178: Why not just store `StatusCodeOperations` as a Set?

extension/agenthealth/config.go
```diff
@@ -12,6 +12,7 @@ import (
 type Config struct {
```
jefchien, line 12: IMO status code stats should have to be enabled. `agenthealth/metrics` and `agenthealth/traces` won't have any allowlisted operations, but are still going to have the handlers created and attached.

extension/agenthealth/handler/stats/provider/statuscode.go
```diff
@@ -0,0 +1,180 @@
...
+	h.resetTimer = time.AfterFunc(statusResetInterval, func() {
+		h.mu.Lock()
+		defer h.mu.Unlock()
+
+		h.statsByOperation.Range(func(key, _ interface{}) bool {
+			h.statsByOperation.Delete(key)
+			return true
+		})
+		log.Println("Status code stats reset.")
+		h.startResetTimer()
```
dricross, line 58: nit - consider using a ticker instead of manually resetting a timer: https://gobyexample.com/tickers

extension/agenthealth/handler/stats/agent/agent.go
```diff
@@ -118,3 +159,39 @@ type StatsConfig struct {
...
+
+	return OperationsFilter{
+		operations: allowed,
+		allowAll:   allowed.Contains(AllowAllOperations),
```
jefchien, line 182: Don't need to do a `Contains` check. This is always going to be false.

extension/agenthealth/handler/stats/provider/statuscode.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+// HandleRequest is a no-op for the StatusCodeHandler.
+func (h *StatusCodeHandler) HandleRequest(ctx context.Context, _ *http.Request) {}
+
+// HandleResponse processes the HTTP response to update status code stats.
+func (h *StatusCodeHandler) HandleResponse(ctx context.Context, r *http.Response) {
+	// Extract the operation name
+	operation := awsmiddleware.GetOperationName(ctx)
+	if operation == "" {
```
dricross, line 69: Would one of these print statements happen for every AWS call? That is probably excessive. Seems like a general trend for the print statements in this struct

extension/agenthealth/handler/stats/provider/statuscode.go
```diff
@@ -0,0 +1,180 @@
...
+)
+
+var (
+	statusCodeSingleton *StatusCodeHandler
+	statusCodeStatsOnce sync.Once
+)
+
+// StatusCodeHandler provides monitoring for status codes per operation.
+type StatusCodeHandler struct {
+	statsByOperation sync.Map
```
dricross, line 30: It looks like you're locking the mutex anytime you access the map so it probably doesn't need to be a sync map

internal/ecsservicediscovery/containerinstanceprocessor.go
```diff
@@ -64,6 +64,7 @@ func splitMapKeys(a map[string]*EC2MetaData, size int) [][]string {
 }
 
 func (p *ContainerInstanceProcessor) handleContainerInstances(cluster string, batch []string, containerInstanceMap map[string]*EC2MetaData) error {
+	log.Println("yooooo we in above describe instances for host")
```
dricross, line 67: debug logs

extension/agenthealth/handler/stats/provider/statuscode.go
```diff
@@ -0,0 +1,180 @@
...
+	case "ListServices":
+		return "ls"
+	case "ListTasks":
+		return "lt"
+	case "CreateLogGroup":
+		return "clg"
+	case "CreateLogStream":
+		return "cls"
+	case "AssumeRole":
+		return "sts"
```
dricross, line 147: I'd recommend following your convention and stick to an abbreviation of the operation rather than use the service (so `ar` instead of `sts`). It would make adding another sts service call confusing otherwise.

extension/agenthealth/config.go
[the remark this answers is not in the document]
Paramadon: the pointer is so that we can emit it if it is empty(nil)

extension/agenthealth/handler/stats/provider/statuscode.go
```diff
@@ -0,0 +1,180 @@
...
+		handler.startResetTimer()
+		statusCodeSingleton = handler
+	})
+	return statusCodeSingleton
+}
+
+// startResetTimer initializes a reset timer to clear stats every 5 minutes.
+func (h *StatusCodeHandler) startResetTimer() {
+	h.resetTimer = time.AfterFunc(statusResetInterval, func() {
+		h.mu.Lock()
```
jefchien, line 50: You don't need a lock. That's the whole point of the `sync.Map`.

extension/agenthealth/handler/stats/provider/statuscode.go
```diff
@@ -0,0 +1,180 @@
...
+		defer h.mu.Unlock()
+
+		h.statsByOperation.Range(func(key, _ interface{}) bool {
```
jefchien, line 53: Use `h.statsByOperation.Clear()` instead.

extension/agenthealth/handler/stats/agent/agent.go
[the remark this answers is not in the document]
Paramadon: yeah I like that better @dricross

extension/agenthealth/handler/stats/provider/statuscode.go
jefchien, line 69: Don't need this check. The filter won't allow an empty string either.

extension/agenthealth/handler/stats/provider/statuscode.go
Paramadon: Yeah I don't think this makes sense, I think we should just return "" and return from the HandleResponse function.

extension/agenthealth/handler/stats/provider/statuscode.go
Paramadon: removed mutex since we have sync.map

extension/agenthealth/handler/stats/handler_test.go
Paramadon: Oops, added this back

extension/agenthealth/handler/stats/handler.go
Paramadon: yes clientstats need to be added as a response handler

extension/agenthealth/handler/stats/agent/agent.go
Paramadon: removed log statement

extension/agenthealth/handler/stats/agent/agent.go
```diff
@@ -80,6 +82,45 @@ func (s *Stats) Merge(other Stats) {
 	if other.EntityRejected != nil {
 		s.EntityRejected = other.EntityRejected
 	}
+	if other.StatusCodes != nil {
```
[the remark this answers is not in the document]
Paramadon: I can do something like this below:

```
	if other.StatusCodes == nil {
		return
	}

	if s.StatusCodes == nil {
		s.StatusCodes = make(map[string][5]int)
	}

	for key, value := range other.StatusCodes {
		if existing, ok := s.StatusCodes[key]; ok {
			s.StatusCodes[key] = [5]int{
				existing[0] + value[0], // 200
				existing[1] + value[1], // 400
				existing[2] + value[2], // 408
				existing[3] + value[3], // 413
				existing[4] + value[4], // 429
			}
		} else {
			s.StatusCodes[key] = value
		}
	}
```

but it will be different from all the other if statements so I think I'm going to leave as is.

extension/agenthealth/config.go
Paramadon: I agree with this and I changed it so that we would have to enable statuscodes

extension/agenthealth/handler/stats/provider/statuscode.go
```diff
@@ -0,0 +1,171 @@
...
+
+	value, loaded := h.statsByOperation.LoadOrStore(operation, &[5]int{})
+	if !loaded {
+		log.Printf("Initializing stats for operation: %s", operation)
+	}
+	stats := value.(*[5]int)
+
+	h.updateStatusCodeCount(stats, statusCode, operation)
+
+	h.statsByOperation.Store(operation, stats)
```
dricross, line 86: Hmm, these lines consist of a read-modify-write operation. Two threads running this at the same time could clobber each others updates. Only the latter thread will have it's updates applied. As an example:

1. Thread A and B are both handling an HTTP response for the same operation. HTTP response is 200 for both threads.
1. Thread A loads current values, let's say stats[0] = 10
1. Thread B loads current values, also sees stats[0] = 10
1. Both threads now have a copy stats array
1. B adds 1 to its copy, stats[0] = 11
1. B stores its result
1. A adds 1 to it's copy, stats[0] = 11
1. A stores its result

We end up with stats[0]=11 for the operation instead of stats[0]=12. Perhaps its not actually an issue though. Is it possible we could be handling more than one HTTP response for the same operation at once? If not, then I wouldn't worry about it. If we don't know, or if it doesn't happen now but could in the future, we should handle it.

extension/agenthealth/handler/stats/provider/statuscode.go
jefchien: Oops. Looks like `sync.Map#Clear` was added in 1.23

extension/agenthealth/handler/stats/handler.go
```diff
@@ -21,12 +21,36 @@ const (
...
+	}
+
+	if agentStatsEnabled {
+		clientStats := client.NewHandler(agent.NewOperationsFilter())
+		statsProviders = append(statsProviders, clientStats, provider.GetProcessStats(), provider.GetFlagsStats())
+		responseHandlers = append(responseHandlers, clientStats)
+		requestHandlers = append(requestHandlers, clientStats)
+
+	}
+	filter := agent.NewStatusCodeAndOtherOperationsFilter(cfg.Operations)
```
jefchien, line 48: Why does the stats handler filter include the status code operations?

Paramadon: because don't we want the status code handler be apart of the statshandler?

extension/agenthealth/handler/stats/agent/agent.go
```diff
@@ -104,6 +127,26 @@ func (of OperationsFilter) IsAllowed(operationName string) bool {
 	return of.allowAll || of.operations.Contains(operationName)
 }
 
+var StatusCodeOperations = []string{ // all the operations that are allowed
```
dricross, line 130: I think you're missing `AssumeRole` here

translator/translate/otel/extension/agenthealth/translator.go
```diff
@@ -59,6 +70,10 @@ func (t *translator) Translate(conf *confmap.Conf) (component.Config, error) {
 	if usageData, ok := common.GetBool(conf, common.ConfigKey(common.AgentKey, usageDataKey)); ok {
 		cfg.IsUsageDataEnabled = cfg.IsUsageDataEnabled && usageData
 	}
+	cfg.IsStatusCodeEnabled = t.statuscodeonly
```
dricross, line 73: Is this saying we only have status code when it's status code only? Why couldn't it do more than one thing? I don't know this package very well, but this behavior is slightly different than how "usage data" is enabled.

translator/translate/otel/extension/agenthealth/translator.go
```diff
@@ -59,6 +70,10 @@ func (t *translator) Translate(conf *confmap.Conf) (component.Config, error) {
...
+	if t.statuscodeonly != false && t.statuscodeonly {
```
dricross, line 74: redundant conditions
```suggestion
	if t.statuscodeonly {
```

extension/agenthealth/handler/stats/provider/statuscode.go
```diff
@@ -0,0 +1,168 @@
...
+		stats[4]++
+	default:
+		return
+	}
+}
+
+func GetShortOperationName(operation string) string {
+	switch operation {
+	case "PutMetricData":
+		return "pmd"
```
jefchien, line 115: Why is `PutMetricData` here?

extension/agenthealth/handler/stats/provider/process_test.go
```diff
@@ -77,6 +75,6 @@ func TestProcessStats(t *testing.T) {
 	mock.mu.Unlock()
 	provider.refresh()
 	assert.Eventually(t, func() bool {
-		return provider.getStats() == agent.Stats{}
+		return len(provider.getStats().StatusCodes) == 0 // map isn't comparable, so we check the length
```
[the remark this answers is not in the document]
Paramadon: This test I think is already flaky and is skipped but I added a function that will check if all attributes were reset.

extension/agenthealth/handler/stats/provider/statuscode.go
```diff
@@ -0,0 +1,168 @@
...
+	statusCodeStatsOnce.Do(func() {
+		handler := &StatusCodeHandler{
+			statsByOperation: make(map[string]*[5]int),
+		}
+
+		if opsFilter, ok := filter.(agent.OperationsFilter); ok {
+			handler.filter = opsFilter
+		}
+		handler.startResetTimer()
+		statusCodeSingleton = handler
```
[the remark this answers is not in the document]
Paramadon: Ahh that makes sense, made stats.Provider only one that is a singleton

.golangci.yml
```diff
@@ -59,4 +59,4 @@ linters:
     - nonamedreturns
 
 issues:
-  new-from-rev: 3221f76
\ No newline at end of file
+  new-from-rev: 9af4477
```
Paramadon, line 62: Had way to many lint problems, need to limit the scope of the linter to decrease linting time in a different pr

extension/agenthealth/handler/stats/provider/statuscode.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+// NewStatusCodeHandler creates a new instance of StatusCodeHandler.
+func NewStatusCodeHandler(filter agent.OperationsFilter) *StatusCodeHandler {
+	provider := GetStatsProvider().(*SingletonStatsProvider) // Get the singleton provider.
+	handler := &StatusCodeHandler{
+		statsProvider: provider,
+		filter:        filter,
+	}
+	handler.startResetTimer()
```
Paramadon, line 48: Reset timer needs to be in provider

extension/agenthealth/handler/stats/handler.go
```diff
@@ -21,12 +21,36 @@ const (
...
+	var requestHandlers []awsmiddleware.RequestHandler
+	var responseHandlers []awsmiddleware.ResponseHandler
+	var statsProviders []agent.StatsProvider
+
+	if !statusCodeEnabled && !agentStatsEnabled {
+		return nil, nil
+	}
+
+	statusCodeFilter := agent.NewStatusCodeOperationsFilter()
+	statusCodeHandler := provider.NewStatusCodeHandler(statusCodeFilter)
```
Paramadon, line 34: put this inside the if

extension/agenthealth/handler/stats/handler.go
```diff
@@ -21,12 +21,36 @@ const (
...
+
+	if statusCodeEnabled {
+		responseHandlers = append(responseHandlers, statusCodeHandler)
+		statsProviders = append(statsProviders, provider.GetStatsProvider())
```
Paramadon, line 38: Name this differently it should be more specific

extension/agenthealth/handler/stats/handler.go
Paramadon, line 34: put status code in it own package

extension/agenthealth/handler/stats/handler.go
```diff
@@ -21,12 +21,36 @@ const (
...
+		statsProviders = append(statsProviders, clientStats, provider.GetProcessStats(), provider.GetFlagsStats())
+		responseHandlers = append(responseHandlers, clientStats)
+		requestHandlers = append(requestHandlers, clientStats)
+
+	}
 	filter := agent.NewOperationsFilter(cfg.Operations...)
-	clientStats := client.NewHandler(filter)
-	stats := newStatsHandler(logger, filter, []agent.StatsProvider{clientStats, provider.GetProcessStats(), provider.GetFlagsStats()})
+	stats := newStatsHandler(logger, filter, statsProviders)
+	requestHandlers = append(requestHandlers, stats)
```
Paramadon, line 50: inside if statement. We only need it if agentStatsEnabled

extension/agenthealth/factory.go
```diff
@@ -28,7 +28,7 @@ func NewFactory() extension.Factory {
 func createDefaultConfig() component.Config {
 	return &Config{
 		IsUsageDataEnabled: true,
-		Stats: agent.StatsConfig{
+		Stats: &agent.StatsConfig{
```
Paramadon, line 31: change this to nil as default

extension/agenthealth/handler/stats/agent/agent.go
```diff
@@ -112,9 +156,12 @@ func NewOperationsFilter(operations ...string) OperationsFilter {
...
 }
 
-type StatsConfig struct {
-	// Operations are the allowed operation names to gather stats for.
-	Operations []string `mapstructure:"operations,omitempty"`
-	// UsageFlags are the usage flags to set on start up.
-	UsageFlags map[Flag]any `mapstructure:"usage_flags,omitempty"`
+// NewStatusCodeOperationsFilter creates a new filter for allowed operations and status codes.
+func NewStatusCodeOperationsFilter() OperationsFilter {
+	allowed := make(map[string]struct{}, len(StatusCodeOperations))
```
Paramadon, line 161: Make sure this is populated

extension/agenthealth/handler/stats/provider/statuscode.go
```diff
@@ -0,0 +1,216 @@
...
+				sp.processStatusCode(entry)
+			case <-sp.stopChan:
+				return
+			}
+		}
+	}()
+}
+
+// Stop signals the StatsProvider to stop processing and waits for cleanup.
+func (sp *StatsProvider) Stop() {
```
jefchien, line 82: This is never called and if it were we would need to ensure that it isn't called multiple times.

extension/agenthealth/handler/stats/provider/statuscode.go
```diff
@@ -0,0 +1,216 @@
...
+	statsProviderOnce.Do(func() {
+		provider := &StatsProvider{
+			statsByOperation: make(map[string]*[5]int),
+			statusCodeChan:   make(chan statusCodeEntry, channelBufferSize),
+			stopChan:         make(chan struct{}),
+		}
+
+		if opsFilter, ok := filter.(agent.OperationsFilter); ok {
+			provider.filter = opsFilter
+		}
```
jefchien, line 57: Why does the provider need the filter? It's never used in the provider.

extension/agenthealth/handler/stats/provider/statuscode.go
```diff
@@ -0,0 +1,216 @@
...
+	channelBufferSize   = 100
+)
+
+var (
+	statusCodeProviderSingleton *StatsProvider
+	statsProviderOnce           sync.Once
+)
+
+// StatsProvider handles processing of status codes and maintains stats.
+type StatsProvider struct {
```
jefchien, line 30: nit: Rename to something more specific
```suggestion
type StatusCodeProvider struct {
```

extension/agenthealth/handler/stats/provider/statuscode.go
```diff
@@ -0,0 +1,216 @@
...
+	case "ListTasks":
+		return "lt"
+	case "CreateLogGroup":
+		return "clg"
+	case "CreateLogStream":
+		return "cls"
+	default:
+		return ""
+	}
+}
```
dricross, line 202: This function is essentially a duplicate of the string array `StatusCodeOperations`, so the two constructs need to be kept in-sync. I'd suggest either moving the function closer to the string array definition, or change the string array to a map which will do this mapping for you. Then, if you do need a list of valid operations, you can get it from the map.

dricross: Also, did we decide to remove AssumeRole? It missing from this list and the StatusCodeOperations

extension/agenthealth/handler/stats/provider/statuscode.go
```diff
@@ -0,0 +1,216 @@
...
+	case 429:
+		stats[4]++
+	default:
+		log.Printf("Ignored status code: %d\n", statusCode)
+	}
+}
+
+// startResetTimer initializes a reset timer to clear stats periodically.
+func (sp *StatsProvider) startResetTimer() {
+	log.Println("Starting stats reset timer...")
```
dricross, line 126: we probably dont want these flooding the logs. there's a couple others too

extension/agenthealth/handler/stats/provider/statuscode.go
Paramadon: We need to get Assume role from the contrib changes so I was thinking of adding it then, but I can add it here too.

extension/agenthealth/handler/stats/provider/statuscode.go
```diff
@@ -0,0 +1,216 @@
...
+		return "clg"
+	case "CreateLogStream":
+		return "cls"
+	default:
+		return ""
+	}
+}
+
+// Stats returns the aggregated stats for operations.
+func (sp *StatsProvider) Stats(_ string) agent.Stats {
```
jefchien, line 170: Right now, this means that each call to `Stats` is going to include the current status code map. This function is called for each request that goes through the `StatsHandler` since it needs to add the data to the header, which means that we're going to get duplicate counts during that time period.

https://github.com/aws/amazon-cloudwatch-agent/blob/a8ef6f778b075e3eaf51e033a15313edaca33d89/extension/agenthealth/handler/stats/handler.go#L72-L74

For example, if we record that "DescribeTags" had a 200, each PMD/PLE/PTS call made until the reset timer goes off would have that count.

Paramadon: Ahh that's right, we need to reset the count that has already been sent so:

statuscode records DescribeTags with a 200: {"dt":[1,0,0,0,0]}
Reset sent map.
PMD sends the statuscodemap as part of the stats header
statuscode records DescribeTags with a 200: statuscodemap{"dt":[1,0,0,0,0]}
PMD sends the statuscodemap again
We'll record 2 200s for DT

extension/agenthealth/handler/stats/provider/statuscode.go
Paramadon: accessing filter from handler. Handle->StatusCodeFilter->Filter. But Filter should be part of handler. So changed it to be part of handler.

dricross (approved):

dricross (approved):

extension/agenthealth/extension.go
```diff
@@ -24,11 +25,30 @@ var _ awsmiddleware.Extension = (*agentHealth)(nil)
...
 	var responseHandlers []awsmiddleware.ResponseHandler
 	requestHandlers := []awsmiddleware.RequestHandler{useragent.NewHandler(ah.cfg.IsUsageDataEnabled)}
-	if ah.cfg.IsUsageDataEnabled {
-		req, res := stats.NewHandlers(ah.logger, ah.cfg.Stats)
-		requestHandlers = append(requestHandlers, req...)
-		responseHandlers = append(responseHandlers, res...)
+
+	if ah.cfg == nil {
+		return nil, nil
+	}
```
jefchien, line 31: nit: Unnecessary check. Cannot be nil at this point.

extension/agenthealth/extension.go
```diff
@@ -24,11 +25,30 @@ var _ awsmiddleware.Extension = (*agentHealth)(nil)
...
+	statusCodeEnabled := ah.cfg.IsStatusCodeEnabled
+
+	var statsResponseHandlers []awsmiddleware.ResponseHandler
+	var statsRequestHandlers []awsmiddleware.RequestHandler
+
+	if ah.cfg.Stats != nil {
+		statsRequestHandlers, statsResponseHandlers = stats.NewHandlers(ah.logger, *ah.cfg.Stats, statusCodeEnabled, true)
+	} else {
+		statsRequestHandlers, statsResponseHandlers = stats.NewHandlers(ah.logger, agent.StatsConfig{}, statusCodeEnabled, false)
+	}
```
jefchien, line 47: nit: It's a bit strange that the stats handler constructor is still being called when the stats configuration is not configured.

extension/agenthealth/handler/stats/handler.go
```diff
@@ -21,12 +21,34 @@ const (
...
+		statusCodeHandler := provider.NewStatusCodeHandler(statusCodeStatsProvider, statusCodeFilter)
+		responseHandlers = append(responseHandlers, statusCodeHandler)
+		statsProviders = append(statsProviders, statusCodeStatsProvider)
+	}
+
+	if agentStatsEnabled {
+		clientStats := client.NewHandler(agent.NewOperationsFilter(cfg.Operations...))
+		statsProviders = append(statsProviders, clientStats, provider.GetProcessStats(), provider.GetFlagsStats())
+		responseHandlers = append(responseHandlers, clientStats)
+		filter := agent.NewOperationsFilter(cfg.Operations...)
```
jefchien, line 45: nit: Move above line 42 so the client handler can use it as well.

extension/agenthealth/handler/stats/provider/statuscode.go
```diff
@@ -0,0 +1,173 @@
...
+		for {
+			select {
+			case entry := <-sp.statusCodeChan:
+				sp.processStatusCode(entry)
+			case <-sp.stopChan:
+				return
+			}
+		}
+	}()
+}
```
jefchien, line 74: What are the wait group and stop channel for now that the `stop` function's been removed? Doesn't look like anything blocks it or closes the channel anymore.

extension/agenthealth/handler/stats/provider/statuscode.go
```diff
@@ -0,0 +1,173 @@
...
+		stats[4]++
+	default:
+		return
+	}
+}
+
+// startResetTimer initializes a reset timer to clear stats periodically.
+func (sp *StatusCodeProvider) startResetTimer() {
+	sp.resetTimer = time.AfterFunc(statusResetInterval, func() {
+		sp.shouldResetStats = true
```
jefchien, line 115: May have a data race where the bool is getting updated at the same time as it's being read. Please add a test for this reset.

dricross (approved):

extension/agenthealth/handler/stats/provider/statuscode.go
```diff
@@ -41,12 +40,19 @@ type statusCodeEntry struct {
...
+
+// StatusCodeProvider methods
+
 func GetStatusCodeStatsProvider() *StatusCodeProvider {
 	StatusCodeProviderOnce.Do(func() {
 		provider := &StatusCodeProvider{
-			statsByOperation: make(map[string]*[5]int),
-			statusCodeChan:   make(chan statusCodeEntry),
+			StatsByOperation: make(map[string]*[5]int),
+			statusCodeChan:   make(chan statusCodeEntry, 1000),
```
jefchien, line 55: How did we determine the 1000 buffer?

extension/agenthealth/handler/stats/provider/statuscode.go
```diff
@@ -104,23 +108,43 @@ func (sp *StatusCodeProvider) updateStatusCodeCount(stats *[5]int, statusCode in
...
+	if sp.ShouldResetStats {
+		for op, stats := range sp.StatsByOperation {
+			statusCodeMap[op] = *stats
+		}
+		sp.StatsByOperation = make(map[string]*[5]int)
+		sp.ShouldResetStats = false
+	} else {
+		for op, stats := range sp.StatsByOperation {
+			statusCodeMap[op] = *stats
+		}
```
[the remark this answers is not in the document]
Paramadon: Yeah I noticed in my testing they were appearing up, just got rid of this.

extension/agenthealth/handler/stats/provider/statuscode.go
```diff
@@ -26,13 +26,12 @@ var (
...
-	resetTimer       *time.Timer
+	StatsByOperation map[string]*[5]int
+	ResetTimer       *time.Timer
 	statusCodeChan   chan statusCodeEntry
 	stopChan         chan struct{}
-	shouldResetStats bool
-	mu               sync.Mutex
-	wg               sync.WaitGroup
+	ShouldResetStats bool
+	Mu               sync.RWMutex
```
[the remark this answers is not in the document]
Paramadon: Made these public to access in test file(which had to be a different package due to cycle imports)

extension/agenthealth/handler/stats/provider/statuscode.go
Paramadon: I thought it would be good to have a big buffer size just in case of high burst of data but it takes microseconds to update a map(which is basically all we are doing) so I think we can change it to 100 or even 10, but just to be safe(not dropping anything) I can change it to 100 if that makes more sense..

extension/agenthealth/extension.go
```diff
@@ -42,8 +37,6 @@ func (ah *agentHealth) Handlers() ([]awsmiddleware.RequestHandler, []awsmiddlewa
 
 	if ah.cfg.Stats != nil {
 		statsRequestHandlers, statsResponseHandlers = stats.NewHandlers(ah.logger, *ah.cfg.Stats, statusCodeEnabled, true)
-	} else {
```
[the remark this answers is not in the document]
Paramadon: Yes I added another statuscodeonly test in extension_test.go to make sure we are getting the right handlers.

dricross (approved):

extension/agenthealth/extension.go
```diff
@@ -24,11 +25,31 @@ var _ awsmiddleware.Extension = (*agentHealth)(nil)
...
+	var agentStatsEnabled bool
+
+	if ah.cfg.Stats != nil {
+		statsConfig = *ah.cfg.Stats
+		agentStatsEnabled = true
+	} else {
+		agentStatsEnabled = false
 	}
+
+	statsRequestHandlers, statsResponseHandlers = stats.NewHandlers(ah.logger, statsConfig, statusCodeEnabled, agentStatsEnabled)
```
jefchien, line 48: nit: Could just pass in the stats pointer directly to determine if it's enabled.
```suggestion
	statsRequestHandlers, statsResponseHandlers = stats.NewHandlers(ah.logger, ah.cfg.Stats, statusCodeEnabled)
```

extension/agenthealth/handler/stats/provider/statuscode.go
```diff
@@ -0,0 +1,163 @@
...
+	for op, stats := range sp.currentStats {
+		newStats.StatusCodes[op] = *stats
+	}
+	sp.currentStats = make(map[string]*[5]int)
+	sp.mu.Unlock()
+
+	select {
+	case existingStats := <-sp.completedStats:
+		existingStats.Merge(newStats)
+		newStats = existingStats
```
jefchien, line 115: nit: newStats.Merge(existingStats) also works.

jefchien (approved):

extension/agenthealth/extension.go
musa-asad, line 48: nit: don't need to define variables
```suggestion
	var statsConfig agent.StatsConfig
	var agentStatsEnabled bool

	if ah.cfg.Stats != nil {
		statsConfig = *ah.cfg.Stats
		agentStatsEnabled = true
	} else {
		agentStatsEnabled = false
	}

	statsRequestHandlers, statsResponseHandlers := stats.NewHandlers(ah.logger, statsConfig, statusCodeEnabled, agentStatsEnabled)
```
(then you can remove the conditionals from @jefchien's changes)

extension/agenthealth/factory.go
```diff
@@ -28,9 +26,7 @@ func NewFactory() extension.Factory {
 func createDefaultConfig() component.Config {
 	return &Config{
 		IsUsageDataEnabled: true,
-		Stats: agent.StatsConfig{
-			Operations: []string{agent.AllowAllOperations},
-		},
+		Stats:              nil,
```
musa-asad, line 29: Do we need to include this?
```suggestion
		IsStatusCodeEnabled: false,
		Stats:              nil,
```

extension/agenthealth/handler/stats/agent/agent.go
```diff
@@ -104,6 +125,29 @@ func (of OperationsFilter) IsAllowed(operationName string) bool {
...
 }
 
+type StatsConfig struct {
+	// Operations are the allowed operation names to gather stats for.
+	Operations []string `mapstructure:"operations,omitempty"`
+	// UsageFlags are the usage flags to set on start up.
+	UsageFlags map[Flag]any `mapstructure:"usage_flags,omitempty"`
+}
+
+var StatusCodeOperations = []string{ // all the operations that are allowed
```
musa-asad, line 135: can you explain the rationale behind these?

translator/translate/otel/extension/agenthealth/translator.go
```diff
@@ -31,14 +31,25 @@ var (
...
+	name                string
+	operations          []string
+	isUsageDataEnabled  bool
+	factory             extension.Factory
+	isStatusCodeEnabled bool
 }
 
 var _ common.Translator[component.Config] = (*translator)(nil)
 
+func NewTranslatorWithStatusCode(name component.DataType, operations []string, isStatusCodeEnabled bool) common.Translator[component.Config] {
```
musa-asad, line 43: this function is unused?

musa-asad, line 43: Why don't we just combine this with the `NewTranslator` function? Is there any case where this function is called when `isStatusCodeEnabled` is false?

translator/translate/otel/extension/agenthealth/translator.go
```diff
@@ -31,14 +31,25 @@ var (
...
+	return &translator{
+		name:                name.String(),
+		operations:          operations,
+		factory:             agenthealth.NewFactory(),
+		isUsageDataEnabled:  envconfig.IsUsageDataEnabled(),
+		isStatusCodeEnabled: isStatusCodeEnabled,
+	}
+}
```
musa-asad, line 51: Why don't we just combine this with the `NewTranslator` function? Is there any case where this function is called when `isStatusCodeEnabled` is false?

Or you can do this instead.
```suggestion
func NewTranslatorWithStatusCode(name component.DataType, operations []string) common.Translator[component.Config] {
	return &translator{
		name:                name.String(),
		operations:          operations,
		factory:             agenthealth.NewFactory(),
		isUsageDataEnabled:  envconfig.IsUsageDataEnabled(),
		isStatusCodeEnabled: true,
	}
}
```

dricross (approved):

PR #1437, opened 2024-11-21
[BugFix] Include entity bytes in PutMetricData metric batch to ensure request payload does not exceed 1MB

lisguo:
# Description of the issue
The agent could make a PMD request of greater than 1MB due to the bytes in the Entity. This can cause RequestEntityTooLarge errors from CloudWatch. 

# Description of changes
* Modify the `payload` function to include the length of the entity string

Not sure if the `String()` method of the `cloudwatch.Entity` struct is good enough...need to confirm

# License

# Tests
Unit tests

# Requirements
2. Run `make lint`

### DIFF
--- a/plugins/outputs/cloudwatch/util.go
+++ b/plugins/outputs/cloudwatch/util.go
@@ -93,7 +93,7 @@ func resize(dist distribution.Distribution, listMaxSize int) (distList []distrib
 	return
 }
 
-func payload(datum *cloudwatch.MetricDatum) (size int) {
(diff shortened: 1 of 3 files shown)


plugins/outputs/cloudwatch/util.go
```diff
@@ -120,6 +120,10 @@ func payload(datum *cloudwatch.MetricDatum) (size int) {
 		size += unitOverheads
 	}
 
+	if entity != nil {
+		size += len(entity.String())
+	}
+
```
agiuliano, line 126: we should account for this size globally instead of each metricDatum

PR #1440, opened 2024-11-21
[Bug Fix] Fix Excessive IMDS related error logging

zhihonl:
# Description of the issue
```
    status code: 404, request id: D! should retry true for imds error : EC2MetadataError: failed to make EC2Metadata request
<?xml version="1.0" encoding="iso-8859-1"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
         "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
  <title>404 - Not Found</title>
 </head>
 <body>
  <h1>404 - Not Found</h1>
 </body>
</html>
```
In version 1.300049.0 and above, the agent will log the above message regardless of log levels on a defined interval. This is because we recently enabled instance tags by default to retrieve ASG name and instance tag name for entity service names. This becomes an issue when instance metadata tags is not enabled which can be majority case since instance metadata tags is an opt-in feature. The issue is especially apparently in EKS since EKS does not support instance metadata tags

# Description of changes
1. Only log the error when debug mode is enabled
2. Only enable instance tag related features in strictly EC2 mode
3. Only retry up to 3 times for instance tag errors.

# License

# Tests
Unit tests

# Requirements
4. Run `make lint`

### DIFF
--- a/extension/entitystore/ec2Info.go
+++ b/extension/entitystore/ec2Info.go
@@ -32,7 +32,8 @@ type EC2Info struct {
 	AutoScalingGroup string
 
 	// region is used while making call to describeTags Ec2 API for AutoScalingGroup
-	Region string
+	Region         string
+	kubernetesMode string
 
 	metadataProvider ec2metadataprovider.MetadataProvider
 	logger           *zap.Logger
@@ -48,8 +49,11 @@ func (ei *EC2Info) initEc2Info() {
 	if err := ei.setInstanceIDAccountID(); err != nil {
 		return
 	}
-	if err := ei.setAutoScalingGroup(); err != nil {
-		return
+	// Instance metadata tags is not usable for EKS nodes
+	// https://github.com/kubernetes/cloud-provider-aws/issues/762
+	if ei.kubernetesMode == "" {
+		limitedRetryer := NewRetryer(false, true, defaultJitterMin, defaultJitterMax, ec2tagger.BackoffSleepArray, maxRetry, ei.done, ei.logger)
+		limitedRetryer.refreshLoop(ei.retrieveAsgName)
 	}
 	ei.logger.Debug("Finished initializing EC2Info")
 }
@@ -99,40 +103,6 @@ func (ei *EC2Info) setInstanceIDAccountID() error {
 	}
 }
 
-func (ei *EC2Info) setAutoScalingGroup() error {
-	retry := 0
-	for {
-		var waitDuration time.Duration
-		if retry < len(ec2tagger.BackoffSleepArray) {
-			waitDuration = ec2tagger.BackoffSleepArray[retry]
-		} else {
-			waitDuration = ec2tagger.BackoffSleepArray[len(ec2tagger.BackoffSleepArray)-1]
-		}
-
-		wait := time.NewTimer(waitDuration)
-		select {
-		case <-ei.done:
-			wait.Stop()
-			return errors.New("shutdown signal received")
-		case <-wait.C:
-		}
-
-		if retry > 0 {
-			ei.logger.Debug("Initial retrieval of tags and volumes", zap.Int("retry", retry))
-		}
-
-		if err := ei.retrieveAsgName(); err != nil {
-			ei.logger.Debug("Unable to fetch instance tags with imds", zap.Int("retry", retry), zap.Error(err))
-		} else {
-			ei.logger.Debug("Retrieval of auto-scaling group tags succeeded")
-			return nil
-		}
-
-		retry++
-	}
-
-}
-
 func (ei *EC2Info) retrieveAsgName() error {
 	tags, err := ei.metadataProvider.InstanceTags(context.Background())
 	if err != nil {
@@ -142,6 +112,7 @@ func (ei *EC2Info) retrieveAsgName() error {
 		asg, err := ei.metadataProvider.InstanceTagValue(context.Background(), ec2tagger.Ec2InstanceTagKeyASG)
 		if err != nil {
 			ei.logger.Error("Failed to get AutoScalingGroup through metadata provider", zap.Error(err))
+			return err
 		} else {
 			ei.logger.Debug("AutoScalingGroup retrieved through IMDS")
 			ei.mutex.Lock()
@@ -156,9 +127,10 @@ func (ei *EC2Info) retrieveAsgName() error {
 	return nil
 }
 
-func newEC2Info(metadataProvider ec2metadataprovider.MetadataProvider, done chan struct{}, region string, logger *zap.Logger) *EC2Info {
+func newEC2Info(metadataProvider ec2metadataprovider.MetadataProvider, kubernetesMode string, done chan struct{}, region string, logger *zap.Logger) *EC2Info {
 	return &EC2Info{
 		metadataProvider: metadataProvider,
+		kubernetesMode:   kubernetesMode,
 		done:             done,
 		Region:           region,
 		logger:           logger,
--- a/extension/entitystore/ec2Info_test.go
+++ b/extension/entitystore/ec2Info_test.go
@@ -5,6 +5,7 @@ package entitystore
 
 import (
 	"bytes"
+	"github.com/aws/amazon-cloudwatch-agent/translator/config"
 	"log"
 	"strings"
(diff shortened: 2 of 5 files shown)


internal/retryer/imdsretryer.go
```diff
@@ -55,3 +56,10 @@ func GetDefaultRetryNumber() int {
 	}
 	return DefaultImdsRetries
 }
+
+// fmtDebugLog logs the content only if the log level is DEBUG
```
zhihonl, line 60: Thought about dropping this in an util folder but I doubt this is going to be reused considering this is strictly used for the retryer. Any other component used for the agent binary uses OTEL or telegraf loggers.

extension/entitystore/ec2Info.go
```diff
@@ -99,40 +103,6 @@ func (ei *EC2Info) setInstanceIDAccountID() error {
 	}
 }
 
-func (ei *EC2Info) setAutoScalingGroup() error {
```
okankoAMZ, line 102: why are we removing autoscaling group all together?

extension/entitystore/ec2Info_test.go
```diff
@@ -236,3 +237,37 @@ func TestNotInitIfMetadataProviderIsEmpty(t *testing.T) {
...
+		kubernetesMode   string
+	}
+	tests := []struct {
+		name    string
+		args    args
+		wantErr bool
+		want    EC2Info
+	}{
+		{
+			name: "happy path",
```
okankoAMZ, line 253: how did we determine this name?

extension/entitystore/extension.go
```diff
@@ -94,9 +94,13 @@ func (e *EntityStore) Start(ctx context.Context, host component.Host) error {
 	e.serviceprovider = newServiceProvider(e.mode, e.config.Region, &e.ec2Info, e.metadataprovider, getEC2Provider, ec2CredentialConfig, e.done, e.logger)
 	switch e.mode {
 	case config.ModeEC2:
-		e.ec2Info = *newEC2Info(e.metadataprovider, e.done, e.config.Region, e.logger)
+		e.ec2Info = *newEC2Info(e.metadataprovider, e.kubernetesMode, e.done, e.config.Region, e.logger)
 		go e.ec2Info.initEc2Info()
-		go e.serviceprovider.startServiceProvider()
+		// Instance metadata tags is not usable for EKS nodes
+		// https://github.com/kubernetes/cloud-provider-aws/issues/762
+		if e.kubernetesMode == "" {
```
zhihonl, line 101: This may seem drastic but based on the codebase serviceprovider isn't really used in Kubernetes scenario. It always gets overriden by either Instrumentation or K8sWorkload service name source.

extension/entitystore/ec2Info.go
zhihonl: This function is basically a retry loop which calls `retrieveAsgName` infinitely until success. We already have helper struct for this type of operation which is why I used [limitedRetryer](https://github.com/aws/amazon-cloudwatch-agent/pull/1440/files#diff-d1bdee0ab819bf459c37069c1cd13cc77105541449f620965e0bfd60c7e100fbR55). Though I just realized original code keeps retrying until successful. This one retries until successful then continues to refresh on an interval. Will flip `oneTime` option to true for the retryer.

extension/entitystore/ec2Info_test.go
zhihonl: updated in new commit

extension/entitystore/ec2Info.go
```diff
@@ -48,8 +49,11 @@ func (ei *EC2Info) initEc2Info() {
 	if err := ei.setInstanceIDAccountID(); err != nil {
 		return
 	}
-	if err := ei.setAutoScalingGroup(); err != nil {
-		return
+	// Instance metadata tags is not usable for EKS nodes
+	// https://github.com/kubernetes/cloud-provider-aws/issues/762
+	if ei.kubernetesMode == "" {
```
lisguo, line 54: we want to keep the retryer on native k8s on ec2 right?

extension/entitystore/extension.go
```diff
@@ -94,9 +94,13 @@ func (e *EntityStore) Start(ctx context.Context, host component.Host) error {
 	e.serviceprovider = newServiceProvider(e.mode, e.config.Region, &e.ec2Info, e.metadataprovider, getEC2Provider, ec2CredentialConfig, e.done, e.logger)
 	switch e.mode {
 	case config.ModeEC2:
-		e.ec2Info = *newEC2Info(e.metadataprovider, e.done, e.config.Region, e.logger)
+		e.ec2Info = *newEC2Info(e.metadataprovider, e.kubernetesMode, e.done, e.config.Region, e.logger)
```
okankoAMZ, line 97: how does the extension get that value?

internal/retryer/imdsretryer.go
lisguo: Isn't there a telegraf logger we can use? Instead of adding a function to check the log level?

extension/entitystore/extension.go
```diff
@@ -94,9 +94,13 @@ func (e *EntityStore) Start(ctx context.Context, host component.Host) error {
...
 		go e.ec2Info.initEc2Info()
-		go e.serviceprovider.startServiceProvider()
+		// Instance metadata tags is not usable for EKS nodes
+		// https://github.com/kubernetes/cloud-provider-aws/issues/762
+		if e.kubernetesMode == "" {
+			go e.serviceprovider.startServiceProvider()
```
lisguo, line 102: Does `e.serviceprovider.startServiceProvider()` do anything beneficial on EKS? Seems strange that we would have to do this just to avoid the IMDS check

extension/entitystore/extension.go
zhihonl: Are you asking about the `kubernetesMode`? it comes from the [translator](https://github.com/aws/amazon-cloudwatch-agent/blob/33ad73e6c22497166db2cab6359a2110b37de7de/translator/translate/otel/extension/entitystore/translator.go#L38) which gets passed as part of the config value

lisguo:
Can you provide logs of the built agent on EC2 and EKS?

extension/entitystore/extension.go
zhihonl: I don't think so. `startServiceProvider` scrapes IAM role or service name from instance tags so they can be used for service names. In Kubernetes environment neither of these will be used because they have lower priority than `Instrumentation` or `K8sWorkload`. 

Code wise, I only see serviceprovider public function called in `awsentity` processor if we are strictly on EC2: https://github.com/aws/amazon-cloudwatch-agent/blob/161cd5ba722e09ab0aa3b0a6d3f40e3a06e6aa1d/plugins/processors/awsentity/processor.go#L214

extension/entitystore/ec2Info.go
zhihonl: Fixed in new commit

extension/entitystore/extension.go
lisguo, line 101: similar to the above -- shouldn't this be `if e.kubernetesMode != modeEKS` ? So that we can account for native k8s?

zhihonl: I disabled it for all kubernetes environment because `serviceprovider` isn't used for any Kubernetes related logics. [GetMetricServiceNameAndSource](https://github.com/aws/amazon-cloudwatch-agent/blob/d2fcc0347e816f31845d1592f646fa53a52da20d/extension/entitystore/extension.go#L167) is the only public function that is used for `awsentity` processor and that is only referenced in EC2 specific processor logic: https://github.com/aws/amazon-cloudwatch-agent/blob/161cd5ba722e09ab0aa3b0a6d3f40e3a06e6aa1d/plugins/processors/awsentity/processor.go#L214

lisguo (dismissed):

lisguo:
pr build failing

lisguo (approved):

okankoAMZ (approved):

PR #1454, opened 2024-12-03
Config Translate Refactor to save >100MB of disk space when the agent is installed

chadpatel:
# Description of the issue
The `config-translator` binary uses an excessive amount of disk space (>100MB) because it brings in a ton of dependencies.

# Description of changes
I turned `config-translator` in to a shim that just calls `amazon-cloudwatch-agent`.  The old `config-translator` code was then moved to `cmdutil`.  I prefixed `ct-` to the arguments to prevent collisions.  

This will save ~114M on disk when installed and about 35M on the RPM

Note: inline diff is VERY hard to follow because of all the moved code, I recommend `split` diff or we can do a code walk-through

# License

# Tests
Manual testing to confirm cloudwatch agent still starts/loads

There were no prior tests that actually tested `config-translate` binary... the old tests basically just tested `cmdutil` so I moved those in to the `cmdutil` tests.  We could write more tests for the shim and for `translate.go`.  Tricky tests to write due to the OS coupling.

Before:
```total 297M
-rwxr-xr-x 1 patchad amazon 129M Dec  2 22:49 amazon-cloudwatch-agent
-rwxr-xr-x 1 patchad amazon  14M Dec  2 22:49 amazon-cloudwatch-agent-config-wizard
-rwxr-xr-x 1 patchad amazon  37M Dec  2 22:44 config-downloader
-rwxr-xr-x 1 patchad amazon 117M Dec  2 22:47 config-translator
-rwxr-xr-x 1 patchad amazon 2.1M Dec  2 22:49 start-amazon-cloudwatch-agent```

```➜  amazon-cloudwatch-agent git:(main) ✗ ls -lh ~/Downloads/amazon-cloudwatch-agent.rpm
Permissions Size User    Date Modified Name
.rw-r--r--@ 113M patchad  2 Dec 15:36  /Users/patchad/Downloads/amazon-cloudwatch-agent.rpm```

After:
```total 183M
-rwxr-xr-x 1 patchad amazon 130M Dec  3 16:07 amazon-cloudwatch-agent
-rwxr-xr-x 1 patchad amazon  14M Dec  3 16:07 amazon-cloudwatch-agent-config-wizard
-rwxr-xr-x 1 patchad amazon  37M Dec  3 16:07 config-downloader
-rwxr-xr-x 1 patchad amazon 1.6M Dec  3 16:07 config-translator
-rwxr-xr-x 1 patchad amazon 2.1M Dec  3 16:07 start-amazon-cloudwatch-agent```

```➜  amazon-cloudwatch-agent git:(main) ✗ ls -lh /local/home/patchad/workplace/cwa/amazon-cloudwatch-agent/build/bin/linux/amd64/amazon-cloudwatch-agent.rpm
-rw-r--r-- 1 patchad amazon 78M Dec  3 16:16 /local/home/patchad/workplace/cwa/amazon-cloudwatch-agent/build/bin/linux/amd64/amazon-cloudwatch-agent.rpm```

# Requirements
2. Run `make lint`

### CODE REVIEW

File: cmd/config-translator/translator.go [hunk 1]
```diff
@@ -1,137 +1,61 @@
-// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
-// SPDX-License-Identifier: MIT
```
chadpatel, line 2: I can bring this back, lost it when moving code around

cmd/amazon-cloudwatch-agent/amazon-cloudwatch-agent.go
```diff
@@ -95,6 +96,16 @@ var fRunAsConsole = flag.Bool("console", false, "run as console application (win
 var fSetEnv = flag.String("setenv", "", "set an env in the configuration file in the format of KEY=VALUE")
 var fStartUpErrorFile = flag.String("startup-error-file", "", "file to touch if agent can't start")
 
+// config-translator
+var fConfigTranslator = flag.Bool("config-translator", false, "run in config-translator mode")
```
chadpatel, line 100: I spent some time trying to reduce the copy/paste aspect.  Some of these constants are defined twice.  There are some things to do to clean it up but they all end up less readable and add a lot of indirection.  Open to ideas tho.

If I was forced to make one change I would put the descriptions in a constant

translator/cmdutil/translatorutil_test.go
```diff
@@ -38,3 +40,195 @@ func TestTranslateJsonMapToEnvConfigFile(t *testing.T) {
 	assert.Equal(t, expectedJson[envconfig.CWAGENT_LOG_LEVEL], actualJson[envconfig.CWAGENT_LOG_LEVEL])
 	assert.Equal(t, expectedJson[envconfig.AWS_SDK_LOG_LEVEL], actualJson[envconfig.AWS_SDK_LOG_LEVEL])
 }
+
+func TestAgentConfig(t *testing.T) {
```
chadpatel, line 44: this code was moved from translator_test.go

translator/cmdutil/translator.go
```diff
@@ -0,0 +1,121 @@
...
+	version            = "1.0"
+	envConfigFileName  = "env-config.json"
+	yamlConfigFileName = "amazon-cloudwatch-agent.yaml"
+)
+
+type ConfigTranslator struct {
+	ctx *context.Context
+}
+
+func NewConfigTranslator(inputOs, inputJsonFile, inputJsonDir, inputTomlFile, inputMode, inputConfig, multiConfig string) (*ConfigTranslator, error) {
```
chadpatel, line 32: this code was moved from the old config-translator/translator.go

tool/cmdwrapper/cmdwrapper.go
```diff
@@ -0,0 +1,67 @@
...
+	for key, value := range flags {
+		if *value != "" {
+			args = append(args, fmt.Sprintf("-%s%s%s", command, delimiter, key), *value)
+		}
+	}
+
+	log.Printf("Executing %s with arguments: %v", paths.AgentBinaryPath, args)
+
+	// Use execCommand instead of exec.Command directly
+	cmd := execCommand(paths.AgentBinaryPath, args...)
```
chadpatel, line 51: TODO - we will need to first check the normal path but if this fails to resolve then we need to check for amazon-cloudwatch-agent in the same directory as the current executing process (i.e. config-translator)

chadpatel: https://pkg.go.dev/os#Executable

tool/cmdwrapper/cmdwrapper.go
```diff
@@ -0,0 +1,99 @@
...
+		return path, nil
+	}
+
+	// if not, check in the current executable's directory
+	execPath, err := os.Executable()
+	if err != nil {
+		return "", fmt.Errorf("failed to get current executable path: %w", err)
+	}
+	execDir := filepath.Dir(execPath)
+	alternatePath := filepath.Join(execDir, "amazon-cloudwatch-agent")
```
chadpatel, line 42: nit - you could use `AgentBinaryName`.  Not sure if that brings in unnecessary binary size tho

tool/cmdwrapper/cmdwrapper.go
```diff
@@ -26,6 +26,32 @@ const delimiter = "-"
...
 var execCommand = exec.Command
 
+// Make findAgentBinary a func variable to replace in tests
+var findAgentBinary = func(path string) (string, error) {
+	// check if the binary is at the normal path
+	if _, err := os.Stat(path); err == nil {
+		return path, nil
+	}
+
+	// if not, check in the current executable's directory
```
chadpatel, line 36: nice! this is a great idea

tool/cmdwrapper/cmdwrapper.go
movence: great point. updated

cmd/amazon-cloudwatch-agent-config-wizard/wizard.go
```diff
@@ -4,139 +4,22 @@
...
-		scanner.Scan()
-		if len(a) > 0 {
-			*a[0].(*string) = scanner.Text()
-			n = len(*a[0].(*string))
-		}
-		err = scanner.Err()
-		return
+	err := cmdwrapper.ExecuteAgentCommand(flags.Command, translatorFlags)
+	if err != nil {
+		os.Exit(1)
```
jefchien: Should we log out the error and exit? Like a `log.Fatalf`? Is the agent already going to log the error?

cmd/amazon-cloudwatch-agent/amazon-cloudwatch-agent.go
```diff
@@ -41,13 +41,20 @@ import (
...
 	"github.com/aws/amazon-cloudwatch-agent/service/configprovider"
 	"github.com/aws/amazon-cloudwatch-agent/service/defaultcomponents"
 	"github.com/aws/amazon-cloudwatch-agent/service/registry"
+	"github.com/aws/amazon-cloudwatch-agent/tool/cmdwrapper"
+	"github.com/aws/amazon-cloudwatch-agent/tool/downloader"
+	downloaderflags "github.com/aws/amazon-cloudwatch-agent/tool/downloader/flags"
 	"github.com/aws/amazon-cloudwatch-agent/tool/paths"
+	"github.com/aws/amazon-cloudwatch-agent/tool/wizard"
+	wizardflags "github.com/aws/amazon-cloudwatch-agent/tool/wizard/flags"
+	"github.com/aws/amazon-cloudwatch-agent/translator/cmdutil"
```
jefchien: nit: Would expect this to be in `tool/translator` for consistency.

tool/wizard/flags/flags.go
```diff
@@ -0,0 +1,30 @@
...
+// this package needs to be separate to keep the binary size of the wizard small
+
+const (
+	Command                             = "config-wizard"
+	DefaultFilePathWindowsConfiguration = "C:\\Program Files\\Amazon\\SSM\\Plugins\\awsCloudWatch\\AWS.EC2.Windows.CloudWatch.json"
+	DefaultFilePathLinuxConfiguration   = "/var/awslogs/etc/awslogs.conf"
+)
+
+var WizardFlags = map[string]cmdwrapper.Flag{
+	"is-non-interactive-windows-migration": {DefaultValue: "false", Description: "If true, it will use command line args to bypass the wizard. Default value is false."},
```
jefchien: Why are we changing the flag names?
```
	isNonInteractiveWindowsMigration = flag.Bool("isNonInteractiveWindowsMigration", false,
		"If true, it will use command line args to bypass the wizard. Default value is false.")

	isNonInteractiveLinuxMigration := flag.Bool("isNonInteractiveLinuxMigration", false,
		"If true, it will do the linux config migration. Default value is false.")

	tracesOnly := flag.Bool("tracesOnly", false, "If true, only trace configuration will be generated")
	useParameterStore := flag.Bool("useParameterStore", false,
		"If true, it will use the parameter store for the migrated config storage.")
```

cmd/amazon-cloudwatch-agent/amazon-cloudwatch-agent.go
```diff
@@ -95,6 +102,11 @@ var fRunAsConsole = flag.Bool("console", false, "run as console application (win
 var fSetEnv = flag.String("setenv", "", "set an env in the configuration file in the format of KEY=VALUE")
 var fStartUpErrorFile = flag.String("startup-error-file", "", "file to touch if agent can't start")
 
+// sub-commands
+var fConfigTranslator = flag.Bool(translatorflags.TranslatorCommand, false, "run in config-translator mode")
+var fConfigDownloader = flag.Bool(downloaderflags.Command, false, "run in config-downloader mode")
+var fConfigWizard = flag.Bool(wizardflags.Command, false, "run in config-wizard mode")
```
jefchien: nit: This works, but the Go idiomatic way is to use a `flag.FlagSet`. This way you don't have to prepend all of the flags with the sub-command name. The flags will only be available if the subcommand is present.

```
amazon-cloudwatch-agent config-wizard --isNonInteractiveWindowsMigration
```

vs

```
amazon-cloudwatch-agent --config-wizard --config-wizard-isNonInteractiveWindowsMigration
```

tool/cmdwrapper/cmdwrapper.go
```diff
@@ -0,0 +1,95 @@
...
+	// Use execCommand instead of exec.Command directly
+	cmd := execCommand(agentPath, args...)
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	cmd.Stdin = os.Stdin
+
+	if err := cmd.Run(); err != nil {
+		var exitErr *exec.ExitError
+		if errors.As(err, &exitErr) {
+			log.Panicf("E! %s process exited with non-zero status: %d, err: %v", command, exitErr.ExitCode(), exitErr)
```
jefchien: nit: Should use `log.Fatalf` instead of `log.Panicf` if it's just an error. Also logging the `exec.ExitError` calls the underlying `os.ProcessState` String() function (https://pkg.go.dev/os#ProcessState.String), which will be "exit status <exit_code>", so this error message will look like

```
E! config-wizard process exited with non-zero status: 1, err: exit status 1
```

which doesn't seem that helpful.

tool/cmdwrapper/cmdwrapper.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+		log.Panicf("E! %s failed. Error: %v", command, err)
+		return err
```
jefchien: nit: Isn't this return unreachable? The panic should result in an exit.

cmd/amazon-cloudwatch-agent/amazon-cloudwatch-agent.go
movence: Sure, but I think the intention here is to have clear distinction between subcommand args by prefixing them with the subcommand itself. I don't have a strong position with the prefix so we can also drop them.

cmd/amazon-cloudwatch-agent-config-wizard/wizard.go
movence: yup error is already logged with `Fatalf` so no need to exit here

translator/flags/flags.go
jefchien, line 1: nit: Move to `tool/translator/flags` to match the other tools.

tool/cmdwrapper/cmdwrapper.go
```diff
@@ -0,0 +1,94 @@
...
+
+// Make findAgentBinary a func variable to replace in tests
+var findAgentBinary = func(path string) (string, error) {
+	// check if the binary is at the normal path
+	if _, err := os.Stat(path); err == nil {
+		return path, nil
+	}
+
+	// if not, check in the current executable's directory
+	execPath, err := os.Executable()
```
okankoAMZ, line 30: Found this in Go's documentation, it sounds a bit concerning but I'm not sure if we are effected:
>There is no guarantee that the path is still pointing to the correct executable. If a symlink was used to start the process, depending on the operating system, the result might be the symlink or the path it pointed to. 

[GoDocs](http://pkg.go.dev/os#Executable)

tool/cmdwrapper/cmdwrapper_test.go
```diff
@@ -0,0 +1,142 @@
...
+		if capturedArgs[i] != expected {
+			t.Errorf("Argument %d: expected %s, got %s", i, expected, capturedArgs[i])
+		}
+	}
+}
+
+// Helper function to create string pointer
+func stringPtr(s string) *string {
+	return &s
+}
```
okankoAMZ: Why do we need this as an helper? Can we just do it inline?

tool/downloader/downloader.go
```diff
@@ -0,0 +1,207 @@
...
+		}
+	}()
+
+	cc := commonconfig.New()
+	if inputConfig != "" {
+		f, err := os.Open(inputConfig)
+		if err != nil {
+			return fmt.Errorf("failed to open Common Config: %v", err)
+		}
+		defer f.Close()
```
okankoAMZ, line 56: Why do we close at the end of the function instead of after parse is complete?

tool/cmdwrapper/cmdwrapper_test.go
movence: well I guess this is a common pattern without adding string vars, but can be removed.

tool/cmdwrapper/cmdwrapper.go
movence: This isn’t a major concern for our case, as it would serve only as a simple fallback.

movence:
> Why are we deleting: cmd/amazon-cloudwatch-agent-config-wizard/wizard_test.go

it's just being moved to `tool/wizard/wizard_test.go`

tool/downloader/downloader.go
movence: think it's fine either way. The previous version didn't even close the file.

cmd/amazon-cloudwatch-agent/amazon-cloudwatch-agent.go
jefchien: I was thinking it would be more inline with something like the AWS CLI or kubectl which have subcommands and doesn't expose the flags for each subcommand until it's selected. We're not building a CLI and don't expect people to call the agent directly for this functionality, so it's not essential that the flags and structure be that clean. Just want to make sure we aren't making it more confusing.

Can you try running `amazon-cloudwatch-agent --help` with the changes? I see that we already have some flags that are seemingly being set by our dependencies.

```
> /opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent --help   
Usage of /opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent:
  -aggregator-filter string
    	filter the aggregators to enable, separator is :
  -allow_dynamic_housekeeping
    	Whether to allow the housekeeping interval to be dynamic (default true)
  -application_metrics_count_limit int
    	Max number of application metrics to store (per container) (default 100)
  -boot_id_file string
    	Comma-separated list of files to check for boot-id. Use the first one that exists. (default "/proc/sys/kernel/random/boot_id")
```

tool/cmdwrapper/cmdwrapper.go
```diff
@@ -0,0 +1,94 @@
...
+	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stderr
+	cmd.Stdin = os.Stdin
+
+	if err := cmd.Run(); err != nil {
+		var exitErr *exec.ExitError
+		if errors.As(err, &exitErr) {
+			log.Fatalf("E! %s process exited with non-zero status: %d", command, exitErr.ExitCode())
+		}
+		log.Fatalf("E! %s failed. Error: %v", command, err)
```
jefchien, line 100: nit: Would rather these returned the error and the caller of `ExecuteAgentCommand` logged it and exited.

tool/downloader/downloader.go
jefchien: Ideally, this block would be in it's own function so the defer-close would happen sooner.

tool/translator/translator.go
```diff
@@ -0,0 +1,140 @@
...
+
+func RunTranslator(flags map[string]*string) error {
+	ct, err := NewConfigTranslator(
+		*flags["os"],
+		*flags["input"],
+		*flags["input-dir"],
+		*flags["output"],
+		*flags["mode"],
+		*flags["config"],
+		*flags["multi-config"],
```
jefchien, line 43: nit: Can we define these flag names in the translatorflags package?

tool/cmdwrapper/cmdwrapper.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func AddFlags(prefix string, flagConfigs map[string]Flag) map[string]*string {
+	flags := make(map[string]*string)
+	for key, flagConfig := range flagConfigs {
+		flagName := key
+		if prefix != "" {
+			flagName = prefix + delimiter + flagName
+		}
+		flags[key] = flag.String(flagName, flagConfig.DefaultValue, flagConfig.Description)
```
jefchien: If they're all string flags now, doesn't this break boolean flags? Would this still work?

```
> /opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent-config-wizard --tracesOnly
Do you have an existing X-Ray Daemon configuration file to import for migration?
```

tool/wizard/wizard.go
```diff
@@ -0,0 +1,147 @@
...
+	return nil
+}
+
+func RunWizardFromFlags(flags map[string]*string) error {
+	params := Params{
+		IsNonInteractiveWindowsMigration: *flags["isNonInteractiveWindowsMigration"] == "true",
+		IsNonInteractiveLinuxMigration:   *flags["isNonInteractiveLinuxMigration"] == "true",
+		TracesOnly:                       *flags["tracesOnly"] == "true",
+		UseParameterStore:                *flags["useParameterStore"] == "true",
+		IsNonInteractiveXrayMigration:    *flags["nonInteractiveXrayMigration"] == "true",
```
jefchien, line 99: This means we expect people to call this with
```
--isNonInteractiveLinuxMigration true
```
now instead of
```
--isNonInteractiveLinuxMigration
```
which currently works.

movence: all three cases will be covered: "--key" only, --key=value and "--key value" where flag.BoolFunc sets string value for these cases.

cmd/amazon-cloudwatch-agent/amazon-cloudwatch-agent.go
movence: Yup, this aligns more closely with other CLIs.

tool/cmdwrapper/cmdwrapper.go
```diff
@@ -48,24 +47,39 @@ var findAgentBinary = func(path string) (string, error) {
...
+			})
+			flags[key] = strPtr
+		} else {
+			flags[key] = fs.String(key, flagConfig.DefaultValue, flagConfig.Description)
 		}
-		flags[key] = flag.String(flagName, flagConfig.DefaultValue, flagConfig.Description)
 	}
-	return flags
+	return fs, flags
 }
```
jefchien: nit: If we're going to use flagsets, one option is to have the flagset be defined and created in the `flags.go` (e.g. `tool/wizard/flags/flags.go`). Something like
```go
var FlagSet = flag.NewFlagSet("<command>", flag.ExitOnError)

func init() {
  FlagSet.Bool("blah", false, "some description")
}
```

Then we don't need to define our own `Flag` type. The neat thing about FlagSets is that the `Visit` function only visits set flags, so we can do something like

```go
func getArgsFromFlagSet(fs *flag.FlagSet) []string {
	var result []string
	fs.Visit(func(f *flag.Flag) {
		result = append(result, formatFlag(f))
	})
	return result
}

func formatFlag(f *flag.Flag) string {
	return fmt.Sprintf("--%s=%s", f.Name, f.Value.String())
}
```

That will give us all the arguments for a flagset. It seems like we don't actually use the map for anything other than recreating the arguments.

Example: https://go.dev/play/p/7qrip8_xZW9

cmd/amazon-cloudwatch-agent/amazon-cloudwatch-agent.go
```diff
@@ -504,9 +499,35 @@ func (p *program) Stop(_ service.Service) error {
...
+	// Override flag.Usage to include subcommand help
+	flag.Usage = func() {
+		fmt.Fprintf(os.Stderr, "Usage of %s:\n", os.Args[0])
+		flag.PrintDefaults()
+		fmt.Fprintf(os.Stderr, "\nAvailable subcommands:\n")
+		fmt.Fprintf(os.Stderr, "  %s\t\tTranslate configuration files\n", translatorflags.TranslatorCommand)
+		fmt.Fprintf(os.Stderr, "  %s\t\tDownload configuration from remote sources\n", downloaderflags.Command)
+		fmt.Fprintf(os.Stderr, "  %s\t\t\tInteractive configuration wizard\n", wizardflags.Command)
+		fmt.Fprintf(os.Stderr, "\nUse '%s <subcommand> --help' for more information about a subcommand.\n", os.Args[0])
+	}
```
jefchien, line 534: This is really nice 👍

jefchien (dismissed):

okankoAMZ (dismissed):

cmd/amazon-cloudwatch-agent/amazon-cloudwatch-agent.go
```diff
@@ -492,6 +499,40 @@ func (p *program) Stop(_ service.Service) error {
 
 func main() {
 	flag.Var(&fOtelConfigs, configprovider.OtelConfigFlagName, "YAML configuration files to run OTel pipeline")
+
+	// Check for subcommands first
+	if len(os.Args) > 1 {
+		subcommand := os.Args[1]
+		if subcommand == translatorflags.TranslatorCommand || subcommand == downloaderflags.Command || subcommand == wizardflags.Command {
```
okankoAMZ, line 506: nit: stylistically I'd rather use a switch statement here but I think this is perfectly fine

tool/cmdwrapper/cmdwrapper.go
```diff
@@ -0,0 +1,127 @@
...
+	return "", fmt.Errorf("amazon-cloudwatch-agent binary not found at default path: %s", paths.AgentBinaryPath)
+}
+
+func CreateFlagSet(command string, flagConfigs map[string]Flag) (*flag.FlagSet, map[string]*string) {
+	fs := flag.NewFlagSet(command, flag.ExitOnError)
+	flags := make(map[string]*string)
+	for key, flagConfig := range flagConfigs {
+		if flagConfig.IsBool {
+			strPtr := new(string)
+			*strPtr = flagConfig.DefaultValue
```
okankoAMZ, line 53: Why can't we just do : `strPtr = &(flagConfig.DefaultValue)`, do specifically don't want to update flagConfig value? 

We can also just do a normal string copy and set the pointer at line 62

tool/cmdwrapper/cmdwrapper.go
```diff
@@ -0,0 +1,127 @@
...
+		}
+	}
+	return fs, flags
+}
+
+func ExecuteSubcommand(command string, flags map[string]*string) error {
+	args := []string{command}
+
+	for key, value := range flags {
+		if *value != "" && *value != "false" {
```
okankoAMZ, line 74: could we make these true and falses const enums?

tool/cmdwrapper/cmdwrapper.go
```diff
@@ -0,0 +1,127 @@
...
+				args = append(args, fmt.Sprintf("-%s", key))
+			} else {
+				args = append(args, fmt.Sprintf("-%s", key), *value)
+			}
+		}
+	}
+
+	agentPath, err := findAgentBinary()
+	if err != nil {
+		// Handle error appropriately
```
okankoAMZ, line 85: how are handling the error here? Just bubble it up?

tool/cmdwrapper/cmdwrapper.go
movence: I think this is standard boolean flag behavior and doesn't need additional complexity.

tool/cmdwrapper/cmdwrapper.go
movence: Yes

tool/cmdwrapper/cmdwrapper.go
movence: we want each flag to have its own mutable copy of the default value

tool/wizard/flags/flags.go
```diff
@@ -0,0 +1,30 @@
...
+	"fmt"
+
+	"github.com/aws/amazon-cloudwatch-agent/tool/cmdwrapper"
+)
+
+// this package needs to be separate to keep the binary size of the wizard small
+
+const (
+	Command                             = "config-wizard"
+	DefaultFilePathWindowsConfiguration = "C:\\Program Files\\Amazon\\SSM\\Plugins\\awsCloudWatch\\AWS.EC2.Windows.CloudWatch.json"
```
dricross, line 16: I know this isn't new, but it's weird that this goes to `C:\Program Files\Amazon\SSM`

tool/cmdwrapper/cmdwrapper.go
```diff
@@ -0,0 +1,127 @@
...
+	"path/filepath"
+
+	"github.com/aws/amazon-cloudwatch-agent/tool/paths"
+)
+
+type Flag struct {
+	DefaultValue string
+	Description  string
+	IsBool       bool
+}
```
dricross, line 22: nit: looks like this can only handle Bool and String flags. How would we handle other flag types if we needed in the future?

tool/cmdwrapper/cmdwrapper.go
```diff
@@ -0,0 +1,127 @@
...
+
+	// fallback to the default path
+	if _, err := os.Stat(paths.AgentBinaryPath); err == nil {
+		return paths.AgentBinaryPath, nil
+	}
+
+	return "", fmt.Errorf("amazon-cloudwatch-agent binary not found at default path: %s", paths.AgentBinaryPath)
+}
+
+func CreateFlagSet(command string, flagConfigs map[string]Flag) (*flag.FlagSet, map[string]*string) {
```
dricross, line 47: nit: could this take in a `map[string]flag.Flag` instead of our own struct so it's extensible to any `flag` type?

cmd/amazon-cloudwatch-agent/amazon-cloudwatch-agent.go
```diff
@@ -492,6 +499,40 @@ func (p *program) Stop(_ service.Service) error {
...
+				downloaderflags.Command:           downloaderflags.DownloaderFlags,
+				wizardflags.Command:               wizardflags.WizardFlags,
+			}
+			handlers := map[string]func(map[string]*string) error{
+				translatorflags.TranslatorCommand: translator.RunTranslator,
+				downloaderflags.Command:           downloader.RunDownloaderFromFlags,
+				wizardflags.Command:               wizard.RunWizardFromFlags,
+			}
+
+			if err := cmdwrapper.HandleSubcommand(subcommands, handlers); err != nil {
```
dricross, line 518: nit: we already did all of the work here to find the subcommand and we already know which handler and flag set to use so the `HandleSubcommand` function seems to duplicate a lot of work. Is it used elsewhere or could we just replace `HandleSubcommand` with `handler[subccommand](subcommands[subcommand]flags)` or something similar?

Not that important to update though since the duplicate work is not going to impact performance in any meaningful way

dricross (approved):
Left some nits/comments, but overall looks good.

okankoAMZ (approved):

cmd/amazon-cloudwatch-agent/amazon-cloudwatch-agent.go
movence: Think this is a fair point. We can improve this in a separate PR as a follow up.

PR #1471, opened 2024-12-13
Support Application Signals .NET runtime metrics exporting

bjrara:
# Description of changes
This PR adds metrics transformation and conversion for Application Signals .NET runtime metrics.

# Related PR:
https://github.com/aws/amazon-cloudwatch-agent-operator/pull/279
https://github.com/aws-observability/helm-charts/pull/148

# License

# Tests

# Requirements
2. Run `make lint`


plugins/processors/awsapplicationsignals/internal/metrichandlers/aggregation_mutator.go
```diff
@@ -0,0 +1,67 @@
...
+// Normally, this issue could be resolved by configuring a view with LastValue metrichandlers within the SDK.
+// However, since the view feature is not fully supported in .NET, this workaround implements the required
+// conversion to LastValue metrichandlers to ensure accurate metric reporting.
+// See https://github.com/open-telemetry/opentelemetry-dotnet/issues/2618.
+type AggregationMutator struct {
+	includes map[string]aggregationType
+}
+
+func NewAggregationMutator() AggregationMutator {
+	return newAggregationMutatorWithConfig(map[string]aggregationType{
```
lisguo, line 36: Currently this aggregation mutator is specific to DotNet. Do we plan on having this interface be generic to support other languages? 

I would have a config mapping specific to each language

translator/tocwconfig/sampleConfig/appsignals_and_eks_config.yaml
```diff
@@ -1175,6 +1175,264 @@ processors:
             experimental_scale: 0
             label: ""
             label_value: ""
+        - action: insert
+          aggregation_type: ""
+          experimental_match_labels:
+            generation: gen0
+          include: process.runtime.dotnet.gc.collections.count
+          match_type: ""
+          new_name: DotNetGCGen0Count
```
lisguo, line 1184: The config is getting quite large...is there a way of having a wildcard for each language?

plugins/processors/awsapplicationsignals/internal/metrichandlers/aggregation_mutator.go
lisguo: Also how does this differ from the current runtime metrics we support? Curious why DotNet requires significant code changes

bjrara: The aggregation is specific to .NET because .NET SDK doesn't provide a mechanism to change the aggregation in the `View` model due to its immaturity. We don't have the same problem in the other languages because their SDKs have provided relatively complete support for the `View` model which allows us to tackle the problem at SDK side.

See https://github.com/open-telemetry/opentelemetry-dotnet/issues/2618.

The aggregation mutator is served as an interim solution before .NET finishes the feature development.

bjrara: See https://github.com/open-telemetry/opentelemetry-python-contrib/issues/2861#issuecomment-2386734618 for how we resolved the same issue in Python SDK.

translator/tocwconfig/sampleConfig/appsignals_and_eks_config.yaml
bjrara: I can give it a try, but by peeking at the code, the chance is low.

We take label from metrics and make it part of the new metric name. On the other hand, the wildcard matching and value assignment in `metricstransform` only takes the original metric name and the new metric name into consideration,

https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/ddfadc2006e53b884f1af833c0ac7cead8ed906e/processor/metricstransformprocessor/metrics_transform_processor_otlp.go#L542

```
if newName := transform.MetricIncludeFilter.expand(transform.NewName, metric.Name()); newName != "" {
  metric.SetName(newName)
} else {
  metric.SetName(transform.NewName)
}
```

We cannot configure the new_name to follow our naming strategy (with label values included).

bjrara: Tested, same conclusion.

translator/translate/otel/processor/metricstransformprocessor/appsignals_runtime_config.yaml
```diff
@@ -277,3 +277,121 @@ transforms:
       - action: add_label
         new_label: Telemetry.Source
         new_value: RuntimeMetric
+  - include: process.runtime.dotnet.gc.collections.count
```
sky333999, line 280: I assume these rules will just be a no-op if runtime_metrics is disabled and these metrics arent sent at all in that case?

plugins/processors/awsapplicationsignals/internal/metrichandlers/aggregation_mutator.go
```diff
@@ -0,0 +1,67 @@
...
+		"DotNetGCGen0HeapSize":    lastValueAggregation,
```
sky333999, line 37: Wonder if we should just make `LastValueAggregation` as a const that can be passed in via the processor config directly from the translator and that way, we dont need any mapping here - and is also extensible for any metric keeping this logic here generic.

translator/translate/otel/processor/metricstransformprocessor/appsignals_runtime_config.yaml
bjrara: The runtime_metrics control is at SDK side. The CloudWatch agent stands by to handle runtime metrics whenever the metrics come. If no metrics is received, the rules won't be applied.

plugins/processors/awsapplicationsignals/internal/metrichandlers/aggregation_mutator.go
bjrara: The aggregation mutator (so far) is an interim solution to work around the issue that the last_value aggregation is not yet supported in `View` model in the .NET OTEL SDK. I was very reluctant to introduce this component but had no other option... We haven't observed such feature disparity in other supported language SDK.

Thus in the long term, the plan is to remove this component when .NET is ready. Making it configurable will make the deprecation harder because we will be faced with more compatibility concerns.

sky333999 (approved):

lisguo (approved):

PR #1474, opened 2024-12-17
Retrieve instance tags at the same time to reduce number of entities in Explore related experience

nathalapooja:
# Description of the issue
_Describe the problem or feature in addition to a link to the issues._
Currently we have two separate threads for reading the AutoScalingGroup tag and the service name tags (service, application, or app). This causes different entities to be created as a result which can be confusing for the customer. We should have a single thread (and retryer) to retrieve all tags needed for entity population for the compass experience.

# Description of changes
_How does this change address the problem?_
**Implementation**

We get the ASG + Service Name + all tags on init
Get all tags on an interval together until success

# License
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

# Tests
_Describe what tests you have done._
Unit Tests
Manual Tests
Before enabling the instance metadata tags
```
Field	Value
@entity.Attributes.AWS.ServiceNameSource	ClientIamRole
@entity.Attributes.EC2.InstanceId	i-0b6fe6aab9e76873b
@entity.Attributes.PlatformType	AWS::EC2
@entity.KeyAttributes.Environment	ec2:default
@entity.KeyAttributes.Name	ec2-compass-iam-role
@entity.KeyAttributes.Type	Service
@ingestionTime	1734465217411
@log	957688854012:agent_log_group
@logStream	agent_log_stream
@message	2024-12-17T19:53:35Z D! [logagent] open file count, 1
@timestamp	1734465216354
```
After enabling the instance metadata tags
```
Field	Value
@entity.Attributes.AWS.ServiceNameSource	ResourceTags
@entity.Attributes.EC2.AutoScalingGroup	asg-compass
@entity.Attributes.EC2.InstanceId	i-0b6fe6aab9e76873b
@entity.Attributes.PlatformType	AWS::EC2
@entity.KeyAttributes.Environment	ec2:asg-compass
@entity.KeyAttributes.Name	poojardy_test_app
@entity.KeyAttributes.Type	Service
@ingestionTime	1734465495685
@log	957688854012:agent_log_group
@logStream	agent_log_stream
@message	2024-12-17T19:58:14Z D! [logagent] open file count, 1
@timestamp	1734465495195
```
Agents unlimited retried until successful retrieval of tags based of BackOffSleepArray. Stopped retying after first success
BackoffSleepArray      = []time.Duration{0, 1 * time.Minute, 1 * time.Minute, 3 * time.Minute, 3 * time.Minute, 3 * time.Minute, 10 * time.Minute}
[...]

PR #1477, opened 2024-12-19
set service provider tags to be case insensitive

chadpatel:
# Description of the issue
The `application`, `app` and `service` service name provider ec2 tags are currently case sensitive, there are also bugs in exactness in how these are being compared due to excessive use of `strings.Contains`

# Description of changes
Update `InstanceTags` to return a `[]string` with the tag keys instead of the raw newline delimited list.  This is a more friendly API to reason about.  We can then convert it to a map with lowercase keys that we use for comparison.

I also fixed the asg lookup to also be an exact match instead.  These matches are relatively efficient as I am converting the arrays to maps for O(1) lookups instead of having to repeatedly loop through the whole strings using Contains or slices.Contains

Related to: https://github.com/aws/amazon-cloudwatch-agent/pull/1474

# License

# Tests
Unit tests updated

# Requirements
2. Run `make lint`

### DIFF
--- a/extension/entitystore/serviceprovider.go
+++ b/extension/entitystore/serviceprovider.go
@@ -245,26 +245,28 @@ func (s *serviceprovider) scrapeIAMRole() error {
 	return nil
 }
 func (s *serviceprovider) scrapeImdsServiceNameAndASG() error {
-	tags, err := s.metadataProvider.InstanceTags(context.Background())
+	tagKeys, err := s.metadataProvider.InstanceTags(context.Background())
 	if err != nil {
 		s.logger.Debug("Failed to get service name from instance tags. This is likely because instance tag is not enabled for IMDS but will not affect agent functionality.")
 		return err
 	}
-	// This will check whether the tags contains SERVICE, APPLICATION, APP, in that order.
-	for _, value := range priorityMap {
-		if strings.Contains(tags, value) {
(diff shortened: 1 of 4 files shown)


extension/entitystore/serviceprovider.go
```diff
@@ -245,26 +245,28 @@ func (s *serviceprovider) scrapeIAMRole() error {
...
 		return err
 	}
-	// This will check whether the tags contains SERVICE, APPLICATION, APP, in that order.
-	for _, value := range priorityMap {
-		if strings.Contains(tags, value) {
-			serviceName, err := s.metadataProvider.InstanceTagValue(context.Background(), value)
+
+	// This will check whether the tags contains SERVICE, APPLICATION, APP, in that order (case insensitive)
+	lowerTagKeys := toLowerKeyMap(tagKeys)
+	for _, potentialServiceNameKey := range priorityMap {
```
lisguo, line 256: nit: I know it's not your code but priorityMap isn't a map, it's a list. Should be `priorityList` or something

lisguo (approved):
lgtm with nit

lisguo:
seems pr build failing

chadpatel:
Replaced with https://github.com/aws/amazon-cloudwatch-agent/pull/1478

PR #1483, opened 2025-01-02
Take entity data into account for payload size

varunch77:
# Description of the issue
Due to the addition of the new `EntityMetricData` field to the metrics, the current calculation of the payload size is not accurate. The agent is undercounting the payload size by not accounting for entity data, which leads to instances where the data appears to be under the 1mb limit but is actually over it. As a result, the metrics are not batched properly and cannot be sent to CloudWatch.

# Description of changes
- Update the overheads and existing metric sizes to account for the entity data
- Add a function to calculate entity size
- Update existing unit tests to account for increase in metrics sizes

# License

# Tests
I added a test that replicates the original scenario that was resulting in a `RequestEntityTooLarge` error:
  - the test creates two identical sets of metrics except one set has entity data and other doesn't
  - the set of metrics without entity data fall just below the 1mb threshold
  - the set of metrics with entity data is pushed over the 1mb limit
  - the test checks that the set of metrics with entity data takes 2 API calls and the one without takes 1 API call

# Requirements
2. Run `make lint` ✅


chadpatel (commented):
It is more complicated than this.  I was 

MetricData / EntityData are not directly related with regards to the PMD request structure
https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_PutMetricData.html#API_PutMetricData_RequestParameters

We have a new top level item EntityMetricData

The current splitting code is here
https://github.com/aws/amazon-cloudwatch-agent/blob/ef15aa4922a86cec0d3f182aa800b079d890a95d/plugins/outputs/cloudwatch/cloudwatch.go#L180

It looks like it increments the payload size based on the MetricDatum and then calls `isFull` to determine if the batch is full

The entity information is appended on line 179

It looks like we only send the datum in to the payload calculation on line 180.

I think we need to either update `payload` to also process the `entityStr/Partition` or `+=` additional `Size`

Lisa's commit might help.  Looks like Partition changed from a list to a map
https://github.com/aws/amazon-cloudwatch-agent/commit/9849658464567a9bba0eeeebcfe2185111470bb8

I am not 100% sure where the serialization happens

Maybe BuildMetricDatum or WriteToCloudWatch.  We need to estimate the payload size for the entity that is getting serialized

chadpatel:
I was expected testing beyond unit testing.  Maybe testing end to end or re-producing the current >1MB payload scenario and demonstrating the payload is batched appropriately.  maybe adding something to cloudwatch-agent-tests

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -174,8 +174,11 @@ func (c *CloudWatch) pushMetricDatum() {
 		case metric := <-c.metricChan:
 			entity, datums := c.BuildMetricDatum(metric)
 			numberOfPartitions := len(datums)
+			entityStr := entityToString(entity)
```
zhihonl, line 177: I don't think using `entityToString` for size calculation is correct. That helper function is mainly used to create a temporary entity mapping and the format is not following a HTTP response, an example would be
```
InstanceId:InstanceId;InstanceType:InstanceType|Environment:Environment;Service:Service
```

I think we would need to look into the request body of what EntityData look like then adjust for special characters like brackets, then the entity data itself. An example HTTP response would be 
```
{
EntityMetricData: [
{
Entity: {
Attributes: {
EC2.InstanceId: "i-xxxxxx",
PlatformType: "AWS::EC2",
AWS.ServiceNameSource: "ClientIamRole"
},
KeyAttributes: {
Type: "Service",
AwsAccountId: "123456789",
Name: "CloudWatchAgentAdminRole",
Environment: "ec2:asg-name",
}
},
MetricData: [{
Dimensions: [{
Name: "host",
Value: "ip-123-45-678"
},{
Name: "metric_type",
Value: "counter"
}],
MetricName: "statsd_counter",
StorageResolution: 1,
Timestamp: 2024-10-16 21:49:48.615462983 +0000 UTC,
Unit: "None",
Value: 1
}]
}],
MetricData: [{
Dimensions: [{
Name: "host",
Value: "ip-123-45-67"
},{
Name: "metric_type",
Value: "counter"
}],
MetricName: "statsd_counter",
StorageResolution: 1,
Timestamp: 2024-10-16 21:48:48.615884662 +0000 UTC,
Unit: "None",
Value: 1
}],
Namespace: "TestEntity10",
StrictEntityValidation: false
}
```

plugins/outputs/cloudwatch/util.go
```diff
@@ -25,27 +25,33 @@ const (
 	pmdActionSize = 20
 	// &Version=2010-08-01
 	versionSize = 19
-	// &MetricData.member.100.StatisticValues.Maximum=1558.3086995967291&MetricData.member.100.StatisticValues.Minimum=1558.3086995967291&MetricData.member.100.StatisticValues.SampleCount=1000&MetricData.member.100.StatisticValues.Sum=1558.3086995967291
-	statisticsSize = 246
-	// &MetricData.member.100.Timestamp=2018-05-29T21%3A14%3A00Z
-	timestampSize = 57
-
-	overallConstPerRequestSize = pmdActionSize + versionSize
+	// &EntityMetricData.member.100.MetricData.member.100.StatisticValues.Maximum=1558.3086995967291&EntityMetricData.member.100.MetricData.member.100.StatisticValues.Minimum=1558.3086995967291&EntityMetricData.member.100.MetricData.member.100.StatisticValues.SampleCount=1000&EntityMetricData.member.100.MetricData.member.100.StatisticValues.Sum=1558.3086995967291
```
chadpatel, line 28: should all the stuff in util.go still change?
like I don't quite get this
```
	// &EntityMetricData.member.100.MetricData.member.100.Timestamp=2018-05-29T21%3A14%3A00Z
```
I didn't think they were chained together like that

plugins/outputs/cloudwatch/util.go
```diff
@@ -93,8 +99,8 @@ func resize(dist distribution.Distribution, listMaxSize int) (distList []distrib
 	return
 }
 
-func payload(datum *cloudwatch.MetricDatum) (size int) {
-	size += timestampSize
+func payload(datum *cloudwatch.MetricDatum) int {
```
chadpatel, line 102: presumably this changed for the linter? The linter doesn't like when you use named return values and then basically treat them as initialized?

plugins/outputs/cloudwatch/cloudwatch_test.go
```diff
@@ -700,3 +703,127 @@ func TestWriteToCloudWatchEntity(t *testing.T) {
...
+		Attributes: map[string]*string{
+			"InstanceID": aws.String("i-1234567890abcdef0"),
+			"Platform":   aws.String("AWS::EC2"),
+		},
+	}
+
+	// this is just for payload calculation purposes
+	metricsWithEntityForCalculation := createLargeTestMetrics(numMetrics, &entity)
+	payloadSizeWithEntity := cw.calculateTotalPayloadSize(metricsWithEntityForCalculation, &entity)
+	log.Printf("Payload size without entity info: %d bytes\n", payloadSizeWithEntity)
```
nathalapooja, line 755: This log statement says without entity info but printing withEntity. Remove log statements and assert for payload sizes.

plugins/outputs/cloudwatch/cloudwatch_test.go
```diff
@@ -482,7 +485,7 @@ func TestPublish(t *testing.T) {
 	interval := 60 * time.Second
 	// The buffer holds 50 batches of 1,000 metrics. So choose 5x.
 	numMetrics := 5 * datumBatchChanBufferSize * defaultMaxDatumsPerCall
-	expectedCalls := numMetrics / defaultMaxDatumsPerCall
+	expectedCalls := 388 // Updated to match the observed number of calls
```
nathalapooja, line 488: Is it possible to write a math instead of number to understand the math behind size calculations?

plugins/outputs/cloudwatch/util.go
```diff
@@ -25,27 +25,33 @@ const (
...
 	// &Version=2010-08-01
 	versionSize = 19
-	// &MetricData.member.100.StatisticValues.Maximum=1558.3086995967291&MetricData.member.100.StatisticValues.Minimum=1558.3086995967291&MetricData.member.100.StatisticValues.SampleCount=1000&MetricData.member.100.StatisticValues.Sum=1558.3086995967291
-	statisticsSize = 246
-	// &MetricData.member.100.Timestamp=2018-05-29T21%3A14%3A00Z
-	timestampSize = 57
-
-	overallConstPerRequestSize = pmdActionSize + versionSize
+	// &EntityMetricData.member.100.MetricData.member.100.StatisticValues.Maximum=1558.3086995967291&EntityMetricData.member.100.MetricData.member.100.StatisticValues.Minimum=1558.3086995967291&EntityMetricData.member.100.MetricData.member.100.StatisticValues.SampleCount=1000&EntityMetricData.member.100.MetricData.member.100.StatisticValues.Sum=1558.3086995967291
+	statisticsSize = 358
```
nathalapooja, line 29: Can you explain how did you come up with these new numbers?

plugins/outputs/cloudwatch/util.go
varunch77: Yup, this is due to the `nonamedreturns` rule

plugins/outputs/cloudwatch/util.go
```diff
@@ -120,7 +126,21 @@ func payload(datum *cloudwatch.MetricDatum) (size int) {
 		size += unitOverheads
 	}
 
-	return
+	return size
+}
+
+func calculateEntitySize(entity cloudwatch.Entity) int {
+	size := strictEntityValidationsize
```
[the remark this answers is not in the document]
varunch77: addressed in [38b022c](https://github.com/aws/amazon-cloudwatch-agent/pull/1483/commits/38b022c8fc5a572e2e9ac086fb30bed52785507f)

plugins/outputs/cloudwatch/util.go
varunch77: This is what was there before:

https://github.com/aws/amazon-cloudwatch-agent/blob/a090dd351cca4c44c4dd3d2a499cb81644f405bc/plugins/outputs/cloudwatch/util.go#L28-L29

The new string takes the old string and adds the "EntityMetricData.member.100." prefix to each of the StatisticValues (Maximum, Minimum, SampleCount, Sum) so we get 246 + (28 * 4) = 358

lisguo:
I would recommend getting consensus on the expected payload size from CW front end for a given metric. Let's make sure our assumptions are correct before merging this.

Also pr builds seem to be failing?

plugins/outputs/cloudwatch/util.go
varunch77: Reverted these changes

plugins/outputs/cloudwatch/util.go
varunch77: Reverted these changes

plugins/outputs/cloudwatch/cloudwatch_test.go
varunch77: Addressed in [4b0a375](https://github.com/aws/amazon-cloudwatch-agent/pull/1483/commits/4b0a375aee5bebeb8f0ec1b699ecc8ecb7e80999)

.golangci.yml
```diff
@@ -19,7 +19,7 @@ run:
 
 output:
   # colored-line-number|line-number|json|tab|checkstyle|code-climate, default is "colored-line-number"
-  format: colored-line-number
+  formats: colored-line-number
```
varunch77, line 22: just gets rid of a warning when we run our linter

chadpatel (approved):

plugins/outputs/cloudwatch/util.go
```diff
@@ -46,6 +46,19 @@ const (
 	valueOverheads = 47
 	// &MetricData.member.1.Unit=Kilobytes/Second
 	unitOverheads = 42
+
```
Paramadon, line 49: Why is everything commented out? Are we waiting for a release?

varunch77: We were originally going to proceed with this change(https://github.com/aws/amazon-cloudwatch-agent/compare/af960d7...459ef7c), but after aligning with all of the metrics/cloudwatch teams we determined that it's best to leave things as is and revisit this in the future after changes to the API/SDK. I'm just leaving some breadcrumbs in the code regarding this change for future purposes.

Paramadon (approved):

PR #1486, opened 2025-01-06
[awsentityprocessor] Use existing attributes from resource detection processor in application signals to populate entity

duhminick:
# Description of the issue
Currently the awsentityprocessor checks IMDS for EC2 instance tags; however, this requires opt-in from the customer for IMDS to have the instance tags.

Instead, if the customer is using application signals (thus resource detector), the agent will make an ec2:DescribeTags call, which will retrieve the instance tags such as the auto scaling group (ASG) name. We can piggyback off this logic and update the entitystore to use the ASG from the DescribeTags call.

# Description of changes

There are essentially three changes made:
1. Config translation changes to add the awsentity processor for EC2 hosts
2. The awsentity processor will now check the resource attributes to see if the [resourcedetection processor](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/resourcedetectionprocessor/README.md) populated `ec2.tag.aws:autoscaling:groupName`. If so, then set the ASG in the entity store
3. The entitystore will now set the ASG at most one time. This is because to update the ASG, we need to hold a write lock. We don't want to hold the write lock every time we see that the ASG is populated. So a sync.Once is used.

# License

# Tests
1. Unit test
2. Manual testing using an EC2 instance that was created from an ASG

## Manual test

The following test has IMDS disabled.

**Before an application signals metric was sent**
image

**The metric that was sent**
image

**Application logs after the application signals metric**

Notice that the ASG is now populated.

image

## Sanity test

As a sanity test, I did the same test but with the latest commit that did not have this change. 

**Application logs after the application signals metric**

Notice that the ASG is **not** populated.

image

# Requirements
6. Run `make lint`


musa-asad:
Can you add the steps and screenshots from manual testing?

extension/entitystore/serviceprovider.go
```diff
@@ -53,13 +53,18 @@ type ServiceAttribute struct {
 type LogGroupName string
 type LogFileGlob string
 
+type autoscalinggroup struct {
```
lisguo, line 56: this is probably fine....wondering if there is an edge case where the ASG could change (ie the tag would get updated). Currently we poll the tag at an interval. 

But in this context maybe its okay to leave as is? Or maybe have a cache where resource detection processor updates the asg if it changes? Might be overkill.

extension/entitystore/serviceprovider.go
```diff
@@ -101,7 +106,21 @@ func (s *serviceprovider) GetIMDSServiceName() string {
 func (s *serviceprovider) getAutoScalingGroup() string {
 	s.mutex.RLock()
 	defer s.mutex.RUnlock()
-	return s.autoScalingGroup
+	return s.autoScalingGroup.name
+}
+
+func (s *serviceprovider) setAutoScalingGroup(asg string) {
+	s.autoScalingGroup.once.Do(func() {
+		s.mutex.Lock()
```
lisguo, line 114: Can this just be a WLock()?

translator/translate/otel/pipeline/applicationsignals/translator.go
```diff
@@ -59,11 +59,12 @@ func (t *translator) Translate(conf *confmap.Conf) (*common.ComponentTranslators
...
 
-	mode := context.CurrentContext().KubernetesMode()
 	translators.Processors.Set(resourcedetection.NewTranslator(resourcedetection.WithDataType(t.dataType)))
 	translators.Processors.Set(awsapplicationsignals.NewTranslator(awsapplicationsignals.WithDataType(t.dataType)))
 
-	if t.dataType == component.DataTypeMetrics && mode != "" {
+	// ECS is not in scope for entity association, so we only add the entity processor in non-ECS platforms
+	isECS := ecsutil.GetECSUtilSingleton().IsECS()
+	if t.dataType == component.DataTypeMetrics && !isECS {
 		translators.Processors.Set(awsentity.NewTranslatorWithEntityType(awsentity.Service, common.AppSignals, false))
```
lisguo, line 68: Can you provide sample AppSignals emf logs on EC2 after this change? We should ensure there's no extra attributes we are sending as a result.

extension/entitystore/serviceprovider.go
duhminick: So turns out, yes, it's possible to move instances around from ASG to ASG: https://docs.aws.amazon.com/autoscaling/ec2/userguide/ec2-auto-scaling-detach-attach-instances.html. 

As for resource detection processor, from what I've gathered by looking at the code (https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/resourcedetectionprocessor/resourcedetection_processor.go#L29-L36), it grabs the tags (thus the ASG) at startup then caches it. The cached tag is then added to the attributes. 

So the issue is here for when we poll for the tags at an interval. What are your thoughts on keeping this as is (i.e. my current implementation)

extension/entitystore/serviceprovider.go
duhminick: Are you referring to just `Lock()`? The code here is getting a lock for writing (which was my goal)

Go doc: https://pkg.go.dev/sync#RWMutex.Lock

extension/entitystore/serviceprovider.go
lisguo: If the resourcedetection processor caches it, then I think it's fine to keep as-is

lisguo (approved):

extension/entitystore/serviceprovider.go
```diff
@@ -101,7 +106,21 @@ func (s *serviceprovider) GetIMDSServiceName() string {
 func (s *serviceprovider) getAutoScalingGroup() string {
 	s.mutex.RLock()
 	defer s.mutex.RUnlock()
-	return s.autoScalingGroup
+	return s.autoScalingGroup.name
+}
+
+func (s *serviceprovider) setAutoScalingGroup(asg string) {
+	s.autoScalingGroup.once.Do(func() {
```
zhihonl, line 113: IIRC `once.Do` only execute once so effectively if `setAutoScalingGroup` was called once, any future calls will not execute anything here right? 

If so, would this be an issue if IMDS metadata tag was not enabled, the initial call was made without getting anything, then we try to get the ASG through AppSignal processor? Since we are using `once.Do` and the call to get ASG from AppSignal processor won't be executed

zhihonl: hmm reading your testing looks like it does work even for the scenario I described, wonder why

duhminick: > IIRC once.Do only execute once so effectively if setAutoScalingGroup was called once, any future calls will not execute anything here right?

Yes

> If so, would this be an issue if IMDS metadata tag was not enabled, the initial call was made without getting anything, then we try to get the ASG through AppSignal processor? Since we are using once.Do and the call to get ASG from AppSignal processor won't be executed

So for the logic where we retrieve from IMDS, if it fails/is empty (i.e. endpoint is unreachable, or tags in IMDS is disabled), then there is no ASG. The call to setAutoScalingGroup in either case is not done. So instead the entity processor will make the *one* call.

Likewise, if IMDS is enabled, then the ASG should have been populated from there, which means that it won't be populated from the entity processor

zhihonl: Oh I see this line is why: https://github.com/aws/amazon-cloudwatch-agent/blob/4a30b31ac5df64138a2b83b858be6a4f81edf31d/extension/entitystore/serviceprovider.go#L297

basically if IMDS tag is not enabled we get empty value and the subsequent calls are never made

zhihonl (approved):

PR #913, opened 2023-10-17
Add New Metrics For Where Customer Are Using The Agent

sethAmazon:
# Description of the issue
CloudWatch Agent wants to collect metrics on where the agent is being run. 

# Description of changes
Add metrics for CloudWatch Agent mode and region provider. I did this by creating global variables for these metrics and shoving them into the config layer to be picked up by the plugins. This is because this information is only known at translation time, but we need this information at plugin runtime. 

# License
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

# Tests
Unit tests

# Requirements
_Before commit the code, please do the following steps._
1. Run `make fmt` and `make fmt-sh`
2. Run `make lint`

PR #1499, opened 2025-01-13
Add worker pool for log pushers

jefchien:
# Description of the issue
One of the bottlenecks for throughput for the agent is that each target (log group/stream) has a single-threaded pusher which blocks until it has successfully sent the request or run out of retries. This limits the amount of requests that can be made and limits the size/rate of increase for the tailed log file.

# Description of changes
By supporting concurrency with the sender, the agent is able to send and prepare batches more efficiently.

- Moved the `pusher.go` into a `pusher` package and split the queue/batch functionality from the sender.
- Added a simple `WorkerPool` that queues up tasks, which are taken by the first available worker.
- Added a `Concurrency` field. If configured, the CloudWatch logs will have a shared pool of workers to send PLE requests.
- Changed the `perEventHeaderBytes` from 200 back to 26. This was arbitrarily changed as part of https://github.com/aws/amazon-cloudwatch-agent/pull/913, but does not match the [PLE specifications](https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_PutLogEvents.html) that it was meant to be taken from.
> The maximum batch size is 1,048,576 bytes. This size is calculated as the sum of all event messages in UTF-8, plus 26 bytes for each log event.

- Added a TargetManager to prevent multiple calls to `CreateLogStream` or `CreateLogGroup` when sends are called concurrently for a new log group/stream. Partially adopted from OTEL implementation https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/internal/aws/cwlogs/pusher.go#L386.
- Created diagram using https://asciiflow.com/#/

# License

# Tests
Added unit tests.

One example in the unit tests for comparison is the `TestPusher`. With a 50ms latency and 100000 events, the sender pool configuration with 5 workers took 1/3 of the time as the single-threaded sender.
```
    --- PASS: TestPusher/WithSender (0.67s)
    --- PASS: TestPusher/WithSenderPool (0.23s)
```

Integration test run: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/12793873872
Similar failing tests as the `main` branch.

*Benchmarks in progress.*

# Requirements
2. Run `make lint`


translator/tocwconfig/sampleConfig/log_filter.conf
```diff
@@ -49,6 +49,7 @@
 [outputs]
 
   [[outputs.cloudwatchlogs]]
+  	concurrency = 10
```
chadpatel, line 52: nit spacing

plugins/outputs/cloudwatchlogs/README.md
```diff
@@ -0,0 +1,34 @@
+# Amazon CloudWatch Logs Output Plugin
+
+For each configured target (log group/stream), the output plugin maintains a queue for log events that it batches.
+Once each batch is full or the flush interval is reached, the current batch is sent using the PutLogEvents API to Amazon CloudWatch.
+
+When concurrency is enabled, the pusher uses a shared worker pool to allow multiple concurrent sends.
```
chadpatel, line 6: why isn't this default behavior?  Default of like min(4, # of cores)

Are we considering it experimental until we get some real world feedback?

plugins/outputs/cloudwatchlogs/internal/pusher/pool.go
```diff
@@ -0,0 +1,110 @@
...
+	Submit(task func())
+	Stop()
+}
+
+type workerPool struct {
+	tasks       chan func()
+	workerCount atomic.Int32
+	wg          sync.WaitGroup
+	stopCh      chan struct{}
+	stopped     atomic.Bool
```
chadpatel, line 22: my initial thought is mixing atomics and channels is kind of weird

jefchien: Do you mean the `stopped` and the `stopCh`? I think you're right. I can probably remove the `atomic.Bool` and just use the channel. I initially just had `stopped` and didn't remove it once I added the channel.

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -138,34 +162,24 @@ func (c *CloudWatchLogs) getDest(t Target, logSrc logs.LogSrc) *cwDest {
...
-	if containerInsightsRegexp.MatchString(t.Group) {
-		useragent.Get().SetContainerInsightsFlag()
-	}
 	client.Handlers.Build.PushBackNamed(handlers.NewRequestCompressionHandler([]string{"PutLogEvents"}))
 	if c.middleware != nil {
 		if err := awsmiddleware.NewConfigurer(c.middleware.Handlers()).Configure(awsmiddleware.SDKv1(&client.Handlers)); err != nil {
 			c.Log.Errorf("Unable to configure middleware on cloudwatch logs client: %v", err)
 		} else {
-			c.Log.Info("Configured middleware on AWS client")
+			c.Log.Debug("Configured middleware on AWS client")
```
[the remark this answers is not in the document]
jefchien: It didn't seem useful as an Info log and the middleware is mostly for us.

Makefile
```diff
@@ -211,7 +211,6 @@ PKG_WITH_DATA_RACE += internal/tls
 PKG_WITH_DATA_RACE += plugins/inputs/logfile
 PKG_WITH_DATA_RACE += plugins/inputs/logfile/tail
 PKG_WITH_DATA_RACE += plugins/outputs/cloudwatch
-PKG_WITH_DATA_RACE += plugins/outputs/cloudwatchlogs
```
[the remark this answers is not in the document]
jefchien: We don't.

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -124,11 +128,31 @@ func (c *CloudWatchLogs) CreateDest(group, stream string, retention int, logGrou
...
+func (c *CloudWatchLogs) getDest(t pusher.Target, logSrc logs.LogSrc) *cwDest {
 	if cwd, ok := c.cwDests[t]; ok {
 		return cwd
 	}
 
+	logThrottleRetryer := retryer.NewLogThrottleRetryer(c.Log)
+	client := c.createClient(logThrottleRetryer)
+	agent.UsageFlags().SetValue(agent.FlagRegionType, c.RegionType)
+	agent.UsageFlags().SetValue(agent.FlagMode, c.Mode)
+	if containerInsightsRegexp.MatchString(t.Group) {
```
lisguo, line 140: I thought container insights doesn't send through cloudwatch output plugin? The agent only uses emf?

lisguo: Oh I see you just moved it from createClient()...

plugins/outputs/cloudwatchlogs/internal/pusher/pool.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// addWorker creates and starts a new worker goroutine.
+func (p *workerPool) addWorker() {
+	p.wg.Add(1)
+	p.workerCount.Add(1)
+	go p.worker()
+}
+
+func (p *workerPool) worker() {
```
lisguo, line 43: super nit: worker() func doesn't have a comment

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
```diff
@@ -0,0 +1,229 @@
...
+	"time"
+
+	"github.com/influxdata/telegraf"
+
+	"github.com/aws/amazon-cloudwatch-agent/logs"
+	"github.com/aws/amazon-cloudwatch-agent/profiler"
+)
+
+type Queue interface {
+	AddEvent(e logs.LogEvent)
```
lisguo, line 18: Can you provide an example of when we would add a blocking event vs non blocking?

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+func (m *targetManager) createLogStream(t Target) error {
+	_, err := m.service.CreateLogStream(&cloudwatchlogs.CreateLogStreamInput{
+		LogGroupName:  &t.Group,
+		LogStreamName: &t.Stream,
+	})
+
+	if err == nil {
+		m.logger.Debugf("successfully created log stream %v", t.Stream)
```
lisguo, line 112: shouldn't we log the error instead?

translator/config/schema.json
```diff
@@ -898,6 +898,11 @@
           "type": "string",
           "minLength": 1,
           "maxLength": 259
+        },
+        "concurrency": {
+          "description": "The number of concurrent workers available for cloudwatch logs export",
+          "type": "integer",
+          "minimum": 1
```
lisguo, line 905: should we define a max?

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
jefchien: Historically, we had functionality where if the log group matched the regex (`^/aws/.*containerinsights/.*/(performance|prometheus)$`), then we'd count it as container insights. I don't think this path is used anymore, but we can verify that and clean it up in a separate PR.

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
jefchien: This is also legacy functionality. The `AddNonBlockingEvent` was used for EMF logs. We've moved that over to OTEL, but don't know if there are any customers using TOML to still send EMF logs this way.

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
jefchien: The error is nil here, so there's nothing to log.

translator/config/schema.json
jefchien: We didn't for the X-Ray concurrency field https://github.com/aws/amazon-cloudwatch-agent/blob/09f8b09a3965f962dc4083c6212b24cec8143491/translator/config/schema.json#L1172-L1176

plugins/outputs/cloudwatchlogs/internal/pusher/pool.go
jefchien: I'll add one. I need to clean up the comments.

chadpatel (approved):

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
```diff
@@ -0,0 +1,140 @@
...
+	return &targetManager{
+		logger:  logger,
+		service: service,
+		cache:   make(map[Target]struct{}),
+	}
+}
+
+// InitTarget initializes a Target if it hasn't been initialized before.
+func (m *targetManager) InitTarget(target Target) error {
+	_, err, _ := m.group.Do(target.String(), func() (any, error) {
```
sky333999, line 49: Does this offer much benefits as opposed to a map with a mutex?

translator/translate/logs/ruleConcurrency.go
```diff
@@ -0,0 +1,24 @@
...
+import "github.com/aws/amazon-cloudwatch-agent/translator"
+
+const ConcurrencySectionKey = "concurrency"
+
+type Concurrency struct {
+}
+
+func (c *Concurrency) ApplyRule(input any) (string, any) {
+	result := map[string]interface{}{}
+	_, val := translator.DefaultCase(ConcurrencySectionKey, float64(0), input)
```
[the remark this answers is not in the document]
jefchien: The translator uses the `json` package to unmarshal the configuration and all numbers are unmarshalled as float64s.

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
jefchien: No that's a good catch. Was missing a mutex.

okankoAMZ (approved):

sky333999 (approved):

chadpatel (approved):

PR #1516, opened 2025-01-23
Use OpenTelemetry configtls for certificate refresh

duhminick:
# Description of the issue
OTel already has a way to refresh certificates when updated, so there's no (real) reason to roll our own atm.

# Description of changes
1. Use OTels' configtls module to handle the certificate retrieval and refresh handling.
a. If using the `ReloadClientCAFile` option, then https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configtls/clientcasfilereloader.go is used in which it uses `fsnotify`, similar to the cert watcher that is currently being used.

2. Remove the logic where we're "reloading" the server. This was not necessary to have.
a. The tls.Config struct has a function as a field 

    GetConfigForClient func(*[ClientHelloInfo](https://pkg.go.dev/crypto/tls#ClientHelloInfo)) (*[Config](https://pkg.go.dev/crypto/tls#Config), [error](https://pkg.go.dev/builtin#error))

    The assignment is done in https://github.com/open-telemetry/opentelemetry-collector/blob/main/config/configtls/configtls.go#L408 which points to `getClientConfig` defined in the reloader. This is what allows for the cert pool to be "dynamic" -- allowing us to not have to "reload" the server. It's usage during the TLS handshake (more info [here](https://www.cloudflare.com/learning/ssl/what-happens-in-a-tls-handshake/)) can be found in https://github.com/golang/go/blob/608acff8479640b00c85371d91280b64f5ec9594/src/crypto/tls/handshake_server.go#L160-L168

# License

# Tests
> [!NOTE]
> This is going to be a bit more detailed for the sake of having the steps defined somewhere in case we want to re-create an environment for this. 

Verified that a newly generated certificate is used and is working with the fluentbit pod.

### Cert-manager
1. Installed cert-manager using Helm
2. Update the agent (if already installed) to use certManager for the agent
```sh
helm upgrade --namespace amazon-cloudwatch amazon-cloudwatch-observability aws-observability/amazon-cloudwatch-observability \
  --set clusterName=CLUSTER_NAME_HERE \
  --set region=us-east-1 \
  --set admissionWebhooks.certManager.enabled=true \
  --set agent.certManager.enabled=true
```

### Sidecar
Added a sidecar container with read-only mounts to the certs by editing
`$ kubectl edit AmazonCloudWatchAgent cloudwatch-agent -n amazon-cloudwatch`
to add a container that has a shell (e.g. Ubuntu image)

 
1. Initially test that the endpoint works by hitting the endpoint
```sh
$ kubectl exec $(kubectl get pods --namespace amazon-cloudwatch | rg fluent | awk '{print $1}' | head -n 1) --namespace amazon-cloudwatch -- cat /etc/amazon-cloudwatch-observability-agent-client-cert/client.crt > client.crt
$ kubectl exec $(kubectl get pods --namespace amazon-cloudwatch | rg fluent | awk '{print $1}' | head -n 1) --namespace amazon-cloudwatch -- cat /etc/amazon-cloudwatch-observability-agent-client-cert/client.key > client.key
$ kubectl port-forward $(kubectl get pods --namespace amazon-cloudwatch | rg fluent | awk '{print $1}' | head -n 1) 4311 4311
$ curl -ivk --cert client.crt --key client.key https://localhost:4311/kubernetes/pod-to-service-env-map
```

2. Open a shell into the sidecar container and `cat` the contents. This is later used to verify if the certificate is updated
3. Rotate the certificates
`$ cmctl renew --namespace amazon-cloudwatch --all`
4. Verify that the certificate is rotated in the sidecar container. This might take a few seconds.
5. I tried hitting the endpoint again with `$ curl -ivk --cert client.crt --key client.key https://localhost:4311/kubernetes/pod-to-service-env-map` and verified that it fails. Meaning that the old certificates are no longer valid.
6. Retrieved the new certs and hit the endpoint successfully
```sh
$ kubectl exec $(kubectl get pods --namespace amazon-cloudwatch | rg fluent | awk '{print $1}' | head -n 1) --namespace amazon-cloudwatch -- cat /etc/amazon-cloudwatch-observability-agent-client-cert/client.crt > client.crt
$ kubectl exec $(kubectl get pods --namespace amazon-cloudwatch | rg fluent | awk '{print $1}' | head -n 1) --namespace amazon-cloudwatch -- cat /etc/amazon-cloudwatch-observability-agent-client-cert/client.key > client.key
$ kubectl port-forward $(kubectl get pods --namespace amazon-cloudwatch | rg fluent | awk '{print $1}' | head -n 1) 4311 4311
$ curl -ivk --cert client.crt --key client.key https://localhost:4311/kubernetes/pod-to-service-env-map
```

# Requirements
8. Run `make lint`


duhminick:
image

Closing the PR. Was using the testing steps that were previously provided, but it looks like there was a gap in that using `-k` in curl is the same as `--insecure`. That's why the testing passed when no `--cacert` was passed. If those options are fixed, then we can see that this option is not viable like I had previously thought.

musa-asad:
You can reference the steps in https://github.com/aws/amazon-cloudwatch-agent-operator/pull/284

PR #1517, opened 2025-01-23
Removing Caching from integ test

Paramadon:
# Description of the issue
Caching was intially put in place to reduce the number of tests that needed to be re-run on partial failures in an effort to speed up testing. It currently is put in place to skip passing tests on re-runs. But since GitHub workflows supports [re-running failed jobs](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/re-running-workflows-and-jobs#re-running-failed-jobs-in-a-workflow), which already covers the purpose of the caching. Caching conflicts have also resulted in false positives for passing tests like the stress tests.

# Description of changes
This pr removes caching to avoid false negatives.

Integration test run: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/12997888061

### DIFF
--- a/.github/workflows/ec2-integration-test.yml
+++ b/.github/workflows/ec2-integration-test.yml
@@ -65,13 +65,6 @@ jobs:
           aws-region: ${{inputs.region}}
           role-duration-seconds: ${{ env.TERRAFORM_AWS_ASSUME_ROLE_DURATION }}
 
-      - name: Cache if success
-        id: cache_if_success
-        uses: actions/cache@v3
-        with:
-          path: go.mod
-          key: ${{inputs.region}}-${{ github.sha }}-${{ matrix.arrays.os }}-${{ matrix.arrays.arc }}-${{ matrix.arrays.test_dir }}
-
       - name: Echo Test Info
         run: |
           echo run cache_if_success os ${{ matrix.arrays.os }} arc ${{ matrix.arrays.arc }} test dir ${{ matrix.arrays.test_dir }}
--- a/.github/workflows/ec2-integration-test.yml
+++ b/.github/workflows/ec2-integration-test.yml
@@ -65,6 +65,13 @@ jobs:
           aws-region: ${{inputs.region}}
           role-duration-seconds: ${{ env.TERRAFORM_AWS_ASSUME_ROLE_DURATION }}
 
+      - name: Cache if success
+        id: cache_if_success
+        uses: actions/cache@v3
+        with:
+          path: go.mod
+          key: ${{inputs.region}}-${{ github.sha }}-${{ matrix.arrays.os }}-${{ matrix.arrays.arc }}-${{ matrix.arrays.test_dir }}
+
       - name: Echo Test Info
         run: |
           echo run cache_if_success os ${{ matrix.arrays.os }} arc ${{ matrix.arrays.arc }} test dir ${{ matrix.arrays.test_dir }}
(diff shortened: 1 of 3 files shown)


.github/workflows/integration-test.yml
```diff
@@ -154,15 +154,7 @@ jobs:
           aws-region: us-west-2
           role-duration-seconds: ${{ env.TERRAFORM_AWS_ASSUME_ROLE_DURATION }}
 
-      - name: Cache if success
-        id: cf-integration-test
-        uses: actions/cache@v2
-        with:
-          path: go.mod
-          key: "cf-integration-${{ github.sha }}-test"
-
```
TravisStark, line 163: Did you need to remove this cache as well?

.github/workflows/integration-test.yml
TravisStark, line 1: Did you need to remove this cache too?
```
      - name: Cache if success
        id: stress-tracking
        uses: actions/cache@v3
        with:
          path: go.mod
          key: stress-tracking-test-${{ github.sha }}-${{ matrix.arrays.os }}-${{ matrix.arrays.arc }}-${{ matrix.arrays.test_dir }}
```

Paramadon: Since EC2IntegrationTest isn't cache_if_success, we can remove it. The reason why it isn't removed for all of them is because we can keep cache for setting up (Login ECR, Intall Terraform, Verify Terroform version) but for testing itself we can remove it.

.github/workflows/eks-e2e-test.yml
musa-asad, line 1: Need to remove:
```
- name: Cache if success
        id: cache_if_success
        uses: actions/cache@v3
        with:
          path: go.mod
          key: ${{inputs.region}}-${{ github.sha }}-${{ matrix.arrays.os }}-${{ matrix.arrays.arc }}-${{ matrix.arrays.test_dir }}
```

.github/workflows/ec2-integration-test.yml
```diff
@@ -65,13 +65,6 @@ jobs:
...
-      - name: Cache if success
-        id: cache_if_success
-        uses: actions/cache@v3
-        with:
-          path: go.mod
-          key: ${{inputs.region}}-${{ github.sha }}-${{ matrix.arrays.os }}-${{ matrix.arrays.arc }}-${{ matrix.arrays.test_dir }}
-
       - name: Echo Test Info
         run: |
           echo run cache_if_success os ${{ matrix.arrays.os }} arc ${{ matrix.arrays.arc }} test dir ${{ matrix.arrays.test_dir }}
```
musa-asad, line 70: this seems to call `cache_if_success`

Paramadon: I added it back, but the test seems to be still passing.
Screenshot 2025-01-24 at 10 57 55 AM

musa-asad (approved):

TravisStark (approved):

.github/workflows/ec2-integration-test.yml
jefchien, line 1: Why aren't we removing this cache?

```
      - name: Cache if success
        id: cache_if_success
        uses: actions/cache@v3
        with:
          path: go.mod
key: ${{inputs.region}}-${{ github.sha }}-${{ matrix.arrays.os }}-${{ matrix.arrays.arc }}-${{ matrix.arrays.test_dir }}
```

jefchien (changes requested):
I'm still seeing some caches https://github.com/aws/amazon-cloudwatch-agent/actions/caches?query=branch%3ARemovingCaching

musa-asad (approved):

jefchien (approved):

PR #1528, opened 2025-02-05
add endpoint slice watcher

pxaws:
# Description of the issue

# Description of changes

To reduce the overhead of ListPod calls on Kubernetes API server at cloudwatch agent startup, we want to switch to use endpoint slice API. The endpoint slice API has much smaller response size. We will use it to figure out the ip to workload mapping for Kubernetes services. 

# License

# Tests

# Requirements
2. Run `make lint`


plugins/processors/awsapplicationsignals/internal/resolver/endpointslicewatcher.go
```diff
@@ -0,0 +1,297 @@
...
+			if endpoint.TargetRef.Kind == "Pod" {
+				podName = endpoint.TargetRef.Name
+			}
+			ns = endpoint.TargetRef.Namespace
+
+			derivedWorkload := inferWorkloadName(podName)
+			if derivedWorkload == "" {
+				w.logger.Warn("failed to infer workload name from Pod name", zap.String("podName", podName))
+				continue
+			}
```
mxiamxia, line 107: not sure the chance of failure on this? if it happens, can we directly use service name from `metadata.name` in EP slice to resolve the ip? My concern, will we skip the IP resolving if something goes wrong here from deployment name?

plugins/processors/awsapplicationsignals/internal/resolver/kubernetes_utils.go
```diff
@@ -101,6 +108,40 @@ func getWorkloadAndNamespace(pod *corev1.Pod) string {
...
+		if rsMatches := deploymentFromReplicaSetPattern.FindStringSubmatch(parentName); rsMatches != nil {
+			return rsMatches[1] // e.g. "nginx-a2b3c4" => "nginx"
+		}
+
+		// Otherwise, it's a "bare" ReplicaSet or DaemonSet—just return parentName.
+		return parentName
+	}
+
+	// 3) If none of the patterns matched, return the entire podName.
+	return podName
```
mxiamxia, line 142: Hope this wont happen. LOL. high cardinality.

pxaws: yeah, I think the chance is low. Even when it happens, we have the metric limiter to mitigate the high cardinality issue.

plugins/processors/awsapplicationsignals/internal/resolver/endpointslicewatcher.go
pxaws: inferWorkloadName return the podName in the worst case. I think it's very unlikely that the podname will be empty.

mxiamxia (dismissed):

plugins/processors/awsapplicationsignals/internal/resolver/endpointslicewatcher.go
```diff
@@ -0,0 +1,297 @@
...
+		// If there's at least one endpoint, derive the workload from it
+		if len(slice.Endpoints) > 0 {
+			if endpoint := slice.Endpoints[0]; endpoint.TargetRef != nil {
+				derived := inferWorkloadName(endpoint.TargetRef.Name)
+				if derived != "" {
+					firstWl := derived + "@" + endpoint.TargetRef.Namespace
+					pairs = append(pairs, kvPair{
+						key:       svcFull,
+						value:     firstWl,
+						isService: true,
```
mxiamxia, line 149: nit: should we follow the same pattern that we used on IP to workload handling?

`if derivedWorkload == "" {
				w.logger.Warn("failed to infer workload name from Pod name", zap.String("podName", podName))
				continue
			}`

pxaws: Right. I will modify it.

plugins/processors/awsapplicationsignals/internal/resolver/endpointslicewatcher.go
```diff
@@ -0,0 +1,297 @@
...
+func (w *endpointSliceWatcher) extractEndpointSliceKeyValuePairs(slice *discv1.EndpointSlice) []kvPair {
+	var pairs []kvPair
+
+	for _, endpoint := range slice.Endpoints {
+		podName := ""
+		ns := ""
+		if endpoint.TargetRef != nil {
+			if endpoint.TargetRef.Kind == "Pod" {
+				podName = endpoint.TargetRef.Name
+			}
```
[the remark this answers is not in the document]
pxaws: right, we should skip if it's not pod

plugins/processors/awsapplicationsignals/internal/resolver/endpointslicewatcher.go
```diff
@@ -0,0 +1,297 @@
...
+//   - The Service name key (isService=true) -> first "workload@ns" found
+//
+// This function does NOT modify ipToWorkload or serviceToWorkload. It's purely for computing
+// the pairs, so it can be reused by both add and update methods.
+func (w *endpointSliceWatcher) extractEndpointSliceKeyValuePairs(slice *discv1.EndpointSlice) []kvPair {
+	var pairs []kvPair
+
+	for _, endpoint := range slice.Endpoints {
+		podName := ""
+		ns := ""
```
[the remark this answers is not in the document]
pxaws: I will move it inside the if block.

plugins/processors/awsapplicationsignals/internal/resolver/endpointslicewatcher.go
```diff
@@ -0,0 +1,297 @@
...
+					firstWl := derived + "@" + endpoint.TargetRef.Namespace
+					pairs = append(pairs, kvPair{
+						key:       svcFull,
+						value:     firstWl,
+						isService: true,
+					})
+				}
+			}
+		}
+	}
```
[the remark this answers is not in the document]
pxaws: Good suggestion. I will merge it to the loop.

plugins/processors/awsapplicationsignals/internal/resolver/endpointslicewatcher.go
```diff
@@ -0,0 +1,297 @@
...
+// in sliceToKeysMap so that we can remove them later upon deletion.
+func (w *endpointSliceWatcher) handleSliceAdd(obj interface{}) {
+	newSlice := obj.(*discv1.EndpointSlice)
+	sliceUID := string(newSlice.UID)
+
+	// Compute all key-value pairs for this new slice
+	pairs := w.extractEndpointSliceKeyValuePairs(newSlice)
+
+	// Insert them into our ipToWorkload / serviceToWorkload, and track the keys.
+	var keys []string
```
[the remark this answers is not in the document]
pxaws: Will update in next commit.

plugins/processors/awsapplicationsignals/internal/resolver/endpointslicewatcher.go
lisguo: are we saying that we will resolve RemoteService to the podname in the worst case? Wondering if this is the user experience we want in the trace map since it has high cardinality? 

wondering if we can do something like remove the uid...not sure what the edge cases are.

plugins/processors/awsapplicationsignals/internal/resolver/kubernetes.go
```diff
@@ -78,47 +91,89 @@ func getKubernetesResolver(platformCode, clusterName string, logger *zap.Logger)
...
-		err = podInformer.SetTransform(minimizePod)
-		if err != nil {
-			logger.Error("failed to minimize Pod objects", zap.Error(err))
-		}
-		serviceInformer := sharedInformerFactory.Core().V1().Services().Informer()
-		err = serviceInformer.SetTransform(minimizeService)
-		if err != nil {
-			logger.Error("failed to minimize Service objects", zap.Error(err))
-		}
+		useListPod := (os.Getenv("USE_LIST_POD") == "true")
```
lisguo, line 94: why use an env variable instead of a config option in the processor?

plugins/processors/awsapplicationsignals/internal/resolver/kubernetes.go
```diff
@@ -78,47 +91,89 @@ func getKubernetesResolver(platformCode, clusterName string, logger *zap.Logger)
...
+			podInformer := sharedInformerFactory.Core().V1().Pods().Informer()
+			err = podInformer.SetTransform(minimizePod)
+			if err != nil {
+				logger.Error("failed to minimize Pod objects", zap.Error(err))
+			}
+			serviceInformer := sharedInformerFactory.Core().V1().Services().Informer()
+			err = serviceInformer.SetTransform(minimizeService)
+			if err != nil {
+				logger.Error("failed to minimize Service objects", zap.Error(err))
+			}
```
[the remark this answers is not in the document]
pxaws: I think it's better to have the pod/service watcher to hide the details of transform inside. For consistency, I will change them accordingly.

plugins/processors/awsapplicationsignals/internal/resolver/kubernetes.go
pxaws: We want to remove the usage of list pods api eventually. So I don't want to make the config change which is more permanent. In the worst case, the old logic is kept here so that customers can quickly switch to it. But I don't want too much publicity for the environmental variable for now.

pxaws: >are we saying that we will resolve RemoteService to the podname in the worst case? Wondering if this is the user experience we want in the trace map since it has high cardinality?

We have the metric limiter that will mitigate the high cardinality for metrics. For traces, the high cardinality issue is handled at backend. I remember that the backend will check how many nodes in the trace and cut off if there are too many. From this perspective, I am not very worried about the worst case. Anyway, the chance for the worst case to happen is pretty low in my opinion. 

>wondering if we can do something like remove the uid...not sure what the edge cases are.

I am not sure what you mean by "remove the uid".

lisguo: > wondering if we can do something like remove the uid...not sure what the edge cases are.

I was wondering if we should cut the auto generated id in the pod name like `cloudwatch-agent-m8ncn` -> `cloudwatch-agent`

pxaws: yes, we do that using the regex pattern `<something>-<5charSuffix>`. I think the edge case is quite unusual. The 5 char uid generated exclude vowels and 0, 1, 3: `bcdfghjklmnpqrstvwxz2456789`. It's unlikely to match the names given manually by users. 

```golang
// InferWorkloadName tries to parse the given podName to find the top-level workload name.
//
// 1) If it matches <statefulset>-<ordinal>, return <statefulset>.
// 2) If it matches <something>-<5charSuffix>:
//   - If <something> is <deployment>-<6–10charSuffix>, return <deployment>.
//   - Else return <something> (likely a bare ReplicaSet or DaemonSet).
//
// 3) If no pattern matches, return the original podName.
//
// Caveat: You can't reliably distinguish DaemonSet vs. bare ReplicaSet by name alone.
func inferWorkloadName(podName string) string {
```

lisguo: Okay makes sense. Going back to the ENV variable a nit would be to use an appsignals specific name, maybe `APP_SIGNALS_USE_LIST_POD` or something

pxaws: Good suggestion for renaming. Will push another commit.

plugins/processors/awsapplicationsignals/internal/resolver/kubernetes.go
```diff
@@ -78,47 +91,74 @@ func getKubernetesResolver(platformCode, clusterName string, logger *zap.Logger)
...
-		err = podInformer.SetTransform(minimizePod)
-		if err != nil {
-			logger.Error("failed to minimize Pod objects", zap.Error(err))
-		}
-		serviceInformer := sharedInformerFactory.Core().V1().Services().Informer()
-		err = serviceInformer.SetTransform(minimizeService)
-		if err != nil {
-			logger.Error("failed to minimize Service objects", zap.Error(err))
-		}
+		useListPod := (os.Getenv("APP_SIGNALS_USE_LIST_POD") == "true")
```
jefchien, line 94: nit: Consider making this a const with a comment for its purpose and specifying that it's temporary. This seems like something that would be a good candidate for a feature gate.

jefchien (dismissed):

plugins/processors/awsapplicationsignals/internal/resolver/endpointslicewatcher.go
```diff
@@ -0,0 +1,290 @@
...
+						})
+					}
+				}
+			}
+
+			// Build service name -> "workload@namespace" pair from the first pod
+			if isFirstPod {
+				svcName := slice.Labels["kubernetes.io/service-name"]
+				if svcName != "" {
+					isFirstPod = false
```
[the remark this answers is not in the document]
pxaws: Make sense. Will change it.

plugins/processors/awsapplicationsignals/internal/resolver/endpointslicewatcher.go
```diff
@@ -0,0 +1,290 @@
...
+	// - eps.UID (from metadata)
+	// - eps.Endpoints: for each endpoint, its Addresses and TargetRef.
+	// - eps.Ports: each port's Port (and optionally Name/Protocol)
+	//
+	// For each endpoint, clear fields that we don’t use.
+	for i := range eps.Endpoints {
+		// We only need Addresses and TargetRef. Hostname, NodeName, and Zone are not used.
+		eps.Endpoints[i].Hostname = nil
+		eps.Endpoints[i].NodeName = nil
+		eps.Endpoints[i].Zone = nil
```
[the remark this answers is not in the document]
pxaws: Let me remove them (though there is not much saving by doing that)

plugins/processors/awsapplicationsignals/internal/resolver/kubernetes.go
pxaws: I will change it accordingly.

plugins/processors/awsapplicationsignals/internal/resolver/endpointslicewatcher.go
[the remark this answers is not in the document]
pxaws: Will keep Conditions anyway. I don't want to complicate the logic. I can't assign nil to it `eps.Endpoints[i].Conditions = nil`. An empty EndpointConditions struct is needed here.

jefchien (approved):

plugins/processors/awsapplicationsignals/internal/resolver/endpointslicewatcher.go
```diff
@@ -0,0 +1,292 @@
...
+		DeleteFunc: func(obj interface{}) {
+			w.handleSliceDelete(obj)
+		},
+	})
+	go w.informer.Run(stopCh)
+}
+
+func (w *endpointSliceWatcher) waitForCacheSync(stopCh chan struct{}) {
+	if !cache.WaitForNamedCacheSync("endpointSliceWatcher", stopCh, w.informer.HasSynced) {
+		w.logger.Fatal("timed out waiting for endpointSliceWatcher cache to sync")
```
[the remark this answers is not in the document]
pxaws: yes. If this failed, it's meaningless to continue. Either there is some permission issue or some time out/throttling, better to kill the agent and let it restart in this case. Metrics/traces with very low qualities will be generated if we continue.

mxiamxia (approved):

lisguo (approved):

PR #1545, opened 2025-02-11
skip PutRetentionPolicy by checking existing policy 

movence:
# Description of the issue
The agent processes PutRetentionPolicy on monitored log groups synchronously every time it restarts, which can lead to API throttling in large deployments. This throttling may cause delays in pushing logs while it is active.

# Description of changes
- Add a new goroutine in TargetManager to process PRP using a retention processing channel
- Make the retention update channel as blocking channel so no updates get dropped when it's full
- Check existing retention policy of a target using `DecribeLogGroups` then skip PRP is there is no change
- Add backoff and jittering which don't exceed max wait time of 30sec

#### Rev1
- Use `rand` when calculating jitter
- Skip DLG for newly created log groups
- Add unit tests to cover this new path
#### Rev2
- Separate channels for DLG and PRP and process them separately
- Use `seededRand` which already exists in the package for jitter
- rebase with upstream
#### Rev3
- Add missing condition in `processDescribeLogGroup`

# License

# Tests
Configure the agent to monitor 120 log groups (20 more than the default channel size). The expectation for the agent is to update retention policies on all log group for the first time, but they are skipped in subsequent agent restarts. 
```
# running the agent for the first time
[logs]# grep -irn "updating retention policy for target" amazon-cloudwatch-agent.log 
6671:2025-02-11T17:32:38Z D! [outputs.cloudwatchlogs] updating retention policy for target {DLG-Test-0 ip-[ip].us-west-2.compute.internal  30}: <nil>
6672:2025-02-11T17:32:38Z D! [outputs.cloudwatchlogs] updating retention policy for target {DLG-Test-1 ip-[ip].us-west-2.compute.internal  30}: <nil>
... (trimmed)

[logs]# grep -irn "updating retention policy for target" amazon-cloudwatch-agent.log | wc -l
120

# delete agent log file then restart the agent
[logs]# grep -irn "updating retention policy for target" amazon-cloudwatch-agent.log | wc -l
0
```

# Requirements
2. Run `make lint`

### DIFF
--- a/plugins/outputs/cloudwatchlogs/internal/pusher/target.go
+++ b/plugins/outputs/cloudwatchlogs/internal/pusher/target.go
@@ -5,7 +5,6 @@ package pusher
 
 import (
 	"fmt"
-	"math/rand"
 	"sync"
 	"time"
 
@@ -209,11 +208,14 @@ func (m *targetManager) updateRetentionPolicy(target Target) error {
 }
 
 func (m *targetManager) calculateBackoff(attempt int) time.Duration {
-	delay := baseDelay * time.Duration(1<<uint(attempt))
+	delay := baseDelay
+	for i := 0; i < attempt && delay < maxDelay; i++ {
+		delay *= 2
+	}
 	if delay > maxDelay {
 		delay = maxDelay
 	}
 
-	jitter := time.Duration(rand.Float64() * float64(delay) * jitterFactor)
+	jitter := time.Duration(float64(delay) * (float64(time.Now().UnixNano()%100) / 500.0 * jitterFactor))
 	return delay + jitter
 }
--- a/plugins/outputs/cloudwatchlogs/internal/pusher/target.go
+++ b/plugins/outputs/cloudwatchlogs/internal/pusher/target.go
@@ -103,7 +103,7 @@ func (m *targetManager) createLogGroupAndStream(t Target) (bool, error) {
 
 		// attempt to create stream again if group created successfully.
 		if err == nil {
-			m.logger.Debugf("successfully created log group %v. Retrying log stream %v", t.Group, t.Stream)
+			m.logger.Debugf("retrying log stream %v", t.Stream)
 			err = m.createLogStream(t)
 		} else {
 			m.logger.Debugf("creating group fail due to : %v", err)
@@ -164,7 +164,6 @@ func (m *targetManager) processDescribeLogGroup() {
 			if currentRetention != target.Retention && target.Retention > 0 {
 				m.logger.Debugf("queueing log group %v to update retention policy", target.Group)
 				m.prp <- target
-				break
 			}
 			break // no change in retention
 		}
--- a/plugins/outputs/cloudwatchlogs/internal/pusher/target.go
+++ b/plugins/outputs/cloudwatchlogs/internal/pusher/target.go
@@ -8,10 +8,11 @@ import (
 	"sync"
 	"time"
 
-	"github.com/aws/amazon-cloudwatch-agent/sdk/service/cloudwatchlogs"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/influxdata/telegraf"
+
+	"github.com/aws/amazon-cloudwatch-agent/sdk/service/cloudwatchlogs"
 )
 
 const (
--- a/plugins/outputs/cloudwatchlogs/internal/pusher/target.go
+++ b/plugins/outputs/cloudwatchlogs/internal/pusher/target.go
@@ -166,6 +166,7 @@ func (m *targetManager) processDescribeLogGroup() {
 				m.prp <- target
 				break
 			}
+			break // no change in retention
 		}
 	}
 }
--- a/plugins/outputs/cloudwatchlogs/internal/pusher/target.go
+++ b/plugins/outputs/cloudwatchlogs/internal/pusher/target.go
@@ -5,25 +5,22 @@ package pusher
 
 import (
 	"fmt"
-	"math/rand"
 	"sync"
 	"time"
 
+	"github.com/aws/amazon-cloudwatch-agent/sdk/service/cloudwatchlogs"
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/influxdata/telegraf"
-
-	"github.com/aws/amazon-cloudwatch-agent/sdk/service/cloudwatchlogs"
 )
 
 const (
 	retentionChannelSize = 100
 	// max wait time with backoff and jittering:
 	// 0 + 2.4 + 4.8 + 9.6 + 10 ~= 26.8 sec
-	baseDelay    = 1 * time.Second  // Increased to 1s
-	maxDelay     = 10 * time.Second // Keep at 10s
-	maxAttempts  = 5                // Keep at 5
-	jitterFactor = 0.2
+	baseRetryDelay      = 1 * time.Second
+	maxRetryDelayTarget = 10 * time.Second
+	numBackoffRetries   = 5
 )
 
 type Target struct {
@@ -40,20 +37,23 @@ type targetManager struct {
 	logger  telegraf.Logger
 	service cloudWatchLogsService
 	// cache of initialized targets
-	cache                  map[Target]struct{}
-	mu                     sync.Mutex
-	retentionPolicyUpdates chan Target
+	cache map[Target]struct{}
+	mu    sync.Mutex
+	dlg   chan Target
+	prp   chan Target
 }
 
 func NewTargetManager(logger telegraf.Logger, service cloudWatchLogsService) TargetManager {
 	tm := &targetManager{
-		logger:                 logger,
-		service:                service,
-		cache:                  make(map[Target]struct{}),
-		retentionPolicyUpdates: make(chan Target, retentionChannelSize),
+		logger:  logger,
+		service: service,
+		cache:   make(map[Target]struct{}),
+		dlg:     make(chan Target, retentionChannelSize),
+		prp:     make(chan Target, retentionChannelSize),
 	}
 
-	go tm.processRetentionUpdates()
+	go tm.processDescribeLogGroup()
+	go tm.processPutRetentionPolicy()
 	return tm
 }
 
@@ -68,11 +68,11 @@ func (m *targetManager) InitTarget(target Target) error {
 		}
 		if target.Retention > 0 {
 			if newGroup {
-				if err := m.processWithRetry(target, true); err != nil {
-					m.logger.Errorf("Failed to set initial retention policy for target %v: %v", target, err)
-				}
+				m.logger.Debugf("sending new log group %v to prp channel", target.Group)
+				m.prp <- target
 			} else {
-				m.PutRetentionPolicy(target)
+				m.logger.Debugf("sending existing log group %v to dlg channel", target.Group)
+				m.dlg <- target
 			}
 		}
 		m.cache[target] = struct{}{}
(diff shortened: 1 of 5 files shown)


plugins/outputs/cloudwatchlogs/internal/pusher/target.go
```diff
@@ -13,6 +15,16 @@ import (
 	"github.com/aws/amazon-cloudwatch-agent/sdk/service/cloudwatchlogs"
 )
 
+const (
+	retentionChannel = 100
```
lisguo, line 19: nit
```suggestion
	retentionChannelSize = 100
```

lisguo: How did we determine the size? I guess we shouldn't be seeing 100 log groups that have a retention policy?

movence: Yeah, there is no science behind this value. We can even go lower than 100 since it's blocked channel which won't drop a request when it's full anyways. 100 just seemed reasonable size that would cover general use cases.

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
```diff
@@ -54,6 +71,13 @@ func (m *targetManager) InitTarget(target Target) error {
 	return nil
 }
 
+func (m *targetManager) PutRetentionPolicy(target Target) {
+	if target.Retention > 0 {
```
jefchien, line 75: Just a note: We have deduping of log groups/retention policy as part of
https://github.com/aws/amazon-cloudwatch-agent/blob/799f0c3bdcfc18fcadf3b5e0c1c9ebaa543e18a8/logs/logs.go#L169-L178
where it sets duplicates to -1 and I believe there's validation to ensure that all retention values are the same for a log group in the configuration.

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
```diff
@@ -109,26 +133,89 @@ func (m *targetManager) createLogStream(t Target) error {
...
+
+		m.logger.Debugf("retrying to update retention policy for target (%v) %v: %v", attempt, target, err)
+		delay := m.calculateBackoff(attempt)
+		time.Sleep(delay)
+	}
+	return fmt.Errorf("failed to update retention policy after %d attempts: %w", maxAttempts, err)
+}
+
+func (m *targetManager) checkAndUpdate(target Target) error {
+	currentRetention, err := m.describeLogGroupRetention(target)
```
jefchien, line 161: Can likely skip this API call when it's a new log group.
https://github.com/aws/amazon-cloudwatch-agent/blob/799f0c3bdcfc18fcadf3b5e0c1c9ebaa543e18a8/plugins/outputs/cloudwatchlogs/internal/pusher/target.go#L43-L55

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
```diff
@@ -109,26 +133,89 @@ func (m *targetManager) createLogStream(t Target) error {
...
-				m.logger.Errorf("Unable to put retention policy for log group %v: %v ", t.Group, err)
+			m.logger.Errorf("Failed to update retention policy for target %v: %v", target, err)
+		}
+	}
+}
+
+func (m *targetManager) processWithRetry(target Target) error {
+	var err error
+	for attempt := 0; attempt < maxAttempts; attempt++ {
+		err = m.checkAndUpdate(target)
```
jefchien, line 148: If we return an error because the log group doesn't exist, do we still want to retry? I don't think we need to.

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
```diff
@@ -109,26 +133,89 @@ func (m *targetManager) createLogStream(t Target) error {
...
+		return 0, fmt.Errorf("describe log groups failed: %w", err)
+	}
+
+	for _, group := range output.LogGroups {
+		if *group.LogGroupName == target.Group {
+			if group.RetentionInDays == nil {
+				return 0, nil
 			}
-		} else {
-			m.logger.Debugf("successfully updated log retention policy for log group %v", t.Group)
```
jefchien, line 131: nit: Can we keep this debug log line?

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
```diff
@@ -109,26 +133,89 @@ func (m *targetManager) createLogStream(t Target) error {
...
+}
+
+func (m *targetManager) describeLogGroupRetention(target Target) (int, error) {
+	input := &cloudwatchlogs.DescribeLogGroupsInput{
+		LogGroupNamePrefix: aws.String(target.Group),
+	}
+
+	output, err := m.service.DescribeLogGroups(input)
+	if err != nil {
+		return 0, fmt.Errorf("describe log groups failed: %w", err)
```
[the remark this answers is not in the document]
movence: 0 is not valid for PRP, but we can use it to compare with configured retention. If no retention is configured (=0), then no update.

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
```diff
@@ -109,26 +133,89 @@ func (m *targetManager) createLogStream(t Target) error {
...
+func (m *targetManager) calculateBackoff(attempt int) time.Duration {
+	delay := baseDelay
+	for i := 0; i < attempt && delay < maxDelay; i++ {
+		delay *= 2
+	}
+	if delay > maxDelay {
+		delay = maxDelay
+	}
+
+	jitter := time.Duration(float64(delay) * (float64(time.Now().UnixNano()%100) / 500.0 * jitterFactor))
```
jefchien, line 219: nit: Would expect jitter to have a degree of randomness. https://aws.amazon.com/builders-library/timeouts-retries-and-backoff-with-jitter/

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
movence: the error can also include throttling event or log group still being created (not even sure if this is possible)

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
jefchien: I think there's a slim chance that someone has 100 distinct log groups in their configuration that each have a retention policy, so we can likely shrink this. If it's full, it'll potentially block startup and sends, so it'd be better if it didn't get full.

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
```diff
@@ -135,17 +146,17 @@ func (m *targetManager) createLogStream(t Target) error {
...
 		}
 	}
 }
 
-func (m *targetManager) processWithRetry(target Target) error {
+func (m *targetManager) processWithRetry(target Target, newGroup bool) error {
 	var err error
 	for attempt := 0; attempt < maxAttempts; attempt++ {
-		err = m.checkAndUpdate(target)
+		err = m.checkAndUpdate(target, newGroup)
```
jefchien, line 159: nit: Would expect to not call `checkAndUpdate` if `newGroup` is true.
```suggestion
        if newGroup {
          err = m.updateRetentionPolicy(target)
        } else {
		  err = m.checkAndUpdate(target)
		}
```

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
```diff
@@ -61,11 +62,19 @@ func (m *targetManager) InitTarget(target Target) error {
...
 	if _, ok := m.cache[target]; !ok {
-		err := m.createLogGroupAndStream(target)
+		newGroup, err := m.createLogGroupAndStream(target)
 		if err != nil {
 			return err
 		}
-		m.PutRetentionPolicy(target)
+		if target.Retention > 0 {
+			if newGroup {
+				if err := m.processWithRetry(target, true); err != nil {
```
jefchien, line 71: This makes one of the branches a blocking call and other not. Would prefer if we could offload it to the other thread and put it in the channel. No need to block sends or reads on whether the put retention policy has been set yet.

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
```diff
@@ -216,6 +233,8 @@ func (m *targetManager) calculateBackoff(attempt int) time.Duration {
 		delay = maxDelay
 	}
 
-	jitter := time.Duration(float64(delay) * (float64(time.Now().UnixNano()%100) / 500.0 * jitterFactor))
+	//nolint:gosec // using math/rand is acceptable for jitter calculation
+	r := rand.New(rand.NewSource(time.Now().UnixNano()))
```
jefchien, line 237: Can reuse random generator from the retry.
https://github.com/aws/amazon-cloudwatch-agent/blob/799f0c3bdcfc18fcadf3b5e0c1c9ebaa543e18a8/plugins/outputs/cloudwatchlogs/internal/pusher/retry.go#L38

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
movence: `checkAndUpdate` will handle that

plugins/outputs/cloudwatchlogs/internal/pusher/target_test.go
```diff
@@ -70,42 +71,83 @@ func TestTargetManager(t *testing.T) {
...
-			Return(&cloudwatchlogs.PutRetentionPolicyOutput{}, &cloudwatchlogs.ResourceNotFoundException{}).Once()
+		mockService.On("DescribeLogGroups", mock.Anything).
+			Return(&cloudwatchlogs.DescribeLogGroupsOutput{}, &cloudwatchlogs.ResourceNotFoundException{}).Times(numBackoffRetries)
 
 		manager := NewTargetManager(logger, mockService)
 		err := manager.InitTarget(target)
-
-		assert.NoError(t, err) // The overall operation should still succeed even if setting retention policy fails
+		assert.NoError(t, err)
+		time.Sleep(30 * time.Second)
```
jefchien, line 125: Can we change the number of retries during the testing to make these shorter? Our unit tests already take a long time and I'd rather we didn't increase them by 1m30s with all these failure case checks. Can set numRetries as a field on the target manager that defaults to 5 and then overwrite that in the test.
```go
manager.numRetries = 2
```

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
```diff
@@ -109,26 +151,89 @@ func (m *targetManager) createLogStream(t Target) error {
...
+			if err != nil {
+				m.logger.Errorf("failed to describe log group retention for target %v: %v", target, err)
+				time.Sleep(m.calculateBackoff(attempt))
+				continue
+			}
+
+			if currentRetention != target.Retention && target.Retention > 0 {
+				m.logger.Debugf("queueing log group %v to update retention policy", target.Group)
+				m.prp <- target
+				break
```
jefchien, line 167: nit: Don't need the break here if we're immediately breaking right after.

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
```diff
@@ -74,24 +111,29 @@ func (m *targetManager) createLogGroupAndStream(t Target) error {
...
 	} else {
-		_, err = m.service.CreateLogGroup(&cloudwatchlogs.CreateLogGroupInput{
+		input = &cloudwatchlogs.CreateLogGroupInput{
 			LogGroupName: &t.Group,
-		})
+		}
+	}
+	_, err := m.service.CreateLogGroup(input)
+	if err == nil {
+		m.logger.Debugf("successfully created log group %v", t.Group)
```
jefchien, line 135: nit: This is already logged in `createLogGroupAndStream`.

plugins/outputs/cloudwatchlogs/internal/pusher/target_test.go
movence: running long tests in parallel now

jefchien (approved):

lisguo (approved):

PR #1564, opened 2025-02-22
Unblock China region integration tests

varunch77:
# Description of the issue
The EC2LinuxCN tests are not able to start up properly due to errors while fetching dependencies. This creates a gap in our testing for this region.

# Description of changes
This change adds a new step to our integration test workflow that zips and uploads the test repo and the necessary dependencies (using `go mod vendor`) to an S3 bucket in the China region. Our tests then pull from this S3 bucket instead of attempting to download them directly for each test. 

This PR is in conjunction with: https://github.com/aws/amazon-cloudwatch-agent-test/pull/469

# License
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice. ✅ 

# Tests
Integration test run: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/13595857669

# Requirements
2. Run `make lint` ✅


.github/workflows/upload-dependencies-cn.yml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: ~1.22.2
+
+      - name: Upload Dependencies to S3
+        run: |
+          go mod tidy
+          go mod vendor
+          aws s3 sync ./vendor s3://${{ inputs.s3_integration_bucket }}/integration-test/vendor/ --delete --quiet
```
dricross, line 58: The s3 path should probably be github sha specific so that tests don't interfere with each other

varunch77:
new integration test run with SHA changes: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/13507749182/job/37741584109

.github/workflows/upload-dependencies-cn.yml
varunch77: Nice catch, updated the code to include this

.github/workflows/upload-dependencies-cn.yml
```diff
@@ -0,0 +1,58 @@
...
+    inputs:
+      region:
+        type: string
+      test_repo_name:
+        required: true
+        type: string
+      test_repo_branch:
+        required: true
+        type: string
+      test_repo_url:
```
musa-asad, line 21: Do we need this input?

.github/workflows/integration-test.yml
```diff
@@ -13,7 +13,7 @@ env:
   ECR_INTEGRATION_TEST_REPO: "cwagent-integration-test"
   CWA_GITHUB_TEST_REPO_NAME: "aws/amazon-cloudwatch-agent-test"
   CWA_GITHUB_TEST_REPO_URL: "https://github.com/aws/amazon-cloudwatch-agent-test.git"
-  CWA_GITHUB_TEST_REPO_BRANCH: "main"
+  CWA_GITHUB_TEST_REPO_BRANCH: "integ-test-CN"
```
musa-asad, line 16: This should be changed back to `main` when ready.

.github/workflows/upload-dependencies-cn.yml
```diff
@@ -0,0 +1,58 @@
...
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: ~1.22.2
+
+      - name: Upload Dependencies to S3
+        run: |
+          go mod tidy
+          go mod vendor
+          aws s3 sync ./vendor s3://${{ inputs.s3_integration_bucket }}/integration-test/vendor/${{ inputs.github_sha }} --delete --quiet
```
musa-asad, line 58: Would it make sense to cache the vendor directory with `actions/cache` so we don't have to pull from S3 each time?

musa-asad (dismissed):

.github/workflows/upload-dependencies-cn.yml
jefchien, line 58: Would recommend zipping the vendor directory so it's easier to manage.

jefchien: I think it'd make sense to cache it, but we'd need to upload it each time if we're relying on the github SHA as a key.

.github/workflows/upload-dependencies-cn.yml
jefchien, line 1: nit: The naming of this file makes it seem China region specific, but it seems generic enough to be used anywhere.

.github/workflows/upload-dependencies-cn.yml
musa-asad: Oh right, good point.

.github/workflows/upload-dependencies-cn.yml
varunch77: Good catch, I've removed it

.github/workflows/upload-dependencies-cn.yml
varunch77: Done

.github/workflows/upload-dependencies-cn.yml
varunch77: Yup, I've renamed so it can be reused for other regions in the future if needed

varunch77:
Integration test run: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/13594721082

.github/workflows/upload-dependencies.yml
```diff
@@ -0,0 +1,62 @@
...
+          mkdir test-repo
+          tar -czf ./test-repo/amazon-cloudwatch-agent-test.tar.gz --exclude='test-repo' .
+          aws s3 cp ./test-repo/amazon-cloudwatch-agent-test.tar.gz s3://${{ inputs.s3_integration_bucket }}/integration-test/cloudwatch-agent-test-repo/${{ inputs.github_sha }}.tar.gz --quiet
+
+      - name: Upload Dependencies to S3
+        run: |
+          go mod tidy
+          go mod vendor
+          tar -czf vendor.tar.gz ./vendor
+          aws s3 cp ./vendor.tar.gz s3://${{ inputs.s3_integration_bucket }}/integration-test/vendor/${{ inputs.github_sha }}.tar.gz --quiet
```
jefchien, line 62: Is there a reason we're separating this into two buckets? Couldn't we zip the entire repo after we've done the go mod vendor? It's not like they will ever be used separately.

varunch77: Yeah this can be combined, I've made that change

varunch77:
New integration test run: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/13595857669/job/38012754850

jefchien (approved):

.github/workflows/upload-dependencies.yml
```diff
@@ -0,0 +1,57 @@
...
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: ~1.22.2
+
+      - name: Upload Dependencies and Test Repo
+        run: |
+          go mod tidy
+          go mod vendor
+          mkdir test-repo
```
jefchien, line 55: nit: What is the point of this?

musa-asad (approved):

.github/workflows/upload-dependencies.yml
varunch77: I wanted to avoid any permissions issues with regards to writing to the parent directory (for storing the zipped file). So when I zipped the checked-out git repo (the current directory), I created `test-repo` to store the zipped file and told the tar command to ignore it to avoid concurrent write/zip issues.

PR #1577, opened 2025-03-03
add fixed size buffer to handle fd in log tailer with feature flag

movence:
# Description of the issue
The current CWA strategy for log collection prioritizes log integrity without consideration for disk space. CWA holds onto open file descriptors until all the log lines have been successfully pushed to CloudWatch Logs. If another process deletes the file, the file descriptor held by the agent will be unaffected. If CWA is unable to publish the logs, it will retry for 14 days. For some customers, this is the expectation and preferred behavior particularly for sensitive files such as security logs.

On Unix, when a file is removed by a process, the file name is unlinked from the underlying file instead of deleted. If there’s an open reference to the file via file descriptor, the file will not be deleted. 

# Description of changes
- Add buffer-based (size of 1) log handling in tailer for file rotation
- Add new runSender goroutine. tailer now sends single event at a time to channel that the sender reads from
- Enable this new path behind feature flag `handle_rotation` 
- Add feature toggle flag that is read either from agent json config or env var with the value from agent config takes precedence. This is to keep the original behavior with `auto_removal` which reads all contents from a rotate log file before deleting them. 
- Feature flag value is then added to `env-config.json` by the translator so it can be added as env var of systemd process of the agent at start up time

# License

# Tests

# Requirements
2. Run `make lint`


plugins/inputs/logfile/tailersrc.go
```diff
@@ -310,6 +326,7 @@ func (ts *tailerSrc) cleanUp() {
 
 	if ts.outputFn != nil {
 		ts.outputFn(nil) // inform logs agent the tailer src's exit, to stop runSrcToDest
+		ts.outputFn = nil
```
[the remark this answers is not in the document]
movence: this is to prevent any unseen race condition due to added go routines trying to call `outputFn`. mutex has been added around this to double protect as well.

translator/tocwconfig/toenvconfig/toEnvConfig_test.go
```diff
@@ -0,0 +1,281 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			// Setup environment variables
+			for k, v := range tt.envVars {
+				os.Setenv(k, v)
+			}
+			defer func() {
+				for k := range tt.envVars {
+					os.Unsetenv(k)
+				}
+			}()
```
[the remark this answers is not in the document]
movence: We might also have leak in our tests by not properly closing files.

plugins/inputs/logfile/tail/tail.go
```diff
@@ -210,6 +212,15 @@ func (tail *Tail) reopen() error {
 		break
 	}
 	OpenFileCount.Add(1)
+
+	tail.openReader()
+	if !resetOffset && tail.curOffset > 0 {
+		err := tail.seekTo(SeekInfo{Offset: tail.curOffset, Whence: io.SeekStart})
+		if err != nil {
+			return fmt.Errorf("unable to restore offset on reopen: %w", err)
```
lisguo, line 220: if we are unable to restore the offset, it kills the tailer in `tailFileSync`. Does that end up just skipping the file and moving onto the next one if available? Or does it just die and not read any other files?

plugins/inputs/logfile/tailersrc.go
```diff
@@ -273,24 +266,102 @@ func (ts *tailerSrc) runTail() {
...
+						select {
+						case <-time.After(closeReopenInterval):
+						case <-ts.done:
+							return
+						}
+					}
+				}
 			}
+		} else {
+			ts.outputFn(e)
```
lisguo, line 327: does this need to be protected by a mutex? is publishEvent concurrent?

translator/tocwconfig/toenvconfig/toEnvConfig.go
```diff
@@ -63,6 +66,20 @@ func ToEnvConfig(jsonConfigValue map[string]interface{}) []byte {
 		envVars[envconfig.AWS_CA_BUNDLE] = sslConfig[commonconfig.CABundlePath]
 	}
 
+	backpressureDrop := false // default value
```
lisguo, line 69: nit: can we move this to a util func so we can just call `util.GetBackpressureDrop(jsonConfigValue)`

plugins/inputs/logfile/tailersrc.go
movence: It is not. This will pretty much get blocked if downstream buffer is full (eg publisher). Keeping the original behavior as is.

plugins/inputs/logfile/logfile.go
```diff
@@ -159,16 +165,19 @@ func (t *LogFile) FindLogSrc() []logs.LogSrc {
...
+		t.Log.Debugf("Looking for files matching %s", fileconfig.FilePath)
+		t.Log.Infof("D! [logfile] Searching for files matching pattern: %s", fileconfig.FilePath)
 		//Add file -> {serviceName,  deploymentEnvironment} mapping to entity store
 		if es != nil {
 			es.AddServiceAttrEntryForLogFile(entitystore.LogFileGlob(fileconfig.FilePath), fileconfig.ServiceName, fileconfig.Environment)
 		}
 
 		targetFiles, err := t.getTargetFiles(fileconfig)
 		if err != nil {
+			t.Log.Errorf("Error getting target files: %v", err)
```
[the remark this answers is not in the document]
movence: Yeah, removed some logging to not clutter and stay close to before changes.

plugins/inputs/logfile/tailersrc.go
```diff
@@ -273,24 +266,102 @@ func (ts *tailerSrc) runTail() {
...
+		if ts.backpressureDrop {
+			select {
+			case ts.buffer <- e:
+				// successfully sent
+			case <-ts.done:
+				return
+			default:
+				// Buffer is full, close file and try again
+				log.Printf("D! [logfile] tailer sender buffer is full, closing file %v", ts.tailer.Filename)
+				ts.tailer.CloseFile()
```
jefchien, line 303: There should be a delay between the buffer being detected as full and the file getting closed. Latency for a PLE request can vary depending on setup, so not having a delay could cause the agent to constantly close/reopen the file.

plugins/inputs/logfile/tailersrc.go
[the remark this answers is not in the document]
jefchien: I think there has to be a better way to do this without setting the outputFn to nil. The race condition is in the outputFn itself right? Wouldn't it be better to fix it there?

translator/tocwconfig/toenvconfig/toEnvConfig.go
jefchien: I'm leaning further towards having this be part of the logs config translator.

plugins/inputs/logfile/tailersrc.go
```diff
@@ -273,24 +266,102 @@ func (ts *tailerSrc) runTail() {
...
-			timestamp, modifiedMsg := ts.timestampFn(msg)
-			e := &LogEvent{
-				msg:    modifiedMsg,
-				t:      timestamp,
-				offset: *fo,
-				src:    ts,
+func (ts *tailerSrc) runSender() {
+	log.Printf("D! [logfile] runSender starting for %s", ts.tailer.Filename)
+	defer func() {
+		ts.bufferCloseOnce.Do(func() {
```
jefchien, line 335: Why are we closing the buffer in a defer?

plugins/inputs/logfile/fileconfig.go
```diff
@@ -60,6 +60,9 @@ type FileConfig struct {
 	// automatically remove the file / symlink after uploading.
 	// This auto removal does not support the case where other log rotation mechanism is already in place.
 	AutoRemoval bool `toml:"auto_removal"`
+	// close current file if processing buffer is full/blocked then reopen when the buffer frees up
+	// AutoRemoval take higher priority than BackpressureDrop
+	BackpressureMode string `toml:"backpressure_mode"`
```
jefchien, line 65: Description of field is outdated.

logs/logs.go
```diff
@@ -144,9 +145,17 @@ func (l *LogAgent) runSrcToDest(src LogSrc, dest LogDest) {
 	eventsCh := make(chan LogEvent)
 	defer src.Stop()
 
+	var mu sync.Mutex
+	closed := false
 	src.SetOutput(func(e LogEvent) {
+		mu.Lock()
```
jefchien, line 151: Having to acquire a lock for every single log event publish seems excessive.

internal/logscommon/const.go
```diff
@@ -26,4 +26,8 @@ const (
 
 	WindowsEventLogPrefix = "Amazon_CloudWatch_WindowsEventLog_"
 	LogType               = "log_type"
+
+	LogBackpressureModeEnvKey = "CWAGENT_LOGS_BACKPRESSURE_MODE"
+	LogBackpressureModeKey    = "backpressure_mode"
+	LogBackpressureModeDrop   = "fd_drop"
```
jefchien, line 32: Can we make this an enum?

internal/logscommon/const.go
```diff
@@ -26,4 +26,8 @@ const (
 
 	WindowsEventLogPrefix = "Amazon_CloudWatch_WindowsEventLog_"
 	LogType               = "log_type"
+
+	LogBackpressureModeEnvKey = "CWAGENT_LOGS_BACKPRESSURE_MODE"
```
jefchien, line 30: nit: All environment variable names should be declared in `envconfig`.

plugins/inputs/logfile/tailersrc.go
```diff
@@ -273,24 +264,86 @@ func (ts *tailerSrc) runTail() {
...
+			default:
+				// sender buffer is full. start timer to close file then retry
+				timer := time.NewTimer(tailCloseInterval)
+				defer timer.Stop()
+
+				for {
+					select {
+					case ts.buffer <- e:
+						// sent event after buffer gets freed up
+						if err := ts.tailer.Reopen(false); err != nil {
```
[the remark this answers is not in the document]
jefchien: What happens if a file is closed and deleted before it tries to reopen the file?

plugins/inputs/logfile/tailersrc.go
```diff
@@ -163,7 +179,12 @@ func (ts *tailerSrc) Done(offset fileOffset) {
 }
 
 func (ts *tailerSrc) Stop() {
-	close(ts.done)
+	ts.stopOnce.Do(func() {
```
lisguo, line 182: do we need this? did you find that in your testing we had multiple `Stop()` invocations?

plugins/inputs/logfile/tailersrc_test.go
```diff
@@ -464,6 +472,90 @@ func resetState(originWaitDuration time.Duration) {
 }
 
 func teardown(resources tailerTestResources) {
+	beforeCount := tail.OpenFileCount.Load()
+	log.Printf("teardown start - OpenFileCount: %d", beforeCount)
```
lisguo, line 476: can remove debug print lines

plugins/inputs/logfile/tailersrc.go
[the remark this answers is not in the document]
movence: added time out to handle deleted file without getting blocked indifinitely

logs/logs.go
movence: removed

jefchien (approved):

translator/tocwconfig/toenvconfig/toEnvConfig.go
```diff
@@ -65,11 +64,9 @@ func ToEnvConfig(jsonConfigValue map[string]interface{}) []byte {
 		envVars[envconfig.AWS_CA_BUNDLE] = sslConfig[commonconfig.CABundlePath]
 	}
 
-	//respect existing flag already set in env
-	backpressureDrop := envconfig.IsBackpressureDropEnabled()
-	// skip adding if not enabled
-	if backpressureDrop {
-		envVars[envconfig.CWAgentBackpressureDrop] = strconv.FormatBool(backpressureDrop)
+	backpressureMode := envconfig.GetLogsBackpressureMode()
+	if len(backpressureMode) > 0 {
```
lisguo, line 67: nit: can just be `if backpressureMode != ""` for readability

lisguo (approved):

plugins/inputs/logfile/fileconfig.go
```diff
@@ -60,6 +60,9 @@ type FileConfig struct {
 	// automatically remove the file / symlink after uploading.
 	// This auto removal does not support the case where other log rotation mechanism is already in place.
 	AutoRemoval bool `toml:"auto_removal"`
+	// strategy during backpressure in log processing where AutoRemoval take higher priority. supported values
```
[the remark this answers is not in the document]
movence: yeah we can update it once there are more to support

plugins/inputs/logfile/tailersrc.go
```diff
@@ -273,24 +265,88 @@ func (ts *tailerSrc) runTail() {
...
+				timer := time.NewTimer(tailCloseThreshold)
+				defer timer.Stop()
+
+				for {
+					select {
+					case ts.buffer <- e:
+						// sent event after buffer gets freed up
+						if ts.tailer.IsFileClosed() { // skip file closing if not already closed
+							if err := ts.tailer.Reopen(false); err != nil {
+								log.Printf("E! [logfile] error reopening file %s: %v", ts.tailer.Filename, err)
```
jefchien, line 310: nit: We might not want to log this if it just wasn't able to find the file anymore. Might be misleading.

jefchien (dismissed):

plugins/inputs/logfile/logfile.go
```diff
@@ -304,18 +304,20 @@ func (t *LogFile) getTargetFiles(fileconfig *FileConfig) ([]string, error) {
 			continue
 		}
 
-		// Add another file blacklist here
 		fileBaseName := filepath.Base(matchedFileName)
 		if blacklistP != nil && blacklistP.MatchString(fileBaseName) {
 			continue
 		}
 		if !fileconfig.PublishMultiLogs {
+			t.Log.Debugf("Single file mode - current target: %s (ModTime: %v), candidate: %s (ModTime: %v)",
```
jefchien, line 312: nit: I think this can become excessive logging as well depending on the number of files that match.

jefchien (approved):

lisguo (approved):

PR #1595, opened 2025-03-14
Fix AMP workspace deletion

varunch77:
# Description of the issue
AMP tests that are cancelled are not properly deleting resources. This leads to an accumulation of unused workspaces over time, causing us to hit the limit for workspaces in a region (currently 25).

# Description of changes

We were previously facing a state lock error when the test attempted to delete resources after a cancellation: 
https://github.com/aws/amazon-cloudwatch-agent/actions/runs/13844537774/job/38787872613#step:8:29

By setting `-lock-timeout`, the test ensures Terraform waits gracefully instead of failing immediately while the state lock is acquired.

# License
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice. ✅ 

# Tests
Cancelled Integration Run (verified that workplaces got deleted from our test account): https://github.com/aws/amazon-cloudwatch-agent/actions/runs/13905078366/job/38909150185#step:8:252

Integration Test Run: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/13906253933

# Requirements
2. Run `make lint` ✅

### DIFF
--- a/.github/workflows/ec2-integration-test.yml
+++ b/.github/workflows/ec2-integration-test.yml
@@ -136,4 +136,4 @@ jobs:
               cd ${{inputs.test_dir}}
             fi
 
-            terraform destroy -lock=false -var="region=${{ inputs.region }}" -var="ami=${{ matrix.arrays.ami }}" --auto-approve
+            terraform destroy -lock-timeout=5m -var="region=${{ inputs.region }}" -var="ami=${{ matrix.arrays.ami }}" --auto-approve
--- a/.github/workflows/ec2-integration-test.yml
+++ b/.github/workflows/ec2-integration-test.yml
@@ -136,4 +136,4 @@ jobs:
               cd ${{inputs.test_dir}}
             fi
 
-            terraform destroy -var="region=${{ inputs.region }}" -var="ami=${{ matrix.arrays.ami }}" --auto-approve
+            terraform destroy -lock=false -var="region=${{ inputs.region }}" -var="ami=${{ matrix.arrays.ami }}" --auto-approve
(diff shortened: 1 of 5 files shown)


.github/workflows/ec2-integration-test.yml
```diff
@@ -136,4 +136,4 @@ jobs:
               cd ${{inputs.test_dir}}
             fi
 
-            terraform destroy -var="region=${{ inputs.region }}" -var="ami=${{ matrix.arrays.ami }}" --auto-approve
+            terraform destroy -lock=false -var="region=${{ inputs.region }}" -var="ami=${{ matrix.arrays.ami }}" --auto-approve
```
jefchien, line 139:
```
│ Terraform acquires a state lock to protect the state from being written
│ by multiple users at the same time. Please resolve the issue above and try
│ again. For most commands, you can disable locking with the "-lock=false"
│ flag, but this is not recommended.
```
Do we know why the state lock is still acquired by the apply?

TravisStark: should `-lock=false` be explicitly added to the `Terraform apply` step as well?

varunch77: @jefchien: i think this issue is specific to the AMP test since it's referring to an external module that may be affecting the state lock: https://github.com/aws/amazon-cloudwatch-agent-test/blob/main/terraform/ec2/linux/main.tf#L113

varunch77: @TravisStark Shouldn't be needed since we only want to do this when tests are cancelled

TravisStark (approved):

PR #1602, opened 2025-03-18
Add SELinux Status using useragent

Paramadon:
# Description of the Issue  
Currently, there is no way to determine how many CloudWatch Agents (CWA) have SELinux enabled. This PR addresses that by introducing an environment variable (`SELINUX`), which is set using the SELinux installation script and stored in the systemd override file for the CloudWatch Agent service.  

# Details  

### Setting SELinux Status  
- A new environment variable, **SELINUX**, is set to `true` or `false` based on whether SELinux is enabled.  
- This variable is written to `/etc/systemd/system/amazon-cloudwatch-agent.service.d/override.conf`.  

### Persistence Across Updates  
- The systemd override file remains unchanged when the CloudWatch Agent is updated, ensuring SELinux status tracking persists across versions.  

### User-Agent Integration  
- The agent reads the **SELINUX** environment variable and includes it in the user-agent string.  
- This provides insight into SELinux usage within CloudWatch Agent deployments.  

### Benefits  
- Enables visibility into SELinux adoption among CloudWatch Agent users.  
- Assists in debugging and understanding system security configurations.  
- Ensures the change is persistent across agent updates with minimal disruption.

# License

# Tests
Tested by running selinux installation script

by adding this to the selinux script
```
# Ensure the directory exists
mkdir -p /etc/systemd/system/amazon-cloudwatch-agent.service.d

# Create the override file with the Environment variable
cat <<EOF | tee /etc/systemd/system/amazon-cloudwatch-agent.service.d/override.conf
[Service]
Environment="SELINUX=true"
EOF

# Reload systemd and restart the CloudWatch agent
systemctl daemon-reload
systemctl restart amazon-cloudwatch-agent
```

### Result
Screenshot 2025-03-17 at 9 43 01 PM

### User Agent output
Screenshot 2025-03-18 at 12 35 53 PM

Selinux flag is in the inputs section of the user agent:

```
User-Agent: CWAgent/1.300052.1 (go1.24.0; linux; amd64) ID/a4f51326-a0f3-47b4-ab21-8492373e11e1 inputs:(logfile mem net procstat selinux) processors:(awsentity cumulativetodelta ec2tagger) outputs:(awscloudwatch cloudwatch cloudwatchlogs) aws-sdk-go/1.48.6 (go1.24.0; linux; amd64)
```

# Requirements
2. Run `make lint`

### DIFF
--- a/cfg/envconfig/envconfig.go
+++ b/cfg/envconfig/envconfig.go
@@ -74,6 +74,6 @@ func IsWindowsHostProcessContainer() bool {
 	return false
 }
 
-func IsSelinuxEnabled() (status bool) {
+func IsSelinuxEnabled() bool {
 	return os.Getenv(RunWithSELinux) == TrueValue
 }
--- a/cfg/envconfig/envconfig.go
+++ b/cfg/envconfig/envconfig.go
@@ -25,6 +25,7 @@ const (
 	RunAsHostProcessContainer = "RUN_AS_HOST_PROCESS_CONTAINER"
 	RunInAWS                  = "RUN_IN_AWS"
 	RunWithIRSA               = "RUN_WITH_IRSA"
+	RunWithSELinux            = "RUN_WITH_SELINUX"
 	UseDefaultConfig          = "USE_DEFAULT_CONFIG"
 	HostName                  = "HOST_NAME"
 	PodName                   = "POD_NAME"
@@ -71,4 +72,8 @@ func IsWindowsHostProcessContainer() bool {
 		return true
 	}
 	return false
+}
+
+func IsSelinuxEnabled() (status bool) {
+	return os.Getenv(RunWithSELinux) == TrueValue
 }
--- a/extension/agenthealth/handler/useragent/useragent.go
+++ b/extension/agenthealth/handler/useragent/useragent.go
@@ -31,6 +31,7 @@ const (
 	flagContainerInsights         = "container_insights"
 	flagAppSignals                = "application_signals"
 	flagEnhancedContainerInsights = "enhanced_container_insights"
+	flagSELinux                   = "selinux"
 
 	separator = " "
 
@@ -76,6 +77,11 @@ func (ua *userAgent) SetComponents(otelCfg *otelcol.Config, telegrafCfg *telegra
 	}
 	for _, output := range telegrafCfg.Outputs {
 		ua.outputs.Add(output.Config.Name)
+	}
+
+	// Adding SELinux status
+	if envconfig.IsSelinuxEnabled() {
+		ua.inputs.Add(flagSELinux)
 	}
 
 	for _, pipeline := range otelCfg.Service.Pipelines {
--- a/cmd/amazon-cloudwatch-agent/amazon-cloudwatch-agent.go
+++ b/cmd/amazon-cloudwatch-agent/amazon-cloudwatch-agent.go
@@ -740,4 +740,3 @@ func checkRightForBinariesFileWithInputPlugins(inputPlugins []string) (string, e
 
 	return "", nil
 }
-
--- a/cmd/amazon-cloudwatch-agent/amazon-cloudwatch-agent.go
+++ b/cmd/amazon-cloudwatch-agent/amazon-cloudwatch-agent.go
@@ -310,6 +310,10 @@ func runAgent(ctx context.Context,
 		}
 	}
 
+	if envconfig.IsSelinuxEnabled() {
+		log.Println("I! SELinux Status: Enabled")
+	}
+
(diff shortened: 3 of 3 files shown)


extension/agenthealth/handler/useragent/useragent.go
```diff
@@ -78,6 +79,15 @@ func (ua *userAgent) SetComponents(otelCfg *otelcol.Config, telegrafCfg *telegra
 		ua.outputs.Add(output.Config.Name)
 	}
 
+	// Adding SELinux status
+	if os.Getenv("SELINUX") == "true" {
```
sky333999, line 83: We need to be consistent with consts - https://github.com/aws/amazon-cloudwatch-agent/blob/fbacc48316a66b73977c01e1891d69229e7b0a2e/cfg/envconfig/envconfig.go#L43

extension/agenthealth/handler/useragent/useragent.go
```diff
@@ -78,6 +79,15 @@ func (ua *userAgent) SetComponents(otelCfg *otelcol.Config, telegrafCfg *telegra
...
+		ua.inputs.Add("SELinux/Enabled")
+		log.Printf("SELinux status: Enabled")
+	} else {
+		ua.inputs.Add("SELinux/Disabled")
```
sky333999, line 87: We probably dont need this. It can be sparse.

extension/agenthealth/handler/useragent/useragent.go
```diff
@@ -78,6 +79,15 @@ func (ua *userAgent) SetComponents(otelCfg *otelcol.Config, telegrafCfg *telegra
 		ua.outputs.Add(output.Config.Name)
 	}
 
+	// Adding SELinux status
+	if os.Getenv("SELINUX") == "true" {
+		ua.inputs.Add("SELinux/Enabled")
+		log.Printf("SELinux status: Enabled")
```
sky333999, line 85: How often does this get logged?

extension/agenthealth/handler/useragent/useragent.go
```diff
@@ -196,6 +206,13 @@ func newUserAgent() *userAgent {
 		outputs:    collections.NewSet[string](),
 	}
 }
+func getSELinuxStatus() string {
```
sky333999: Why do we need this if its never invoked?

extension/agenthealth/handler/useragent/useragent.go
sky333999: Define a const. Look how and where `RUN_WITH_IRSA` is defined.

sky333999: nit: Probably call this `RUN_WITH_SELINUX`.

extension/agenthealth/handler/useragent/useragent.go
```diff
@@ -78,6 +79,15 @@ func (ua *userAgent) SetComponents(otelCfg *otelcol.Config, telegrafCfg *telegra
 		ua.outputs.Add(output.Config.Name)
 	}
 
+	// Adding SELinux status
+	if os.Getenv("SELINUX") == "true" {
+		ua.inputs.Add("SELinux/Enabled")
```
sky333999, line 84: Can you enable http logging and post a screenshot of the UserAgent header to the PR's description

cfg/envconfig/envconfig.go
```diff
@@ -25,6 +25,7 @@ const (
 	RunAsHostProcessContainer = "RUN_AS_HOST_PROCESS_CONTAINER"
 	RunInAWS                  = "RUN_IN_AWS"
 	RunWithIRSA               = "RUN_WITH_IRSA"
+	RunWithSELinux            = "RUN_WITH_SELINUX"
```
Paramadon, line 28: We will have a seperate pr for ROSA due to testing

extension/agenthealth/handler/useragent/useragent.go
Paramadon: Good catch, I was thinking of using this in runAgent for logging selinux status (needed to make G capital to export) but SetComponents seems to be called just once in runAgent flow so to reduce code I removed this function and added a log in SetComponents. But if you think we should still move this log statment out that's fine too, just depends on what we prefer.

extension/agenthealth/handler/useragent/useragent.go
okankoAMZ: Is there a way to make it only run once?

okankoAMZ (approved):
Small comment not a blocker

sky333999 (approved):

okankoAMZ (approved):

okankoAMZ (approved):

TravisStark (approved):

PR #1609, opened 2025-03-19
Bumping eks retries integ test

Paramadon:
# Description of the issue
Test fails in some occasions where it should not, test might be flaky so increasing number of retries will allow for more accurate test. (only increases time duration of test by ~10 minutes-15 if all tests fail).

# Description of changes

# License

# Tests

# Requirements
2. Run `make lint`

### DIFF
--- a/.github/workflows/test-artifacts.yml
+++ b/.github/workflows/test-artifacts.yml
@@ -847,7 +847,7 @@ jobs:
       - name: Terraform apply
         uses: nick-fields/retry@v2
         with:
-          max_attempts: 2
+          max_attempts: 4
           timeout_minutes: 90 # EKS takes about 20 minutes to spin up a cluster and service on the cluster
           retry_wait_seconds: 5
           command: |


.github/workflows/test-artifacts.yml
```diff
@@ -847,7 +847,7 @@ jobs:
       - name: Terraform apply
         uses: nick-fields/retry@v2
         with:
-          max_attempts: 2
+          max_attempts: 4
           timeout_minutes: 90 # EKS takes about 20 minutes to spin up a cluster and service on the cluster
```
dricross, line 851: Do we need to increase the timeout to account for more attempts? comment says 20 minutes to spin up a cluster per attempt, so thats 80 minutes already. There's some time to delete the clusters as well so I'd guess we'd hit the timeout before we finish the 4th attempt

movence:
This change is not needed anymore with https://github.com/aws/amazon-cloudwatch-agent-test/pull/483#pullrequestreview-2699006048

PR #1624, opened 2025-03-27
Add deltatocumulativeprocessor for AMP destinations

dricross:
# Description of the issue
Currently, the CloudWatch Agent uses the [prometheusremotewrite exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/exporter/prometheusremotewriteexporter) to send metrics to Amazon Managed Prometheus. It does not support delta metrics so the metrics are dropped:

> Non-cumulative monotonic, histogram, and summary OTLP metrics are dropped by this exporter.

This issue was discovered while updating the AMP integration test: https://github.com/aws/amazon-cloudwatch-agent-test/pull/465#issue-2864562335

# Description of changes
To remediate, the agent will use [deltatocumulativeprocessor](https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/deltatocumulativeprocessor) in AMP destination pipelines to convert delta metrics to cumulative metrics before being processed by the exporter.

The processor has two settings:
* `max_stale`: how long until a series not receiving new samples is removed
* `max_streams`: upper limit of streams to track. new streams exceeding this limit will be dropped

The `max_stale` setting defaults to 5m which may not be ideal for all users. If users emit hourly delta metrics, for example, then the ingested datapoints will always start a new stream. The [`cumulativetodeltaprocessor`](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/cumulativetodeltaprocessor/README.md) is very similar processor which has a similar setting. By default, the processor uses infinite staleness so metric streams will never be dropped. Ideally, we'd configure the deltatocumulativeprocessor in the same manner the be consistent, but infinite staleness is not an option for the deltatocumulativeprocessor. Instead, we'll follow AWS's [PutMetricData API](https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_PutMetricData.html) convention of supporting metrics up two weeks:

> You can specify time stamps that are as much as two weeks before the current date, and as much as 2 hours after the current day and time.

The `max_streams` setting defaults to 64-bit integer maximum (9223372036854775807) which is effectively infinite. This is consistent with the cumulativetodeltaprocessor which uses an unbounded map to keep track of metrics.

# License

# Tests
Ran the AMP integration test with agent updates and then locate delta metrics in AMG:
snap 250326182535 Z7tJrCzY

Updated AMP integration test to check for delta counter and delta histogram metrics now that they are supported.

# Requirements
2. Run `make lint`

### DIFF
--- a/translator/translate/otel/processor/deltatocumulativeprocessor/translator_test.go
+++ b/translator/translate/otel/processor/deltatocumulativeprocessor/translator_test.go
@@ -1,0 +1,70 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+package deltatocumulativeprocessor
+
+// func TestTranslator(t *testing.T) {
+// 	cdpTranslator := NewTranslator(common.WithName("test"), WithDefaultKeys())
+// 	require.EqualValues(t, "cumulativetodelta/test", cdpTranslator.ID().String())
+// 	testCases := map[string]struct {
+// 		input   map[string]any
+// 		want    map[string]any
+// 		wantErr error
+// 	}{
+// 		"GenerateDeltaProcessorConfigWithCPU": {
+// 			input: map[string]any{
+// 				"metrics": map[string]any{
+// 					"metrics_collected": map[string]any{
+// 						"cpu": map[string]any{},
+// 					},
+// 				},
+// 			},
+// 			wantErr: &common.MissingKeyError{ID: cdpTranslator.ID(), JsonKey: fmt.Sprint(diskioKey, " or ", netKey, " or ", otlpKey, " or ", otlpEmfKey)},
+// 		},
+// 		"GenerateDeltaProcessorConfigWithNet": {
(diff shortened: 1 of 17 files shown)


translator/translate/otel/processor/deltatocumulativeprocessor/translator_test.go
```diff
@@ -0,0 +1,70 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+package deltatocumulativeprocessor
+
+// func TestTranslator(t *testing.T) {
```
jefchien, line 6: Why is this commented out?

dricross: Looks like I coped over the tests from cumulativetodeltaprocessor, commented this out to get something building, and forgot to actually fill in the unit tests. I will fix it

jefchien (approved):

PR #1629, opened 2025-03-31
Add EC2 integration test job to PR requests

Paramadon:
# Description of the issue
Pr adds another step in the PR-build that ensures integ test passes otherwise pr will fail since the agent change is causing an integration test failure. 
 
# Description of changes

# License

# Tests

# Requirements
2. Run `make lint`


.github/workflows/PR-build.yml
```diff
@@ -2,22 +2,30 @@
...
       - main*
       - feature*
-    types: 
+    types:
       - opened
       - synchronize
       - reopened
       - ready_for_review
 
-concurrency:
```
Paramadon, line 17: This is being removed since it is part of build-test-artifacts.yml which causes a loop error to occur.

varunch77 (approved):

TravisStark (approved):

PR #1634, opened 2025-04-02
Add translation for awsebsnvmereceiver

duhminick:
# Description of the issue

When customers configure ebs volume metrics under the diskio plugin, we should include the ebsnvme receiver in the host metrics pipeline:

```
"metrics_collected": {
  "diskio" {
    ...
    "measurement": ["diskio_ebs_total_read_ops"]
  }
}
```

# Description of changes
1. A new translator for the awsebsnvmereceiver component was created.
  - Retrieves collection interval from the diskio section, then the agent section. If not set, then defaults to 1 minute.
  - Uses the resources field to determine which devices the receiver should be scraping metrics for
  - Enables the metrics on the receiver (as opposed to using a filterprocessor). Only one metric is enabled by default since that is a requirement of mdatagen. The one default metric is disabled in the translation logic so that the configuration from the user is then used.
  - Logic to not include Telegraf metrics for the NVMe receiver since we are sharing the same diskio section
2. Update to Telegraf translator logic to not have it be configured for EBS NVMe metrics
3. Update so that the host metrics pipeline can have both the EBS NVMe receiver and Telegraf adapter receiver. Only if the metric from the specific receiver is used will it then be added to the pipeline.

# License

# Tests
1. Integration tests: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/14135365040/job/39606042755, https://github.com/aws/amazon-cloudwatch-agent-test/tree/dominic-ebs-nvme
4. tocwconfig unit tests
5. Manual test
image

#### Agent Config
```
"metrics": {
  "metrics_collected": {
     "diskio": {
       "resources": ["nvme0n1"],
       "measurements": ["diskio_ebs_total_write_bytes", "diskio_ebs_total_read_bytes"]
     }
  }
}
```
#### Output
```
receivers:
    awsebsnvmereceiver:
        collection_interval: 1m0s
        initial_delay: 1s
        metrics:
            diskio_ebs_ec2_instance_performance_exceeded_iops:
                enabled: false
            diskio_ebs_ec2_instance_performance_exceeded_tp:
                enabled: false
            diskio_ebs_total_read_bytes:
                enabled: true
            diskio_ebs_total_read_ops:
                enabled: false
            diskio_ebs_total_read_time:
                enabled: false
            diskio_ebs_total_write_bytes:
                enabled: true
            diskio_ebs_total_write_ops:
                enabled: false
            diskio_ebs_total_write_time:
                enabled: false
            diskio_ebs_volume_performance_exceeded_iops:
                enabled: false
            diskio_ebs_volume_performance_exceeded_tp:
                enabled: false
            diskio_ebs_volume_queue_length:
                enabled: false
        resource_attributes:
            VolumeId:
                enabled: true
        devices:
            - nvme0n1
        timeout: 0s
```

# Requirements
7. Run `make lint`


translator/translate/otel/receiver/adapter/translators.go
```diff
@@ -159,10 +161,27 @@ func fromInputs(conf *confmap.Conf, validInputs map[string]bool, baseKey string)
 					continue
 				}
 			}
+
 			cfgKey := common.ConfigKey(baseKey, inputName)
-			if measurement := common.GetArray[any](conf, common.ConfigKey(cfgKey, common.MeasurementKey)); measurement != nil && len(measurement) == 0 {
+
+			hasMeasurements := true
+			hasOnlyNonAdaptedMetrics := false
```
lisguo, line 168: I would rename this to something simpler like `noTelegrafReceivers`
```suggestion
			noTelegrafReceivers := false
```

translator/tocwconfig/sampleConfig/advanced_config_darwin.yaml
```diff
@@ -38,6 +38,8 @@ processors:
             metrics:
                 - iops_in_progress
                 - diskio_iops_in_progress
+                - ebs_volume_queue_length
+                - diskio_ebs_volume_queue_length
```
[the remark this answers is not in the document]
duhminick: Yeah, that's what I was thinking as well. I was just following the pattern above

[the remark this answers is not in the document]
zhihonl: If these two metrics are same metrics but different name then we should only have one of them. Which name is the product aligned on?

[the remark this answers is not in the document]
zhihonl: Also curious, does this mean current implementation is emitting two metrics that are the same thing? If so we should fix that. Haven't read the rest of PRs so maybe not an issue.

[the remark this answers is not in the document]
duhminick: It's on `diskio_ebs_volume_queue_length`. I can remove it. The current implementation will only emit `diskio_ebs_volume_queue_length`

translator/translate/otel/receiver/awsebsnvme/translator.go
```diff
@@ -0,0 +1,102 @@
...
+
+	resources := common.GetArray[string](conf, common.ConfigKey(baseKey, common.ResourcesKey))
+	if resources == nil {
+		// Was not set by the user, so collect all devices by default
+		cfg.Devices = []string{"*"}
+	} else {
+		cfg.Devices = resources
+	}
+
+	// Total Read Ops is the only metric enabled by default. Disable it so that
```
zhihonl, line 69: Would it be cleaner to make it false by default instead of doing this weird logic here or is there a reason to make it enabled by default?

translator/translate/otel/receiver/adapter/translators.go
```diff
@@ -159,10 +161,27 @@ func fromInputs(conf *confmap.Conf, validInputs map[string]bool, baseKey string)
...
+			if conf.IsSet(common.ConfigKey(cfgKey, common.MeasurementKey)) {
+				inputConf := conf.Get(cfgKey).(map[string]any)
```
zhihonl, line 170: Do we need a nil check for `conf.Get(cfgKey)`?

translator/translate/otel/receiver/awsebsnvme/testdata/non_ebs_metrics.json
```diff
@@ -0,0 +1,14 @@
+{
+  "metrics": {
+    "metrics_collected": {
+      "diskio": {
+        "measurement": [
+          "ebs_total_read_ops",
```
zhihonl, line 6: Isn't this case covered in `mixed_metrics.json`?

translator/translate/metrics/util/measurementutil.go
```diff
@@ -58,7 +61,9 @@ func ApplyMeasurementRule(inputs interface{}, pluginName string, targetOs string
 		}
 
 		if formatted_metricName := GetValidMetric(targetOs, pluginName, inputMetricName.(string)); formatted_metricName != "" {
-			returnVal = append(returnVal, formatted_metricName)
+			if !shouldFilterPluginSpecificMetrics(pluginName, formatted_metricName) {
```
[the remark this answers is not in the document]
duhminick: Yeah, we do not want any EBS metrics in returnVal. returnVal will return the measurements used for Telegraf.

As for why I pulled it out into a function: https://github.com/aws/amazon-cloudwatch-agent/blob/11067e59c827a957d0fcd608a4e3390f74f7947b/translator/translate/metrics/util/measurementutil.go#L191-L205

Was just trying to maintain the already existing pattern of doing it per-plugin

translator/translate/otel/receiver/awsebsnvme/testdata/non_ebs_metrics.json
duhminick: Oh, huh. I'm just gonna remove this file. NIce catch

translator/translate/otel/receiver/adapter/translators.go
duhminick: I think realistically, no. But I'll do it anyway, it won't hurt anything

zhihonl (approved):

PR #1654, opened 2025-04-15
Removing unwanted os's test for selinux in pr-test 

Paramadon:
# Description of the issue
Currently we have a run that tests for multiple different linux os's but we only want to test for al2 and al2023. Here is the an example run. https://github.com/aws/amazon-cloudwatch-agent/actions/runs/14477540108/job/40606719207?pr=1628

This pr removes other os's from test and just keeps al2 and al2023.

# Requirements
2. Run `make lint`


musa-asad:
Can you share an older run just for comparison?

TravisStark (approved):

PR #1672, opened 2025-05-06
Dropped .NET 6 E2E Tests

AsakerMohd:
# Description of the issue
We are upgrading our DotNet support so removing the workflow that runs against .NET Version 6 since we are no longer going to support it. 

# License

# Tests
This is a workflow change. I tested it on another repo and the tests for version 6 aren't running anymore. 

# Requirements
2. Run `make lint`


vastin (approved):
LGTM

PR #1677, opened 2025-05-08
Add exponential histogram support to CloudWatch PMD Exporter

dricross:
# Description of the issue
The Cloudwatch/PMD exporter currently drops all exponential histogram metrics.

# Description of changes
* Add support for exponential histogram to the ec2tagger processor 
* Add support for exponential histogram to the CloudWatch/PMD exporter

> [!NOTE]
> See companion PR for updating cumulativetodelta processor: https://github.com/amazon-contributing/opentelemetry-collector-contrib/pull/331

# License

# Tests
Current histogram test fails for cumulative histograms as the current agent behavior for cumulative histograms is incorrect and it being fixed in this PR. Specifically, the agent does not convert cumulative histograms to delta before pushing to cloudwatch. With these changes, the agent will. So the existing test fails, but the updated tests pass.

> [!NOTE]
> See companion PR for integration test: https://github.com/aws/amazon-cloudwatch-agent-test/pull/558

Integration test run: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/16839298907
Histogram test: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/16839298907/job/47706510747

The data we are sending as part of the test is defined here: https://github.com/aws/amazon-cloudwatch-agent-test/blob/dricross/exphistograms/test/histograms/resources/otlp_metrics.json#L188. where CUMULATIVE_HIST_COUNT and CUMULATIVE_HIST_SUM are increasing by 2 every 10 seconds. Important things to note:

* Scale=0
* IndexOffset = 1
* Buckets = [0, 2] (where 2 is growing by 2 every 10 seconds)
* With Scale=0, Base = 2 (Base = 2^(2^(-scale)) = 2^(2^0) = 2^1)

Formula for mapping bucket index to value: `(Base^(index+offset), Base^(index+offset+1]`

The first bucket (with 0 count) maps to range (2,4] (`(2^(0+1), 2^(1+1+1]`)
The second bucket (with a count of 2) maps to range (4, 8] (`(2^(1+1), 2^(1+1+1] = (4, 8]`)

Since we take the midpoints of the buckets when sending to exponential histograms to CloudWatch (consistent with the EMF exporter), the values/counts we send to CloudWatch is:
```
Values=[6,3]
Counts[2,0]
```

That list of values and counts is sent 6 times per minute

CloudWatch metrics:
image

EMF Logs from test as verification (this is existing behavior but demonstrates the intended output)
```
{
    "AWS.ServiceNameSource": "Instrumentation",
    "CloudWatchMetrics": [
        {
            "Namespace": "CWAgent",
            "Dimensions": [
                [
                    "service.name",
                    "my.exponential.histogram.attr"
                ]
            ],
            "Metrics": [
                {
                    "Name": "my.delta.exponential.histogram",
                    "Unit": "",
                    "StorageResolution": 60
                }
            ]
        }
    ],
    "EC2.InstanceId": "i-04282f146a9432a17",
    "Environment": "ec2:default",
    "PlatformType": "AWS::EC2",
    "Service": "my.service",
    "Timestamp": "1753710063060",
    "Version": "0",
    "my.exponential.histogram.attr": "some value",
    "service.name": "my.service",
    "my.delta.exponential.histogram": {
        "Values": [
            6,
            0
        ],
        "Counts": [
            2,
            1
        ],
        "Max": 5,
        "Min": 0,
        "Count": 3,
        "Sum": 10
    }
}
{
    "AWS.ServiceNameSource": "Instrumentation",
    "CloudWatchMetrics": [
        {
            "Namespace": "CWAgent",
            "Dimensions": [
                [
                    "my.cumulative.exponential.histogram.attr",
                    "service.name"
                ]
            ],
            "Metrics": [
                {
                    "Name": "my.cumulative.exponential.histogram",
                    "Unit": "",
                    "StorageResolution": 60
                }
            ]
        }
    ],
    "EC2.InstanceId": "i-04282f146a9432a17",
    "Environment": "ec2:default",
    "PlatformType": "AWS::EC2",
    "Service": "my.service",
    "Timestamp": "1753710063060",
    "Version": "0",
    "my.cumulative.exponential.histogram.attr": "some value",
    "service.name": "my.service",
    "my.cumulative.exponential.histogram": {
        "Values": [
            6
        ],
        "Counts": [
            2
        ],
        "Max": 0,
        "Min": 0,
        "Count": 2,
        "Sum": 2
    }
}
```

# Requirements
2. Run `make lint`


metric/distribution/exph/exph.go
```diff
@@ -0,0 +1,289 @@
...
+	size := len(d.negativeBuckets) + len(d.positiveBuckets)
+	if d.zeroCount > 0 {
+		size++
+	}
+	return size
+}
+
+// ValuesAndCounts outputs two arrays representing the midpoints of each exponential histogram bucket and the
+// counter of datapoints within the corresponding exponential histogram buckets
+func (d *ExpHistogramDistribution) ValuesAndCounts() ([]float64, []float64) {
```
jefchien, line 75: Does the order of the values matter? Does it have to be positive to negative?

metric/distribution/exph/exph.go
```diff
@@ -0,0 +1,289 @@
...
+		bucketBegin := LowerBoundary(offsetIndex, int(d.scale))
+		bucketEnd := LowerBoundary(offsetIndex+1, int(d.scale))
+		value := (bucketBegin + bucketEnd) / 2.0
+		values = append(values, value)
+		counts = append(counts, float64(counter))
+	}
+
+	if d.zeroCount > 0 {
+		values = append(values, 0)
+		counts = append(counts, float64(d.zeroCount))
```
jefchien, line 97: nit: If we need to convert all of these to float64s, why not just store it as a float64 instead of a uint64?

metric/distribution/exph/exph.go
```diff
@@ -0,0 +1,289 @@
...
+	values := []float64{}
+	counts := []float64{}
+
+	// iterate through positive buckets in descending order
+	posOffsetIndicies := make([]int, 0, len(d.positiveBuckets))
+	for offsetIndex := range d.positiveBuckets {
+		posOffsetIndicies = append(posOffsetIndicies, offsetIndex)
+	}
+	slices.Sort(posOffsetIndicies)
+	slices.Reverse(posOffsetIndicies)
```
jefchien, line 85: Instead of sorting and then reversing, we can use `maps.Keys` which returns an iterator and a sort function to reverse it while doing the sort.

https://pkg.go.dev/maps#Keys
https://pkg.go.dev/slices#SortedFunc

```suggestion
	posOffsetIndicies := slices.SortedFunc(maps.Keys(d.positiveBuckets), func(a, b int) int {
	  return cmp.Compare(b, a)
	})
```

metric/distribution/exph/exph.go
```diff
@@ -0,0 +1,289 @@
...
+	}
+	scaleFactor := math.Ldexp(math.Log2E, scale)
+	// The use of math.Log() to calculate the bucket index is not guaranteed to be exactly correct near powers of two.
+	return int(math.Floor(math.Log(math.Abs(value)) * scaleFactor))
+}
+
+// LowerBoundaryNegativeScale computes the lower boundary for index
+// with scales <= 0.
+func LowerBoundaryNegativeScale(index int, scale int) float64 {
+	return math.Ldexp(1, index<<-scale)
```
jefchien, line 273: Why are we using `math.Ldexp` instead of just `math.Exp2` if frac is always 1? 

https://pkg.go.dev/math#Ldexp
> ```
>func Ldexp(frac float64, exp int) float64
>```
> Ldexp is the inverse of [Frexp](https://pkg.go.dev/math#Frexp). It returns frac × 2**exp.

metric/distribution/exph/exph.go
```diff
@@ -0,0 +1,289 @@
...
+	negativeOffset := negativeBuckets.Offset()
+	negBucketCounts := negativeBuckets.BucketCounts().AsRaw()
+	for negBucketIndex := range negBucketCounts {
+		offsetIndex := negBucketIndex + int(negativeOffset)
+		d.negativeBuckets[offsetIndex] = negBucketCounts[negBucketIndex]
+	}
+}
+
+func (d *ExpHistogramDistribution) Resize(_ int) []*ExpHistogramDistribution {
+	// for now, do not split data points into separate PMD requests
```
jefchien, line 231: When we say "for now" does that mean this should be a `TODO`?

metric/distribution/exph/exph.go
```diff
@@ -0,0 +1,289 @@
...
+	return []*ExpHistogramDistribution{d}
+}
+
+// MapToIndexScale0 computes a bucket index at scale 0.
+func MapToIndexScale0(value float64) int {
```
jefchien, line 236: It looks like these were all taken from https://opentelemetry.io/docs/specs/otel/metrics/data-model/#negative-scale-extract-and-shift-the-exponent. Consider putting in a separate file in the package (like `mapping.go` or `math.go`) and indicating the source at the top.

metric/distribution/exph/exph.go
dricross: It's something that needs to be done later, yes. Does marking it as `TODO` specifically do something special?

metric/distribution/exph/exph.go
dricross: As you mentioned later, it was pulled directly from the OTLP documentation: https://opentelemetry.io/docs/specs/otel/metrics/data-model/#negative-scale-extract-and-shift-the-exponent

metric/distribution/exph/exph.go
dricross: Sure, I can move those

metric/distribution/exph/exph.go
dricross: very nice, I'll update

metric/distribution/exph/exph.go
dricross: I don't believe so. See PutMetricData API documentation here: https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_PutMetricData.html. The order of values just needs to match the order of counts.

I selected this order based on the behavior of the awsemfexporter for exponential histograms, e.g. https://github.com/amazon-contributing/opentelemetry-collector-contrib/blame/aws-cwa-dev/exporter/awsemfexporter/datapoint_test.go#L1168

plugins/outputs/cloudwatch/convert_otel.go
```diff
@@ -144,6 +145,42 @@ func ConvertOtelHistogramDataPoints(
...
+				Unit:              aws.String(unit),
+				Timestamp:         aws.Time(dp.Timestamp().AsTime()),
+				StorageResolution: aws.Int64(storageResolution),
+			},
+			aggregationInterval: aggregationInterval,
+			entity:              entity,
+		}
+		// Assume function pointer is valid.
+		ad.expHistDistribution = exph.NewExpHistogramDistribution()
+		ad.expHistDistribution.ConvertFromOtel(dp, unit)
```
jefchien, line 178: nit: If we store the unit in the MetricDatum already, why do we need to store it in the distribution?

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -460,34 +454,106 @@ func (c *CloudWatch) BuildMetricDatum(metric *aggregationDatum) (cloudwatch.Enti
...
+				Counts:            aws.Float64Slice(counts),
+				StatisticValues:   &s,
+			}
+			datums = append(datums, datum)
+		}
+	}
+	return datums
+}
+
+func (c *CloudWatch) buildMetricDataumExph(metric *aggregationDatum, dimensionsList [][]*cloudwatch.Dimension) []*cloudwatch.MetricDatum {
```
jefchien, line 510: nit: Typo 
```suggestion
func (c *CloudWatch) buildMetricDatumExph(metric *aggregationDatum, dimensionsList [][]*cloudwatch.Dimension) []*cloudwatch.MetricDatum {
```

metric/distribution/exph/exph.go
```diff
@@ -0,0 +1,289 @@
...
+	}
+	return nil
+}
+
+// weight is 1/samplingRate
+func (d *ExpHistogramDistribution) AddEntry(value float64, weight float64) error {
+	return d.AddEntryWithUnit(value, weight, "")
+}
+
+func (d *ExpHistogramDistribution) AddDistribution(other *ExpHistogramDistribution) {
```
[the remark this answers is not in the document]
dricross: Yeah, you're right it doesn't satisfy the interface. Initially I was trying to do so but it ended up not working out because the Distribution interface is specific too `HistogramDataPoint`. In hindsight, I don't think it's even possible to combine exp histograms and classic histograms anyways so not really necessary at all. I already had this code though and didn't think it was worth refactoring everything to make it fit nicely together so I left it as is for now

plugins/outputs/cloudwatch/convert_otel.go
```diff
@@ -82,7 +83,7 @@ func ConvertOtelNumberDataPoints(
 	unit string,
 	scale float64,
 	entity cloudwatch.Entity,
-) []*aggregationDatum {
+) []*aggregationDatum { //nolint:revive
```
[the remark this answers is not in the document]
dricross: yeah that would be better. will update

go.mod
```diff
@@ -4,6 +4,10 @@ go 1.24.4
 
 replace github.com/influxdata/telegraf => github.com/aws/telegraf v0.10.2-0.20250113150713-a2dfaa4cdf6d
 
+replace collectd.org v0.4.0 => github.com/collectd/go-collectd v0.4.0
```
lisguo, line 7: are these dependencies of cumulativetodeltaprocessor? I don't see `go-collectd` or `clock` used in the new code

metric/distribution/exph/exph.go
```diff
@@ -0,0 +1,228 @@
...
+	return d.sum
+}
+
+func (d *ExpHistogramDistribution) Unit() string {
+	return d.unit
+}
+
+func (d *ExpHistogramDistribution) Size() int {
+	size := len(d.negativeBuckets) + len(d.positiveBuckets)
+	if d.zeroCount > 0 {
```
lisguo, line 72: what is zeroCount? the number of datapoints with 0?

metric/distribution/exph/exph.go
```diff
@@ -0,0 +1,228 @@
...
+	negativeOffset := negativeBuckets.Offset()
+	negBucketCounts := negativeBuckets.BucketCounts().AsRaw()
+	for negBucketIndex := range negBucketCounts {
+		offsetIndex := negBucketIndex + int(negativeOffset)
+		d.negativeBuckets[offsetIndex] = negBucketCounts[negBucketIndex]
+	}
+}
+
+func (d *ExpHistogramDistribution) Resize(_ int) []*ExpHistogramDistribution {
+	// TODO: split data points into separate PMD requests if the number of buckets exceeds the API limit
```
lisguo, line 209: what happens if we exceed the API limit?

metric/distribution/exph/exph.go
```diff
@@ -0,0 +1,228 @@
...
+	size := len(d.negativeBuckets) + len(d.positiveBuckets)
+	if d.zeroCount > 0 {
+		size++
+	}
+	return size
+}
+
+// ValuesAndCounts outputs two arrays representing the midpoints of each exponential histogram bucket and the
+// counter of datapoints within the corresponding exponential histogram buckets
+func (d *ExpHistogramDistribution) ValuesAndCounts() ([]float64, []float64) {
```
lisguo, line 80: nit: we can make the name of the function more descriptive. Maybe GetMidpointsAndCounts?

metric/distribution/exph/exph.go
```diff
@@ -0,0 +1,228 @@
...
+	if !distribution.IsSupportedValue(value, 0, distribution.MaxValue) {
+		return fmt.Errorf("unsupported value %v: %w", value, distribution.ErrUnsupportedValue)
+	}
+
+	d.sampleCount += weight
+	d.sum += value * weight
+	d.min = min(d.min, value)
+	d.max = max(d.max, value)
+
+	if math.Abs(value) > d.zeroThreshold {
```
lisguo: so if my zero threshold is 1, then both values 2 and -2 will be counted in the zero bucket?

metric/distribution/exph/exph.go
```diff
@@ -0,0 +1,228 @@
...
+		d.zeroCount += uint64(weight)
+	} else if value > d.zeroThreshold {
```
lisguo: would this ever happen? Wouldn't we always hit the first case?

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -419,28 +419,22 @@ func (c *CloudWatch) WriteToCloudWatch(req interface{}) {
...
-			log.Printf("E! metric has a distribution with no entries, %s", *metric.MetricName)
-			return metric.entity, datums
-		}
-		if metric.distribution.Unit() != "" {
-			metric.SetUnit(metric.distribution.Unit())
-		}
-		distList = resize(metric.distribution, c.config.MaxValuesPerDatum)
-	}
+		datums = c.buildMetricDatumDist(metric, dimensionsList)
+	} else if metric.expHistDistribution != nil {
```
lisguo: is there ever a chance we will have both `distribution` and `expHistDistribution` not nil? Probably shouldn't happen...

go.mod
dricross: I was hitting some issues with downloading the collectd dependency after clearing my local go cache and I needed to redirect to github to pick it up. I'm not exactly sure what happened but it looks like collectd.org stopped vending their package via collectd.org. The issue may have been resolved by now though, so I can try again.
```
../../../.gvm/pkgsets/go1.22.7/global/pkg/mod/github.com/aws/telegraf@v0.10.2-0.20250113150713-a2dfaa4cdf6d/plugins/parsers/collectd/parser.go:8:2: unrecognized import path "collectd.org": https fetch: Get "https://collectd.org/?go-get=1": dial tcp: lookup collectd.org on 10.4.4.10:53: read udp 10.169.109.191:52627->10.4.4.10:53: i/o timeout
```

metric/distribution/exph/exph.go
dricross: Yes, pretty much. OTLP exponential histograms splits the data into three section: negative values, zero values, positive values. Positive and negative buckets are defined separately by a series of buckets+counts. The zero values don't have any buckets so its just a counter stored in the histogram structure. Side note: the definition of "0" in OTLP exponential histograms is loose. Datapoints with a magnitude less than the configurable "zero threshold" is treated as 0.

metric/distribution/exph/exph.go
dricross: Good catch, this logic is incorrect. In fact this function is left over work from trying to adhere to the [Distribution](https://github.com/aws/amazon-cloudwatch-agent/blob/main/metric/distribution/distribution.go#L20) interface that's not used nor needed. I'll remove this function entirely.

metric/distribution/exph/exph.go
dricross: Incorrect logic, but will be removing this function.

metric/distribution/exph/exph.go
dricross: Based on the [API documentation](https://docs.aws.amazon.com/AmazonCloudWatch/latest/APIReference/API_PutMetricData.html), I believe the PMD request will be rejected with error code 400 InvalidParameterValue, though I haven't tried.

plugins/outputs/cloudwatch/cloudwatch.go
dricross: No (at least they shouldn't...). The agent will convert the different OTLP datatypes into aggregationDatums in [ConvertOtelMetrics](https://github.com/aws/amazon-cloudwatch-agent/blob/main/plugins/outputs/cloudwatch/convert_otel.go#L151). The existing ConvertOtelHistogramDataPoints sets `distribution` field and the new ConvertOtelExponentialHistogramDataPoints sets the `expHistDistribution` field.

metric/distribution/exph/exph.go
dricross: This naming scheme was based off of the existing function in the [Distribution](https://github.com/aws/amazon-cloudwatch-agent/blob/main/metric/distribution/distribution.go#L20) interface. I think ValuesAndCounts is a fairly descriptive name as that's what is actually pushed to CloudWatch in the PMD request (an array of values and an array of counts).

plugins/outputs/cloudwatch/convert_otel.go
dricross: This is based off of how it currently works for the regular histograms. The unit is given to the distribution so that it can check that similar distributions are being combined. If the unit for distribution does match the unit for the incoming distribution to combine it with, a debug log is printed.

plugins/outputs/cloudwatch/util.go
```diff
@@ -78,34 +77,6 @@ func setNewDistributionFunc(maxValuesPerDatumLimit int) {
 	}
 }
 
-func resize(dist distribution.Distribution, listMaxSize int) (distList []distribution.Distribution) {
```
dricross, line 81: refactored as functions on each distribution

metric/distribution/exph/exph.go
```diff
@@ -0,0 +1,188 @@
...
+		bucketEnd := LowerBoundary(offsetIndex+1, int(d.scale))
+		value := -(bucketBegin + bucketEnd) / 2.0
+		values = append(values, value)
+		counts = append(counts, float64(counter))
+	}
+
+	return values, counts
+}
+
+func (d *ExpHistogramDistribution) AddDistribution(other *ExpHistogramDistribution) {
```
lisguo: can probably remove this since the weights are all 1

metric/distribution/exph/mapping_test.go
```diff
@@ -0,0 +1,210 @@
...
+		values   []float64
+		expected []int
+	}{
+		{
+			name:     "positive value inside bucket",
+			values:   []float64{1.5, 3.0, 6, 12, 18},
+			expected: []int{0, 1, 2, 3, 4},
+		},
+		{
+			// for negative values, histogram buckets use lower-inclusive boundaries
```
lisguo: confusing considering the test case uses positive values

metric/distribution/exph/exph.go
```diff
@@ -0,0 +1,187 @@
...
+
+	"go.opentelemetry.io/collector/pdata/pmetric"
+)
+
+type ExpHistogramDistribution struct {
+	max             float64
+	min             float64
+	sampleCount     float64
+	sum             float64
+	scale           int32
```
jefchien: nit: Why not just store `scale` as an int if we're going to be casting it to an int everywhere it's used?

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -419,28 +419,22 @@ func (c *CloudWatch) WriteToCloudWatch(req interface{}) {
 // Or it might expand it into many datums due to dimension aggregation.
 // There may also be more datums due to resize() on a distribution.
 func (c *CloudWatch) BuildMetricDatum(metric *aggregationDatum) (cloudwatch.Entity, []*cloudwatch.MetricDatum) {
-	var datums []*cloudwatch.MetricDatum
-	var distList []distribution.Distribution
+
+	datums := []*cloudwatch.MetricDatum{}
```
jefchien: nit: Can initialize this as a nil slice to avoid allocation. 
```suggestion
	var datums []*cloudwatch.MetricDatum
```

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -460,34 +454,157 @@ func (c *CloudWatch) BuildMetricDatum(metric *aggregationDatum) (cloudwatch.Enti
...
+				Counts:            aws.Float64Slice(counts),
+				StatisticValues:   &s,
+			}
+			datums = append(datums, datum)
+		}
+	}
+	return datums
+}
+
+func (c *CloudWatch) buildMetricDatumExph(metric *aggregationDatum, dimensionsList [][]*cloudwatch.Dimension) []*cloudwatch.MetricDatum {
```
jefchien: nit: What is the difference between `buildMetricDatumExph` and `buildMetricDatumDist`? They seem to have the exact same code? Could we just pass in the distribution instead of the aggregationDatum?

```go
buildMetricDatumDist(metric, metric.expHistDistribution, dimensionList)
```

Think it might be worth making the exph distribution fit an interface. Something like

```go
type Distribution interface {
	Maximum() float64

	Minimum() float64

	SampleCount() float64

	Sum() float64

	ValuesAndCounts() ([]float64, []float64)

	Unit() string

	Size() int
}

type ResizableDistribution[T Distribution] interface {
    Distribution
    Resize(int) []T
}
```

jefchien (approved):

plugins/outputs/cloudwatch/cloudwatch.go
dricross: They are basically the same exact code. I tried making them share an interface but I couldn't get them to fit the same interface due to the `ConvertToOtel` and `ConvertFromOtel` functions which are specific to `pmetric.HistogramDataPoint`. It was looking like quite a lot of rework to get them to fit. Let me try your suggestion though

go.mod
jefchien: I think it's been resolved. Was able to pull vendor files from `collectd.org` directly the other day. Maybe their GitHub page (https://github.com/collectd/collectd.github.io) was down for a bit. Not opposed to this though.

metric/distribution/distribution.go
```diff
@@ -38,16 +38,24 @@ type Distribution interface {
...
 
 	ConvertFromOtel(dp pmetric.HistogramDataPoint, unit string)
 }
 
-var NewDistribution func() Distribution
+type ExponentialDistribution interface {
+	Distribution
+
+	ConvertFromOtel(dp pmetric.ExponentialHistogramDataPoint, unit string)
+}
```
jefchien, line 56: Nice! This is a lot cleaner than the generics I had in mind.

metric/distribution/exph/exph.go
```diff
@@ -107,43 +113,60 @@ func (d *ExpHistogramDistribution) ValuesAndCounts() ([]float64, []float64) {
...
-func (d *ExpHistogramDistribution) AddDistribution(from *ExpHistogramDistribution) {
+// weight is 1/samplingRate
+func (d *ExpHistogramDistribution) AddEntry(value float64, weight float64) error {
+	return d.AddEntryWithUnit(value, weight, "")
+}
+
+// weight is 1/samplingRate
+func (d *ExpHistogramDistribution) AddEntryWithUnit(value float64, weight float64, unit string) error {
+	return nil
+}
```
jefchien: nit: If these functions aren't implemented, would it make sense to move them out of the `Distribution` interface and into the `ClassicDistribution`?

dricross: I don't think we can do that as the `Distribution` interface used by the aggregator actually uses those functions when the incoming metric isn't already a distribution.

jefchien (approved):

lisguo (approved):

PR #1682, opened 2025-05-09
[processor/awsentity] Add configurable entity transform option 

zhihonl:
# Description of the issue
We need a way to send metrics with entity as UserConfiguration. We can do this by adding support for entity attribute overrides in the AWS Entity processor. This enhancement allows for manual override of service attributes through configuration, providing more flexibility in service naming and attribute management.

# Description of changes
- Implemented new `EntityOverrider` component that can override resource attributes based on configuration
- Added configuration support for key attributes and additional attributes overriding
- Override only occurs when not instrumentation

# License

# Tests
Unit tests

### Sample YAML
```
exporters:
    debug:
        verbosity: detailed
    awscloudwatch:
        force_flush_interval: 1m0s
        max_datums_per_call: 1000
        max_values_per_datum: 150
        middleware: agenthealth/metrics
        namespace: CWAgent
        region: us-west-2
        resource_to_telemetry_conversion:
            enabled: true
extensions:
    agenthealth/metrics:
        is_usage_data_enabled: true
        stats:
            operations:
                - PutMetricData
            usage_flags:
                mode: EC2
                region_type: EC2M
    agenthealth/statuscode:
        is_status_code_enabled: true
        is_usage_data_enabled: true
        stats:
            usage_flags:
                mode: EC2
                region_type: EC2M
    entitystore:
        mode: ec2
        region: us-west-2
processors:
    awsentity/service/otlp:
        entity_type: Service
        platform: ec2
        transform_entity:
            key_attributes:
                - key: Name
                  value: config-service-nam
                - key: Environment
                  value: config-environment
            attributes:
                - key: AWS.ServiceNameSource
                  value: UserConfiguration
    cumulativetodelta/hostOtlpMetrics:
        exclude:
            match_type: ""
        include:
            match_type: ""
        initial_value: 2
        max_staleness: 0s
receivers:
    otlp/metrics:
        protocols:
            grpc:
                dialer:
                    timeout: 0s
                endpoint: 127.0.0.1:4317
                include_metadata: false
                max_concurrent_streams: 0
                max_recv_msg_size_mib: 0
                read_buffer_size: 524288
                transport: tcp
                write_buffer_size: 0
            http:
                endpoint: 127.0.0.1:4318
                idle_timeout: 0s
                include_metadata: false
                logs_url_path: /v1/logs
                max_request_body_size: 0
                metrics_url_path: /v1/metrics
                read_header_timeout: 0s
                read_timeout: 0s
                traces_url_path: /v1/traces
                write_timeout: 0s
service:
    extensions:
        - agenthealth/metrics
        - agenthealth/statuscode
        - entitystore
    pipelines:
        metrics/hostOtlpMetrics:
            exporters:
                - debug
                - awscloudwatch
            processors:
                - cumulativetodelta/hostOtlpMetrics
                - awsentity/service/otlp
            receivers:
                - otlp/metrics
    telemetry:
        logs:
            development: false
            disable_caller: false
            disable_stacktrace: false
            encoding: console
            level: debug
            output_paths:
                - /opt/aws/amazon-cloudwatch-agent/logs/amazon-cloudwatch-agent.log
            sampling:
                enabled: true
                initial: 2
                thereafter: 500
                tick: 10s
        metrics:
            address: ""
            level: None
        traces:
            level: None
```

### ListEntitiesForMetric Output
```
{"Entities":[{"__type":"com.amazonaws.observability#Entity","Attributes":{"AWS.ServiceNameSource":"UserConfiguration"},"KeyAttributes":{"Environment":"config-environment","Type":"Service","Name":"config-service-nam"}}]}
```

### UI Snapshot
Screenshot 2025-05-19 at 2 46 14 PM

### Sample YAML
Same as Scenario 1

### ListEntitiesForMetric Output
```
{"Entities":[{"__type":"com.amazonaws.observability#Entity","Attributes":{"AWS.ServiceNameSource":"Instrumentation"},"KeyAttributes":{"Environment":"instrumented-environment","Type":"Service","Name":"instrumented-service"}}]}
```

### UI Snapshot
Screenshot 2025-05-19 at 2 51 14 PM

# Requirements
2. Run `make lint`


musa-asad (approved):

plugins/processors/awsentity/config.go
```diff
@@ -23,11 +28,34 @@ type Config struct {
 	// EntityType determines the type of entity processing done for
 	// telemetry. Possible values are Service and Resource
 	EntityType string `mapstructure:"entity_type,omitempty"`
+	// OverrideEntity contains configuration for overriding entity attributes
+	OverrideEntity *entityoverrider.EntityOverride `mapstructure:"override_entity,omitempty"`
```
lisguo, line 32: its kind of misleading to say its an override when the `service.name` otel attributes still take precedence. 

It might be fine to just say "EntityConfiguration" so that the user knows this is for "UserConfiguration"

plugins/processors/awsentity/entityattributes/entityattributes.go
```diff
@@ -89,6 +89,62 @@ var attributeEntityToShortNameMap = map[string]string{
...
+	Workload:          AttributeEntityWorkload,
+	Node:              AttributeEntityNode,
+	Platform:          AttributeEntityPlatformType,
+	InstanceID:        AttributeEntityInstanceID,
+	AutoscalingGroup:  AttributeEntityAutoScalingGroup,
+	ServiceNameSource: AttributeEntityServiceNameSource,
+}
+
+// GetFullAttributeName returns the full attribute name for a given short name
+func GetFullAttributeName(shortName string) (string, bool) {
```
lisguo, line 114: Instead of doing all these conversions -- is it worth it to just use the full attribute names directly in the configuration? Our translator can do the work of converting the user input into the full `awsentity.internal...` attributes. This way we don't have to convert during runtime

plugins/processors/awsentity/internal/entityoverrider/entityoverrider.go
```diff
@@ -0,0 +1,80 @@
...
+
+	var serviceName, source string
+	for _, keyAttr := range p.overrides.KeyAttributes {
+		if keyAttr.Key == entityattributes.ServiceName {
+			serviceName = keyAttr.Value
+		}
+	}
+
+	for _, attr := range p.overrides.Attributes {
+		if attr.Key == entityattributes.ServiceNameSource {
```
lisguo, line 74: Don't we want this to always be "UserConfiguration" We probably shouldn't expose `ServiceNameSource` to be overridden

plugins/processors/awsentity/processor.go
```diff
@@ -126,16 +127,18 @@ var getPodMeta = func(ctx context.Context) k8sclient.PodMetadata {
 // deployment.environment resource attributes set, then adds the association between the log group(s) and the
 // service/environment names to the entitystore extension.
 type awsEntityProcessor struct {
-	config     *Config
-	k8sscraper scraper
-	logger     *zap.Logger
+	config         *Config
+	k8sscraper     scraper
+	entityReplacer *entityoverrider.EntityOverrider
```
lisguo, line 132: nit: make variable name consistent with the object name

plugins/processors/awsentity/processor.go
```diff
@@ -305,6 +308,13 @@ func (p *awsEntityProcessor) processMetrics(ctx context.Context, md pmetric.Metr
 					AddAttributeIfNonEmpty(resourceAttrs, entityattributes.AttributeEntityInstanceID, ec2Attributes.InstanceId)
 					AddAttributeIfNonEmpty(resourceAttrs, entityattributes.AttributeEntityAutoScalingGroup, ec2Attributes.AutoScalingGroup)
 					AddAttributeIfNonEmpty(resourceAttrs, entityattributes.AttributeEntityServiceNameSource, ec2Attributes.ServiceNameSource)
+					if ec2Attributes.ServiceNameSource != entitystore.ServiceNameSourceInstrumentation {
```
lisguo, line 311: If the `service.name` is set to `unknown_service` will we apply an override? Feel like we should

plugins/processors/awsentity/entityattributes/entityattributes.go
musa-asad: Yeah, I prefer the shorter name in the config for simplicity.

plugins/processors/awsentity/internal/entityoverrider/entityoverrider.go
zhihonl: I think it's more natural to keep the code generic this way. Then in our translator we can add logic to configure key `AWS.ServiceNameSource` and value `UserConfiguration`.

plugins/processors/awsentity/processor.go
zhihonl: Will fix in new commit

plugins/processors/awsentity/processor.go
zhihonl: I don't think we will ever hit that case because of this line: https://github.com/aws/amazon-cloudwatch-agent/blob/84254e3c299bbc0cc460067ee418bc7d60851f1d/plugins/processors/awsentity/processor.go#L279

musa-asad: Maybe a comment to make it clear?

musa-asad (dismissed):

internal/entity/entity.go
```diff
@@ -0,0 +1,16 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+package entity
+
+// KeyPair represents a key-value pair for entity attributes
+type KeyPair struct {
+	Key   string `mapstructure:"key"`
```
zhihonl, line 8: i had to move these to their own package because they are referenced in multiple packages and it's better to make it public.

If we place them in `entitytransformer.go`, translator package won't have access to them due to it being internal to `awsentity` package. If we move to `config.go`, it will cause cyclic dependency between `entitytransformer` and `awsentity` package. Alternative is move `entitytransformer` out of internal folder for `awsentity` package but it ruins the encapsulation

plugins/processors/awsentity/processor.go
lisguo: So we are confident that at this point that if ServiceNameSource=Instrumentation, then the `service.name` is legitimately populated?

musa-asad (approved):

plugins/processors/awsentity/processor.go
zhihonl: > So we are confident that at this point that if ServiceNameSource=Instrumentation, then the service.name is legitimately populated?

That's right

plugins/processors/awsentity/entityattributes/entityattributes.go
```diff
@@ -89,6 +89,62 @@ var attributeEntityToShortNameMap = map[string]string{
...
+	if fullName, ok := attributeEntityToLongNameMap[shortName]; ok {
+		return fullName, true
+	}
+	return "", false
+}
+
+// IsAllowedKeyAttribute checks if the given key is an allowed entity key attribute name
+func IsAllowedKeyAttribute(key string) bool {
+	// Check if the key matches any of the values in keyAttributeEntityToShortNameMap
+	for _, shortName := range keyAttributeEntityToShortNameMap {
```
lisguo, line 129: nit: instead of iterating over the values of the map, can we use `keyAttributeEntityToLongNameMap` whiich is a reverse map of `keyAttributeEntityToShortNameMap` to index on the key? 

Otherwise we can use `util.Set` or something

plugins/processors/awsentity/entityattributes/entityattributes.go
```diff
@@ -89,6 +89,62 @@ var attributeEntityToShortNameMap = map[string]string{
...
+			return true
+		}
+	}
+	return false
+}
+
+// IsAllowedAttribute checks if the given key is an allowed attribute name
+func IsAllowedAttribute(key string) bool {
+	// Check if the key matches any of the values in attributeEntityToShortNameMap
+	for _, shortName := range attributeEntityToShortNameMap {
```
lisguo, line 140: same comment as above

lisguo (dismissed):

plugins/processors/awsentity/entityattributes/entityattributes.go
zhihonl: oh good point, re-using existing map is better. will fix

lisguo (approved):

musa-asad (approved):

PR #1701, opened 2025-05-23
Centralize log state management into shared package

jefchien:
# Description of the issue
Both the log tailing input plugin ([logfile](https://github.com/aws/amazon-cloudwatch-agent/blob/93ae77359bc5cfce19a1869de33f31e1cd56ae48/plugins/inputs/logfile/README.md)) and the windows log events input plugin ([windows_event_log](https://github.com/aws/amazon-cloudwatch-agent/tree/93ae77359bc5cfce19a1869de33f31e1cd56ae48/plugins/inputs/windows_event_log)) maintain state files. The plugins create one state file for each tailed log file and each subscribed windows event respectively. The state files serve as a way to keep track of progress (byte offset) and a recovery mechanism to restart from the last known position if the agent restarts. This helps to maintain log integrity and prevent duplicate processing of already sent logs.

Since these are maintained in two separate packages, there is overlap in how the behavior is defined and makes it more difficult to maintain or make changes to.

# Description of changes
Centralizes log state management into a single shared package `state` which has a `FileOffsetManager` that manages the update/save loop for the state file.

# License

# Tests
Added unit tests and update existing unit tests to use state manager package.

# Requirements
2. Run `make lint`

### DIFF
--- a/internal/state/offset.go
+++ b/internal/state/offset.go
@@ -1,0 +1,211 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+package state
+
+import (
+	"bytes"
+	"encoding"
+	"errors"
+	"log"
+	"math"
+	"os"
+	"strconv"
+	"time"
+)
+
+const (
+	// defaultSaveInterval is the default duration between state file saves
+	defaultSaveInterval = 100 * time.Millisecond
+	// defaultQueueSize is the default capacity of the offset queue
+	defaultQueueSize = 2000
+)
+
+// FileOffset represents a position within a file. It handles truncation detection through
+// sequence numbers.
+type FileOffset struct {
+	// seq handles file truncation, when file is truncated, we increase the seq
+	seq    uint64
+	offset uint64
+}
+
+var _ encoding.TextMarshaler = (*FileOffset)(nil)
+var _ encoding.TextUnmarshaler = (*FileOffset)(nil)
+var _ Comparator[FileOffset] = (*FileOffset)(nil)
+
+// NewFileOffset creates a new FileOffset with the specified value.
+func NewFileOffset(v uint64) FileOffset {
+	return FileOffset{offset: v}
+}
+
+// Set updates the offset. Increments the sequence number when a smaller offset is given to indicate possible
+// file truncation.
+func (o *FileOffset) Set(v uint64) {
+	if v < o.offset {
+		o.seq++
+	}
+	o.offset = v
+}
+
+// SetInt64 updates the offset. Ignores negative values.
+func (o *FileOffset) SetInt64(v int64) {
+	if v >= 0 {
+		o.Set(uint64(v))
+	}
+}
+
+// Get returns the offset.
+func (o FileOffset) Get() uint64 {
+	return o.offset
+}
+
+// GetInt64 returns the offset. Returns 0 if offset exceeds MaxInt64.
+func (o FileOffset) GetInt64() int64 {
+	if o.offset > math.MaxInt64 {
+		return 0
+	}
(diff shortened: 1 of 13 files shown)


internal/state/offset.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+	return &fileOffsetManager{
+		name:          cfg.Name,
+		stateFilePath: cfg.StateFilePath(),
+		offsetCh:      make(chan FileOffset, cfg.QueueSize),
+		saveInterval:  cfg.SaveInterval,
+	}
+}
+
+// Enqueue the offset. Will drop the oldest in the queue if full.
```
duhminick, line 134: Should we add some type of log message for when the offset is dropped?

plugins/inputs/logfile/tailersrc.go
```diff
@@ -206,7 +190,8 @@ func (ts *tailerSrc) runTail() {
 	var init string
 	var msgBuf bytes.Buffer
 	var cnt int
-	fo := &fileOffset{}
+	//fo := &fileOffset{}
```
duhminick, line 193: nit: remove?

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
```diff
@@ -78,19 +76,19 @@ func NewEventLog(name string, levels []string, logGroupName, logStreamName, rend
...
 		resubscribeCh: make(chan struct{}),
 	}
 	return eventLog
 }
 
 func (w *windowsEventLog) Init() error {
-	go w.runSaveState()
-	w.eventOffset = w.loadState()
+	go w.stateManager.Run(state.Notification{Done: w.done})
+	offset, _ := w.stateManager.Restore()
```
duhminick, line 90: Just for my knowledge, if the Restore returns an error, then the offset should just be 0 -- so the error can be safely ignored?

internal/state/offset_test.go
```diff
@@ -0,0 +1,267 @@
...
+		offset := NewFileOffset(100)
+
+		data, err := offset.MarshalText()
+		assert.NoError(t, err)
+
+		var restored FileOffset
+		assert.NoError(t, restored.UnmarshalText(data))
+		assert.EqualValues(t, offset.Get(), restored.Get())
+	})
+	t.Run("Unmarshal/Invalid", func(t *testing.T) {
```
duhminick, line 49: Should we add a test case where the content of the file is `test\ninvalid`? Just so we can test the number parsing part when it's invalid in case of corrupted files

internal/state/offset.go
```diff
@@ -0,0 +1,212 @@
...
+// FileOffsetManager is a state manager that handles the FileOffset.
+type FileOffsetManager Manager[FileOffset]
+
+var _ FileOffsetManager = (*fileOffsetManager)(nil)
+
+func NewFileOffsetManager(cfg ManagerConfig) FileOffsetManager {
+	if cfg.QueueSize <= 0 {
+		cfg.QueueSize = defaultQueueSize
+	}
+	if cfg.SaveInterval <= 0 {
```
movence, line 123: any plan to make this configurable instead always using the default? maybe just remove from ManagerConfig for now?

internal/state/offset.go
```diff
@@ -0,0 +1,212 @@
...
+	"os"
+	"strconv"
+	"time"
+)
+
+const (
+	// defaultSaveInterval is the default duration between state file saves
+	defaultSaveInterval = 100 * time.Millisecond
+	// defaultQueueSize is the default capacity of the offset queue
+	defaultQueueSize = 2000
```
movence, line 21: do we need to keep this many in the queue? Is the purpose to reduce the likelihood of older items being dropped when the queue becomes full?

PR #1715, opened 2025-06-03
tjstark/github pr ruleset

TravisStark:
- **Create ruleset to enforce integration tests on PR merge**
- **add branch rule**

# Description of the issue

# Description of changes

# License

# Tests

# Requirements
2. Run `make lint`

### DIFF
--- a/.github/rulesets/rulesets.json
+++ b/.github/rulesets/rulesets.json
@@ -19,6 +19,29 @@
         },
         "allow_force_pushes": false
       }
+    },
+    {
+      "type": "branch",
+      "name": "main branch protection",
+      "target": [
+        "main",
+        "tjstark/github-pr-ruleset"
+      ],
+      "enforcement": "active",
+      "conditions": {
+        "ref_name": {
+          "include": ["refs/heads/main"],
+          "exclude": []
+        }
+      },
+      "rules": {
+        "required_status_checks": {
+          "strict": true,
+          "contexts": ["PR-test"]
+        },
+        "allow_force_pushes": false,
+        "allow_deletions": false
+      }
     }
   ]
 }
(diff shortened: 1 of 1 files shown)


.github/workflows/PR-test.yml
```diff
@@ -212,3 +212,61 @@ jobs:
...
+
+          echo "Successfully validated jobs:"
+          if [ ${#successful_jobs[@]} -eq 0 ]; then
+            echo "None"
+          else
+            printf '%s\n' "${successful_jobs[@]}"
+          fi
+
+          echo -e "\nSkipped jobs:"
+          if [ ${#skipped_jobs[@]} -eq 0 ]; then
```
agarakan: Curious why we allow skipping for PR test but not PR builld

.github/workflows/PR-build.yml
```diff
@@ -158,3 +158,41 @@ jobs:
...
+  verify-all:
+    name: Verify All PR Build Jobs
+    needs: [changes, lint, build, test-data-race]
+    runs-on: ubuntu-latest
+    if: always()
+    steps:
+      - name: Check Job Status
+        run: |
+          declare -A job_statuses=(
+            ["changes"]="${{ needs.changes.result }}"
```
agarakan: Can this be changed into a for loop to make it reusable? It seems like the rest of the validation is identical here and in the other file

TravisStark: I'll try that out.

.github/workflows/PR-test.yml
TravisStark: Good catch, added that for now because some of the tests are unreliable at the moment.  That can be removed in a follow-up PR.

agarakan (dismissed):
All comments minor, logic looks good. approved

agarakan (approved):
Looks clean, maybe a future iteration can be about consolidating these into a common function that both workflows invoke.

Paramadon (approved):

PR #1717, opened 2025-06-04
Batch DescribeLogGroups calls

duhminick:
# Description of the issue

# Description of changes

# License

# Tests

# Requirements
2. Run `make lint`


plugins/outputs/cloudwatchlogs/internal/pusher/target.go
```diff
@@ -174,44 +176,55 @@ func (m *targetManager) createLogStream(t Target) error {
...
 	}
-
-	output, err := m.service.DescribeLogGroups(input)
-	if err != nil {
-		return 0, fmt.Errorf("describe log groups failed: %w", err)
+	describeLogGroupsInput := &cloudwatchlogs.DescribeLogGroupsInput{
+		LogGroupIdentifiers: identifiers,
 	}
+	for attempt := 0; attempt < numBackoffRetries; attempt++ {
+		output, err := m.service.DescribeLogGroups(describeLogGroupsInput)
```
movence, line 212: I would assume this call is all-or-nothing with no partial results?

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
```diff
@@ -52,8 +54,8 @@ func NewTargetManager(logger telegraf.Logger, service cloudWatchLogsService) Tar
 		service:  service,
 		cache:    make(map[Target]time.Time),
 		cacheTTL: cacheTTL,
-		dlg:      make(chan Target, retentionChannelSize),
-		prp:      make(chan Target, retentionChannelSize),
+		dlg:      make(chan Target, describeLogGroupChannelSize),
```
movence: I don't think we need to limit the channel size to same size of batch api call limit

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
duhminick: It works with partial results :)

```
$ aws logs describe-log-groups --log-group-identifiers "log.txt" "notarealloggroup"
```

works with log.txt in the result

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
duhminick: Updated 👍

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
```diff
@@ -174,44 +175,55 @@ func (m *targetManager) createLogStream(t Target) error {
...
+			continue
+		}
 
-	for _, group := range output.LogGroups {
-		if *group.LogGroupName == target.Group {
-			if group.RetentionInDays == nil {
-				return 0, nil
+		for _, logGroups := range output.LogGroups {
+			target := targets[*logGroups.LogGroupName]
+			if target.Retention != int(*logGroups.RetentionInDays) && target.Retention > 0 {
```
jefchien: What does `DescribeLogGroups` return in the `RetentionInDays` field if retention is not set? Is it nil? Will this panic?

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
```diff
@@ -174,44 +175,55 @@ func (m *targetManager) createLogStream(t Target) error {
...
-			break // no change in retention
 		}
 	}
 }
 
-func (m *targetManager) getRetention(target Target) (int, error) {
-	input := &cloudwatchlogs.DescribeLogGroupsInput{
-		LogGroupNamePrefix: aws.String(target.Group),
+func (m *targetManager) updateTargetBatch(targets map[string]Target) {
+	var identifiers []*string
```
jefchien: nit: Can pre-allocate the size since you know the number of targets.
```suggestion
	identifiers := make([]*string, 0, len(targets))
```

plugins/outputs/cloudwatchlogs/internal/pusher/target_test.go
```diff
@@ -344,6 +254,147 @@ func TestTargetManager(t *testing.T) {
...
+		mockService.On("DescribeLogGroups", mock.MatchedBy(func(input *cloudwatchlogs.DescribeLogGroupsInput) bool {
+			return len(input.LogGroupIdentifiers) == logGroupIdentifierLimit
+		})).Return(&cloudwatchlogs.DescribeLogGroupsOutput{
+			LogGroups: []*cloudwatchlogs.LogGroup{},
+		}, nil).Once()
+
+		manager := NewTargetManager(logger, mockService)
+		tm := manager.(*targetManager)
+
+		for i := 0; i < logGroupIdentifierLimit; i++ {
```
jefchien, line 368: This only goes up to the limit. Can you have it exceed the limit to make sure this is splitting at 50.

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
duhminick: Great catch, the value is indeed nil. It doesn't panic though

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
```diff
@@ -174,44 +175,55 @@ func (m *targetManager) createLogStream(t Target) error {
...
+			continue
+		}
 
-	for _, group := range output.LogGroups {
-		if *group.LogGroupName == target.Group {
-			if group.RetentionInDays == nil {
-				return 0, nil
+		for _, logGroups := range output.LogGroups {
+			target := targets[*logGroups.LogGroupName]
+			if (logGroups.RetentionInDays == nil || target.Retention != int(*logGroups.RetentionInDays)) && target.Retention > 0 {
```
jefchien, line 221: If we're ignoring all `target.Retention` set to -1, should we just not even include it in our lookup?

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
```diff
@@ -174,44 +175,58 @@ func (m *targetManager) createLogStream(t Target) error {
...
 }
 
-func (m *targetManager) getRetention(target Target) (int, error) {
-	input := &cloudwatchlogs.DescribeLogGroupsInput{
-		LogGroupNamePrefix: aws.String(target.Group),
+func (m *targetManager) updateTargetBatch(targets map[string]Target) {
+	identifiers := make([]*string, 0, len(targets))
+	for logGroup, target := range targets {
+		// It's documented that only values >= 1 are accepted by the agent config
+		if target.Retention >= 1 {
```
jefchien: It looks like it already is ignored in the places that `m.dlg <- target` gets called.

https://github.com/aws/amazon-cloudwatch-agent/blob/6f37decc741c5932359eff4a32e1028a6718ac54/plugins/outputs/cloudwatchlogs/internal/pusher/target.go#L76-L84

https://github.com/aws/amazon-cloudwatch-agent/blob/6f37decc741c5932359eff4a32e1028a6718ac54/plugins/outputs/cloudwatchlogs/internal/pusher/target.go#L90-L96

So we don't need to add another check. Everything that gets to this point should have a retention policy > 0. My bad.

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
jefchien: Do we need to handle pagination of the results?

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
duhminick: Np. I was thinking about that after I logged off last night too. Should've checked in the first place.

duhminick: https://github.com/aws/amazon-cloudwatch-agent/blob/6f37decc741c5932359eff4a32e1028a6718ac54/sdk/service/cloudwatchlogs/api.go#L15286-L15288

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
duhminick: Technically, no. The default limit is 50. I can add that explicitly to make it clear

https://github.com/aws/amazon-cloudwatch-agent/blob/6f37decc741c5932359eff4a32e1028a6718ac54/sdk/service/cloudwatchlogs/api.go#L15286-L15288

plugins/outputs/cloudwatchlogs/internal/pusher/target_test.go
```diff
@@ -120,97 +121,6 @@ func TestTargetManager(t *testing.T) {
 		assertCacheLen(t, manager, 0)
 	})
 
-	t.Run("SetRetentionPolicy", func(t *testing.T) {
```
jefchien, line 123: Why did we delete these? Are they fully covered by the new tests?

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
```diff
@@ -174,44 +175,56 @@ func (m *targetManager) createLogStream(t Target) error {
...
 
 func (m *targetManager) processDescribeLogGroup() {
-	for target := range m.dlg {
-		for attempt := 0; attempt < numBackoffRetries; attempt++ {
-			currentRetention, err := m.getRetention(target)
-			if err != nil {
-				m.logger.Errorf("failed to describe log group retention for target %v: %v", target, err)
-				time.Sleep(m.calculateBackoff(attempt))
-				continue
+	t := time.NewTicker(5 * time.Second)
```
jefchien, line 178: I'm sure it's fine, but what's the reasoning behind the ticker vs timer. Are we anticipating`DescribeLogGroup` calls minutes after start up?

plugins/outputs/cloudwatchlogs/internal/pusher/target_test.go
```diff
@@ -344,6 +254,181 @@ func TestTargetManager(t *testing.T) {
 	})
 }
 
+func TestDescribeLogGroupsBatching(t *testing.T) {
```
jefchien, line 352: nit: Can we run these in parallel? The multiple 5+ second sleeps make this seem like a long test.

plugins/outputs/cloudwatchlogs/internal/pusher/target_test.go
```diff
@@ -344,6 +254,181 @@ func TestTargetManager(t *testing.T) {
...
+		manager := NewTargetManager(logger, mockService)
+		tm := manager.(*targetManager)
+
+		for i := 0; i < logGroupIdentifierLimit; i++ {
+			target := Target{
+				Group:     fmt.Sprintf("group-%d", i),
+				Stream:    "stream",
+				Retention: 7,
+			}
+			tm.dlg <- target
```
jefchien: nit: If possible, I would prefer if these tests used the exposed functions.

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
duhminick: I think it is possible, but maybe this covers the scenario I'm thinking of: https://github.com/aws/amazon-cloudwatch-agent/blob/47683ec2f0f1c8a624a707f43d36608be933d9d6/logs/logs.go#L174


Another scenario - my thinking is that it's not too safe to assume that it will only need to be called once with the timer. The system could be slow to initialize the targets so having this on a timer could potentially miss those log groups

plugins/outputs/cloudwatchlogs/internal/pusher/target_test.go
duhminick: I kind of preferred to do it this way to isolate the logic since InitTarget is the only way we expose this channel which brings in CreateLogStream, and PutRetentionPolicy

jefchien: That's not quite true. The `PutRetentionPolicy` is the main way we reach this channel.

https://github.com/aws/amazon-cloudwatch-agent/blob/47683ec2f0f1c8a624a707f43d36608be933d9d6/plugins/outputs/cloudwatchlogs/internal/pusher/target.go#L90-L96

and I would rather this call that function in case we ever make changes to `PutRetentionPolicy`. I want to know that this will cover it.

jefchien (approved):

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
```diff
@@ -16,8 +16,9 @@ import (
 )
 
 const (
-	cacheTTL             = 5 * time.Second
-	retentionChannelSize = 100
+	retentionChannelSize    = 100
+	cacheTTL                = 5 * time.Second
+	logGroupIdentifierLimit = 50
```
Paramadon, line 21: Why 50? I'm guessing this is the api limi?

Paramadon (approved):

plugins/outputs/cloudwatchlogs/internal/pusher/target.go
duhminick: Yep, it's the API limit :/

PR #1718, opened 2025-06-05
Add support for range state management

jefchien:
# Description of the issue
Currently, the state file management stores the max offset. With the introduction of multi-threaded log publishers (https://github.com/aws/amazon-cloudwatch-agent/pull/1499), if multi-threading is enabled, each chunk of a tailed file can finish sending out of order. If the chunks non-sequential, then the stored offset could advance ahead of unsent logs leading to data loss on restart.

# Description of changes
Replaces the offset state manager with a ranged one, which accepts ranges and merges them if they are adjacent or overlapping. Uses `github.com/google/btree` as the backing store of the ranges. The store is bounded and will collapse/merge the smallest two ranges if it grows beyond the specified capacity. At the moment, these are both configured to 1 to maintain current behavior.

## State file format change

Old
```
1030
/tmp/test.log
```
New
```
1030
/tmp/test.log
0-900,970-1030
```

By keeping the max offset and file name in that order, the generated state file is still backwards compatible.

## Merging behavior
The merge is done on each insert if overlapping or adjacent ranges are found.

### Example 1: Adjacent ranges
Suppose the state contains
```
0-100,200-250
```
If a new range `100-150` is put in the queue, the state manager will merge it with `0-100`. Results in
```
0-150,200-250
```

### Example 2: Overlapping ranges
This is less likely to happen because the log plugins should only process distinct chunks.

Suppose the state contains
```
0-100,200-250
```
If a new range `50-200` is put in the queue, the state manager will merge it with `0-100` and `200-250`. Results in
```
0-250
```

### Example 3: Non-overlapping ranges
Suppose the state contains
```
0-100,200-250
```
If a new range `125-175` is put in the queue, it will not merge any of them. Results in
```
0-100,125-175,200-250
```

If the `maxPersistItems` is less than the number of ranges stored in the tree, it will collapse/merge them. In this case if `maxPersistItems` was 2, this would result in
```
0-175,200-250
```
even though there was a gap from `100-125`. This capacity limit prevents the tree from growing unbounded at the expense of dropping tracking of older gaps.

# License

# Tests
Added and updated unit tests. Integration tests (`EC2LinuxTests / al2:log_state_logfile_test`) show that the single threaded case still works.

# Requirements
2. Run `make lint`

### DIFF
--- a/internal/state/manager.go
+++ b/internal/state/manager.go
@@ -99,25 +99,25 @@ func (m *rangeManager) Run(notification Notification) {
 	defer t.Stop()
 
 	var lastSeq uint64
-	current := newRangeTracker(m.name, m.maxPersistItems)
+	currentTracker := newRangeTracker(m.name, m.maxPersistItems)
 	shouldSave := false
 	for {
 		select {
-		case replace := <-m.replaceTrackerCh:
-			current = replace
+		case replaceTracker := <-m.replaceTrackerCh:
+			currentTracker = replaceTracker
 		case item := <-m.queue:
 			// truncation detected, clear tree
 			if item.seq > lastSeq {
 				lastSeq = item.seq
-				current.Clear()
+				currentTracker.Clear()
 			}
-			changed := current.Insert(item)
+			changed := currentTracker.Insert(item)
 			shouldSave = shouldSave || changed
 		case <-t.C:
 			if !shouldSave {
 				continue
 			}
-			if err := m.save(current); err != nil {
+			if err := m.save(currentTracker); err != nil {
 				log.Printf("E! Error happened when saving state file (%s): %v", m.stateFilePath, err)
 				continue
 			}
@@ -129,7 +129,7 @@ func (m *rangeManager) Run(notification Notification) {
 			}
 			return
 		case <-notification.Done:
-			if err := m.save(current); err != nil {
+			if err := m.save(currentTracker); err != nil {
 				log.Printf("E! Error happened during final state file (%s) save, duplicate log maybe sent at next start: %v", m.stateFilePath, err)
 			}
 			return
--- a/internal/state/range.go
+++ b/internal/state/range.go
@@ -107,14 +107,19 @@ func (r Range) Contains(other Range) bool {
 
 // String returns a string representation of the range "start-end". If the end is unbounded, returns "start-".
 func (r Range) String() string {
-	if r.IsEndOffsetUnbounded() {
-		return fmt.Sprintf("%v-", r.start)
-	}
-	return fmt.Sprintf("%d-%d", r.start, r.end)
+	b, _ := r.MarshalText()
+	return string(b)
 }
 
+// MarshalText serializes the range into "start-end". If the end is unbounded, the format is "start-".
 func (r Range) MarshalText() ([]byte, error) {
-	return []byte(r.String()), nil
+	var buf bytes.Buffer
+	buf.WriteString(strconv.FormatUint(r.start, 10))
+	buf.WriteByte('-')
+	if !r.IsEndOffsetUnbounded() {
+		buf.WriteString(strconv.FormatUint(r.end, 10))
+	}
+	return buf.Bytes(), nil
 }
 
 // UnmarshalText supports unmarshalling both the "start-end" and "start-" formats.
@@ -161,11 +166,26 @@ func (r RangeList) OnlyUseMaxOffset() bool {
 	return len(r) == 0 || (len(r) == 1 && r[0].StartOffset() == 0)
 }
 
+// String returns a string representation of all stored ranges (e.g. "[0-5,10-15]").
+func (r RangeList) String() string {
+	var builder strings.Builder
+	first := true
+	builder.WriteByte('[')
+	for _, item := range r {
+		if !first {
+			builder.WriteByte(',')
+		}
+		first = false
+		builder.WriteString(item.String())
+	}
+	builder.WriteByte(']')
+	return builder.String()
+}
+
 // RangeTracker manages a collection of ranges. Handles insertion, retrieval, and serialization.
 type RangeTracker interface {
 	encoding.TextMarshaler
 	encoding.TextUnmarshaler
-	fmt.Stringer
 	// Insert a Range into the store. Returns false if the Range is already contained by another Range in the store or is
 	// invalid.
 	Insert(Range) bool
@@ -177,29 +197,31 @@ type RangeTracker interface {
 	Clear()
 }
 
-// newRangeTracker creates a RangeTracker based on the capacity. If the capacity is 1, it creates a maxOffsetTracker.
+// newRangeTracker creates a RangeTracker based on the capacity. If the capacity is 1, it creates a singleRangeTracker.
 // Otherwise, it will create a multiRangeTracker configured with the capacity.
 func newRangeTracker(name string, capacity int) RangeTracker {
 	if capacity == 1 {
-		return newMaxOffsetTracker(name)
+		return newSingleRangeTracker(name)
 	}
 	return newMultiRangeTrackerWithCap(name, capacity)
 }
 
-// maxOffsetTracker only keeps track of the maximum offset. It stores a single range starting at 0 and ending at the
+// singleRangeTracker only keeps track of the maximum offset. It stores a single range starting at 0 and ending at the
 // maximum observed offset.
-type maxOffsetTracker struct {
+type singleRangeTracker struct {
 	name string
 	r    Range
 }
 
-var _ RangeTracker = (*maxOffsetTracker)(nil)
+var _ RangeTracker = (*singleRangeTracker)(nil)
 
-func newMaxOffsetTracker(name string) RangeTracker {
-	return &maxOffsetTracker{name: name}
+func newSingleRangeTracker(name string) RangeTracker {
+	return &singleRangeTracker{name: name}
 }
 
-func (t *maxOffsetTracker) Insert(r Range) bool {
+// Insert updates the tracked maximum offset if the provided Range has a larger end or sequence number. Returns false
+// if the tracker was not updated by the insert.
+func (t *singleRangeTracker) Insert(r Range) bool {
 	if !r.IsValid() {
 		return false
 	}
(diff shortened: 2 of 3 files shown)


go.mod
jefchien, line 1: The only addition is `github.com/google/btree`. All other changes are from merging in the other direct require block.

internal/state/manager.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+	if cfg.SaveInterval <= 0 {
+		cfg.SaveInterval = defaultSaveInterval
+	}
+	return &rangeManager{
+		name:            cfg.Name,
+		stateFilePath:   cfg.StateFilePath(),
+		queue:           make(chan Range, cfg.QueueSize),
+		saveInterval:    cfg.SaveInterval,
+		maxPersistItems: cfg.MaxPersistItems,
```
jefchien: For the case where `maxPersistItems` is 1, I'm thinking about changing the logic so it doesn't actually insert it into the tree and just keeps track of the maxOffset. Should be more performant.

internal/state/manager.go
```diff
@@ -0,0 +1,138 @@
...
+	"log"
+	"os"
+	"time"
+)
+
+const (
+	// defaultSaveInterval is the default duration between state file saves
+	defaultSaveInterval = 100 * time.Millisecond
+	// defaultQueueSize is the default capacity of the offset queue
+	defaultQueueSize = 2000
```
okankoAMZ, line 17: we should comment why it is 2000

internal/state/manager.go
```diff
@@ -0,0 +1,138 @@
...
+	replaceTreeCh   chan *rangeTree
+}
+
+// FileRangeManager is a state manager that handles the Range.
+type FileRangeManager Manager[Range, RangeList]
+
+var _ FileRangeManager = (*rangeManager)(nil)
+
+func NewFileRangeManager(cfg ManagerConfig) FileRangeManager {
+	if cfg.QueueSize <= 0 {
```
okankoAMZ, line 35: so queue size is configurable? If so it is going to be stuck until translator changes

internal/state/manager.go
```diff
@@ -0,0 +1,138 @@
...
+		case <-t.C:
+			if !shouldSave {
+				continue
+			}
+			if err := m.save(current); err != nil {
+				log.Printf("E! Error happened when saving state file (%s): %v", m.stateFilePath, err)
+				continue
+			}
+			shouldSave = false
+		case <-notification.Delete:
```
okankoAMZ, line 125: What happens if state file is deleted while CloudWatch Agent is being ran?

internal/state/manager.go
jefchien: The queue size is not configurable through the agent JSON. It's mainly there because the windows event logs and logfile plugins had different queue sizes.

internal/state/manager.go
jefchien: Nothing. The state file is only used during agent start up. If the state file is deleted, but the file still exists, the agent will write a new state file in the next tick.

jefchien:
> Did you run a go --race test on it?

The tests for this package are included in the `Test data race` PR check.

```
ok  	github.com/aws/amazon-cloudwatch-agent/internal/state	1.821s
```

internal/state/manager.go
movence: this is inherited from the curent behavior so guess there wound't be clear why https://github.com/aws/amazon-cloudwatch-agent/pull/1701/files#r2116116009

internal/state/range_test.go
```diff
@@ -0,0 +1,352 @@
...
+		assert.True(t, tree.Insert(Range{start: 15, end: 20}))
+		assert.Equal(t, 2, tree.Len())
+		assert.Equal(t, "[0-10,15-30]", tree.String())
+	})
+	t.Run("Merge/Overlap/Single", func(t *testing.T) {
+		t.Parallel()
+		tree := newRangeTree("test")
+		assert.True(t, tree.Insert(Range{start: 0, end: 5}))
+		assert.True(t, tree.Insert(Range{start: 20, end: 30}))
+		assert.True(t, tree.Insert(Range{start: 15, end: 25}))
```
movence: maybe this is just a test scenario but is this possible? this basically means duplciated logs?

okankoAMZ (dismissed):

plugins/inputs/logfile/logfile.go
```diff
@@ -187,15 +187,16 @@ func (t *LogFile) FindLogSrc() []logs.LogSrc {
 				}
 			}
 
-			stateManager := state.NewFileOffsetManager(state.ManagerConfig{
-				StateFileDir: t.FileStateFolder,
-				Name:         filename,
+			stateManager := state.NewFileRangeManager(state.ManagerConfig{
+				StateFileDir:    t.FileStateFolder,
+				Name:            filename,
+				MaxPersistItems: 1, // TODO: Base this on the number of threads
```
movence, line 193: will this be worked on in another pr?

plugins/inputs/logfile/logfile_test.go
```diff
@@ -191,33 +190,6 @@ func TestCompressedFile(t *testing.T) {
...
-	// Test negative offset.
-	err = os.WriteFile(
-		state.FilePath(tmpfolder, logFileStateFileName),
-		[]byte(strconv.FormatInt(-8675, 10)+"\n"+logFilePath),
-		state.FileMode)
-	require.NoError(t, err)
-	roffset, err = m.Restore()
-	require.Error(t, err)
-	assert.Equal(t, uint64(0), roffset.Get(), fmt.Sprintf("The actual offset is %d, different from the expected offset %d.", roffset.Get(), 0))
-}
```
movence, line 219: why remove? is this just out of scope of logfile and more appropriate for the manager?

movence: maybe it's covered under `Restore/Valid/BackwardsCompatible`

internal/state/state.go
```diff
@@ -39,13 +32,22 @@ type Notification struct {
...
 	StateFilePrefix string
-	QueueSize       int
-	SaveInterval    time.Duration
+	// QueueSize determines the size of the internal buffer for pending state changes.
+	QueueSize int
+	// SaveInterval determines how often the state is persisted.
+	SaveInterval time.Duration
+	// MaxPersistItems is the maximum number of items to persist in the saved state. If zero or negative, the
+	// persistence is unbounded.
+	MaxPersistItems int
```
movence, line 47: vars with desc :+1:

plugins/inputs/logfile/logfile.go
jefchien: Yep. It'll be done in a follow up.

internal/state/range_test.go
jefchien: It's not possible unless our log input plugins were broken, but just wanted to show that these edge cases are handled.

plugins/inputs/logfile/logfile_test.go
jefchien: Yeah it's just out of scope for the logfile and ended up require access to functions that aren't accessible outside of the state manager package.

internal/state/range.go
```diff
@@ -0,0 +1,537 @@
...
+		t.collapseOldest()
+	}
+	return true
+}
+
+// collapseOldest takes the two oldest (i.e. smallest start) ranges and merges them.
+func (t *multiRangeTracker) collapseOldest() {
+	var first, second *Range
+	var count int
+	t.tree.Ascend(func(item Range) bool {
```
movence, line 384: check Len for early exit?

internal/state/range_test.go
```diff
@@ -142,29 +142,29 @@ func TestNewRangeTracker(t *testing.T) {
...
+func TestSingleRangeTracker_Insert(t *testing.T) {
 	t.Run("NonOverlapping", func(t *testing.T) {
 		t.Parallel()
-		tracker := newMaxOffsetTracker("test")
+		tracker := newSingleRangeTracker("test")
 		assert.True(t, tracker.Insert(Range{start: 0, end: 5}))
 		assert.True(t, tracker.Insert(Range{start: 20, end: 30}))
 		assert.Equal(t, 1, tracker.Len())
-		assert.Equal(t, "[0-30]", tracker.String())
+		assert.Equal(t, "[0-30]", tracker.Ranges().String())
```
movence, line 152: is this the same expected behavior? looks like we are assuming the missing log portion inbetween those 2 ranges will be pushed eventually? I think the previous behavior is the current offset doesn't get written to the state file unless it's processed for sure and sequentially.

internal/state/manager.go
```diff
@@ -0,0 +1,138 @@
...
+	default:
+		old := <-m.queue
+		log.Printf("D! Offset range queue is full for %s. Dropping oldest offset range: %s", m.stateFilePath, old)
+		m.queue <- item
+	}
+}
+
+// Restore the ranges if the state file exists.
+func (m *rangeManager) Restore() (RangeList, error) {
+	content, err := os.ReadFile(m.stateFilePath)
```
lisguo, line 64: will there ever be more than 1 manager? You mentioned above that we'd have one for windows log events vs app logs...do they have the same state file path? Is there a chance of a race condition?

internal/state/manager.go
```diff
@@ -0,0 +1,138 @@
...
+	if cfg.SaveInterval <= 0 {
+		cfg.SaveInterval = defaultSaveInterval
+	}
+	return &rangeManager{
+		name:             cfg.Name,
+		stateFilePath:    cfg.StateFilePath(),
+		queue:            make(chan Range, cfg.QueueSize),
+		saveInterval:     cfg.SaveInterval,
+		maxPersistItems:  cfg.MaxPersistItems,
+		replaceTrackerCh: make(chan RangeTracker, 1),
```
lisguo, line 47: Seems like the tracker only gets replaced when we restore from the state file correct? This realistically should only happen on restart? In that case do we really need this?

lisguo:
During your testing, have you seen any additional latency when restarting the agent with state files?

internal/state/range_test.go
jefchien: The previous behavior is that we just store the largest offset that we get. That's what we're trying to fix with the multi-range tracker changes.

internal/state/manager.go
jefchien: Each tailed file has its own. The state file path for logfile is based on the tailed log

https://github.com/aws/amazon-cloudwatch-agent/blob/5ef3dba446cb56a4c2306878592b5d14300ae82f/plugins/inputs/logfile/logfile.go#L190

while the windows log event is combination of the log group, stream, and windows event name

https://github.com/aws/amazon-cloudwatch-agent/blob/5ef3dba446cb56a4c2306878592b5d14300ae82f/plugins/inputs/windows_event_log/windows_event_log.go#L139-L144

internal/state/manager.go
jefchien: That's right. It's just to make sure that there's no possible race condition. It shouldn't be an issue, but it's guarding against it.

jefchien:
> During your testing, have you seen any additional latency when restarting the agent with state files?

Did not see anything like that. There will be another PR where we actually read the gaps and try to send them.

internal/state/manager.go
jefchien: They should be distinct enough to prevent race conditions. It's the same number of managers as before all of these changes.

lisguo (approved):

PR #1721, opened 2025-06-11
Restore and send gaps

duhminick:
# Description of the issue
With the addition of multi-threaded log sending support, there is a chance that certain ranges of logs were not sent. The state file will have been updated with ranges that were successfully sent. As such, the agent on recovery/startup should attempt to send the missing ranges.

For example given:
```
a
b
c
d
e
```

with ranges `0-1,3-4`, the missing gap is `2-3`. The agent will send the line `b` first then continue to tail the file at the offset.

# Description of changes
1. If gaps are found, then the given ranges will be "inverted". Meaning that the gaps in the ranges will be constructed. The gaps are then used in `tailFileSync()`.
2. `tailFileSync` will first check to see if the inverted ranges are present (i.e. len is greater than 0). It will iterate through the array of ranges
  a. In the range, it will seek to the beginning offset. It will then read and send lines until the `curOffset` is greater than or equal to the ending offset in the range. Once that is complete, it will move on to the next range.

The logic is largely similar to the existing loop for the tailer minus some error handling and EOF. That is because it's better to just let the main loop handle the errors.

# License

# Tests
1. Unit test
2. Existing integration tests
3. Tested locally

### Example

Given the state file with log.txt configured for log group log.txt
```
36
/home/ec2-user/log.txt
0-15,25-36
```

with log.txt having content
```
a
b
c
d
e
f
... (cont)
x
y
z
```

Resulted in:
**Notice that `[a-h], [o-s]` are missing**
image

# Requirements
6. Run `make lint`

-------
### Integration Tests

### DIFF
--- a/plugins/inputs/logfile/logfile_test.go
+++ b/plugins/inputs/logfile/logfile_test.go
@@ -700,6 +700,7 @@ func TestLogsFileWithRangeNoGaps(t *testing.T) {
 	tt.FileConfig = []FileConfig{{FilePath: tmpfile.Name(), FromBeginning: true}}
 	tt.FileConfig[0].init()
 	tt.started = true
+	tt.MaxPersistState = 2
 
 	lsrcs := tt.FindLogSrc()
 	if len(lsrcs) != 1 {
@@ -751,6 +752,7 @@ func TestLogsFileWithRangeGaps(t *testing.T) {
 	tt.FileConfig = []FileConfig{{FilePath: tmpfile.Name(), FromBeginning: true}}
 	tt.FileConfig[0].init()
 	tt.started = true
+	tt.MaxPersistState = 2
 
 	lsrcs := tt.FindLogSrc()
 	if len(lsrcs) != 1 {
@@ -809,6 +811,7 @@ func TestLogsFileWithEOFRangeGaps(t *testing.T) {
 	tt.FileConfig = []FileConfig{{FilePath: tmpfile.Name(), FromBeginning: true}}
 	tt.FileConfig[0].init()
 	tt.started = true
+	tt.MaxPersistState = 2
 
 	lsrcs := tt.FindLogSrc()
 	if len(lsrcs) != 1 {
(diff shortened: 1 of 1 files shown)


plugins/inputs/logfile/logfile.go
```diff
@@ -201,6 +201,10 @@ func (t *LogFile) FindLogSrc() []logs.LogSrc {
 				seekFile = &tail.SeekInfo{Whence: io.SeekEnd, Offset: 0}
 			}
 
+			var rangeList state.RangeList
+			if !restored.OnlyUseMaxOffset() {
```
Paramadon, line 207: So if the restored statei s not emtpy we invert the ranges. You mentioned that if we find the gaps we'll invert it and then construct, but I'm wondering how this finds the gaps exaclty...

duhminick: https://github.com/aws/amazon-cloudwatch-agent/blob/47683ec2f0f1c8a624a707f43d36608be933d9d6/internal/state/range.go#L528-L543

So inverting the ranges is the same as finding the gaps. I think this unit test might show case a good example:
https://github.com/aws/amazon-cloudwatch-agent/blob/47683ec2f0f1c8a624a707f43d36608be933d9d6/internal/state/range_test.go#L412-L429

So in the test, the ranges `5-10,20-25` were sent. That means that `0-5,10-20` still need to be sent before we begin the main loop of tailing the log file

Paramadon (dismissed):
Reviewed pr on call with duhminick@, approved!

plugins/inputs/logfile/logfile.go
```diff
@@ -201,6 +201,10 @@ func (t *LogFile) FindLogSrc() []logs.LogSrc {
 				seekFile = &tail.SeekInfo{Whence: io.SeekEnd, Offset: 0}
 			}
 
+			var rangeList state.RangeList
```
jefchien: nit: Recommend making the variable name more descriptive like `gapsToRead` or `pendingRanges` or `retryRanges` or something.

plugins/inputs/logfile/logfile.go
```diff
@@ -190,7 +190,7 @@ func (t *LogFile) FindLogSrc() []logs.LogSrc {
 			stateManager := state.NewFileRangeManager(state.ManagerConfig{
 				StateFileDir:    t.FileStateFolder,
 				Name:            filename,
-				MaxPersistItems: 1, // TODO: Base this on the number of threads
+				MaxPersistItems: 2, // TODO: Base this on the number of threads
```
jefchien: Going to have a merge conflict here with https://github.com/aws/amazon-cloudwatch-agent/pull/1724

jefchien (dismissed):

jefchien (approved):

sky333999 (approved):

PR #1723, opened 2025-06-12
Improve batching for range state

jefchien:
# Description of the issue
With the range state management added as part of https://github.com/aws/amazon-cloudwatch-agent/pull/1718, if multi-threading is enabled, the agent tries to keep track of distinct, non-overlapping ranges. The multi-range tracker will merge any overlapping ranges on insert. It was built with capacity limits and when capacity is exceeded, it will collapse the lowest ranges.

The current logic on successful send of a log event batch is to call all the `Done` callbacks for each individual log event in the batch. This enqueues each of the ranges onto the tracker for insertion/merging. With a maximum limit of 10000 events in a batch, this quickly becomes too much unnecessary work.

# Description of changes
Instead of sending each individual range, pre-merge the ranges in the batch and send them as a single range. The log tailing logic is still sequential/single-threaded, so the log events in a batch will be continuous and can be naively merged. This reduces the amount of work the tracker needs to do and helps to keep the number of tracked ranges below the the capacity threshold.

# License

# Tests
Added unit tests. Ran the changes manually with some additional logging in the `RangeQueueBatcher`.
```
	log.Printf("D! Done in RangeQueueBatcher: %s became %s", b.rs, b.r)
```

```
D! Done in RangeQueueBatcher: [19204063-19204117,19204117-19204183,19204183-19204237,19204237-19204254,19204254-19204290,19204290-19204597,19204597-19204619,19204619-19204925,19204925-19204949,19204949-19204991,19204991-19205043,19205043-19205073,19205073-19206413,19206413-19206455,19206480-19206481,19206481-19206535,19206535-19206602,19206602-19206656,19206656-19206673,19206673-19206693,19206693-19206735,19206735-19206772,19206830-19206831,19206831-19206885,19206885-19207737,19207737-19207930,19207930-19207984,19207984-19208038,19208038-19208092,19208092-19208146,19208146-19208200,19208200-19208254,19208254-19208308,19208308-19208362,19208362-19208416,19208416-19208470,19208470-19208524,19208524-19208578,19208578-19208632,19208632-19208686] became 19204063-19208686
```

# Requirements
2. Run `make lint`

-------
### Integration Tests

### DIFF
--- a/internal/state/manager_test.go
+++ b/internal/state/manager_test.go
@@ -58,7 +58,7 @@ func TestFileRangeManager(t *testing.T) {
 		for name, testCase := range testCases {
 			t.Run(name, func(t *testing.T) {
 				got := NewFileRangeManager(testCase.cfg).(*rangeManager)
-				assert.Equal(t, testCase.cfg.Name, got.name)
+				assert.Equal(t, testCase.cfg.Name, got.ID())
 				assert.Equal(t, testCase.wantFilePath, got.stateFilePath)
 				assert.Equal(t, testCase.wantQueueSize, cap(got.queue))
 				assert.Equal(t, testCase.wantSaveInterval, got.saveInterval)
@@ -120,8 +120,8 @@ func TestFileRangeManager(t *testing.T) {
 		tmpDir := t.TempDir()
 		manager := NewFileRangeManager(ManagerConfig{
 			StateFileDir: tmpDir,
-			Name:         "replace.log"},
-		).(*rangeManager)
+			Name:         "replace.log",
+		}).(*rangeManager)
 
 		notification := Notification{
 			Delete: make(chan struct{}),
@@ -139,7 +139,7 @@ func TestFileRangeManager(t *testing.T) {
 
 		time.Sleep(2 * defaultSaveInterval)
 
-		tree := newMultiRangeTracker("replace.log")
+		tree := newRangeTracker("replace.log", 10)
 		tree.Insert(Range{start: 500, end: 600})
 		assert.NoError(t, manager.save(tree))
 		time.Sleep(2 * defaultSaveInterval)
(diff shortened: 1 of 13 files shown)


internal/state/range.go
```diff
@@ -306,14 +311,9 @@ type multiRangeTracker struct {
 
 var _ RangeTracker = (*multiRangeTracker)(nil)
 
-// newMultiRangeTracker creates an unbounded multiRangeTracker.
-func newMultiRangeTracker(name string) RangeTracker {
```
jefchien, line 310: Removed the unbounded constructor, since it could be misleading and improperly used.

internal/state/manager_test.go
```diff
@@ -374,6 +379,69 @@ func TestFileRangeManager(t *testing.T) {
...
+	})
+	t.Run("MultipleRanges/Distinct", func(t *testing.T) {
+		q := &mockFileRangeQueue{}
+		b := NewRangeQueueBatcher(q)
+		b.Merge(Range{start: 100, end: 200})
+		b.Merge(Range{start: 20, end: 30})
+		b.Merge(Range{start: 5, end: 10})
+		b.Done()
+		assert.Len(t, q.enqueued, 1)
+		assert.Equal(t, Range{start: 5, end: 200}, q.enqueued[0])
```
the-mann, line 444: help me understand this one... 5-10, 20-30, and 100-200 don't overlap, so why is it getting merged into a single range?

jefchien: It's showing the behavior of the `RangeQueueBatcher` when the ranges are non-overlapping. In our normal usage of this (in the log pusher), they will always be continuous because we read and batch log lines sequentially.

I need to find a better name for it, but it's basically a super naive merger and only retains the min/max of all the ranges that get merged in. This is all in an effort to reduce the amount of merging the actual state manager will need to do. If you look in the description under the test, you can see an example of all the ranges in a single batch.

jefchien: I can make it stricter so the assumption is it will always be in order or without gaps.

the-mann: gotcha – I think a better name would be helpful but it can be done as a follow up

the-mann (dismissed):

Paramadon (approved):
LGTM!

duhminick (approved):

duhminick (approved):

sky333999 (approved):

PR #1740, opened 2025-06-23
Debugger tool initial commit

James-Hou22:
# Description of the issue
Currently, the agent does not have any local debugging capabilities. This commit kicks off the creation of a new debugger tool binary that helps customers view their agent config and debug common issues.

# Description of changes
- Created debugger binary file
- Added debugger binary to RPM
- Added key file existence and readability checks
- Added instance info display
- Added getMergedJSONConfig function to get merged JSON

# License

# Tests
N/A

# Requirements
2. Run `make lint`

-------
### Integration Tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,6 +75,7 @@ amazon-cloudwatch-agent-linux: copy-version-file
 	$(LINUX_ARM64_BUILD)/start-amazon-cloudwatch-agent github.com/aws/amazon-cloudwatch-agent/cmd/start-amazon-cloudwatch-agent
 	$(LINUX_AMD64_BUILD)/amazon-cloudwatch-agent-config-wizard github.com/aws/amazon-cloudwatch-agent/cmd/amazon-cloudwatch-agent-config-wizard
 	$(LINUX_ARM64_BUILD)/amazon-cloudwatch-agent-config-wizard github.com/aws/amazon-cloudwatch-agent/cmd/amazon-cloudwatch-agent-config-wizard
+	$(LINUX_ARM64_BUILD)/amazon-cloudwatch-agent-debugger github.com/aws/amazon-cloudwatch-agent/cmd/amazon-cloudwatch-agent-debugger
 
 
 amazon-cloudwatch-agent-darwin: copy-version-file
--- a/cmd/amazon-cloudwatch-agent-debugger/debugger.go
+++ b/cmd/amazon-cloudwatch-agent-debugger/debugger.go
@@ -1,3 +1,6 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
 package main
 
 import (
--- a/cmd/amazon-cloudwatch-agent-debugger/debugger.go
+++ b/cmd/amazon-cloudwatch-agent-debugger/debugger.go
@@ -6,7 +6,10 @@ import (
 	"log"
 
 	"github.com/aws/amazon-cloudwatch-agent/internal/debugger"
+	"github.com/aws/amazon-cloudwatch-agent/translator/cmdutil"
 )
+
+var mergedConfig map[string]interface{}
 
 func main() {
 	ctx := context.Background()
@@ -19,6 +22,18 @@ func main() {
 	printInstanceInfo(info)
 
 	debugger.CheckConfigFiles()
+
+	//Load merged config, this is the same logic that the translator uses
+	config, err := cmdutil.GetMergedConfig("/opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.json", "/opt/aws/amazon-cloudwatch-agent/etc/amazon-cloudwatch-agent.d", "ec2", info.OS)
+	if err != nil {
+		log.Printf("Failed to load config: %v", err)
+	} else {
+		mergedConfig = config
+		fmt.Println("\n=== Configuration Loaded ===")
+		parseEndpoints()
+	}
+
+	debugger.CheckLogs(mergedConfig)
 }
 
 func printHeader() {
@@ -36,4 +51,27 @@ func printInstanceInfo(info *debugger.InstanceInfo) {
 	fmt.Printf("Availability Zone: %s\n", info.AvailabilityZone)
 	fmt.Printf("Architecture:      %s\n", info.Architecture)
 	fmt.Printf("OS:				   %s\n", info.OS)
+}
+
+func parseEndpoints() {
+	if mergedConfig == nil {
+		fmt.Println("No configuration available")
+		return
+	}
+
+	if metrics, ok := mergedConfig["metrics"].(map[string]interface{}); ok {
+		if endpoint, ok := metrics["endpoint_override"].(string); ok {
+			fmt.Printf("Metrics Endpoint: %s\n", endpoint)
+		} else {
+			fmt.Println("Metrics Endpoint: Default CloudWatch endpoint (no override)")
+		}
+	}
+
+	if logs, ok := mergedConfig["logs"].(map[string]interface{}); ok {
+		if endpoint, ok := logs["endpoint_override"].(string); ok {
+			fmt.Printf("Logs Endpoint: %s\n", endpoint)
+		} else {
+			fmt.Println("Logs Endpoint: Default CloudWatch Logs endpoint (no override)")
+		}
+	}
 }
--- a/cmd/amazon-cloudwatch-agent-debugger/debugger.go
+++ b/cmd/amazon-cloudwatch-agent-debugger/debugger.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+
 	"github.com/aws/amazon-cloudwatch-agent/internal/debugger"
 )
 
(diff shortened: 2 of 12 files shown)


internal/debugger/fileCheck.go
```diff
@@ -0,0 +1,131 @@
...
+			Description: "Log configuration file",
+			Required:    false,
+		},
+		{
+			Path:        "/opt/aws/amazon-cloudwatch-agent/etc/env-config.json",
+			Description: "Environment configuration file",
+			Required:    false,
+		},
+		{
+			Path:        "/opt/aws/amazon-cloudwatch-agent/logs/amazon-cloudwatch-agent.log",
```
Paramadon: I think there are constants for these defined in a maybe the common package of the repo? But I rather see those here instead of these strings.

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
```diff
@@ -313,6 +309,7 @@ func (le LogEvent) Time() time.Time {
 }
 
 func (le LogEvent) Done() {
+	le.RangeQueue().Enqueue(le.Range())
```
Paramadon: Why are we changing this as part of this pr?

plugins/inputs/logfile/tailersrc.go
```diff
@@ -47,6 +47,7 @@ func (le LogEvent) Time() time.Time {
 }
 
 func (le LogEvent) Done() {
+	le.RangeQueue().Enqueue(le.Range())
```
Paramadon: Why is this changing?

Makefile
Paramadon, line 1: Why do we have logfile changes as part of this cr? Also could you add some images/vidoe of how the debugger tool looks? It would be nice to see the flow of the debugger tool to see how it would look in front of users.

plugins/inputs/logfile/tailersrc.go
James-Hou22: This was part of Jeffrey's rollback commit, I pulled it over the weekend

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
James-Hou22: Part of Jeffrey's commit as well

Makefile
James-Hou22: The logfile changes were because the feature branch that the PR is going to didn't get the rollback pull, so those are a part of Jeffrey's rollback commit. I also added an example of running the debugger above

internal/debugger/fileCheck.go
```diff
@@ -0,0 +1,133 @@
...
+			Description: "YAML configuration file",
+			Required:    true,
+		},
+		{
+			Path:        paths.CommonConfigPath,
+			Description: "Common configuration file",
+			Required:    false,
+		},
+		{
+			Path:        "/opt/aws/amazon-cloudwatch-agent/etc/log-config.json",
```
the-mann: why is this not a constant like the rest of the config files?

cmd/amazon-cloudwatch-agent-debugger/debugger.go
```diff
@@ -0,0 +1,80 @@
...
+func printInstanceInfo(info *debugger.InstanceInfo) {
+	fmt.Println("")
+	fmt.Printf("Instance ID:       %s\n", info.InstanceID)
+	fmt.Printf("Account ID:        %s\n", info.AccountID)
+	fmt.Printf("Region:            %s\n", info.Region)
+	fmt.Printf("Instance Type:     %s\n", info.InstanceType)
+	fmt.Printf("AMI:               %s\n", info.ImageID)
+	fmt.Printf("Availability Zone: %s\n", info.AvailabilityZone)
+	fmt.Printf("Architecture:      %s\n", info.Architecture)
+	fmt.Printf("OS:				   %s\n", info.OS)
```
the-mann: nit: can you use logging instead of fmt.Print?

internal/debugger/fileCheck.go
```diff
@@ -0,0 +1,133 @@
...
+
+	for _, file := range configFiles {
+		if strings.HasSuffix(file.Path, ".d") {
+			// Special handling for .d directory
+			entries, err := os.ReadDir(file.Path)
+			status := "✓ Present"
+
+			if err != nil {
+				if os.IsNotExist(err) {
+					status = "✗ Missing"
```
the-mann: hmm, status should probably be an enum

internal/debugger/fileCheck.go
```diff
@@ -0,0 +1,133 @@
...
+		},
+	}
+
+	for _, file := range configFiles {
+		if strings.HasSuffix(file.Path, ".d") {
+			// Special handling for .d directory
+			entries, err := os.ReadDir(file.Path)
+			status := "✓ Present"
+
+			if err != nil {
```
the-mann: i'd recommend using a guard statement instead of nesting your conditionals... might make the code a bit more readable

cmd/amazon-cloudwatch-agent-debugger/debugger.go
the-mann, line 1: overall i'd try and prioritize guard statements to make this a bit more readable.

internal/debugger/instanceInfo.go
```diff
@@ -0,0 +1,47 @@
...
+type InstanceInfo struct {
+	InstanceID       string `json:"instanceId"`
+	AccountID        string `json:"accountId"`
+	Region           string `json:"region"`
+	InstanceType     string `json:"InstanceType"`
+	ImageID          string `json:"imageId"`
+	AvailabilityZone string `json:"availabilityZone"`
+	Architecture     string `json:"architecture"`
+	OS               string `json:"os"`
+}
```
the-mann: nit: this feels redundant/like there should be an object that already stores this... but maybe not.

internal/debugger/logCheck.go
```diff
@@ -0,0 +1,53 @@
...
+func checkLogPermissions(filePath string) {
+	name := filepath.Base(filePath)
+	if _, err := os.Stat(filePath); err != nil {
+		fmt.Printf("Configured log file %s does not exist.\n", name)
+		return
+	}
+
+	if file, err := os.Open(filePath); err != nil {
+		fmt.Printf("Agent does not have read permission for log file %s\n", name)
+		fmt.Printf("Try: sudo chmod 644 %s\n", filePath)
```
the-mann: we might not want to blanket recommend this... not sure if there's a better alternative... maybe run debugger as sudo? that's also dangerous, but at least temporary

translator/cmdutil/translatorutil.go
```diff
@@ -255,3 +255,27 @@ func ConfigToYamlFile(config interface{}, yamlConfigFilePath string) error {
 	}
 	return os.WriteFile(yamlConfigFilePath, []byte(res), fileMode)
 }
+
+// GetMergedConfig creates a merged JSON config map from the given parameters
+// This allows reusing the translator logic without setting up external context
+func GetMergedConfig(configPath, configDir, mode, osType string) (map[string]interface{}, error) {
```
the-mann, line 261: would like to see some tests for this

the-mann (changes requested):
Overall good first pass – I'd like to see some unit tests. This might inform how you write your code. Maybe consider if you can write it in a way such that your functions have outputs that you can test the output of without 
1. excessive mocking or 
2. testing for specific strings to be printed

internal/debugger/fileCheck.go
James-Hou22: I'll double check but I don't think there is a constant for this

internal/debugger/instanceInfo.go
James-Hou22: This struct holds info from 2 different objects but most of them are from the identity document. But because the identity document holds a bunch more stuff I thought it would make more sense to take just what I needed

internal/debugger/fileCheck_test.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+}
+
+func TestCheckFileContentStatus(t *testing.T) {
+	// Non-existent file
+	status := checkFileContentStatus("/non/existent/file.json")
+	assert.Equal(t, StatusMissing, status, "checkFileContentStatus() for non-existent file returned unexpected status")
+
+	// Valid JSON
+	tmpFile, err := os.CreateTemp("", "test-*.json")
```
the-mann: what does the * do? seems weird cause that's usually used for a [glob](https://en.wikipedia.org/wiki/Glob_(programming)) wildcard

the-mann (changes requested):
Looks great! all that's left is a small nit of renaming from camelCase to snake_case (ie fileCheck to file_check and logCheck to log_check), but besides that this looks good :)

internal/debugger/fileCheck_test.go
James-Hou22: A random string will replace the * character: https://pkg.go.dev/os#CreateTemp. I used it to make specific different file names every time the test is ran to ensure no overlap or issues with file names. I couldn't really imagine a time where this would be a problem but I thought I'd take advantage of the feature.

the-mann (approved):

internal/debugger/log_check_test.go
```diff
@@ -0,0 +1,209 @@
...
+				assert.Error(t, err, "getCollectListFromConfig() should return an error")
+			} else {
+				assert.NoError(t, err, "getCollectListFromConfig() should not return an error")
+				assert.Len(t, collectList, tt.expectedLen, "getCollectListFromConfig() returned unexpected number of items")
+			}
+		})
+	}
+}
+
+func TestCheckLogs(t *testing.T) {
```
Paramadon, line 104: I would like to see a test where the agent does not have permissions to open the file

Paramadon (approved):

the-mann (approved):

PR #1745, opened 2025-06-27
Add tarball integration for local and SSM calling

James-Hou22:
# Description of the issue

A common pain point of customers when cutting a ticket is gathering information regarding their agent for the support team. They must sift through their agent, gather relevant files, and provide them to the support engineer. Adding a feature to gather all necessary files and compress them allows users to provide support engineers with relevant context through one command.

# Description of changes

* Created tarball generator
* Added unit tests for generator

It is not possible to provide the triaging interactivity if the agent tarball generator is called via SSM, so if the agent detects that it is being called through SSM, it will not ask for triaging.

# Output
The tarball will have the following structure:
```
support-bundle-YYYY-MM-DD-HHMMSS.tar.gz
├── debug-info.txt
├── etc/
│   ├── amazon-cloudwatch-agent.toml
│   ├── amazon-cloudwatch-agent.d/config.json #agent config
│   ├── common-config.toml
│   ├── log-config.json
│   └── env-config.json
├── logs/
    └── amazon-cloudwatch-agent.log
```
If SSM is used, the tarball will not have the `debug-info.txt` file.

image

5bigbooms

# License

### DIFF
--- a/internal/debugger/error_collector.go
+++ b/internal/debugger/error_collector.go
@@ -37,10 +37,12 @@ func (ec *ErrorCollector) printErrorMessages(messages []string, title string) {
 
 	if len(messages) == 0 {
 		fmt.Printf("No %s\n", title)
-	} else {
-		fmt.Printf("%s (%d):\n", title, len(messages))
-		for _, msg := range messages {
-			fmt.Printf("  %s\n", msg)
-		}
+		return
 	}
+
+	fmt.Printf("%s (%d):\n", title, len(messages))
+	for _, msg := range messages {
+		fmt.Printf("  %s\n", msg)
+	}
+
 }


cmd/amazon-cloudwatch-agent-debugger/debugger.go
```diff
@@ -16,6 +17,17 @@ var mergedConfig map[string]interface{}
...
+		switch os.Args[1] {
+		case "--tarball":
+			debugger.CreateTarball(false)
+			return
+		case "--tarballssm":
+			debugger.CreateTarball(true)
+			return
+		}
+	}
+
```
the-mann: should use [flag package](https://pkg.go.dev/flag) instead

internal/debugger/tarball.go
```diff
@@ -0,0 +1,341 @@
...
+	"github.com/aws/amazon-cloudwatch-agent/tool/paths"
+)
+
+type Answers struct {
+	Occurence             string
+	EnvironmentChange     string
+	EnvironmentChangeDesc string
+}
+
+func Triage() Answers {
```
the-mann: lol, great name

internal/debugger/tarball.go
```diff
@@ -0,0 +1,341 @@
...
+		return
+	}
+	defer file.Close()
+
+	gzipWriter := gzip.NewWriter(file)
+	defer gzipWriter.Close()
+	tarWriter := tar.NewWriter(gzipWriter)
+	defer tarWriter.Close()
+
+	if err := addFileToTarball(tarWriter, paths.AgentLogFilePath, "logs/amazon-cloudwatch-agent.log", 500); err != nil {
```
the-mann: why? we should include all of it

internal/debugger/tarball.go
```diff
@@ -0,0 +1,341 @@
...
+	EnvironmentChangeDesc string
+}
+
+func Triage() Answers {
+	answers := Answers{}
+	reader := bufio.NewReader(os.Stdin)
+
+	fmt.Println("Please answer these questions to better assist with your issue:")
+
+	fmt.Println("Is this issue once off, intermittent, or consistently happening right now? (o/i/c): ")
```
the-mann: i would let them write anything, give us as much info as possible

internal/debugger/tarball.go
```diff
@@ -0,0 +1,341 @@
...
+	occurence, err := reader.ReadString('\n')
+	if err == nil {
+		answers.Occurence = strings.TrimSpace(occurence)
+	}
+
+	fmt.Print("Has anything changed in the environment recently? (y/n): ")
```
the-mann: same thing. we should let them share when it changed, etc by just letting them type whatever.

James-Hou22: If the user says yes they have the opportunity to write about what did change after.

internal/debugger/tarball.go
James-Hou22: True, I'll add a section to let them write anything they want. FYI this section is a WIP with the support team. They're going to request some more questions to be added here in the future.

internal/debugger/tarball.go
James-Hou22: Going to go up to 50,000 lines, this attributes to about ~800KB-2MB of compressed data.

James-Hou22: Also going to add an option for customers to turn on debug mode for logging and wait for some `x` seconds of gathering those logs before they get gathered and grouped to the tarball.

Paramadon (approved):

cmd/amazon-cloudwatch-agent-debugger/debugger.go
```diff
@@ -15,67 +18,133 @@ import (
 var mergedConfig map[string]interface{}
 
 func main() {
+	ssm := flag.Bool("ssm", false, "Run debugger with limited formatting for SSM")
```
okankoAMZ: nit: the naming can be more descriptive like: isRunForSSM, ssmModeOn, etc.

cmd/amazon-cloudwatch-agent-debugger/debugger.go
```diff
@@ -15,67 +18,133 @@ import (
...
+	tarball := flag.Bool("tarball", false, "Create tarball")
```
okankoAMZ: Same as my last comment it could be more descriptive

cmd/amazon-cloudwatch-agent-debugger/debugger.go
```diff
@@ -15,67 +18,133 @@ import (
 var mergedConfig map[string]interface{}
```
okankoAMZ, line 20: I think you define your own struct here and NOT use use `map[string]interface{}`

internal/debugger/endpoint_check.go
```diff
@@ -0,0 +1,72 @@
...
+		return &EndpointInfo{
+			MetricsEndpoint: "No configuration available",
+			LogsEndpoint:    "No configuration available",
+		}, nil
+	}
+
+	fmt.Fprintln(w, "\n=== Endpoint Configuration ===")
+
+	var metricsEndpoint, logsEndpoint string
+	if metrics, ok := config["metrics"].(map[string]interface{}); ok {
```
okankoAMZ: nit: I am personally a fan of [Never Nesting](https://medium.com/@MBCube/i-am-a-never-nester-8c88baf7a218) and I highly recommend it.

internal/debugger/endpoint_check.go
```diff
@@ -0,0 +1,72 @@
...
+			logsEndpoint = "Default CloudWatch Logs endpoint (no override)"
+		}
+	} else {
+		logsEndpoint = "No logs configuration found"
+	}
+
+	if ssm {
+		fmt.Fprintf(w, "Metrics: %s\n", metricsEndpoint)
+		fmt.Fprintf(w, "Logs:    %s\n", logsEndpoint)
+	} else {
```
okankoAMZ: why use a else here when you can just return inside that if statement?

internal/debugger/error_collector.go
```diff
@@ -0,0 +1,41 @@
...
+	DebuggerErrorCollector.ConfigErrors = append(DebuggerErrorCollector.ConfigErrors, msg)
+}
+
+func AddConfigWarning(msg string) {
+	DebuggerErrorCollector.ConfigWarnings = append(DebuggerErrorCollector.ConfigWarnings, msg)
+}
+
+func PrintAggregatedErrors() {
+	fmt.Println("=== Errors & Warnings Summary ===")
+	if len(DebuggerErrorCollector.ConfigErrors) > 0 || len(DebuggerErrorCollector.ConfigWarnings) > 0 {
```
okankoAMZ: nit: Like I mentioned before I think if you utilise, return statements with your if statements it will be easier to read this.

internal/debugger/error_collector.go
```diff
@@ -0,0 +1,41 @@
...
+
+		if len(DebuggerErrorCollector.ConfigErrors) > 0 {
```
okankoAMZ: nit: you can save len as a variable since you already called it once.

internal/debugger/file_check.go
```diff
@@ -127,3 +232,12 @@ func getDisplayName(path string) string {
 	}
 	return filepath.Base(path)
 }
+
+// Using runes to support "─"
+func repeatChar(char rune, count int) string {
```
okankoAMZ: I think you defined this function twice, i'd just have utils.go and save it there and re-use it

internal/debugger/error_collector.go
James-Hou22: Yeah ur right. I think I've been caught up in the main idea of printing for all that I've been doing and forgetting to utilize returning more.

internal/debugger/endpoint_check.go
James-Hou22: I'm going to refactor this whole file honestly

cmd/amazon-cloudwatch-agent-debugger/debugger.go
James-Hou22: `map[string]interface{}` is the return value of what we're storing in this variable and its from `translator.util` so I'd rather keep this and not explicitly define a struct for consistency purposes.

okankoAMZ: I still think we should have a data structure here and we can convert `map[string]interface{}` from translator to that

cmd/amazon-cloudwatch-agent-debugger/debugger.go
```diff
@@ -5,77 +5,141 @@ package main
...
 	"github.com/aws/amazon-cloudwatch-agent/translator/cmdutil"
 )
 
 var mergedConfig map[string]interface{}
 
 func main() {
+	compact := flag.Bool("compact", false, "Run debugger with compact formatting")
+	createtarball := flag.Bool("tarball", false, "Create tarball")
+	createtarballssm := flag.Bool("tarballssm", false, "Create tarball with SSM")
+	startmcpserver := flag.Bool("mcp", false, "Start MCP server for IDE integration")
```
okankoAMZ: nit: 
```suggestion
	startmcpserver := flag.Bool("mcp", false, "Start MCP server for AI integration")
```

cmd/amazon-cloudwatch-agent-debugger/debugger.go
```diff
@@ -5,77 +5,141 @@ package main
...
+	"github.com/aws/amazon-cloudwatch-agent/internal/debugger/utils"
 	"github.com/aws/amazon-cloudwatch-agent/tool/paths"
 	"github.com/aws/amazon-cloudwatch-agent/translator/cmdutil"
 )
 
 var mergedConfig map[string]interface{}
 
 func main() {
+	compact := flag.Bool("compact", false, "Run debugger with compact formatting")
+	createtarball := flag.Bool("tarball", false, "Create tarball")
```
okankoAMZ: can you use camelCase for these vars

cmd/amazon-cloudwatch-agent-debugger/debugger.go
```diff
@@ -5,77 +5,141 @@ package main
...
 
-	//Load merged config, this is the same logic that the translator uses
-	config, err := cmdutil.GetMergedConfig(paths.JsonConfigPath, paths.ConfigDirPath, "ec2", info.OS)
-	if err != nil {
-		log.Printf("Failed to load config: %v", err)
+	// We provide a stream because MCP uses a buffer. This is so when MCP calls tools it will not print to stdout.
+	// There are better ways of doing this but not without significant refactoring overhead.
+	if !debugger.CheckConfigFiles(os.Stdout, *compact) {
+		fmt.Println("⚠️  ERROR: Required configuration files are missing - cannot conduct log checks.")
+		return
```
okankoAMZ, line 62: if you are returning here, why do you need the else?

cmd/amazon-cloudwatch-agent-debugger/debugger.go
```diff
@@ -5,77 +5,141 @@ package main
...
 	} else {
-		mergedConfig = config
-		log.Println("\n=== Configuration Loaded ===")
-		parseEndpoints()
+		config, err := cmdutil.GetMergedConfig(paths.JsonConfigPath, paths.ConfigDirPath, "ec2", info.OS)
+		if err != nil {
+			fmt.Printf("Failed to load config: %v\n", err)
+			return
+		} else {
```
okankoAMZ: if you are returning here, why do you need the else?

cmd/amazon-cloudwatch-agent-debugger/debugger.go
```diff
@@ -5,77 +5,141 @@ package main
...
+			mergedConfig = config
+		}
+		_, err = debugger.CheckEndpoints(os.Stdout, mergedConfig, *compact)
+		if err != nil {
+			fmt.Printf("Failed to check endpoints: %v\n", err)
```
okankoAMZ: We keep going even thought we cannot check endpoints?

cmd/amazon-cloudwatch-agent-debugger/debugger.go
```diff
@@ -5,77 +5,141 @@ package main
...
 	}
 
-	if metrics, ok := mergedConfig["metrics"].(map[string]interface{}); ok {
-		if endpoint, ok := metrics["endpoint_override"].(string); ok {
-			log.Printf("Metrics Endpoint: %s\n", endpoint)
-		} else {
-			log.Println("Metrics Endpoint: Default CloudWatch endpoint (no override)")
-		}
+	labelWidth := 18
+
```
okankoAMZ, line 102: Naming of these two vars a bit unclear

internal/debugger/endpoint_check.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"github.com/aws/amazon-cloudwatch-agent/internal/debugger/utils"
+)
+
+type EndpointInfo struct {
+	MetricsEndpoint string `json:"metrics_endpoint"`
+	LogsEndpoint    string `json:"logs_endpoint"`
+}
+
+func CheckEndpoints(w io.Writer, config map[string]interface{}, compact bool) (*EndpointInfo, error) {
```
okankoAMZ: What endpoint? The naming of the function does not distinguish that

internal/debugger/endpoint_check.go
```diff
@@ -0,0 +1,75 @@
...
+			LogsEndpoint:    "No configuration available",
+		}, nil
+	}
+
+	fmt.Fprintln(w, "\n=== Endpoint Configuration ===")
+
+	metricsEndpoint := getEndpoint(config, "metrics", "Default CloudWatch endpoint (no override)", "No metrics configuration found")
+	logsEndpoint := getEndpoint(config, "logs", "Default CloudWatch Logs endpoint (no override)", "No logs configuration found")
+
+	if compact {
```
okankoAMZ, line 32: So compact flag is actually if it is ssm or not?

internal/debugger/endpoint_check.go
okankoAMZ: When do you ever return an error with this function?

internal/debugger/endpoint_check.go
```diff
@@ -0,0 +1,75 @@
...
+
+	return &EndpointInfo{
+		MetricsEndpoint: metricsEndpoint,
+		LogsEndpoint:    logsEndpoint,
+	}, nil
+}
+
+func getEndpoint(config map[string]interface{}, section, defaultMsg, notFoundMsg string) string {
+	sectionConfig, ok := config[section].(map[string]interface{})
+	if !ok {
```
okankoAMZ, line 46: Let's use a consistent format for these checks - either all one-liners or all multi-line.

internal/debugger/error_collector.go
```diff
@@ -0,0 +1,37 @@
...
+package debugger
+
+import "fmt"
+
+type ErrorCollector struct {
+	ConfigErrors   []string
+	ConfigWarnings []string
+}
+
+var DebuggerErrorCollector = &ErrorCollector{}
```
okankoAMZ: why do we define it globally here and not main?

internal/debugger/file_check.go
```diff
@@ -30,17 +30,88 @@ type ConfigFile struct {
...
 	Description string
 	Required    bool
+	Purpose     string
+	MissingMsg  string
 }
 
-// Checks for existence and readability of key configuration files
-func CheckConfigFiles() {
-	log.Println("Checking Configuration Files:")
+func CheckConfigFiles(w io.Writer, compact bool) bool {
```
okankoAMZ: what are you checking this function name is not clear. You can rename it to something like:
`isConfigFilesPresent`

internal/debugger/file_check.go
```diff
@@ -30,17 +30,88 @@ type ConfigFile struct {
...
+	}
+
+	fmt.Fprintf(w, "└%s┴%s┴%s┘\n",
+		utils.RepeatChar('─', fileNameWidth+2),
+		utils.RepeatChar('─', statusWidth+2),
+		utils.RepeatChar('─', 50))
+}
+
+func handleFileStatus(file ConfigFile, status FileStatus) {
+	if status == StatusMissing || status == StatusPresentNotReadable || status == StatusNoFile {
```
okankoAMZ: nit: I think you should use a switch statement here or work backwards.

internal/debugger/log_check.go
```diff
@@ -5,132 +5,237 @@ package debugger
...
 
-func CheckLogs(config map[string]interface{}) LogCheckResult {
-	log.Println("Running CheckLogs...")
-	defer func() {
-		if r := recover(); r != nil {
-			log.Printf("Panicked with error: %v", r)
-			return
-		}
-	}()
+func CheckLogs(w io.Writer, config map[string]interface{}, compact bool) ([]AgentLogConfig, error) {
```
okankoAMZ: Its not clear what log you are trying to "check" 
```suggestion
func CheckConfiguredLogs(w io.Writer, config map[string]interface{}, compact bool) ([]AgentLogConfig, error) {
```

internal/debugger/log_check.go
```diff
@@ -5,132 +5,237 @@ package debugger
...
+				fileConfig.FilePath = matchedFile
+				checkLogFileStatus(&fileConfig)
+				logConfigs = append(logConfigs, fileConfig)
+			}
+		}
+	}
+
+	totalConfigs := len(logConfigs)
+	accessibleConfigs := 0
+	issueConfigs := 0
```
okankoAMZ: naming here is unclear

internal/debugger/tarball.go
```diff
@@ -0,0 +1,441 @@
...
+		fmt.Println("\nTo capture more detailed logs, DEBUG level logging will be temporarily enabled")
+		fmt.Println("for the CloudWatch Agent and LogDebug level logging for the AWS SDK.")
+		fmt.Println("This will help diagnose issues more effectively.")
+		fmt.Print("\nIs this okay? (y/n, default: y): ")
+
+		reader := bufio.NewReader(os.Stdin)
+		response, _ := reader.ReadString('\n')
+		response = strings.TrimSpace(strings.ToLower(response))
+
+		if response == "" || response == "y" {
```
okankoAMZ, line 61: I think we have a library for this that checks yes and no combinations

internal/debugger/tarball.go
```diff
@@ -0,0 +1,441 @@
...
+		return
+	}
+	defer file.Close()
+
+	gzipWriter := gzip.NewWriter(file)
+	defer gzipWriter.Close()
+	tarWriter := tar.NewWriter(gzipWriter)
+	defer tarWriter.Close()
+
+	if err := addFileToTarball(tarWriter, paths.AgentLogFilePath, "logs/amazon-cloudwatch-agent.log", 50000); err != nil {
```
okankoAMZ, line 113: magic number, what is "50000"

internal/debugger/tarball.go
```diff
@@ -0,0 +1,441 @@
...
+		fmt.Println("Error, unable to add log file to tarball: ", err)
+	} else {
+		fmt.Println("Warning: Only the last 50,000 lines of the log file are included.")
```
okankoAMZ, line 116: Why only 50,000?

cmd/amazon-cloudwatch-agent-debugger/debugger.go
```diff
@@ -5,77 +5,141 @@ package main
...
+		debugger.CreateTarball(true)
+		return
+	case *startmcpserver:
+		mcp.StartMCPServer()
+		return
+	}
+
+	defer func() {
+		debugger.PrintAggregatedErrors()
+		fmt.Println()
```
Paramadon, line 45: Nit: url is constant so maybe do add url as a constant and you can use \n instead of having an extra print statement.

internal/debugger/tarball.go
```diff
@@ -0,0 +1,441 @@
...
+	answers := Triage()
+
+	// Create answers content
+	answersContent := "CloudWatch Agent Debugging Information\n"
+	answersContent += "===================================\n\n"
+	answersContent += "Q: Is this issue once off, intermittent, or consistently happening right now?\n"
+
+	// Formatting answers
+	occurrenceAnswer := ""
+	switch strings.ToLower(answers.Occurence) {
```
okankoAMZ: I think you can break these question answer portions to a different file and different functions and jsut call them to get occurenceAnswer , or xAnswer. Maybe you can store the questions and answer choices in a map, then just for loop to collect all the information. I think that would be more scalable in the long run

internal/debugger/tarball.go
```diff
@@ -0,0 +1,441 @@
...
+	}
+	fileSize := stat.Size()
+
+	if fileSize == 0 {
+		return 0, []byte{}, nil
+	}
+
+	pos := fileSize
+	lineCount := 0
+	chunkSize := int64(4096) // 4KB chunks
```
okankoAMZ: why 4KB?

cmd/amazon-cloudwatch-agent-debugger/debugger.go
```diff
@@ -5,77 +5,141 @@ package main
...
+	printInstanceInfo(info, *compact)
 
-	//Load merged config, this is the same logic that the translator uses
-	config, err := cmdutil.GetMergedConfig(paths.JsonConfigPath, paths.ConfigDirPath, "ec2", info.OS)
-	if err != nil {
-		log.Printf("Failed to load config: %v", err)
+	// We provide a stream because MCP uses a buffer. This is so when MCP calls tools it will not print to stdout.
+	// There are better ways of doing this but not without significant refactoring overhead.
+	if !debugger.CheckConfigFiles(os.Stdout, *compact) {
+		fmt.Println("⚠️  ERROR: Required configuration files are missing - cannot conduct log checks.")
```
Paramadon: nit: not sure if we should add ⚠️ as part of the print, we don't do that anywhere else in the agent code so this seems different. But just a preference, you can keep this.

translator/cmdutil/translatorutil.go
```diff
@@ -259,10 +260,18 @@ func ConfigToYamlFile(config interface{}, yamlConfigFilePath string) error {
 // GetMergedConfig creates a merged JSON config map from the given parameters
 // This allows reusing the translator logic without setting up external context
 func GetMergedConfig(configPath, configDir, mode, osType string) (map[string]interface{}, error) {
-	// Temporarily redirect stdout to suppress output
+	// Suppress log output and stdout during debugger config generation
+	origLogOutput := log.Writer()
```
okankoAMZ, line 265: how do you determine when is debugger run? This looks like you are suppressing log output for every case.

internal/debugger/error_collector.go
```diff
@@ -0,0 +1,37 @@
...
+	DebuggerErrorCollector.ConfigWarnings = append(DebuggerErrorCollector.ConfigWarnings, msg)
+}
+
+func PrintAggregatedErrors() {
+	fmt.Println("=== Errors & Warnings Summary ===")
+	printMessages(DebuggerErrorCollector.ConfigErrors, "Errors", "❌")
+	printMessages(DebuggerErrorCollector.ConfigWarnings, "Warnings", "⚠️")
+}
+
+func printMessages(messages []string, title, icon string) {
```
Paramadon: Vague function name

cmd/amazon-cloudwatch-agent-debugger/debugger.go
Paramadon:
```suggestion
	createTarball := flag.Bool("tarball", false, "Create tarball")
    createTarballSSM := flag.Bool("tarballssm", false, "Create tarball with SSM")

```

cmd/amazon-cloudwatch-agent-debugger/debugger.go
James-Hou22: I'm going to remove all MCP related stuff for now. Once the feature branch is done I'll fork from it and add the MCP stuff but I'll edit this and stash it.

cmd/amazon-cloudwatch-agent-debugger/debugger.go
James-Hou22: nothing relies on endpoints being able to be checked or not so if we can't check them then it's fine, we can move on to logs.

internal/debugger/endpoint_check.go
James-Hou22: It's also for MCP. I need to take out the stuff related to MCP but I think this should still just be called compact since it literally just a compact output.

internal/debugger/endpoint_check.go
James-Hou22: U right, I think I was just so used to adding an error return. I removed this. I reflect the "error" in what I put in the returned struct.

internal/debugger/endpoint_check.go
James-Hou22: I mean technically they are both multi line. One of them checks for `!ok` and the other checks for `ok` so they naturally will look different.

James-Hou22:
```
func getEndpoint(config map[string]interface{}, section, defaultMsg, notFoundMsg string) string {
	sectionConfig, ok := config[section].(map[string]interface{})
	if !ok {
		return notFoundMsg
	}
	endpoint, ok := sectionConfig["endpoint_override"].(string)
	if ok {
		return endpoint
	}
	return defaultMsg
}
```
I'm going to postulate this is what you were referring to

internal/debugger/file_check.go
James-Hou22: Going to rename it to `IsConfigFilesPresentAndReadable` cuz we check for readability too

translator/cmdutil/translatorutil.go
James-Hou22: It's not, this function is run only by the debugger

internal/debugger/tarball.go
James-Hou22: 50,000 log lines equate to about 2MB of zipped data; 50,000 was chosen due to the 5MB file limit of a ticket. I'll make this a const and provide my reasoning

internal/debugger/tarball.go
James-Hou22: I couldn't find one that we use. Maybe I'm missing something?

internal/debugger/tarball.go
James-Hou22: We use chunks in order to not load the entire log file into memory which can be a lot. 4KB is chosen because the normal disk block size on Linux is 4096 bytes so chunking by that much can provide better I/O performance

James-Hou22: I'll make this a const and provide my reasoning for it though, good catch

cmd/amazon-cloudwatch-agent-debugger/debugger.go
James-Hou22: Yeah, going to remove all emojis

internal/debugger/error_collector.go
James-Hou22: Made this object oriented

internal/debugger/file_check.go
```diff
@@ -106,11 +106,21 @@ func printConfigFilesTable(w io.Writer, configFiles []ConfigFile) {
...
+		switch status {
+		case StatusMultipleFiles:
+			message = fmt.Sprintf("%s: Multiple configuration files found in directory - %s", getDisplayName(file.Path), file.Purpose)
+		case StatusInvalidJSONFormat:
+			message = fmt.Sprintf("%s: Invalid JSON format - %s", getDisplayName(file.Path), file.Purpose)
+		}
+
 		if file.Required {
-			AddConfigError(fmt.Sprintf("%s: %s - %s", getDisplayName(file.Path), file.MissingMsg, file.Purpose))
+			GetErrorCollector().AddError(message)
```
okankoAMZ: I think you should pass ErrorCollector as a pointer as part of the parameters instead of using GetErrorCollector

internal/debugger/log_check.go
```diff
@@ -26,7 +26,7 @@ type AgentLogConfig struct {
 	Message         string `json:"-"`
 }
 
-func CheckLogs(w io.Writer, config map[string]interface{}, compact bool) ([]AgentLogConfig, error) {
+func CheckConfiguredLogs(w io.Writer, config map[string]interface{}, compact bool) ([]AgentLogConfig, error) {
```
okankoAMZ: What are we checking?

internal/debugger/tarball.go
```diff
@@ -66,8 +82,8 @@ func CreateTarball(ssm bool) {
 				fmt.Println("Debug logging enabled successfully.")
 			}
 
-			fmt.Println("\nWaiting 60 seconds to collect debug logs...")
-			showProgressBar(60)
+			fmt.Println("\nWaiting 90 seconds to collect debug logs...")
```
okankoAMZ: I would pass the constant so that it will not be misleading if we update the constant
```suggestion
			fmt.Printf("\nWaiting %d seconds to collect debug logs...",logWaitTime)
```

internal/debugger/tarball.go
```diff
@@ -105,100 +121,27 @@ func CreateTarball(ssm bool) {
...
-
-		if strings.ToLower(answers.EnvironmentChange) == "y" {
-			fmt.Print("Please describe what changed and when: ")
-			envChangeDesc, err := reader.ReadString('\n')
-			if err == nil {
-				answers.EnvironmentChangeDesc = strings.TrimSpace(envChangeDesc)
-			}
-		}
-	}
+func Triage() string {
```
okankoAMZ: nit: Triage what? I think the naming can be more detailed

internal/debugger/tarball.go
```diff
@@ -105,100 +121,27 @@ func CreateTarball(ssm bool) {
...
-
-	return answersContent
+	answers := utils.RunTriage()
+	formattedAnswers := utils.FormatReport(answers)
+	return formattedAnswers
 
 }
 
-func addStringToTarball(tarWriter *tar.Writer, content, tarPath string) error {
+func addTriageToTarball(tarWriter *tar.Writer, content, tarPath string) error {
```
okankoAMZ: you are saying addTriage but its not part of the parameter, I don't understand what this function does

internal/debugger/tarball.go
```diff
@@ -26,12 +27,27 @@ type Answers struct {
...
+	logWaitTime = 90
+
+	// Specifies how many lines are taken from the logs.
+	// A ticket has a maximum file capacity of 5MB.
+	// 50,000 lines compressed equates to ~2MB of data, allowing for a large buffer for other files.
+	logLinesTaken = 50000
+
+	// Specifies the size of logs we process at a time.
+	// Chunking by 4KB can improve I/O efficiency because Linux's disk block size is 4096 bytes.
+	chunkSize = 4096
```
okankoAMZ: nit: 
```suggestion
	logChunkSize = 4096
```

internal/debugger/tarball.go
```diff
@@ -293,7 +235,7 @@ func findTailContent(file *os.File, maxLines int) (int64, []byte, error) {
 
 	pos := fileSize
 	lineCount := 0
-	chunkSize := int64(4096) // 4KB chunks
+	chunkSize := int64(chunkSize)
```
okankoAMZ: nit: why not just cast int64 when you are passing it to the function below?

internal/debugger/utils/triage_questions.go
```diff
@@ -0,0 +1,109 @@
...
+	"bufio"
+	"fmt"
+	"os"
+	"strings"
+)
+
+type Question struct {
+	Text          string
+	AnswerOptions map[string]string
+	Condition     func(answers map[string]string) bool
```
okankoAMZ: what is the bool here, like confirmed they want it, or something else? Can you clarify?

internal/debugger/utils/triage_questions.go
```diff
@@ -0,0 +1,109 @@
...
+	},
+	"additional_info": {
+		Text: "Is there any additional information you would like to add?",
+	},
+}
+
+var QuestionOrder = []string{"occurrence", "env_change", "env_desc", "additional_info"}
+
+func (question *Question) AskQuestion(reader *bufio.Reader) string {
+	fmt.Print("\n", question.Text, " ")
```
okankoAMZ: nit: I recommend using printf if you want to format your output

internal/debugger/utils/triage_questions.go
```diff
@@ -0,0 +1,109 @@
...
+	answer := strings.TrimSpace(strings.ToLower(input))
+
+	if formatted, ok := question.AnswerOptions[answer]; ok {
+		return formatted
+	}
+
+	return answer
+}
+
+func RunTriage() map[string]string {
```
okankoAMZ: why not take QuestionOrder as a parameter?

okankoAMZ (commented):
Left few comments

internal/debugger/tarball.go
James-Hou22: `content` is the content of the triage, I'll make this more clear by renaming this to something like `triageContent`

internal/debugger/tarball.go
James-Hou22: It gets used outside of that function like in the for loop, specifically for the `pos` variable which is defined as `int64` because it effectively is the return value of `stat.Size()` which returns `int64`

internal/debugger/log_check.go
James-Hou22: We're checking if the configured logs exist and are readable. I'm going to change this to `CheckConfiguredLogsExistAndReadable`

internal/debugger/utils/triage_questions.go
James-Hou22: The boolean determines if the question should be asked based on the function's implementation against the past answers, which is in `answers`. I'll write a comment about how this works.

internal/debugger/tarball.go
James-Hou22: Triaging the customer issue. I'll rename to `TriageCustomerIssue`

internal/debugger/utils/triage_questions.go
James-Hou22: Ended up refactoring this whole file

internal/debugger/error_collector.go
```diff
@@ -34,11 +34,13 @@ func (ec *ErrorCollector) PrintErrors() {
...
 
 func (ec *ErrorCollector) printErrorMessages(messages []string, title string) {
+
 	if len(messages) == 0 {
-		return
-	}
-	fmt.Printf("%s (%d):\n", title, len(messages))
-	for _, msg := range messages {
-		fmt.Printf("  %s\n", msg)
+		fmt.Printf("No %s\n", title)
```
okankoAMZ, line 39: why not return here

internal/debugger/file_check.go
```diff
@@ -87,6 +88,7 @@ func printConfigFilesTable(w io.Writer, configFiles []ConfigFile) {
 		utils.RepeatChar('─', statusWidth+2),
 		utils.RepeatChar('─', 50))
 
+	errorCollector := GetErrorCollector()
```
okankoAMZ, line 91: why define it twice?

internal/debugger/tarball.go
```diff
@@ -133,18 +133,10 @@ func CreateTarball(ssm bool) {
...
-	return formattedAnswers
-
-}
-
-func addTriageToTarball(tarWriter *tar.Writer, content, tarPath string) error {
+func addTriageToTarball(tarWriter *tar.Writer, triageContent, tarPath string) error {
 	header := &tar.Header{
 		Name:    tarPath,
-		Size:    int64(len(content)),
+		Size:    int64(len(triageContent)),
```
okankoAMZ, line 139: why int64?

okankoAMZ (commented):
very small comments

internal/debugger/tarball.go
James-Hou22: The tar header requires int64

okankoAMZ:
Could you add screenshots of the wizard changes working?

okankoAMZ (approved):
LGTM!

PR #1747, opened 2025-06-28
Dominic gap windows pr

duhminick:
# Description of the issue

# Description of changes

# License

# Tests

# Requirements
2. Run `make lint`

-------
### Integration Tests


duhminick:
Seems like the Windows range-based offset is inclusive unlike the Linux one. Going to make an update for that.

plugins/inputs/windows_event_log/wineventlog/mock_windows_event_api.go
jefchien, line 1: Wow, this mock is a lot more in-depth than I expected.

nit: Can you make this part of the test file or in a separate package (e.g. `plugins/inputs/windows_event_log/wineventlog/mock`)? Want to make sure this isn't getting built into the agent.

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
```diff
@@ -79,16 +83,26 @@ func NewEventLog(name string, levels []string, logGroupName, logStreamName, rend
 		stateManager:  stateManager,
 		retention:     retention,
 
+		gapsToRead: nil,
+
 		done:          make(chan struct{}),
 		resubscribeCh: make(chan struct{}),
+
+		eventAPI: NewWindowsEventAPI(),
```
jefchien: nit: Would it make sense for this to reference a singleton? It doesn't maintain state right? Could be as simple as
```
var winEventAPI = NewWindowsEventAPI()
```

plugins/inputs/windows_event_log/wineventlog/utils.go
```diff
@@ -50,39 +54,54 @@ func RenderEventXML(eventHandle EvtHandle, renderBuf []byte) ([]byte, error) {
...
+		filters = append(filters, levelsFilter)
+	}
+	filters = append(filters, ignoreOlderThanTwoWeeksFilter)
+	// Add any additional filters (e.g. record IDs)
+	if len(additionalFilters) > 0 {
+		filters = append(filters, strings.Join(additionalFilters, " and "))
 	}
 
-	xml := fmt.Sprintf(eventLogQueryTemplate, path, filterLevels)
+	xml := fmt.Sprintf(eventLogQueryTemplate, path, strings.Join(filters, " and "))
```
jefchien, line 104: Is there a limit to the query length? We're not going to hit it with the range filters, but can see it being a problem later on.

plugins/inputs/windows_event_log/wineventlog/utils.go
```diff
@@ -219,3 +238,17 @@ func insertPlaceholderValues(rawMessage string, evtDataValues []Datum) string {
 	}
 	return sb.String()
 }
+
+func utf16PtrToString(ptr *uint16) string {
```
jefchien, line 242: nit: Seems like this is only used in the mock/tests. Consider moving it there.

plugins/inputs/windows_event_log/wineventlog/utils.go
```diff
@@ -50,39 +54,54 @@ func RenderEventXML(eventHandle EvtHandle, renderBuf []byte) ([]byte, error) {
...
+
+func createWindowsEventFilter(path string, levels []string, additionalFilters ...string) (*uint16, error) {
+	// Add log levels
+	var levelsFilter string
+	formattedLevels := make([]string, len(levels))
+	for i, level := range levels {
+		formattedLevels[i] = fmt.Sprintf(eventLogLevelFilter, level)
+	}
+	if len(formattedLevels) > 0 {
+		levelsFilter = fmt.Sprintf("(%s)", strings.Join(formattedLevels, " or "))
```
jefchien, line 87: nit: Could use a `string.Builder` instead of creating the slice and string. Not a big deal since this isn't called that frequently and there aren't that many levels/filters.

plugins/inputs/windows_event_log/wineventlog/utils.go
```diff
@@ -15,16 +15,20 @@ import (
...
 	"golang.org/x/text/encoding/unicode"
 	"golang.org/x/text/transform"
+
+	"github.com/aws/amazon-cloudwatch-agent/internal/state"
 )
 
 const (
 	bookmarkTemplate         = `<BookmarkList><Bookmark Channel="%s" RecordId="%d" IsCurrent="True"/></BookmarkList>`
-	eventLogQueryTemplate    = `<QueryList><Query Id="0"><Select Path="%s">%s</Select></Query></QueryList>`
+	eventLogQueryTemplate    = `<QueryList><Query Id="0"><Select Path="%s">*[System[%s]]</Select></Query></QueryList>`
```
jefchien, line 28: Why is this template changing?

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
```diff
@@ -100,6 +114,10 @@ func (w *windowsEventLog) SetOutput(fn func(logs.LogEvent)) {
 	w.startOnce.Do(func() { go w.run() })
 }
 
+func (w *windowsEventLog) SetEventAPI(api WindowsEventAPI) {
```
jefchien: nit: Don't need to expose this function since it's only ever used in this package. Could also just set the `eventAPI` field directly.

plugins/inputs/windows_event_log/wineventlog/utils.go
```diff
@@ -15,16 +15,20 @@ import (
...
 	eventLogLevelFilter      = "Level='%s'"
 	eventIgnoreOldFilter     = "TimeCreated[timediff(@SystemTime) &lt;= %d]"
+	eventRangeFilter         = "EventRecordID &gt; %d and EventRecordID &lt; %d"
```
jefchien: So this doesn't include the start or end record IDs?

plugins/inputs/windows_event_log/wineventlog/utils.go
duhminick: I'm using it in two separate tests, utils_test.go and wineventlog_test.go - I figured it would be okay to have it here

plugins/inputs/windows_event_log/wineventlog/utils.go
duhminick: Yeah, it won't include the start or end IDs. I used to have it include the start, but it would've led to one duplicate event

plugins/inputs/windows_event_log/wineventlog/utils.go
duhminick: When building the queries, it always included `*[System[%s]]` so I figured to just move it to the template. It made it a bit easier on building out the filter string too

I wrote the unit tests before making those changes in utils.go to make sure I wasn't going to break anything. That's also why the filters are built in the order they are still

plugins/inputs/windows_event_log/wineventlog/utils.go
duhminick: From what I've read, there isn't a specific character limit, *but* there is a limit to how many expressions you can have. 

> Each XPath that you specify is limited to 32 expressions

(source: https://learn.microsoft.com/en-us/windows/win32/wes/queryschema-schema)

At most, there can be 6 level filters, 1 filter for older than two weeks, 2 filters for the event record ID. In total we have 9. We should be good.

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
```diff
@@ -160,15 +181,28 @@ func (w *windowsEventLog) run() {
...
 					continue
 				}
 				recordNumber, _ := strconv.ParseUint(record.System.EventRecordID, 10, 64)
-				r.Shift(recordNumber)
+				// Need to shift by 1 because the range logic assumes [start, end) whereas Windows events is [start, end]
+				if readingFromGap {
+					r.Shift(recordNumber + 1)
+				} else {
+					r.Shift(recordNumber)
+				}
```
jefchien: Since this is just setting the number for the record that was just read, that means the end has to be inclusive. So each range that's sent is really just 1-wide and is actually just `(start, end]` and the equivalent of `[end, end]`. For example, if we just read ID 100, the range would look like `(99, 100]`. It represents knowing that 100 was processed, but not knowing it 99 was. This is fine when subscribing because of the `EvtSubscribeStartAfterBookmark` flag as mentioned above.

I'm not sure I understand why it won't work the same for the gaps. Isn't it going to try to restore from the state file that was generated during the subscribe? Is this just a matter of the RangeQuery not being upper bound inclusive?

plugins/inputs/windows_event_log/wineventlog/utils.go
jefchien: Should the upper bound be lte instead? From what I can tell, if the state file is
```
0-10,20-30
``` 
It means that 10 was sent, but 20 was not. So when we invert it to get the gaps, it would be (10, 20].

plugins/inputs/windows_event_log/wineventlog/utils.go
duhminick: Hmm, looks like I have an import cycle. What are your thoughts on just moving the mock to the test file?

plugins/inputs/windows_event_log/wineventlog/mock_windows_event_api.go
duhminick: Hmm, looks like I have an import cycle. What are your thoughts on just moving the mock to the test file?

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
```diff
@@ -258,19 +295,48 @@ func (w *windowsEventLog) ResubscribeCh() chan struct{} {
...
+		}
+
+		handle, err := w.openAtRange(r)
+		defer func() {
+			winEventAPI.EvtClose(handle)
+		}()
+		if err != nil {
+			continue
+		}
+		readRecords := w.readFromHandle(handle)
```
jefchien, line 312: nit: Right now, the defer call will happen at the end of `readGaps`. Consider moving this into a separate function so the close happens after each read. 

```go
func (w *windowsEventLog) readGap(r state.Range) []*windowsEventLogRecord
```

plugins/inputs/windows_event_log/wineventlog/utils.go
jefchien: If it's only used in tests, then IMO it should be in a test file.

plugins/inputs/windows_event_log/wineventlog/wineventlog_test.go
```diff
@@ -171,6 +326,69 @@ func checkEvents(t *testing.T, records []*windowsEventLogRecord, substring strin
...
+		text, _ := r.MarshalText()
+		marshalledRanges = append(marshalledRanges, string(text))
+	}
+	return strings.Join(marshalledRanges, ",")
+}
+
+func assertStateFileRange(t *testing.T, fileName string, rl state.RangeList) {
+	content, _ := os.ReadFile(fileName)
+	assert.Contains(t, string(content), marshalRangeList(rl))
+}
```
jefchien, line 341: nit: Instead of marshaling, we could also just call the `Restore` function of the state manager.

plugins/inputs/windows_event_log/wineventlog/wineventlog_test.go
```diff
@@ -108,6 +115,154 @@ func TestReadWithBothSources(t *testing.T) {
...
+			state.NewRange(0, 2),
+			state.NewRange(4, 5),
+		}
+		elog, stateFileName := newTestEventLogWithState(t, NAME, LEVELS, rl)
+		mockAPI := NewMockWindowsEventAPI()
+		winEventAPI = mockAPI
+
+		// This is per EvtHandle hence the necessity to break up these calls
+		// 0, 1, 4 were "sent" previously (should be skipped)
+		mockAPI.AddMockEventsForQuery(createMockEventRecords(0, 2, 5))
```
jefchien, line 186: nit: Think this is incorrect. Should be 1, 2, and 5 were previously sent and should be skipped.

plugins/inputs/windows_event_log/wineventlog/wineventlog_test.go
```diff
@@ -108,6 +115,154 @@ func TestReadWithBothSources(t *testing.T) {
...
+		// Gap records (should be read by gap reading)
+		mockAPI.AddMockEventsForQuery(createMockEventRecords(3, 4))
+
+		elog.Init()
+
+		// Simulate new subscription events arriving
+		mockAPI.SimulateSubscriptionEvents(createMockEventRecords(5, 6, 7, 8))
```
jefchien, line 193: nit: The subscription event should not have 5 in it if we follow our logic and 5 was already sent.

jefchien (approved):

plugins/inputs/windows_event_log/wineventlog/wineventlog.go
movence: This is a good point, and there will be another PR to address this.

duhminick:
Going to follow up on the nits

duhminick:
Follow up PR: https://github.com/aws/amazon-cloudwatch-agent/pull/1777

PR #1751, opened 2025-07-01
Implement prometheus adapter primary functionality

dricross:
# Description of the issue
As part of the histogram support, the OpenTelemetry Prometheus receiver will be outfitted with an adapter processor which translates the metrics into the legacy format output by the legacy Telegraf-based Prometheus receiver.

# Description of changes
This PR adds additional functionality to the adapter receiver including:
* attribute relabeling
* adding additional metric labels
* dropping untyped metrics
* dropping extraneous default metrics
* dropping extraneous attributes
* outputting summaries w/ "detailed metrics"
* converting counters and summaries to deltas

# License

# Tests
Unit tests. Integration will not pass until environment variable expansion issue is resolved and ecsobserver processor is integrated

# Requirements
2. Run `make lint`

-------
### Integration Tests


plugins/processors/prometheusadapter/internal/delta_calculator.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+func NewDeltaCalculator() *DeltaCalculator {
+	return &DeltaCalculator{preDataPoints: mapWithExpiry.NewMapWithExpiry(CacheTTL), lastCleanUpTime: 0}
+}
+
+func isValueValid(v float64) bool {
+	//treat NaN and +/-Inf values as invalid as emf log doesn't support them
+	return !value.IsStaleNaN(v) && !math.IsNaN(v) && !math.IsInf(v, 0)
+}
```
dricross, line 186: This logic was not part of the delta calculator in the receiver, but it was part of the receiver itself: https://github.com/aws/amazon-cloudwatch-agent/blob/main/plugins/inputs/prometheus/metrics_receiver.go#L36

plugins/processors/prometheusadapter/internal/identity.go
```diff
@@ -0,0 +1,27 @@
...
+func (mi *MetricIdentity) getKey() string {
+	b := new(bytes.Buffer)
+	keys := slices.Sorted(maps.Keys(mi.tags))
+	for _, k := range keys {
+		_, _ = fmt.Fprintf(b, "%s=%s,", k, mi.tags[k])
+	}
+	// We assume there won't be same metricName+tags with different metricType, so that it is not necessary to add metricType into uniqKey.
+	_, _ = fmt.Fprintf(b, "metricName=%s,", mi.name)
+	return b.String()
+}
```
dricross, line 27: Replicated logic from `utils` file: https://github.com/aws/amazon-cloudwatch-agent/blob/main/plugins/inputs/prometheus/utils.go#L32-L37

translator/tocwconfig/sampleConfig/prometheus_combined_config_linux.yaml
```diff
@@ -182,11 +182,13 @@ receivers:
                 scrape_protocols:
                     - OpenMetricsText1.0.0
                     - OpenMetricsText0.0.1
+                    - PrometheusText1.0.0
```
dricross, line 185: config file additions are due to the OTel bump in main and rebase onto main for `replace-telegraf-prom-plugin` branch missed this

translator/translate/otel/receiver/prometheus/translator_test.go
```diff
@@ -212,7 +212,9 @@ func TestMetricsEmfTranslator(t *testing.T) {
 					ClientConfig: confighttp.ClientConfig{
 						TLSSetting: configtls.ClientConfig{
 							Config: configtls.Config{
-								CAFile:         defaultTlsCaPath,
+								CAFile:         defaultTLSCaPath,
+								CertFile:       defaultTLSCertPath,
+								KeyFile:        defaultTLSKeyPath,
```
dricross, line 217: rebase onto main for `replace-telegraf-prom-plugin` branch missed this and tests fail

plugins/processors/prometheusadapter/processor.go
```diff
@@ -5,24 +5,195 @@ package prometheusadapter
...
+		sms := rm.ScopeMetrics()
+		sms.RemoveIf(func(sm pmetric.ScopeMetrics) bool {
+			metrics := sm.Metrics()
+			metrics.RemoveIf(func(m pmetric.Metric) bool {
+				return d.processMetric(m, rma)
+			})
+			return metrics.Len() == 0
+		})
+		return sms.Len() == 0
+	})
```
jefchien: Took a while to find it, but I remembered that there was a memory improvement done a while back that involved not creating references to the intermediate metric types https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/14729

```suggestion
	md.ResourceMetrics().RemoveIf(func(rm pmetric.ResourceMetrics) bool {
		rma := rm.Resource().Attributes()
		rm.ScopeMetrics().RemoveIf(func(sm pmetric.ScopeMetrics) bool {
			sm.Metrics().RemoveIf(func(m pmetric.Metric) bool {
				return d.processMetric(m, rma)
			})
			return sm.Metrics().Len() == 0
		})
		return rm.ScopeMetrics().Len() == 0
	})
```

Follows the pattern used by the filterprocessor as well https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/processor/filterprocessor/metrics.go#L173

plugins/processors/prometheusadapter/processor.go
```diff
@@ -5,24 +5,195 @@ package prometheusadapter
...
+
+	d.postprocessFilter(md)
+
 	return md, nil
 }
+
+func (d *prometheusAdapterProcessor) preprocessFilter(md pmetric.Metrics) {
```
jefchien, line 76: nit: Why doesn't this follow the same `RemoveIf` pattern that the main `processMetrics` does? Seems like it'd be useful to remove the resource metrics if none of the metrics are kept.

plugins/processors/prometheusadapter/processor.go
```diff
@@ -5,24 +5,195 @@ package prometheusadapter
...
+
+			const (
+				keep = false
+				drop = true
+			)
+
+			metrics.RemoveIf(func(m pmetric.Metric) bool {
+
+				// for backwards compatibility with legacy Telegraf receiver, we want to drop some extraneous metrics
+				extraneousMetrics := map[string]struct{}{
```
jefchien: Define this set outside of the function. Consider using a `collections.Set[string]`.

plugins/processors/prometheusadapter/processor.go
```diff
@@ -5,24 +5,195 @@ package prometheusadapter
...
+func (d *prometheusAdapterProcessor) postprocessFilter(md pmetric.Metrics) {
+	rms := md.ResourceMetrics()
+	for i := 0; i < rms.Len(); i++ {
+		rm := rms.At(i)
+		rma := rm.Resource().Attributes()
+
+		// Remove extraneous resource attributes
+		// This must be done after processing metrics so that the resource attributes can be moved to datapoint attributes
+		rma.RemoveIf(func(key string, value pcommon.Value) bool {
+			extraneousAttributes := map[string]struct{}{
```
jefchien: This set as well. Can we use semconv variables instead of defining them ourselves?

plugins/processors/prometheusadapter/processor.go
dricross: Looks better too, thanks.

plugins/processors/prometheusadapter/processor.go
dricross: Makes sense

plugins/processors/prometheusadapter/processor.go
dricross: Does it make a performance difference? I wouldn't expect so. I like having the map close to where we use it. Makes it easy to read rather than jumping around the file.

Didn't know we had our own set. Very cool.

plugins/processors/prometheusadapter/processor.go
dricross: Sure thing

plugins/processors/prometheusadapter/processor.go
```diff
@@ -5,24 +5,194 @@ package prometheusadapter
...
+				)
+
+				// for backwards compatibility with legacy Telegraf receiver, we want to drop some extraneous metrics
+				extraneousMetrics := collections.Set[string]{
+					"scrape_duration_seconds":               {},
+					"scrape_samples_post_metric_relabeling": {},
+					"scrape_samples_scraped":                {},
+					"scrape_series_added":                   {},
+					"up":                                    {},
+				}
```
jefchien: I meant as a package level variable. This is a static set, so we shouldn't define and allocate a new map on each loop.

```suggestion
	var scrapeStatMetrics = collections.NewSet("scrape_duration_seconds", "scrape_samples_post_metric_relabeling", "scrape_samples_scraped", "scrape_series_added", "up")
```

plugins/processors/prometheusadapter/processor.go
```diff
@@ -5,24 +5,194 @@ package prometheusadapter
...
+				if _, ok := extraneousMetrics[m.Name()]; ok {
```
jefchien: nit: Can optionally use the `Contains` function, which basically does the same thing.
```suggestion
				if extraneousMetrics.Contains(m.Name()) {
```

plugins/processors/prometheusadapter/processor.go
dricross: Won't the compiler make that optimization for us and not actually allocate a new map each loop?

plugins/processors/prometheusadapter/internal/delta_calculator.go
```diff
@@ -0,0 +1,186 @@
...
+			}
+
+			dc.preDataPoints.Set(metricKey, dataPoint{value: curVal, timestamp: curTime})
+
+			return dropValue
+		})
+
+	case pmetric.MetricTypeSummary:
+		dps := m.Summary().DataPoints()
+		dps.RemoveIf(func(dp pmetric.SummaryDataPoint) bool {
```
Paramadon, line 104: nit: Idk if I like the RemoveIf pattern, but I guess it needs to be like that...

PR #1762, opened 2025-07-08
Adjust log input and output plugins to support 1MB log lines and resolve issues with truncation logic

TravisStark:
# Description of the issue
CloudWatch Logs has increased the maximum log event size from 256KB to 1MB. The CloudWatch Agent needs to be updated to handle this new limit correctly. Previously, logs exceeding 256KB were either truncated with a "[Truncated...]" suffix or split inconsistently between the input and output plugins.

# Description of changes
This PR updates the CloudWatch Agent to properly handle the new 1MB log event size limit:

1. **Increased Default Size Limit**:
   - Changed the default maximum log event size from 256KB to 1MB in both input and output plugins
   - Kept the per-event header size at 200 bytes to avoid regressions.

2. **Improved Log Splitting**:
   - Enhanced the input plugin to cleanly split oversized logs into multiple events
   - All parts maintain the same timestamp for correlation
  
Oversized logs are now properly split into appropriately sized chunks and sent as separate events, preserving all the original content.

# License

# Tests
Tested the changes with logs of various sizes:
- Small logs (<1KB)
- Medium logs (10KB, 50KB, 100KB)
- Logs near the old limit (250KB, 256KB, 260KB)
- Logs between old and new limits (512KB, 768KB)
- Logs near the new limit (1MB-1KB, 1MB-100B, exactly 1MB)
- Logs exceeding the new limit (1MB+100B, 1MB+1KB, 2MB)

All logs were properly processed and sent to CloudWatch Logs. Logs exceeding 1MB were correctly split into multiple events with no data loss.

Prevents this case:   
>    - Customer has 500KiB log line
>    - With a limit of 256KiB, the input plugin reads in
>      - The first 256KiB of the line as one log event
>      - The second 244KiB as another log event
>    - Both are sent to the output plugin
>    - The output plugin has a slightly different limit: 256KiB - per event header
>      - The first 256KiB event is larger than the limit, so the message is truncated
>      - The second 244KiB event is sent as is
>    - Results in CloudWatch Logs getting the 500KiB log line in two events with 
      [Truncated...] at the end of the first one

### Sample logs with fix
Screenshot 2025-07-14 at 3 56 58 PM
Screenshot 2025-07-16 at 4 13 36 PM

File used:
[precise_test_logs.txt](https://github.com/user-attachments/files/21221783/precise_test_logs.txt)

Screenshot 2025-07-09 at 9 51 21 PM

### 2-character issue (before fix):
Screenshot 2025-07-14 at 4 01 00 PM

### Stress Test Runs
- https://github.com/aws/amazon-cloudwatch-agent/actions/runs/16325250972/job/46118735455
- https://github.com/aws/amazon-cloudwatch-agent/actions/runs/16325662037/job/46118745268

# Requirements
2. Run `make lint`

-------
### Integration Tests

### DIFF
--- a/plugins/inputs/logfile/constants/constants.go
+++ b/plugins/inputs/logfile/constants/constants.go
@@ -9,7 +9,4 @@ const (
 
 	// PerEventHeaderBytes is the bytes required for metadata for each log event
 	PerEventHeaderBytes = 200
-
-	// DefaultTruncateSuffix is the suffix added to truncated log messages
-	DefaultTruncateSuffix = "[Truncated...]"
 )
--- a/plugins/inputs/logfile/tailersrc.go
+++ b/plugins/inputs/logfile/tailersrc.go
@@ -70,7 +70,6 @@ type tailerSrc struct {
 	timestampFn     func(string) (time.Time, string)
 	enc             encoding.Encoding
 	maxEventSize    int
-	truncateSuffix  string
 	retentionInDays int
 
 	outputFn           func(logs.LogEvent)
@@ -98,7 +97,6 @@ func NewTailerSrc(
 	timestampFn func(string) (time.Time, string),
 	enc encoding.Encoding,
 	maxEventSize int,
-	truncateSuffix string,
 	retentionInDays int,
 	backpressureMode logscommon.BackpressureMode,
 ) *tailerSrc {
@@ -116,7 +114,6 @@ func NewTailerSrc(
 		timestampFn:        timestampFn,
 		enc:                enc,
 		maxEventSize:       maxEventSize,
-		truncateSuffix:     truncateSuffix,
 		retentionInDays:    retentionInDays,
 		backpressureFdDrop: !autoRemoval && backpressureMode == logscommon.LogBackpressureModeFDRelease,
 		done:               make(chan struct{}),
--- a/plugins/inputs/logfile/tailersrc_test.go
+++ b/plugins/inputs/logfile/tailersrc_test.go
@@ -83,7 +83,6 @@ func TestTailerSrc(t *testing.T) {
 		parseRFC3339Timestamp,
 		nil, // encoding
 		constants.DefaultMaxEventSize,
-		constants.DefaultTruncateSuffix,
 		1,
 		"",
 	)
@@ -194,7 +193,6 @@ func TestEventDoneCallback(t *testing.T) {
 		parseRFC3339Timestamp,
 		nil, // encoding
 		constants.DefaultMaxEventSize,
-		constants.DefaultTruncateSuffix,
 		1,
 		"",
 	)
@@ -425,7 +423,6 @@ func setupTailer(t *testing.T, multiLineFn func(string) bool, maxEventSize int, 
 		parseRFC3339Timestamp,
 		nil, // encoding
 		maxEventSize,
-		constants.DefaultTruncateSuffix,
 		1,
 		backpressureDrop,
 	)
--- a/plugins/outputs/cloudwatchlogs/internal/pusher/batch.go
+++ b/plugins/outputs/cloudwatchlogs/internal/pusher/batch.go
@@ -30,7 +30,7 @@ const (
 	// A batch of log events in a single request cannot span more than 24 hours. Otherwise, the operation fails.
 	batchTimeRangeLimit = 24 * time.Hour
 	// Suffix to indicate that a message has been truncated
-	truncationSuffix = constants.DefaultTruncateSuffix
+	defaultTruncationSuffix = "[Truncated...]"
 )
 
 // validateAndTruncateMessage ensures events don't exceed limit before we send to CloudWatch
@@ -42,7 +42,7 @@ func validateAndTruncateMessage(message string) string {
 	}
 
 	// Truncate the message and add a suffix to indicate truncation
-	truncatedMessage := message[:maxMessageSize-len(truncationSuffix)] + truncationSuffix
+	truncatedMessage := message[:maxMessageSize-len(defaultTruncationSuffix)] + defaultTruncationSuffix
 	return truncatedMessage
 }
 
--- a/plugins/inputs/logfile/fileconfig.go
+++ b/plugins/inputs/logfile/fileconfig.go
@@ -152,9 +152,6 @@ func (config *FileConfig) init() error {
 	if config.MaxEventSize == 0 {
 		config.MaxEventSize = constants.DefaultMaxEventSize - constants.PerEventHeaderBytes
 	}
-	if config.TruncateSuffix == "" {
-		config.TruncateSuffix = constants.DefaultTruncateSuffix
-	}
 	if config.RetentionInDays == 0 {
 		config.RetentionInDays = -1
 	}
--- a/plugins/inputs/logfile/logfile.go
+++ b/plugins/inputs/logfile/logfile.go
@@ -266,7 +266,6 @@ func (t *LogFile) FindLogSrc() []logs.LogSrc {
 				fileconfig.timestampFromLogLine,
 				fileconfig.Enc,
 				fileconfig.MaxEventSize,
-				fileconfig.TruncateSuffix,
 				fileconfig.RetentionInDays,
 				fileconfig.BackpressureMode,
 			)
--- a/plugins/outputs/cloudwatchlogs/internal/pusher/batch_test.go
+++ b/plugins/outputs/cloudwatchlogs/internal/pusher/batch_test.go
@@ -230,7 +230,7 @@ func TestEventValidation_Over1MB(t *testing.T) {
 	event := newStatefulLogEvent(time.Now(), oversizeMessage, nil, nil)
 	// The total length should still be maxMessageSize
 	assert.Equal(t, maxMessageSize, len(event.message))
-	assert.Equal(t, oversizeMessage[:maxMessageSize-len(truncationSuffix)]+truncationSuffix, event.message)
+	assert.Equal(t, oversizeMessage[:maxMessageSize-len(defaultTruncationSuffix)]+defaultTruncationSuffix, event.message)
 }
 
 func TestEventValidation_Between256KBand1MB(t *testing.T) {
@@ -262,7 +262,7 @@ func TestValidateAndTruncateMessage(t *testing.T) {
 		{
 			name:           "Over limit",
 			input:          strings.Repeat("a", maxMessageSize+1000),
-			expectedOutput: strings.Repeat("a", maxMessageSize-len(truncationSuffix)) + truncationSuffix,
+			expectedOutput: strings.Repeat("a", maxMessageSize-len(defaultTruncationSuffix)) + defaultTruncationSuffix,
 		},
 	}
 


plugins/inputs/logfile/tail/tail_test.go
```diff
@@ -197,3 +199,83 @@ func tearDown(tmpfile *os.File) {
 	exitOnDeletionCheckDuration = time.Minute
 	exitOnDeletionWaitDuration = 5 * time.Minute
 }
+
+func TestUtf16LineSize(t *testing.T) {
+	tmpfile, err := os.CreateTemp("", "")
+	if err != nil {
+		t.Fatal(err)
+	}
```
dricross: nit: use `require.NoError(t, err)` instead. Couple of other places too

plugins/inputs/logfile/tail/tail.go
```diff
@@ -306,8 +320,12 @@ func (tail *Tail) readlineUtf16() (string, error) {
 			}
 			cur = append(cur, nextByte)
 		}
-		// 262144 => 256KB
-		if resSize+len(cur) >= 262144 {
+		// Use MaxLineSize if configured, otherwise default to 1MB
+		maxSize := 1024 * 1024 // 1MB default
```
dricross: Can this pick up the default from defaultMaxEventSize? Same for a couple of other places where we specify 1024 * 1024

plugins/inputs/logfile/tail/tail.go
```diff
@@ -244,12 +244,26 @@ func (tail *Tail) readLine() (string, error) {
...
+		// Truncate to MaxLineSize if the line is too long
+		maxSize := tail.MaxLineSize
+		if maxSize <= 0 {
+			// Default to 1MB (1024 * 1024 bytes)
+			maxSize = 1024 * 1024
+		}
+
+		if len(line) > maxSize {
+			line = line[:maxSize]
+		}
```
dricross: Is the truncated part lost? I thought we wanted to split it into separate log events?

plugins/inputs/logfile/tail/tail.go
```diff
@@ -244,12 +244,26 @@ func (tail *Tail) readLine() (string, error) {
 
 	line, err := tail.readSlice('\n')
 	if err == bufio.ErrBufferFull {
+		// Truncate to MaxLineSize if the line is too long
+		maxSize := tail.MaxLineSize
+		if maxSize <= 0 {
+			// Default to 1MB (1024 * 1024 bytes)
+			maxSize = 1024 * 1024
```
dricross: I think the log event really needs to be capped at 1MB - 24 bytes for the overhead

plugins/inputs/logfile/tail/tail.go
```diff
@@ -533,12 +551,13 @@ func (tail *Tail) waitForChanges() error {
...
-		// add 2 to account for newline characters
-		tail.reader = bufio.NewReaderSize(tail.file, tail.MaxLineSize+2)
-	} else {
-		tail.reader = bufio.NewReader(tail.file)
+	maxSize := tail.MaxLineSize
+	if maxSize <= 0 {
+		// Default to 1MB (1024 * 1024 bytes)
+		maxSize = 1024 * 1024
 	}
+	// Use exactly maxSize for the buffer to avoid the trailing 2 characters issue
```
dricross: Do we know what the "trailing 2 characters issue" is? Do we need to keep 2 extra characters for some reason?

plugins/inputs/logfile/tail/tail.go
```diff
@@ -533,12 +551,13 @@ func (tail *Tail) waitForChanges() error {
...
-	if tail.MaxLineSize > 0 {
-		// add 2 to account for newline characters
-		tail.reader = bufio.NewReaderSize(tail.file, tail.MaxLineSize+2)
-	} else {
-		tail.reader = bufio.NewReader(tail.file)
+	maxSize := tail.MaxLineSize
+	if maxSize <= 0 {
+		// Default to 1MB (1024 * 1024 bytes)
+		maxSize = 1024 * 1024
 	}
```
dricross: There's a slight change in behavior here that I don't know the consequence of. Previously, if `tail.MaxLinesize` was 0, the bufio reader wouldn't have a pre-defined size. With these changes, it will always have a predefined size. What is the consequence or need of this?

plugins/inputs/logfile/tail/tail.go
```diff
@@ -563,7 +582,7 @@ func (tail *Tail) sendLine(line string, offset int64) bool {
 	now := time.Now()
 	lines := []string{line}
 
-	// Split longer lines
+	// Split longer lines using partitionString
```
dricross: nit: this addition isn't super helpful

plugins/inputs/logfile/tail/tail.go
TravisStark: Will add a screenshot of this issue.  It was to protect against an edge case, but causes another issue where the initial split is too large and causes the line to be 2 chars too long.

plugins/outputs/cloudwatchlogs/internal/pusher/batch.go
```diff
@@ -22,12 +22,33 @@ const (
 	reqSizeLimit = 1024 * 1024
 	// The maximum number of log events in a batch.
 	reqEventsLimit = 10000
-	// The bytes required for metadata for each log event.
-	perEventHeaderBytes = 200
```
dricross: Do we know why we had 200 before? That seems very weird that we'd get that so wrong. Perhaps it used to be 200 and FE optimized it to 26?

plugins/outputs/cloudwatchlogs/internal/pusher/batch.go
```diff
@@ -22,12 +22,33 @@ const (
...
+	// Safety buffer to prevent edge cases with logs at exactly the size limit
+	// Using the same size as perEventHeaderBytes for consistency
+	safetyBufferBytes = perEventHeaderBytes
 	// A batch of log events in a single request cannot span more than 24 hours. Otherwise, the operation fails.
 	batchTimeRangeLimit = 24 * time.Hour
+	// Suffix to indicate that a message has been truncated
+	truncationSuffix = "[Truncated...]"
 )
 
+// validateAndTruncateMessage ensures events don't exceed 1MB limit
```
dricross: nit: in case 1MB limit ever changes again, don't mention the specific number here

plugins/outputs/cloudwatchlogs/internal/pusher/batch.go
TravisStark: Will leave this at 200 for this PR

plugins/inputs/logfile/tailersrc.go
```diff
@@ -239,9 +239,10 @@ func (ts *tailerSrc) runTail() {
 				msgBuf.WriteString("\n")
 				msgBuf.WriteString(text)
 				if msgBuf.Len() > ts.maxEventSize {
-					msgBuf.Truncate(ts.maxEventSize - len(ts.truncateSuffix))
-					msgBuf.WriteString(ts.truncateSuffix)
+					fo.ShiftInt64(line.Offset)
+					continue
```
movence: This seems to be the right approach to me, but it's a change in the exsting behavior at the same time IMO. Can we double check this is acceptable?

movence: is len check required if shift by `line.Offset` is the dafult behavior for this else block?

plugins/outputs/cloudwatchlogs/internal/pusher/batch.go
```diff
@@ -22,12 +23,29 @@ const (
 	reqSizeLimit = 1024 * 1024
 	// The maximum number of log events in a batch.
 	reqEventsLimit = 10000
-	// The bytes required for metadata for each log event.
+	// The bytes required for metadata for each log event
 	perEventHeaderBytes = 200
+	// Maximum size for individual log events (1MB)
+	maxEventPayloadBytes = 1024 * 1024
```
movence: nit: use `constants.DefaultMaxEventSize` instead?

plugins/outputs/cloudwatchlogs/internal/pusher/batch.go
```diff
@@ -22,12 +23,29 @@ const (
...
 	// A batch of log events in a single request cannot span more than 24 hours. Otherwise, the operation fails.
 	batchTimeRangeLimit = 24 * time.Hour
+	// Suffix to indicate that a message has been truncated
+	truncationSuffix = constants.DefaultTruncateSuffix
 )
 
+// validateAndTruncateMessage ensures events don't exceed limit before we send to CloudWatch
+func validateAndTruncateMessage(message string) string {
+	maxMessageSize := maxEventPayloadBytes - perEventHeaderBytes
```
movence, line 38: guess this could be another const?

plugins/inputs/logfile/tailersrc.go
```diff
@@ -351,6 +352,27 @@ func (ts *tailerSrc) cleanUp() {
...
+					ts.outputFn(e)
+				}
+			default:
+				break
+			}
+		}
+	}
+
+	// Ensure file is closed
+	if ts.tailer != nil && !ts.tailer.IsFileClosed() {
```
movence: why close a file here when `tailerSrc` already takes care of it with `cleanUp()`?

plugins/inputs/logfile/tailersrc.go
```diff
@@ -351,6 +352,27 @@ func (ts *tailerSrc) cleanUp() {
 			log.Printf("I! [logfile] Successfully removed file %v with auto_removal feature", ts.tailer.Filename)
 		}
 	}
+
+	// Ensure all buffers are flushed
+	if ts.buffer != nil {
```
movence: I'm curious why we are flushing logs here?

TravisStark: The changes in the cleanUp function are from some testing I was doing based on the stress test runs this morning

plugins/inputs/logfile/tailersrc.go
TravisStark: removed this change

plugins/inputs/logfile/tailersrc.go
TravisStark: removed changes here

plugins/outputs/cloudwatchlogs/internal/pusher/batch.go
TravisStark: I'll leave this as a calculated variable for now

plugins/inputs/logfile/tailersrc_test.go
```diff
@@ -81,75 +82,71 @@ func TestTailerSrc(t *testing.T) {
...
+	// Verify the content of the events
+	for i, evt := range receivedEvents {
+		msg := evt.Message()
+		expectedMsg := lines[i]
+
+		require.Equal(t, expectedMsg, msg, fmt.Sprintf("Log Event %d doesn't match exactly", i))
 	}
 
+	// Verify we processed the expected number of events
+	require.Equal(t, len(lines), len(receivedEvents), "Should have processed all events")
```
dricross: nit: this is a duplicate of line 133

dricross (approved):

plugins/inputs/logfile/tailersrc_test.go
TravisStark: removed duplicate line

dricross (dismissed):

dricross (dismissed):

dricross (approved):

plugins/inputs/logfile/tail/tail.go
```diff
@@ -306,8 +307,12 @@ func (tail *Tail) readlineUtf16() (string, error) {
 			}
 			cur = append(cur, nextByte)
 		}
-		// 262144 => 256KB
-		if resSize+len(cur) >= 262144 {
+		// Use MaxLineSize if configured, otherwise use DefaultMaxEventSize
+		maxSize := constants.DefaultMaxEventSize // Use the constant defined in constants package
```
okankoAMZ, line 311: why comment here?

plugins/inputs/logfile/constants/constants.go
```diff
@@ -0,0 +1,12 @@
...
+// SPDX-License-Identifier: MIT
+
+package constants
+
+const (
+	// DefaultMaxEventSize is the default maximum size for log events (1MB)
+	DefaultMaxEventSize = 1024 * 1024
+
+	// PerEventHeaderBytes is the bytes required for metadata for each log event
+	PerEventHeaderBytes = 200
```
okankoAMZ, line 11: How did you decide on this number?

plugins/inputs/logfile/tail/tail.go
TravisStark: Will remove in another CR.

plugins/inputs/logfile/constants/constants.go
TravisStark: There was memory increase when it was set to 26 a couple months ago

PR #1768, opened 2025-07-10
Fixing prometheus env configuration expansion problem

Paramadon:
## Description:

OpenTelemetry Prometheus receiver's regex replacement syntax ($1, ${1}, etc.) conflicts with the CloudWatch Agent's environment variable expansion, causing startup errors when processing Prometheus configurations with metric relabeling rules.

## Description of change
Modified the Prometheus translator to escape $ characters (replacing $ with $$$$) in the configuration before parsing. This double escaping handles the two-step variable expansion process in the agent while preserving the functionality of both regex capture groups and environment variables.

### DIFF
--- a/translator/tocwconfig/tocwconfig_test.go
+++ b/translator/tocwconfig/tocwconfig_test.go
@@ -54,6 +54,9 @@ var prometheusConfig string
 //go:embed sampleConfig/prometheus_cwa_config.yaml
 var prometheusPMDConfig string
 
+//go:embed sampleConfig/prometheus_expansion.yaml
+var prometheusExpansionConfig string
+
 type testCase struct {
 	filename        string
 	targetPlatform  string
@@ -431,7 +434,6 @@ func TestPrometheusConfig(t *testing.T) {
 	checkTranslation(t, "prometheus_config_windows", "windows", nil, "", tokenReplacements)
 }
 
-
 func TestPrometheusPMDConfig(t *testing.T) {
 	resetContext(t)
 	context.CurrentContext().SetRunInContainer(true)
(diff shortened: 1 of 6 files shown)


translator/tocwconfig/tocwconfig_test.go
```diff
@@ -946,7 +967,7 @@ func verifyToYamlTranslation(t *testing.T, input interface{}, expectedYamlFilePa
 		yamlStr := toyamlconfig.ToYamlConfig(yamlConfig)
 		require.NoError(t, yaml.Unmarshal([]byte(yamlStr), &actual))
 
-		// assert.NoError(t, os.WriteFile(expectedYamlFilePath, []byte(yamlStr), 0644)) // useful for regenerating YAML
+		//assert.NoError(t, os.WriteFile(expectedYamlFilePath, []byte(yamlStr), 0644)) // useful for regenerating YAML
```
musa-asad: nit: avoid touching these

translator/translate/otel/receiver/prometheus/translator.go
```diff
@@ -117,3 +126,7 @@ func (t *translator) Translate(conf *confmap.Conf) (component.Config, error) {
 
 	return cfg, nil
 }
+
+func escapeRegexReplacements(content []byte) []byte {
+	return []byte(strings.ReplaceAll(string(content), "$", "$$$$"))
```
musa-asad: I think this makes sense. We did confirm that in any case `$$$$` is present, that it will always return back to `$`, right? Just want to be aware of possible edge cases where `expandconverter` or `confmap` don't cover.

translator/translate/otel/receiver/prometheus/translator.go
```diff
@@ -64,10 +64,15 @@ func (t *translator) ID() component.ID {
 
 func (t *translator) Translate(conf *confmap.Conf) (component.Config, error) {
 	cfg := t.factory.CreateDefaultConfig().(*prometheusreceiver.Config)
-	configPathKey := common.ConfigKey(t.configKey, common.PrometheusConfigPathKey)
-
-	if !conf.IsSet(configPathKey) {
-		return nil, &common.MissingKeyError{ID: t.ID(), JsonKey: configPathKey}
+	var configPathKey string
+	configPathKeyLogs := common.ConfigKey(common.LogsKey, common.MetricsCollectedKey, common.PrometheusKey, common.PrometheusConfigPathKey)
+	configPathKeyMetrics := common.ConfigKey(common.MetricsKey, common.MetricsCollectedKey, common.PrometheusKey, common.PrometheusConfigPathKey)
```
musa-asad, line 69: For clarity, are we supporting the metrics path for prometheus now? Doesn't prometheus only ship EMF metrics?

translator/translate/otel/receiver/prometheus/translator.go
```diff
@@ -64,10 +64,15 @@ func (t *translator) ID() component.ID {
...
+	if conf.IsSet(configPathKeyMetrics) {
+		configPathKey = configPathKeyMetrics
+	} else if conf.IsSet(configPathKeyLogs) {
+		configPathKey = configPathKeyLogs
+	} else {
+		return nil, &common.MissingKeyError{ID: t.ID(), JsonKey: common.PrometheusConfigPathKey}
```
musa-asad: Why do we need to do change this? Shouldn't this be taken care of under https://github.com/aws/amazon-cloudwatch-agent/blob/64059fc9115fb799540fe3678750982733bb46b3/translator/translate/otel/pipeline/prometheus/translator.go#L120?

translator/translate/otel/receiver/prometheus/translator.go
dricross: I am slightly concerned about replacing `$` in the `regex` field, although the only valid place is at end of the regular expression and adding additional `$` symbols won't change the behavior of the regular expression. So it'll look weird in the agent yaml configuration but shouldn't have any impact. Not sure if there are some other edge cases though.

translator/translate/otel/receiver/prometheus/translator.go
Paramadon: It is broken currently in this branch. Also, we can have prometheus receiver for either logs of metrics section so we need to account for both.

dricross (approved):

TravisStark (approved):

PR #1776, opened 2025-07-15
[EKS Performance Testing] Implement github action for scaling up and down eks performance cluster

JayPolanco:
# Description of the issue
We are working towards implementing EKS performance testing on a large scale EKS cluster. The cluster is already created and static.

# Description of changes
Implemented a github workflow that does the following:
1. Connects to EKS cluster
2. Uses aws eks cli to scale the nodegroups up and down. 

# License

# Tests
Success Case: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/16332007802/job/46136527659
Failure Case: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/16331406386/job/46134643323

Defaults and variables work for both.

# Requirements
4. Run `make lint`

-------
### Integration Tests

### DIFF
--- a/.github/workflows/eks-performance-cluster-scaling.yml
+++ b/.github/workflows/eks-performance-cluster-scaling.yml
@@ -23,10 +23,15 @@ on:
         required: true
         type: number
         default: 500
+      node_group_count:
+        description: 'Count of node groups'
+        type: number
+        default: 10
 
 env:
   AWS_REGION: ${{ inputs.region || 'us-west-2' }}
   CLUSTER_NAME: ${{ inputs.cluster_name || 'eks-performance' }}
+  NODE_GROUP_COUNT: ${{ inputs.node_group_count || 10 }}
   DESIRED_CAPACITY_PER_NODEGROUP: ${{ inputs.desired_capacity_per_nodegroup || 500 }}
   TERRAFORM_AWS_ASSUME_ROLE: ${{ vars.TERRAFORM_AWS_ASSUME_ROLE }}
   TERRAFORM_AWS_ASSUME_ROLE_DURATION: 3600  # 1 hour duration
@@ -68,7 +73,7 @@ jobs:
         run: |
           echo "Starting scale UP operation with desired capacity: 500"
           
-          for i in {1..10}; do
+          for i in $(seq 1 $NODE_GROUP_COUNT); do
             echo "Scaling node group: eks-performance-node-${i} to 500"
             aws eks update-nodegroup-config \
               --cluster-name eks-performance \
@@ -85,7 +90,7 @@ jobs:
         run: |
           echo "Starting scale DOWN operation with desired capacity: 0"
           
-          for i in {1..10}; do
+          for i in $(seq 1 $NODE_GROUP_COUNT); do
             echo "Scaling node group: eks-performance-node-${i} to 0"
             aws eks update-nodegroup-config \
               --cluster-name eks-performance \
@@ -102,7 +107,7 @@ jobs:
         run: |
           echo "Starting manual scaling operation with desired capacity: ${{ env.DESIRED_CAPACITY_PER_NODEGROUP }}"
           
-          for i in {1..10}; do
+          for i in $(seq 1 $NODE_GROUP_COUNT); do
             echo "Scaling node group: eks-performance-node-${i} to ${{ env.DESIRED_CAPACITY_PER_NODEGROUP }}"
             aws eks update-nodegroup-config \
               --cluster-name eks-performance \
@@ -122,9 +127,9 @@ jobs:
           
           # Determine expected count based on trigger type
           if [ "${{ github.event.schedule }}" = "0 21 * * 1" ]; then
-            EXPECTED_NODE_COUNT=$((10 * 0))
+            EXPECTED_NODE_COUNT=$((${{ env.NODE_GROUP_COUNT }} * 0))
           else
-            EXPECTED_NODE_COUNT=$((10 * ${{ env.DESIRED_CAPACITY_PER_NODEGROUP }}))
+            EXPECTED_NODE_COUNT=$((${{ env.NODE_GROUP_COUNT }} * ${{ env.DESIRED_CAPACITY_PER_NODEGROUP }}))
           fi
           
           echo "Expected total nodes: ${EXPECTED_NODE_COUNT}"
--- a/.github/workflows/eks-performance-cluster-scaling.yml
+++ b/.github/workflows/eks-performance-cluster-scaling.yml
@@ -18,7 +18,7 @@ on:
         required: true
         type: string
         default: 'eks-performance'
-      desired_capacity:
+      desired_capacity_per_nodegroup:
         description: 'Desired capacity for each node group'
         required: true
         type: number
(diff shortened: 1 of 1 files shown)


.github/workflows/eks-performance-cluster-scaling.yml
```diff
@@ -0,0 +1,140 @@
...
+        description: 'AWS Region'
+        required: true
+        type: string
+        default: 'us-west-2'
+      cluster_name:
+        description: 'EKS Cluster Name'
+        required: true
+        type: string
+        default: 'eks-performance'
+      desired_capacity:
```
okankoAMZ: do we want this as env instead? do we ever want to configure this?

.github/workflows/eks-performance-cluster-scaling.yml
```diff
@@ -0,0 +1,140 @@
...
+      - uses: actions/checkout@v3
+        with:
+          repository: ${{ env.CWA_GITHUB_TEST_REPO_NAME }}
+          ref: ${{ env.CWA_GITHUB_TEST_REPO_BRANCH }}
+
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{ env.TERRAFORM_AWS_ASSUME_ROLE}}
+          aws-region: ${{ inputs.region || 'us-west-2' }}
```
okankoAMZ, line 58: we don't need the `|| "us-west-2"` we should just get it from region

TravisStark (approved):

.github/workflows/eks-performance-cluster-scaling.yml
```diff
@@ -0,0 +1,140 @@
...
+              --nodegroup-name eks-performance-node-${i} \
+              --region us-west-2 \
+              --scaling-config desiredSize=0
+          
+            echo "Waiting 1 minute before scaling next node group..."
+            sleep 60
+          done
+
+      - name: Scale node groups (Manual)
+        if: github.event_name == 'workflow_dispatch'
```
okankoAMZ, line 105: we might be able to use github caches to scale up and down instead of schedule if flag

.github/workflows/eks-performance-cluster-scaling.yml
```diff
@@ -0,0 +1,140 @@
...
+        run: |
+          aws eks update-kubeconfig --name ${{ env.CLUSTER_NAME }} --region ${{ env.AWS_REGION }}
+          kubectl get nodes -o wide
+
+      - name: Scale up node groups (Sunday)
+        if: github.event.schedule == '0 9 * * 0'
+        run: |
+          echo "Starting scale UP operation with desired capacity: 500"
+          
+          for i in {1..10}; do
```
okankoAMZ: can we make the `10` here a variable so that we can change it later if we need to scale more.

okankoAMZ (dismissed):
No blocker, but left some comments

.github/workflows/eks-performance-cluster-scaling.yml
musa-asad: Does it have a default?

.github/workflows/eks-performance-cluster-scaling.yml
```diff
@@ -0,0 +1,140 @@
...
+          # Determine expected count based on trigger type
+          if [ "${{ github.event.schedule }}" = "0 9 * * 0" ]; then
+            EXPECTED_NODE_COUNT=$((10 * 500))
+          elif [ "${{ github.event.schedule }}" = "0 21 * * 1" ]; then
+            EXPECTED_NODE_COUNT=$((10 * 0))
+          else
+            EXPECTED_NODE_COUNT=$((10 * ${{ env.DESIRED_CAPACITY }}))
+          fi
+          
+          echo "Expected total nodes: ${EXPECTED_NODE_COUNT}"
```
musa-asad: This is just here to halt testing in case there's a set up issue?

.github/workflows/eks-performance-cluster-scaling.yml
```diff
@@ -0,0 +1,140 @@
...
+        run: |
+          echo "Waiting 20 minutes for scaling operations to complete..."
+          sleep 1200
+          
+          echo "Validating total number of nodes in the cluster..."
+          ACTUAL_NODE_COUNT=$(kubectl get nodes --no-headers | wc -l)
+          
+          # Determine expected count based on trigger type
+          if [ "${{ github.event.schedule }}" = "0 9 * * 0" ]; then
+            EXPECTED_NODE_COUNT=$((10 * 500))
```
musa-asad: Should this 500 be hard coded?

movence: think this can be simplified to check `desired=0` for the scale down schedule then must match `env.DESIRED_CAPACITY * 10` all other cases.

.github/workflows/eks-performance-cluster-scaling.yml
movence: nit: `desired_capacity_per_nodegroup` for clarity

.github/workflows/eks-performance-cluster-scaling.yml
JayPolanco: I guess we could apply this logic to other regions, but we would have to upgrade our service quotas in those region. Since the quotas aren't in place having a default was useful

.github/workflows/eks-performance-cluster-scaling.yml
JayPolanco: How so?

.github/workflows/eks-performance-cluster-scaling.yml
JayPolanco: There are only 10 node groups. As this is a static cluster it can't have more than 10 unless we make changes to the cluster itself. If we decide that we need more int he future, it's an easy change

.github/workflows/eks-performance-cluster-scaling.yml
JayPolanco: Yes, because this test is specifically for large 5k node clusters. at 10 node groups, 500 is the most we can do for our service quotas. If for w.e reason we need to increase it int he future, it would require service quota increases as well as approval from finaance as the bigger the cluster the more expensive it gets

.github/workflows/eks-performance-cluster-scaling.yml
JayPolanco: Not sure what you mean. This is here to validate that the scaling worked as expected. If an error occurs with managed node groups, there's a change that not all nodes get created, which would not be great for our tests

.github/workflows/eks-performance-cluster-scaling.yml
JayPolanco: Nice catch, updating

.github/workflows/eks-performance-cluster-scaling.yml
JayPolanco: > do we want this as env instead? do we ever want to configure this?

desired_capacity is changed to an environment varibale as well

movence (dismissed):

.github/workflows/eks-performance-cluster-scaling.yml
okankoAMZ: I would like that being a variable like `node_group_count` it make the code self documenting

.github/workflows/eks-performance-cluster-scaling.yml
okankoAMZ: cache with the commit_id if there is another run with the same commit_id scale it down

.github/workflows/eks-performance-cluster-scaling.yml
JayPolanco: Sure, I'll make this small change as part of the test PR

TravisStark (approved):

PR #1786, opened 2025-07-21
Implement workflow to install addon/helm on performance cluster

JayPolanco:
# Description of the issue
To enable performance testing on the large scale testing EKS cluster created [here](https://github.com/aws/amazon-cloudwatch-agent/pull/1776), we need to ensure that the AWS Observability addon gets installed upon cluster sclaing action. This task implements a github action that will do this via helm.

# Description of changes
- Created GitHub action for installing addon with helm
- Builds the latest versions of the operator and agent and uplaods it to ECR
- Installs it using helm command
- Triggers upon completion of the `EKS Cluster Scaling` workflow
- Can also be ran manually with workflow_dispatch
- Cleans/removes helm chart on 2 conditions: If the task fails/is canceled or node count is 0 which means a scale down was completed

# License

# Tests
- https://github.com/aws/amazon-cloudwatch-agent/actions/runs/16380906842

# Requirements
2. Run `make lint`

-------
### Integration Tests

### DIFF
--- a/.github/workflows/eks-performance-cluster-addon-install.yml
+++ b/.github/workflows/eks-performance-cluster-addon-install.yml
@@ -1,0 +1,246 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT
+name: "Install/Remove Helm Charts After Scaling"
+on:
+  # Use workflow_run to trigger this workflow after the scaling workflow completes
+  workflow_run:
+    workflows: ["EKS Cluster Scaling"]
+    types:
+      - completed
+  #    branches:
+  #      - main  # Adjust this if your default branch is different
+
+  # Keep the manual trigger option
+  workflow_dispatch:
+    inputs:
+      # Required Core Settings
+      cluster_name:
+        description: 'EKS Cluster Name'
+        required: true
+        type: string
+        default: 'eks-performance'
+      region:
+        description: 'AWS Region'
+        required: true
+        type: string
+        default: 'us-west-2'
+
+      # Optional Settings
+      cloudwatch_agent_repository:
+        description: 'CloudWatch Agent Repository'
+        type: string
+      cloudwatch_agent_tag:
+        description: 'CloudWatch Agent Tag'
+        type: string
+      cloudwatch_agent_operator_repository:
+        description: 'CloudWatch Agent Operator Repository'
+        type: string
+      cloudwatch_agent_operator_tag:
+        description: 'CloudWatch Agent Operator Tag'
+        type: string
+      helm-charts-branch:
+        description: 'Branch of the helm charts to test'
+        type: string
+        default: 'main'
+      operator-branch:
+        description: 'Branch of the operator to test'
+        type: string
+        default: 'main'
+      terraform_assume_role:
+        description: 'AWS IAM Role to assume'
+        type: string
+
+concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/eks-performance-cluster-addon-install.yml
```diff
@@ -0,0 +1,246 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT
+name: "Install/Remove Helm Charts After Scaling"
+on:
+  # Use workflow_run to trigger this workflow after the scaling workflow completes
+  workflow_run:
+    workflows: ["EKS Cluster Scaling"]
+    types:
+      - completed
+  #    branches:
```
movence: is this meant to be commented out? if yes, you can just remove

.github/workflows/eks-performance-cluster-addon-install.yml
```diff
@@ -0,0 +1,246 @@
+# Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+# SPDX-License-Identifier: MIT
+name: "Install/Remove Helm Charts After Scaling"
+on:
+  # Use workflow_run to trigger this workflow after the scaling workflow completes
+  workflow_run:
+    workflows: ["EKS Cluster Scaling"]
+    types:
+      - completed
```
movence, line 9: does it make this action to run regardless of scale up and down cases? just curious if this gets executed when the cluster is scaled down.

.github/workflows/eks-performance-cluster-addon-install.yml
```diff
@@ -0,0 +1,246 @@
...
+      region:
+        description: 'AWS Region'
+        required: true
+        type: string
+        default: 'us-west-2'
+
+      # Optional Settings
+      cloudwatch_agent_repository:
+        description: 'CloudWatch Agent Repository'
+        type: string
```
movence, line 31: no default for repos?

movence: oh I see the defaults being used during the execution

.github/workflows/eks-performance-cluster-addon-install.yml
movence: I see there is node count check later on, but is there a better way to check what action the parent workflow/action did for an early exit?

.github/workflows/eks-performance-cluster-addon-install.yml
JayPolanco: Nope, meant to uncomment that. By default it already uses main, but this is in case we ever wanted to change that

.github/workflows/eks-performance-cluster-addon-install.yml
JayPolanco: Normally yes, this ensures that this workflow executes whenever the scaling workflow completes.

Additional checks were done to make it so this only happens automatically on the schedule workflow. If you were to run the workflow manually it wouldn't auto trigger.

When a scale down is automatically down it's always back to zero. This is why we do this check

.github/workflows/eks-performance-cluster-addon-install.yml
JayPolanco: The defaults are in the command because workflow runs don't allow you to have inputs. So it wouldn't pickup the defaults otherwise

PR #1789, opened 2025-07-22
Update state when batch dropped

duhminick:
# Description of the issue

# Description of changes

# License

# Tests

# Requirements
2. Run `make lint`

-------
### Integration Tests


plugins/outputs/cloudwatchlogs/internal/pusher/sender.go
```diff
@@ -87,18 +87,21 @@ func (s *sender) Send(batch *logEventBatch) {
...
 			s.logger.Errorf("Non aws error received when sending logs to %v/%v: %v. CloudWatch agent will not retry and logs will be missing!", batch.Group, batch.Stream, err)
+			batch.updateState()
 			return
 		}
 
 		switch e := awsErr.(type) {
 		case *cloudwatchlogs.ResourceNotFoundException:
 			if targetErr := s.targetManager.InitTarget(batch.Target); targetErr != nil {
 				s.logger.Errorf("Unable to create log stream %v/%v: %v", batch.Group, batch.Stream, targetErr)
+				batch.updateState()
```
duhminick: Actually, after thinking about it some more, we should remove the update here. This definitely is a valid error that can and should be retried

plugins/outputs/cloudwatchlogs/internal/pusher/batch.go
```diff
@@ -164,14 +166,36 @@ func (b *logEventBatch) addDoneCallback(callback func()) {
...
 		done()
 	}
 }
 
+// updateState runs only the state callbacks to update the state file
+// without executing other success-related callbacks. This is used when a batch
+// fails after exhausting all retry attempts to prevent reprocessing the same
+// batch after restart.
+func (b *logEventBatch) updateState() {
+	for i := len(b.stateCallbacks) - 1; i >= 0; i-- {
```
the-mann, line 193: nit – why not use [range](https://go.dev/tour/moretypes/16)?

plugins/outputs/cloudwatchlogs/internal/pusher/batch.go
```diff
@@ -164,14 +166,36 @@ func (b *logEventBatch) addDoneCallback(callback func()) {
...
 		done := b.doneCallbacks[i]
 		done()
 	}
 }
 
+// updateState runs only the state callbacks to update the state file
+// without executing other success-related callbacks. This is used when a batch
+// fails after exhausting all retry attempts to prevent reprocessing the same
+// batch after restart.
+func (b *logEventBatch) updateState() {
```
okankoAMZ, line 192: why are we iterating backwards?

duhminick: Just following the same pattern for the done() function

https://github.com/aws/amazon-cloudwatch-agent/blob/73da7a4e81d60471b74bc3fd67bc6a82a6f310c7/plugins/outputs/cloudwatchlogs/internal/pusher/batch.go#L179-L186

plugins/outputs/cloudwatchlogs/internal/pusher/batch.go
duhminick: Existing integ tests and replied to the same question above

PR #1790, opened 2025-07-23
Adding default error_mode to jmx drop config yaml to fix unmarshalling issue

Paramadon:
# Description of the issue
Currently there the translator for jmx drop config yaml does not have error_mode set which means that it will be set to "" by default which can cause unmarshalling issues in the agent. 

# Description of changes
Adding error_mode: propagate to jmx drop config yaml to make sure there are no unmarshaling issues. We have the same behavior in the metrics decorator translator code.

### DIFF
--- a/.github/workflows/eks-performance-cluster-addon-install.yml
+++ b/.github/workflows/eks-performance-cluster-addon-install.yml
@@ -67,6 +67,8 @@ env:
 
   # Github repository environment variables
   OPERATOR_GITHUB_REPO_NAME: "aws/amazon-cloudwatch-agent-operator"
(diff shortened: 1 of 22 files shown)


translator/tocwconfig/sampleConfig/jmx_eks_config_linux.yaml
```diff
@@ -188,7 +188,7 @@ processors:
         log_statements: []
         metric_statements:
             - context: resource
-              error_mode: ""
+              error_mode: propagate
```
sky333999, line 191: What about ones such as https://github.com/aws/amazon-cloudwatch-agent/blob/92115e8212aab0939e02f2a8cc1183e29aedacf5/translator/tocwconfig/sampleConfig/container_insights_jmx.yaml#L481?

PR #1794, opened 2025-07-25
[EKS Scalability Testing] Implement workflows to execute large scale performance test

JayPolanco:
# Description of the issue
As part of an initiative to improve our performance testing on large scale eks clusters, we need to implement workflows that scale up, install addon and execute test that runs on our 5k node cluster.

# Description of changes
1. eks-performance-cluster-scaling.yml - Responsible for scaling eks-performance cluster (5k node cluster). 
   - Updated to also scale up leader node group dynamically. This node group has max of 1 node t3.xlarge
   - Scales up on Sunday 9am, scales down Monday 9pm
2.  eks-performance-cluster-addon-install.yml - Responsible for installing the addon upon successful completion of scaling workflow
   - adds the `cwagent-target="true"` label to all t3.medium instances in the cluster (non leader node). This ensures that cloudwatch-agent non leader pods are installed on these nodes
   - gets the hostname (ip-x-x-x-x.<region>.compute.internal) and substitutes it into the <hostname> tag in the override file in the test repo under test/performance/eks/resources/leader_election_overrides/base-overrides.yml
   - installs the addon with helm, with the override file --values
3. eks-performance-cluster-tests.yml - responsible for executing the test in the test repo (eks_performance_test.go)
   - Updated check-trigger to ensure tests are only ran if the cluster has active nodes
   - installs sample applications use kubectl (petclinic-sample-app in test repo)
   - cleans up once test is complete
 
# License

# Tests
Cluster Scaling Workflow - https://github.com/aws/amazon-cloudwatch-agent/actions/runs/16512847744/job/46697932764
Addon Install Workflow - https://github.com/aws/amazon-cloudwatch-agent/actions/runs/16523571905/job/46731027008
Performance Test Workflow - https://github.com/aws/amazon-cloudwatch-agent/actions/runs/16526410774/job/46740785992

# Requirements
5. Run `make lint`

-------
### Integration Tests

### DIFF
--- a/.github/workflows/eks-performance-cluster-addon-install.yml
+++ b/.github/workflows/eks-performance-cluster-addon-install.yml
@@ -68,7 +68,7 @@ env:
   # Github repository environment variables
   OPERATOR_GITHUB_REPO_NAME: "aws/amazon-cloudwatch-agent-operator"
   CWA_GITHUB_TEST_REPO_NAME: "aws/amazon-cloudwatch-agent-test"
-  CWA_GITHUB_TEST_REPO_BRANCH: "performance-test-testrun"
+  CWA_GITHUB_TEST_REPO_BRANCH: "main"
 
 jobs:
   # Check if this workflow should run
--- a/.github/workflows/eks-performance-cluster-addon-install.yml
+++ b/.github/workflows/eks-performance-cluster-addon-install.yml
@@ -173,10 +173,11 @@ jobs:
         run: |
           aws eks update-kubeconfig --name $CLUSTER_NAME --region $AWS_REGION
 
+      # TODO: Revert to using main helm branch when changes from leader-election are merged in
       - name: Clone Helm Charts Repository
         run: |
           rm -rf ./helm-charts
-          git clone -b ${{ inputs.helm-charts-branch || 'main' }} https://github.com/aws-observability/helm-charts.git ./helm-charts
+          git clone -b ${{ inputs.helm-charts-branch || 'sky333999/leader-election' }} https://github.com/aws-observability/helm-charts.git ./helm-charts
 
       - name: Clone Test Repo
         uses: actions/checkout@v4
@@ -185,10 +186,6 @@ jobs:
           ref: ${{ env.CWA_GITHUB_TEST_REPO_BRANCH }}
           path: ./test-repo
 
-      - name: Label nodes for CloudWatch Agent targeting
-        run: |
-          kubectl label nodes -l 'node.kubernetes.io/instance-type=t3.medium' cwagent-target="true"
-
       - name: Replace hostname in override files
         run: |
           HOSTNAME=$(kubectl get nodes -l eks.amazonaws.com/nodegroup=$CLUSTER_NAME-leader-node -o jsonpath='{.items[0].metadata.name}')
@@ -196,6 +193,7 @@ jobs:
             sed -i "s/<hostname>/$HOSTNAME/g" "$file"
           done
 
+      # TODO: Revert to using workflow built agent image once required changes are made on main branch
       - name: Check node count and manage Helm chart
         run: |
           NODE_COUNT=$(kubectl get nodes --no-headers | wc -l)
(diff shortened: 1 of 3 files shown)


.github/workflows/eks-performance-cluster-scaling.yml
```diff
@@ -89,6 +105,17 @@ jobs:
         run: |
           echo "Starting scale DOWN operation with desired capacity: 0"
           
+          # Scale leader node to 0
+          echo "Scaling leader node group: $CLUSTER_NAME-leader-node to 0"
```
TravisStark, line 109: Nit: Might be helpful for log readability to add "Scaling down" here and "Scaling up" in the log below

JayPolanco: Log right above for Starting scale DOWN operation with desired capacity: 0

.github/workflows/eks-performance-cluster-scaling.yml
```diff
@@ -105,6 +132,18 @@ jobs:
         if: github.event_name == 'workflow_dispatch'
         run: |
           echo "Starting manual scaling operation with desired capacity: $DESIRED_CAPACITY_PER_NODEGROUP"
+          echo "Leader node desired capacity: $LEADER_NODE_DESIRED_CAPACITY"
```
movence, line 135: does the current leader model support more than 1 instance? Im just curious what scenario we would want to test more than 1 leader case.

.github/workflows/eks-performance-cluster-scaling.yml
```diff
@@ -118,17 +157,17 @@ jobs:
 
       - name: Validate total node count
         run: |
-          echo "Waiting 20 minutes for scaling operations to complete..."
-          sleep 1200
+          echo "Waiting 30 minutes for scaling operations to complete and stabilize..."
+          sleep 1800
```
movence, line 161: 30m sleep seems a bit excessive. can we do a loop instead with a shorter sleep time?

.github/workflows/eks-performance-cluster-addon-install.yml
```diff
@@ -176,6 +178,24 @@ jobs:
...
+      - name: Clone Test Repo
+        uses: actions/checkout@v4
+        with:
+          repository: ${{ env.CWA_GITHUB_TEST_REPO_NAME }}
+          ref: ${{ env.CWA_GITHUB_TEST_REPO_BRANCH }}
+          path: ./test-repo
+
+      - name: Label nodes for CloudWatch Agent targeting
+        run: |
+          kubectl label nodes -l 'node.kubernetes.io/instance-type=t3.medium' cwagent-target="true"
```
movence: why are we labeling nodes to schedule non-leader instances? wouldn't it better to label node(s) to target where the leader(s) get deployed?

.github/workflows/eks-performance-cluster-scaling.yml
JayPolanco: The current model does not, but I left it open just in case. No real impact in terms of behavior

.github/workflows/eks-performance-cluster-scaling.yml
JayPolanco: This was left over from some manual testing, since I was messing around it was taking slightly longer. Will drop back to 20

.github/workflows/eks-performance-cluster-addon-install.yml
JayPolanco: During testing, there were constant leader elections occurring. This was due to the agent pods (daemonset and deployment) competing to become the leader. No matter what was doen, this was always the case. To fix this I mimicked the exact thing we do for the titan cluster which was to label the nodes.

JayPolanco: Also from my experience, if you don't label the non-leader pod nodes, the override does not install the agent pods in the non-leader nodes

JayPolanco: Fixed to use hardcoded images until we merge in the right changes to the main branches

.github/workflows/eks-performance-cluster-scaling.yml
TravisStark: Looks like this one is still 30 mins

TravisStark (approved):

PR #1797, opened 2025-07-28
Add suggestions feature to debugger tool

James-Hou22:
# Description of the issue
Previously, the debugger does not have any feature to suggest fixes to common issues.

# Description of changes
Added a feature that parses agent logs for common issues that customers get and provides suggestions for them.

# License

# Tests
Unit tests in `error_finder_test.go`

# Requirements
2. Run `make lint`

# Output
This is an example output of the suggestions section that notifies the user that they don't permissions to the PutMetricData action, and we suggest adding the action or using the default CWAgent policy.
```
=== Log Errors Summary ===
Found 2 error entries since last startup
Issues from logs (1):
. Problem: CloudWatch Agent is trying to push metrics via the PutMetricData endpoint but it does not have necessary permissions to do so.
  Fix: Your IAM role requires 'cloudwatch:PutMetricData' action allowed. Add CloudWatchCustomServerPolicy to your IAM role or if you are using a custom role add 'cloudwatch:PutMetricData'.
```

### DIFF
--- a/internal/debugger/common_errors.go
+++ b/internal/debugger/common_errors.go
@@ -8,9 +8,9 @@ import (
 )
 
 type DiagnosticSuggestion struct {
-	Pattern     *regexp.Regexp
-	Issue       string
-	Possibility string
+	Pattern *regexp.Regexp
+	Problem string
+	Fix     string
 }
 
 // Metric related errors
@@ -94,7 +94,7 @@ var networkErrors = []DiagnosticSuggestion{
 	{
 		regexp.MustCompile(`lookup .*\.amazonaws\.com.*: no such host`),
 		"Problem: CWAgent failed to resolve an AWS service hostname. This often happens when a custom endpoint override is misconfigured or if DNS resolution is blocked or unavailable on the instance.",
-		"Fix: If using a custom endpoint_override, verify that the hostname is correct and resolvable. Otherwise, check your instance's DNS configuration, VPC settings, and ensure outbound DNS (UPD port 53) is not blocked.",
+		"Fix: If using a custom endpoint_override, verify that the hostname is correct and resolvable. Otherwise, check your instance's DNS configuration, VPC settings, and ensure outbound DNS (UDP port 53) is not blocked.",
 	},
 	{
 		regexp.MustCompile(`dial tcp 10\..*:443: connect: connection refused`),
--- a/internal/debugger/error_finder.go
+++ b/internal/debugger/error_finder.go
@@ -12,10 +12,9 @@ import (
 )
 
 const (
-	TimeFormat = "2006-01-02T15:04:05Z"
-	DateRegex  = `\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z E!`
+	DateRegex = `\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z E!`
 	// Finder looks for errors from each agent restart
-	StartUp = `^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z\s+I!\s+Starting AmazonCloudWatchAgent.*with log file`
+	StartUpPattern = `^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}Z\s+I!\s+Starting AmazonCloudWatchAgent.*with log file`
 )
 
 func GetErrorSuggestions(logFilePath string) []DiagnosticSuggestion {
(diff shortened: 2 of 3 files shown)


internal/debugger/common_errors.go
```diff
@@ -0,0 +1,120 @@
...
+	},
+	{
+		regexp.MustCompile(`proxyconnect tcp: dial tcp .*: connect: connection refused`),
+		"Problem: CWAgent is configured to use a proxy but is unable to connect to the proxy server. This prevents the agent from reaching AWS services such as CloudWatch, CloudWatch Logs or the IMDS.",
+		"Fix: Ensure proxy is correctly configured and reachable. Also, check if the AWS endpoints are reachable through your proxy.",
+	},
+	{
+		regexp.MustCompile(`lookup .*\.amazonaws\.com.*: no such host`),
+		"Problem: CWAgent failed to resolve an AWS service hostname. This often happens when a custom endpoint override is misconfigured or if DNS resolution is blocked or unavailable on the instance.",
+		"Fix: If using a custom endpoint_override, verify that the hostname is correct and resolvable. Otherwise, check your instance's DNS configuration, VPC settings, and ensure outbound DNS (UPD port 53) is not blocked.",
```
Paramadon: typo: should be UDP port 53*

internal/debugger/error_finder.go
```diff
@@ -0,0 +1,191 @@
...
+
+	return suggestions
+}
+
+func deduplicateSuggestions(suggestions []DiagnosticSuggestion) []DiagnosticSuggestion {
+	seen := make(map[string]bool)
+	var uniqueErrors []DiagnosticSuggestion
+
+	for _, suggestion := range suggestions {
+		if !seen[suggestion.Possibility] {
```
Paramadon: Should we also do it based off of Issue instead of just possibility, just thinking based off of if two issues have the same fix.

internal/debugger/error_finder.go
```diff
@@ -0,0 +1,191 @@
...
+import (
+	"fmt"
+	"io"
+	"os"
+	"regexp"
+	"strings"
+)
+
+const (
+	TimeFormat = "2006-01-02T15:04:05Z"
```
the-mann: maybe get rid of this if unused?

internal/debugger/error_finder.go
```diff
@@ -0,0 +1,191 @@
...
+}
+
+func checkAgentLogPermissions(logFilePath string) error {
+	_, err := os.Stat(logFilePath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return fmt.Errorf("CloudWatch Agent log file not found at %s. Ensure the CloudWatch Agent is running", logFilePath)
+		}
+		return fmt.Errorf("Cannot access log file %s: %v. Try running as cwagent user or with sudo", logFilePath, err)
+	}
```
the-mann: maybe redundant?

internal/debugger/error_finder.go
```diff
@@ -0,0 +1,191 @@
...
+	file, err := os.Open(logFilePath)
+	if err != nil {
+		return fmt.Errorf("Cannot read log file %s: %v. Try running as cwagent user or with sudo", logFilePath, err)
+	}
+	file.Close()
+
+	return nil
+}
+
+func getLogEntries(logFilePath string, startupPattern string) ([]string, error) {
```
the-mann: make startupPattern a constant/use it directly? why pass in as parameter?

internal/debugger/error_finder.go
```diff
@@ -0,0 +1,191 @@
...
+	logFile, err := os.Open(logFilePath)
+	if err != nil {
+		return nil, err
+	}
+	defer logFile.Close()
+
+	logFileContent, err := io.ReadAll(logFile)
```
the-mann, line 66: could be dangerous but i think fine for now. if we were to productionize it, maybe something to keep in mind?

internal/debugger/common_errors.go
```diff
@@ -0,0 +1,120 @@
...
+package debugger
+
+import (
+	"regexp"
+)
+
+type DiagnosticSuggestion struct {
+	Pattern     *regexp.Regexp
+	Issue       string
+	Possibility string
```
the-mann: rename to Solution or something of the like

the-mann: or Fix: and then remove that from the prompt and add it when you log to simplify adding a new suggestion

the-mann: or rename it to suggestion or something idk

the-mann (approved):

internal/debugger/error_finder.go
James-Hou22: We enforce decoupling with problems and fixes just to keep everything organized, so even if two issues have similar fixes we will keep it separate.

internal/debugger/common_errors.go
James-Hou22: Changing it to be consistent with the frontend, which is just Problem and Fix.

James-Hou22: "frontend"

PR #1809, opened 2025-08-04
Dynamic reader buffer for tail

duhminick:
# Description of the issue

# Description of changes

# License

# Tests

# Requirements
2. Run `make lint`

-------
### Integration Tests

### DIFF
--- a/plugins/inputs/logfile/tail/tail.go
+++ b/plugins/inputs/logfile/tail/tail.go
@@ -73,8 +73,9 @@ type Tail struct {
 	Lines    chan *Line
 	Config
 
-	file   *os.File
-	reader *bufio.Reader
+	file           *os.File
+	reader         *bufio.Reader
+	useLargeBuffer bool
 
 	watcher watch.FileWatcher
 	changes *watch.FileChanges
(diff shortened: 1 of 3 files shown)


TravisStark:
Were you able to test this with actual logs published to CloudWatch as well?  Sometimes the behavior was not as expected even with tests passing.

Edit: Discussed offline, logs output looks correct for logs >= 1MB

duhminick:
> Were you able to test this with actual logs published to CloudWatch as well? Sometimes the behavior was not as expected even with tests passing.

Yep! Here is an example with small lines, then a big line, then small lines again.

image

Size for that one line with 'aaaa...aaaa'
image

I then added another line with that's 800KB in size
image

Size:
image

TravisStark (approved):

plugins/inputs/logfile/tail/tail.go
```diff
@@ -544,14 +545,52 @@ func (tail *Tail) waitForChanges() error {
...
+		peek, _ := tail.reader.Peek(unread)
+		buf = append(buf, peek...)
+		tail.reader.Discard(unread)
+	}
+
+	// Switch to a bigger buffer
+	tail.reader = bufio.NewReaderSize(tail.file, tail.MaxLineSize)
+	// In the event that the tail is re-opened, we don't want to have to do this
+	// re-sizing of the buffer again. The reader should just re-open with the larger buffer
+	tail.useLargeBuffer = true
```
lisguo, line 585: We don't ever set this back to false right? So it'll only take 1 occurrence to use the larger buffer forever in the tailer?

lisguo (approved):
Approving since it's still better than what we have now,  but I'd wonder if we should ever go back from 1MB -> 256MB after upgrading once.

plugins/inputs/logfile/tail/tail.go
TravisStark: Good catch, that will stay true.  I'll create a ticket to adjust that.

PR #1817, opened 2025-08-12
Refactor AWS NVMe Scraper for Easier Device Extensibility

Paramadon:
### Description
This PR refactors the awsnvmereceiver by introducing the DeviceInfoProvider interface to abstract device-specific operations. The scraper now interacts with NVMe devices through this interface instead of hardcoded switch logic, making it easier to add new device types, improve testability, and maintain the code.

### Interface Description:
```
Model() string – Returns the device model name.

DeviceType() string – Returns a unique type identifier for the device.

Identifier(serial string) (string, error) – Converts a device serial to a unique identifier; validates input.

SetResourceAttribute(rb *metadata.ResourceBuilder, identifier string) – Sets the device-specific resource attributes for metrics.

RecordMetrics(recordMetric RecordMetricFunc, mb *metadata.MetricsBuilder, ts pcommon.Timestamp, metrics Metrics) – Records metrics using a typed function and device-specific metrics.

IsEnabled(m *metadata.MetricsConfig) bool – Indicates whether scraping is enabled for this device type.

ParseRawData(data []byte) (Metrics, error) – Parses raw NVMe data into a strongly typed Metrics object.
```

In order to add a new nvme device we would just need to implement this interface and we should be able to get those nvme metrics.

### Testing
Manually tested and was able to get metrics:  
Screenshot 2025-08-12 at 2 13 19 PM


receiver/awsnvmereceiver/internal/ebs/scraper.go
```diff
@@ -0,0 +1,102 @@
...
+	"go.opentelemetry.io/collector/pdata/pcommon"
+
+	"github.com/aws/amazon-cloudwatch-agent/receiver/awsnvmereceiver/internal/metadata"
+	"github.com/aws/amazon-cloudwatch-agent/receiver/awsnvmereceiver/internal/nvme"
+)
+
+// scraper implements DeviceTypeScraper for EBS devices.
+type scraper struct{}
+
+func NewScraper() *scraper {
```
jefchien:
```suggestion
func NewScraper() nvme.DeviceTypeScraper {
```

receiver/awsnvmereceiver/scraper.go
```diff
@@ -17,35 +17,42 @@ import (
...
+type DeviceTypeScraper interface {
+	Model() string
+	DeviceType() string
+	Identifier(serial string) (string, error)
+	SetResourceAttribute(rb *metadata.ResourceBuilder, identifier string)
+	HandlesMetrics(metricsAny any) bool
+	RecordMetrics(recordMetric func(func(pcommon.Timestamp, int64), pcommon.Timestamp, uint64), mb *metadata.MetricsBuilder, ts pcommon.Timestamp, metricsAny any)
+	IsEnabled(m *metadata.MetricsConfig) bool
+	ParseRawData(data []byte) (any, error)
+}
```
jefchien: nit: Consider moving into `nvme` package to reduce the chance of circular dependency.

receiver/awsnvmereceiver/scraper.go
```diff
@@ -187,33 +177,27 @@ func (s *nvmeScraper) getNVMeDevicesByController() (map[int]*nvmeDevices, error)
...
-			}
-			identifier = fmt.Sprintf("vol-%s", serial[3:])
-		case instanceStoreModel:
-			if !s.collectInstanceStore {
-				s.logger.Debug("skipping Instance Store device as no IS metrics enabled", zap.String("device", deviceName))
-				continue
+		var matchedScraper DeviceTypeScraper
+		var identifier string
+		for _, scraper := range s.typeScrapers {
+			if model == scraper.Model() {
```
jefchien: Since these are being accessed via the model, it would reduce the lookup if the scrapers were stored in a map.

receiver/awsnvmereceiver/scraper.go
```diff
@@ -17,35 +17,42 @@ import (
...
 type nvmeScraper struct {
 	logger *zap.Logger
 	mb     *metadata.MetricsBuilder
 	nvme   nvme.DeviceInfoProvider
 
-	allowedDevices       collections.Set[string]
-	collectEbs           bool
-	collectInstanceStore bool
+	allowedDevices collections.Set[string]
+	typeScrapers   []DeviceTypeScraper
```
jefchien: nit: Makes the contents of this clearer.
```suggestion
	enabledScrapers   []DeviceTypeScraper
```

receiver/awsnvmereceiver/scraper.go
```diff
@@ -17,35 +16,31 @@ import (
...
 type nvmeScraper struct {
 	logger *zap.Logger
 	mb     *metadata.MetricsBuilder
 	nvme   nvme.DeviceInfoProvider
 
-	allowedDevices       collections.Set[string]
-	collectEbs           bool
-	collectInstanceStore bool
+	allowedDevices  collections.Set[string]
+	enabledScrapers []nvme.DeviceTypeScraper
```
jefchien: Why do we need to store the slice? It's never used outside of populating the `scrapersByModel`. Would suggest renaming `scrapersByModel` to `enabledScrapers` and removing this field.

receiver/awsnvmereceiver/internal/instancestore/scraper.go
```diff
@@ -0,0 +1,131 @@
...
+	Read  [32]uint64
+	Write [32]uint64
+}
+
+func (Metrics) IsNVMeMetrics() {}
+
+// Scraper implements DeviceTypeScraper for Instance Store devices.
+type Scraper struct{}
+
+func NewScraper() *Scraper {
```
jefchien: nit: Would return the `nvme.DeviceTypeScraper` and avoid exporting the `instancestore` specific implementation. This enforces the interface if it ever changes and makes it clearer that it's of that type.

receiver/awsnvmereceiver/internal/nvme/metrics_unix_test.go
jefchien, line 1: `metrics_unix.go` still has functions that need to be tested like `GetRawData`.

receiver/awsnvmereceiver/scraper.go
```diff
@@ -187,33 +159,21 @@ func (s *nvmeScraper) getNVMeDevicesByController() (map[int]*nvmeDevices, error)
...
-			if !s.collectInstanceStore {
-				s.logger.Debug("skipping Instance Store device as no IS metrics enabled", zap.String("device", deviceName))
-				continue
-			}
-			deviceType = "instance_store"
-			identifier = serial
-		default:
+		scraper, ok := s.scrapersByModel[model]
+		if !ok {
 			s.logger.Debug("skipping unknown device model", zap.String("device", deviceName), zap.String("model", model))
```
jefchien: nit: This isn't necessarily skipping unknown device models. If the EBS metrics aren't configured, then this would log out for all EBS devices. Are we sure we want to log out this message at all?

receiver/awsnvmereceiver/internal/instancestore/scraper.go
```diff
@@ -0,0 +1,131 @@
...
+
+	"go.opentelemetry.io/collector/pdata/pcommon"
+
+	"github.com/aws/amazon-cloudwatch-agent/receiver/awsnvmereceiver/internal/metadata"
+	"github.com/aws/amazon-cloudwatch-agent/receiver/awsnvmereceiver/internal/nvme"
+)
+
+const InstanceStoreMagic uint32 = 0xEC2C0D7E
+
+var ErrInvalidInstanceStoreMagic = errors.New("invalid Instance Store magic number")
```
jefchien: nit: Neither of these are used outside of this package. Don't export them.

receiver/awsnvmereceiver/internal/instancestore/scraper.go
```diff
@@ -0,0 +1,131 @@
...
+
+func (s *Scraper) DeviceType() string {
+	return "instance_store"
+}
+
+func (s *Scraper) Identifier(serial string) (string, error) {
+	if strings.TrimSpace(serial) == "" {
+		return "", fmt.Errorf("empty serial for instance store")
+	}
+	return serial, nil
```
jefchien: Trimming the space is fine, but it looks like we don't actually return the modified value, so if the serial is "    serial", then we'll be sending that as is. Is that intentional?

receiver/awsnvmereceiver/scraper.go
```diff
@@ -224,21 +184,38 @@ func (s *nvmeScraper) getNVMeDevicesByController() (map[int]*nvmeDevices, error)
...
+		if ts.IsEnabled(&cfg.MetricsBuilderConfig.Metrics) {
+			enabledScrapers = append(enabledScrapers, ts)
+		}
+	}
+	scraper.enabledScrapers = enabledScrapers
+
+	scraper.scrapersByModel = make(map[string]nvme.DeviceTypeScraper, len(enabledScrapers))
+	for _, ts := range enabledScrapers {
+		scraper.scrapersByModel[ts.Model()] = ts
+	}
```
jefchien: Why do we need the slice? Why not create the map from the beginning?

```suggestion
	scraper.enabledScrapers = make(map[string]nvme.DeviceTypeScraper)
	for _, ts := range allScrapers {
		if ts.IsEnabled(&cfg.MetricsBuilderConfig.Metrics) {
			scraper.enabledScrapers[ts.Model()] = ts
		}
	}
```

receiver/awsnvmereceiver/internal/instancestore/scraper.go
Paramadon: Updated to just return serial as that is what we are doing previously, same for ebs Identifier func.

receiver/awsnvmereceiver/internal/ebs/scraper.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	return fmt.Sprintf("vol-%s", serial[3:]), nil
+}
+
+func (s *scraper) SetResourceAttribute(rb *metadata.ResourceBuilder, identifier string) {
+	rb.SetVolumeID(identifier)
+}
+
+func (s *scraper) ParseRawData(data []byte) (nvme.Metrics, error) {
+	log.Println("Parsing Raw Data EBS3")
```
zhihonl: this is too noisy in the logs, let's remove this

zhihonl (dismissed):

receiver/awsnvmereceiver/internal/nvme/metrics_unix_test.go
```diff
@@ -19,216 +19,51 @@
 
 //go:build linux
 
-package nvme
+package nvme // Replace with the actual package name
```
jefchien: nit: Comment seems like it shouldn't be here.

receiver/awsnvmereceiver/scraper.go
```diff
@@ -224,40 +183,43 @@ func (s *nvmeScraper) getNVMeDevicesByController() (map[int]*nvmeDevices, error)
...
+
+	return scraper
 }
 
-// computeCollectFlags computes whether to collect for EBS and Instance Store based on enabled metrics
-func computeCollectFlags(cfg *Config) (bool, bool) {
-	m := cfg.MetricsBuilderConfig.Metrics
-	collectEbs := m.DiskioEbsTotalReadOps.Enabled ||
+// ComputeNVMeFeatures returns feature flags for NVMe adoption metrics based on enabled metrics.
+func ComputeNVMeFeatures(m *metadata.MetricsConfig) []string {
```
jefchien: This doesn't seem like it should be part of this PR.

PR #1375, opened 2024-10-07
Extend OTLP receiver to publish via EMF logs

sky333999:
# Description of changes
https://github.com/aws/amazon-cloudwatch-agent/pull/1361 introduced the support for an OTLP plugin for publishing metrics via PutMetricData to CloudWatch.
This PR extends the ability to use the OTLP plugin to publish metrics as EMF logs via PutLogEvents to CloudWatch.

* Introduces `otlp` section under `logs -> metrics_collected` that can either be a json blob or an array of json blobs for multiple listeners.
* OTLP Metrics are setup to flow through our a new `hostDeltaMetrics/cloudwatchlogs` pipeline using:
  * `otlpreceiver` - to receive the metrics on both HTTP and gRPC endpoints
  * `cummulativetodelta` processor - to convert cumulatives to deltas by default when publishing to CloudWatch
  * `batch` processor - to respect the `force_flush_interval` in the `logs` section of the agent config
  * `awsemfexporter` - to publish as EMF logs via PutLogEvents API calls to CloudWatch
* Serves as a passthrough plugin where all OTLP metrics sent are passed along as CloudWatch metrics without the need for explicit EMF metric declarations. Similarly, all resource and datapoint attributes are converted to dimensions on the metrics without any rollups i.e. they are sent along at the same granularity at which they are received.
* EMF logs are sent to `aws/cwagent` log group containing a log stream per unique run of the agent i.e. the log stream is based off a UUID generated on each run. Example log stream: `otel-stream-52215dd6-abfa-44cb-a23d-25d79d1278fd`.
* Metrics from EMF logs are extracted into the `CWAgent` namespace.
* Supports TLS configuration.

# License
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

# Tests
* Updated unit tests
* Manually tested using a OTLP sample app

Sample config:
```
{
  "logs": {
    "metrics_collected": {
      "otlp": {
        "grpc_endpoint": "0.0.0.0:4317",
        "http_endpoint": "0.0.0.0:4318"
      }
    }
  }
}
```
Sample EMF log event:
```
{
    "CloudWatchMetrics": [
        {
            "Namespace": "CWAgent",
            "Dimensions": [
                [
                    "service.name",
                    "roll.value",
                    "telemetry.sdk.language",
                    "telemetry.sdk.name",
                    "telemetry.sdk.version",
                    "telemetry.auto.version"
                ]
            ],
            "Metrics": [
                {
[...]

PR #1361, opened 2024-09-30
Add support for OTLP Receiver for Metrics

sky333999:
# Description of changes
Introduce support for the OTLP receiver for Metrics. This change adds the OTLP receiver to the existing `hostDeltaMetrics` pipeline that also uses a Cumulative to Delta processor and an AWS CloudWatch exporter.

* Introduces new `otlp` section under `metrics -> metrics_collected` that can either be a json blob or an array of json blobs for multiple listeners.
* OTLP Metrics are setup to flow through our current `hostDeltaMetrics` pipeline using:
  * `otlpreceiver` - to receive the metrics on both HTTP and gRPC endpoints
  * `cummulativetodelta` processor - to convert cumulatives to deltas by default when publishing to CloudWatch
  * AWS `cloudwatch` exporter - to publish via PutMetricData API calls to CloudWatch
* Supports `append_dimensions` as expected.
* We treat this as a passthrough plugin where the agent does minimal modifications. Accordingly:
  * Does NOT support agent features such as measurement renames, units, aggregation_dimensions or drop_original_metrics at this time.
  * Does NOT append `host` as a default dimension even if `InstanceId` is not present in `append_dimensions`.
* Supports TLS configuration.

Sample config:
```
{
  "metrics": {
    "metrics_collected": {
      "otlp": {
        "grpc_endpoint": "0.0.0.0:1234",
        "http_endpoint": "0.0.0.0:2345",
        "tls": {
          "cert_file": "/path/to/cert.pem",
          "key_file": "/path/to/key.pem"
        }
      }
    }
  }
}
```

# License
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

# Tests
* Updated unit tests
* Test using a sample app as per https://opentelemetry.io/docs/languages/python/getting-started/
image
* 

# Requirements
_Before commit the code, please do the following steps._
1. Run `make fmt` and `make fmt-sh`
2. Run `make lint`

PR #1821, opened 2025-08-14
Introduce shared receiver model for OTLP to avoid port conflicts

movence:
# Description of the issue
With this changes ([PR #1361](https://github.com/aws/amazon-cloudwatch-agent/pull/1361), [PR #1375](https://github.com/aws/amazon-cloudwatch-agent/pull/1375)), if multiple use cases are configured on the same port, this can cause port binding errors and crash the agent.

# Description of changes
Introduce shared receiver model at config translation so that the same receiver can be used for different pipelines (traces, metrics and applicationSignals)

# License

# Tests
TBD

# Requirements
2. Run `make lint`

-------
### Integration Tests

### DIFF
--- a/translator/translate/otel/pipeline/applicationsignals/translator.go
+++ b/translator/translate/otel/pipeline/applicationsignals/translator.go
@@ -52,7 +52,7 @@ func (t *translator) Translate(conf *confmap.Conf) (*common.ComponentTranslators
 	}
 
 	translators := &common.ComponentTranslators{
-		Receivers:  common.NewTranslatorMap(otlp.NewTranslator(common.WithName(common.AppSignals), otlp.WithSignal(t.signal))),
+		Receivers:  common.NewTranslatorMap(otlp.NewSharedTranslator(conf, common.WithName(common.AppSignals), otlp.WithSignal(t.signal))),
 		Processors: common.NewTranslatorMap[component.Config, component.ID](),
 		Exporters:  common.NewTranslatorMap[component.Config, component.ID](),
 		Extensions: common.NewTranslatorMap[component.Config, component.ID](),
--- a/translator/translate/otel/receiver/otlp/translator.go
+++ b/translator/translate/otel/receiver/otlp/translator.go
@@ -7,6 +7,7 @@ import (
 	_ "embed"
 	"fmt"
 	"strconv"
+	"sync"
 
 	"go.opentelemetry.io/collector/component"
 	"go.opentelemetry.io/collector/config/configtls"
@@ -73,6 +74,97 @@ func (t *translator) ID() component.ID {
 	return component.NewIDWithName(t.factory.Type(), t.Name())
 }
 
+var (
+	registryMu sync.RWMutex
+	registry   = make(map[string]common.ComponentTranslator)
+	counter    int
+)
+
+func getEndpointKey(conf *confmap.Conf, configKey string, index int, signal pipeline.Signal) (string, error) {
+	grpcEndpoint := defaultGrpcEndpoint
+	httpEndpoint := defaultHttpEndpoint
+	tlsCert, tlsKey := "", ""
+
+	// Handle application signals defaults
+	if appSignalsConfigKeys, ok := common.AppSignalsConfigKeys[signal]; ok {
+		for _, appKey := range appSignalsConfigKeys {
+			if configKey == appKey {
+				grpcEndpoint = defaultAppSignalsGrpcEndpoint
+				httpEndpoint = defaultAppSignalsHttpEndpoint
+				break
+			}
+		}
+	}
+
+	if conf != nil && conf.IsSet(configKey) {
+		otlpMap := common.GetIndexedMap(conf, configKey, index)
+		if grpc, ok := otlpMap["grpc_endpoint"]; ok {
+			grpcEndpoint = grpc.(string)
+		}
+		if http, ok := otlpMap["http_endpoint"]; ok {
+			httpEndpoint = http.(string)
+		}
+		if tls, ok := otlpMap["tls"].(map[string]interface{}); ok {
+			if cert, ok := tls["cert_file"].(string); ok {
+				tlsCert = cert
+			}
+			if key, ok := tls["key_file"].(string); ok {
+				tlsKey = key
+			}
+		}
+	}
+
+	portKey := fmt.Sprintf("%s|%s", grpcEndpoint, httpEndpoint)
+	fullKey := fmt.Sprintf("%s|%s|%s", portKey, tlsCert, tlsKey)
+
+	// Check for port conflicts with different TLS configs
+	for existingKey := range registry {
+		if existingKey != fullKey && existingKey[:len(portKey)] == portKey && len(existingKey) > len(portKey) {
+			return "", fmt.Errorf("port conflict: endpoints %s already in use with different TLS configuration", portKey)
+		}
+	}
+
+	return fullKey, nil
+}
+
+func NewSharedTranslator(conf *confmap.Conf, opts ...common.TranslatorOption) common.ComponentTranslator {
+	t := &translator{factory: otlpreceiver.NewFactory()}
+	t.SetIndex(-1)
+	for _, opt := range opts {
+		opt(t)
+	}
+
+	registryMu.Lock()
+	defer registryMu.Unlock()
+
+	key, err := getEndpointKey(conf, t.configKey, t.Index(), t.signal)
+	if err != nil {
+		// Return error translator that will fail during Translate()
+		return &errorTranslator{err: err}
+	}
+
+	if existing, exists := registry[key]; exists {
+		return existing
+	}
+
+	counter++
+	t.SetName(fmt.Sprintf("otlp_shared_%d", counter))
+	registry[key] = t
+	return t
+}
+
+type errorTranslator struct {
+	err error
+}
+
+func (e *errorTranslator) ID() component.ID {
+	return component.NewID(component.MustNewType("otlp"))
+}
+
+func (e *errorTranslator) Translate(*confmap.Conf) (component.Config, error) {
+	return nil, e.err
+}
+
 func (t *translator) Translate(conf *confmap.Conf) (component.Config, error) {
 	cfg := t.factory.CreateDefaultConfig().(*otlpreceiver.Config)
 
@@ -124,4 +216,12 @@ func (t *translator) Translate(conf *confmap.Conf) (component.Config, error) {
 		cfg.HTTP.ServerConfig.Endpoint = httpEndpoint.(string)
 	}
 	return cfg, nil
+}
+
+// ResetRegistry clears the shared receiver registry for testing
+func ResetRegistry() {
+	registryMu.Lock()
+	defer registryMu.Unlock()
+	registry = make(map[string]common.ComponentTranslator)
+	counter = 0
 }
--- a/translator/translate/otel/pipeline/host/translators.go
+++ b/translator/translate/otel/pipeline/host/translators.go
@@ -54,14 +54,14 @@ func NewTranslators(conf *confmap.Conf, configSection, os string) (common.Transl
 	switch v := conf.Get(common.ConfigKey(configSection, common.OtlpKey)).(type) {
 	case []any:
 		for index := range v {
-			otlpReceivers.Set(otlpreceiver.NewTranslator(
+			otlpReceivers.Set(otlpreceiver.NewSharedTranslator(conf,
 				otlpreceiver.WithSignal(pipeline.SignalMetrics),
 				otlpreceiver.WithConfigKey(common.ConfigKey(configSection, common.OtlpKey)),
 				common.WithIndex(index),
 			))
 		}
 	case map[string]any:
-		otlpReceivers.Set(otlpreceiver.NewTranslator(
+		otlpReceivers.Set(otlpreceiver.NewSharedTranslator(conf,
 			otlpreceiver.WithSignal(pipeline.SignalMetrics),
 			otlpreceiver.WithConfigKey(common.ConfigKey(configSection, common.OtlpKey)),
 		))
--- a/translator/translate/otel/pipeline/xray/translator.go
+++ b/translator/translate/otel/pipeline/xray/translator.go
@@ -56,7 +56,7 @@ func (t *translator) Translate(conf *confmap.Conf) (*common.ComponentTranslators
 		translators.Receivers.Set(awsxrayreceiver.NewTranslator())
 	}
 	if conf.IsSet(otlpKey) {
-		translators.Receivers.Set(otlp.NewTranslator(
+		translators.Receivers.Set(otlp.NewSharedTranslator(conf,
 			otlp.WithSignal(pipeline.SignalTraces),
 			otlp.WithConfigKey(otlpKey)),
 		)
--- a/translator/translate/otel/receiver/otlp/translator_test.go
+++ b/translator/translate/otel/receiver/otlp/translator_test.go
@@ -424,3 +424,180 @@ func TestTranslateJMX(t *testing.T) {
 	assert.NotNil(t, gotCfg.HTTP)
 	assert.Equal(t, "0.0.0.0:4314", gotCfg.HTTP.ServerConfig.Endpoint)
 }
+
+func TestSharedTranslatorDeduplication(t *testing.T) {
+	ResetRegistry()
+
+	config := map[string]interface{}{
+		"metrics": map[string]interface{}{
+			"metrics_collected": map[string]interface{}{
+				"otlp": map[string]interface{}{
+					"grpc_endpoint": "127.0.0.1:4317",
+					"http_endpoint": "127.0.0.1:4318",
+				},
+			},
+		},
+		"traces": map[string]interface{}{
+			"traces_collected": map[string]interface{}{
+				"otlp": map[string]interface{}{
+					"grpc_endpoint": "127.0.0.1:4317",
+					"http_endpoint": "127.0.0.1:4318",
+				},
+			},
+		},
+	}
+
+	conf := confmap.NewFromStringMap(config)
+
+	metricsTranslator := NewSharedTranslator(conf,
+		WithSignal(pipeline.SignalMetrics),
+		WithConfigKey(common.ConfigKey(common.MetricsKey, common.MetricsCollectedKey, common.OtlpKey)),
+	)
+	tracesTranslator := NewSharedTranslator(conf,
+		WithSignal(pipeline.SignalTraces),
+		WithConfigKey(common.ConfigKey(common.TracesKey, common.TracesCollectedKey, common.OtlpKey)),
+	)
+
+	assert.Equal(t, metricsTranslator.ID(), tracesTranslator.ID(), "Same endpoints should share receiver")
+
+	_, err1 := metricsTranslator.Translate(conf)
+	_, err2 := tracesTranslator.Translate(conf)
+	assert.NoError(t, err1)
+	assert.NoError(t, err2)
+}
+
+func TestArrayConfigDeduplication(t *testing.T) {
+	ResetRegistry()
+
+	config := map[string]interface{}{
+		"metrics": map[string]interface{}{
+			"metrics_collected": map[string]interface{}{
+				"otlp": []interface{}{
+					map[string]interface{}{
+						"grpc_endpoint": "127.0.0.1:4317",
+						"http_endpoint": "127.0.0.1:4318",
+					},
+					map[string]interface{}{
+						"grpc_endpoint": "127.0.0.1:4317",
+						"http_endpoint": "127.0.0.1:4318",
+					},
+					map[string]interface{}{
+						"grpc_endpoint": "127.0.0.1:5317",
+						"http_endpoint": "127.0.0.1:5318",
+					},
+				},
+			},
+		},
+	}
+
+	conf := confmap.NewFromStringMap(config)
+	configKey := common.ConfigKey(common.MetricsKey, common.MetricsCollectedKey, common.OtlpKey)
+
+	translator1 := NewSharedTranslator(conf, WithSignal(pipeline.SignalMetrics), WithConfigKey(configKey), common.WithIndex(0))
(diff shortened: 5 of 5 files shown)


translator/translate/otel/receiver/otlp/translator.go
```diff
@@ -125,3 +217,11 @@ func (t *translator) Translate(conf *confmap.Conf) (component.Config, error) {
...
 	return cfg, nil
 }
+
+// ResetRegistry clears the shared receiver registry for testing
+func ResetRegistry() {
+	registryMu.Lock()
+	defer registryMu.Unlock()
+	registry = make(map[string]common.ComponentTranslator)
+	counter = 0
+}
```
jefchien: nit: If it's only used for testing, move it into `translator_test.go`.

translator/translate/otel/receiver/otlp/translator.go
```diff
@@ -73,6 +74,97 @@ func (t *translator) ID() component.ID {
...
+	for existingKey := range registry {
+		if existingKey != fullKey && existingKey[:len(portKey)] == portKey && len(existingKey) > len(portKey) {
+			return "", fmt.Errorf("port conflict: endpoints %s already in use with different TLS configuration", portKey)
+		}
+	}
+
+	return fullKey, nil
+}
+
+func NewSharedTranslator(conf *confmap.Conf, opts ...common.TranslatorOption) common.ComponentTranslator {
```
jefchien: Is there ever a case where we don't use the shared translator?

nit: Would rather this just returned the error and failed here.

translator/translate/otel/receiver/otlp/translator.go
```diff
@@ -73,6 +74,97 @@ func (t *translator) ID() component.ID {
 	return component.NewIDWithName(t.factory.Type(), t.Name())
 }
 
+var (
+	registryMu sync.RWMutex
```
jefchien: Do we need synchronization? I thought all translation was done on a single thread.

translator/translate/otel/receiver/otlp/translator.go
```diff
@@ -73,6 +74,97 @@ func (t *translator) ID() component.ID {
...
+	registry   = make(map[string]common.ComponentTranslator)
+	counter    int
+)
+
+func getEndpointKey(conf *confmap.Conf, configKey string, index int, signal pipeline.Signal) (string, error) {
```
jefchien: Doesn't take into account the JMX default endpoint.

translator/translate/otel/receiver/otlp/translator.go
```diff
@@ -73,6 +74,97 @@ func (t *translator) ID() component.ID {
...
+				tlsCert = cert
+			}
+			if key, ok := tls["key_file"].(string); ok {
+				tlsKey = key
+			}
+		}
+	}
+
+	portKey := fmt.Sprintf("%s|%s", grpcEndpoint, httpEndpoint)
+	fullKey := fmt.Sprintf("%s|%s|%s", portKey, tlsCert, tlsKey)
```
jefchien: Rather than a string key, would it make sense to have a struct with string fields? It's still comparable, but would make accessing the fields easier. We're basically doing all of the translation parsing logic here. Maybe we could just pass this struct in when creating the translator. That way we don't have to do it again.

jefchien:
Can we add sample JSON configurations with port conflicts for testing?

translator/translate/otel/receiver/otlp/translator.go
movence: you might be right, and this is just a safety net since we have seen concurrent map issues in the agent.

translator/translate/otel/receiver/otlp/translator.go
sky333999: Yeah that might be better.
Similar to https://github.com/amazon-contributing/opentelemetry-collector-contrib/blob/af455c296233f12589cdb5774b6f19ad62b16cfc/exporter/awsemfexporter/grouped_metric.go#L17

translator/translate/otel/pipeline/applicationsignals/translator.go
```diff
@@ -52,7 +52,7 @@ func (t *translator) Translate(conf *confmap.Conf) (*common.ComponentTranslators
 	}
 
 	translators := &common.ComponentTranslators{
-		Receivers:  common.NewTranslatorMap(otlp.NewTranslator(common.WithName(common.AppSignals), otlp.WithSignal(t.signal))),
+		Receivers:  common.NewTranslatorMap(otlp.NewSharedTranslator(conf, common.WithName(common.AppSignals), otlp.WithSignal(t.signal))),
```
sky333999: Mostly just personal preference.. but dont think we need to call it `NewSharedTranslator` explicitly. Think its name can remain what it used to be, just that the logic for `otlp` ensures its "shareable".

translator/translate/otel/receiver/otlp/translator.go
sky333999: +1 dont think anything should use the non-shared `NewTranslator` anymore?

translator/translate/otel/receiver/otlp/translator.go
```diff
@@ -73,6 +74,97 @@ func (t *translator) ID() component.ID {
...
+		// Return error translator that will fail during Translate()
+		return &errorTranslator{err: err}
+	}
+
+	if existing, exists := registry[key]; exists {
+		return existing
+	}
+
+	counter++
+	t.SetName(fmt.Sprintf("otlp_shared_%d", counter))
```
sky333999: Id follow the pattern we have today:
```
t.SetName(t.Name() + "/" + strconv.Itoa(counter))
```

translator/tocwconfig/sampleConfig/appsignals_and_eks_config.json
```diff
@@ -29,7 +29,12 @@
   },
   "traces": {
     "traces_collected": {
-      "application_signals": {}
+      "application_signals": {
```
movence: This is now enforced since partial configs are not supported for the same endpoint.

translator/translate/otel/pipeline/applicationsignals/translator.go
```diff
@@ -52,12 +52,24 @@ func (t *translator) Translate(conf *confmap.Conf) (*common.ComponentTranslators
...
 	translators := &common.ComponentTranslators{
-		Receivers:  common.NewTranslatorMap(otlp.NewTranslator(common.WithName(common.AppSignals), otlp.WithSignal(t.signal))),
+		Receivers:  common.NewTranslatorMap[component.Config, component.ID](),
 		Processors: common.NewTranslatorMap[component.Config, component.ID](),
 		Exporters:  common.NewTranslatorMap[component.Config, component.ID](),
 		Extensions: common.NewTranslatorMap[component.Config, component.ID](),
 	}
 
+	// Add OTLP receivers
+	otlps, err := otlp.ParseOtlpConfig(conf, common.AppSignals, "", t.signal, -1)
```
movence: This block is kind of redundant in pipeline translators so could add a help function or something

translator/translate/otel/receiver/otlp/translator.go
```diff
@@ -29,99 +34,161 @@ const (
...
+		cfg.HTTP.ServerConfig.Endpoint = otlpConfig.endpoint
+		cfg.HTTP.ServerConfig.TLSSetting = tlsSettings
+	} else {
+		cfg.HTTP = nil
+		cfg.GRPC.NetAddr.Endpoint = otlpConfig.endpoint
+		cfg.GRPC.TLSSetting = tlsSettings
+	}
+
+	configCache[otlpConfig] = cfg
+	t.cfg = cfg
```
jefchien: nit: Could we store the `EndpointConfig` on the translator and move this to the `Translate` function? It can still use the cache, but you wouldn't need to store the `cfg` or `err`.

translator/translate/otel/receiver/otlp/translator.go
```diff
@@ -29,99 +34,161 @@ const (
 type translator struct {
 	common.NameProvider
 	common.IndexProvider
-	configKey string
-	signal    pipeline.Signal
-	factory   receiver.Factory
+	signal  pipeline.Signal
```
jefchien: Do we still need this field and the `WithSignal` function? It doesn't seem like they're used anymore since the `ParseOtlpConfig` function already finds the right config section.

translator/translate/otel/pipeline/applicationsignals/translator.go
```diff
@@ -52,12 +52,24 @@ func (t *translator) Translate(conf *confmap.Conf) (*common.ComponentTranslators
...
+	if err == nil {
+		for _, otlpConfig := range otlps {
+			translators.Receivers.Set(otlp.NewTranslator(
+				otlpConfig,
+				otlp.WithSignal(t.signal),
+				common.WithName(common.AppSignals)),
```
jefchien: nit: I don't think these are actually used, so it might be a bit misleading to keep them.

translator/tocwconfig/tocwconfig_test.go
```diff
@@ -880,6 +887,9 @@ func resetContext(t *testing.T) {
 	ecsutil.GetECSUtilSingleton().Region = ""
 	context.ResetContext()
 
+	// Clear OTLP config cache to avoid conflicts between tests
+	otlp.ClearConfigCache()
```
jefchien, line 911: Can we move this into the `resetContext` function? We may as well always clear the cache before each test.

translator/tocwconfig/sampleConfig/appsignals_and_eks_config.json
jefchien: We can't change this. We have to make an exception for application signals or this will break existing customers.

translator/translate/otel/receiver/otlp/translator.go
movence: absolutely! this was leftover before adding pre-parsing at the pipeline level. Moved the logic into `Translate` and removed those instance vars.

translator/translate/otel/receiver/otlp/translator.go
movence: Nope, removed

translator/tocwconfig/tocwconfig_test.go
movence: This is already in `resetContext` 🙃

jefchien: Oops you're right. Not sure how I misread that.

translator/config/sampleSchema/validOTLPShared.json
jefchien, line 1: nit: Can we add an invalid case as well?

translator/translate/otel/receiver/otlp/translator.go
```diff
@@ -29,99 +34,157 @@ const (
...
-	return cfg, nil
+
+	// If no specific endpoints configured, return defaults
+	if len(configs) == 0 {
+		configs = append(configs, EndpointConfig{
+			protocol: GRPC, endpoint: grpcDefault, certFile: certFile, keyFile: keyFile,
+		})
+		configs = append(configs, EndpointConfig{
+			protocol: HTTP, endpoint: httpDefault, certFile: certFile, keyFile: keyFile,
+		})
```
jefchien: nit: Can combine the append.
```suggestion
		configs = append(configs, EndpointConfig{
			protocol: GRPC, endpoint: grpcDefault, certFile: certFile, keyFile: keyFile,
		}, EndpointConfig{
			protocol: HTTP, endpoint: httpDefault, certFile: certFile, keyFile: keyFile,
		})
```

translator/translate/otel/receiver/otlp/translator.go
```diff
@@ -29,99 +34,157 @@ const (
...
-	if t.Name() == "" && t.signal.String() != "" {
-		t.SetName(t.signal.String())
-		if t.Index() != -1 {
-			t.SetName(t.Name() + "/" + strconv.Itoa(t.Index()))
-		}
+
+	t.pipelineName = t.Name()
+	// set name as "{type - http or grpc}" then appends "_{port}" if available
+	t.SetName(string(otlpConfig.protocol))
+	if parts := strings.Split(otlpConfig.endpoint, ":"); len(parts) > 1 {
```
jefchien: nit: This will probably have an unexpected name for IPv6, but as long as it's consistent, it's fine.

translator/translate/otel/receiver/otlp/translator.go
```diff
@@ -29,99 +34,157 @@ const (
...
 
-// WithSignal determines where the translator should look to find
-// the configuration.
-func WithSignal(signal pipeline.Signal) common.TranslatorOption {
-	return func(target any) {
-		if t, ok := target.(*translator); ok {
-			t.signal = signal
-		}
-	}
+type EndpointConfig struct {
```
jefchien, line 40: nit: Type does not need to be exported.
```suggestion
type endpointConfig struct {
```

jefchien (approved):

translator/translate/otel/receiver/otlp/translator.go
```diff
@@ -29,99 +34,154 @@ const (
...
 	for _, opt := range opts {
 		opt(t)
 	}
-	if t.Name() == "" && t.signal.String() != "" {
-		t.SetName(t.signal.String())
-		if t.Index() != -1 {
-			t.SetName(t.Name() + "/" + strconv.Itoa(t.Index()))
-		}
+
+	t.pipelineName = t.Name()
```
sky333999, line 73: At this line specifically, what is setting the name (t.Name()) by the time we get here?

Because i see it getting created few lines above and there's no calls to SetName().

translator/translate/otel/receiver/otlp/translator.go
```diff
@@ -29,99 +34,154 @@ const (
...
+	if existingCfg, exists := configCache[t.endpointConfig]; exists {
+		return existingCfg, nil
+	}
+
+	for cachedConfig := range configCache {
+		if cachedConfig.protocol == t.endpointConfig.protocol && cachedConfig.endpoint == t.endpointConfig.endpoint &&
+			(cachedConfig.certFile != t.endpointConfig.certFile || cachedConfig.keyFile != t.endpointConfig.keyFile) {
+			// ignores (missing) TLS conflict for application signals pipelines https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Agent-Application_Signals.html
+			if t.pipelineName == common.AppSignals {
+				return configCache[cachedConfig], nil
```
sky333999: So for this case, we assume two sections in the agent config defined - one that has TLS and one that doesnt - so a dense and a spare config.
If the sparse config is the one stored in `configCache`, and we are now running this line of code corresponding to the dense config, will we just return the cached sparse config?

translator/translate/otel/receiver/otlp/translator.go
sky333999: Disregard this.. i see we do `otlp.NewTranslator(common.WithName(common.PipelineNameJmx))` so its one of the `opts`.

translator/translate/otel/receiver/otlp/translator.go
```diff
@@ -29,99 +34,154 @@ const (
...
-	if t.Name() == "" && t.signal.String() != "" {
-		t.SetName(t.signal.String())
-		if t.Index() != -1 {
-			t.SetName(t.Name() + "/" + strconv.Itoa(t.Index()))
-		}
+
+	t.pipelineName = t.Name()
+	// set name as "{type - http or grpc}" then appends "_{port}" if available
+	t.SetName(string(otlpConfig.protocol))
+	if parts := strings.Split(otlpConfig.endpoint, ":"); len(parts) > 1 {
```
sky333999: nit: also an edge case of two otlp receivers with diff dns host names, but same port - we will create them with the same name.
I dont think we need to handle that for now.

translator/translate/otel/pipeline/applicationsignals/translator.go
```diff
@@ -52,12 +52,23 @@ func (t *translator) Translate(conf *confmap.Conf) (*common.ComponentTranslators
...
 	translators := &common.ComponentTranslators{
-		Receivers:  common.NewTranslatorMap(otlp.NewTranslator(common.WithName(common.AppSignals), otlp.WithSignal(t.signal))),
+		Receivers:  common.NewTranslatorMap[component.Config, component.ID](),
 		Processors: common.NewTranslatorMap[component.Config, component.ID](),
 		Exporters:  common.NewTranslatorMap[component.Config, component.ID](),
 		Extensions: common.NewTranslatorMap[component.Config, component.ID](),
 	}
 
+	// Add OTLP receivers
+	otlps, err := otlp.ParseOtlpConfig(conf, common.AppSignals, "", t.signal, -1)
```
sky333999: Im wondering if this contract here should still be the same i.e. we just call `otlp.NewTranslator` here and then we just tweak `NewTranslator` to internally have this logic and also invoke `ParseOtlpConfig` internally.

For reference - https://github.com/aws/amazon-cloudwatch-agent/blob/sky333999/otlp/translator/translate/otel/receiver/otlp/translator.go. In this file, my `translate` is the equivalent of your `ParseOtlpConfig`.
So we still keep things like `WithSignal` and none of these pipeline translators would actually change (except passing in conf).

jefchien: I think the difference is that `ParseOtlpConfig` can generate multiple OTLP receivers, so would have to be a `NewTranslators` function. I like that. I think that would make it clearer what the purpose is. Could just have it return `[]common.ComponentTranslator`.

jefchien (approved):

translator/translate/otel/receiver/otlp/translator.go
```diff
@@ -29,99 +33,141 @@ const (
...
-			return nil, fmt.Errorf("no application_signals config key defined for signal: %s", t.signal)
-		}
-		if conf.IsSet(appSignalsConfigKeys[0]) {
-			configKey = appSignalsConfigKeys[0]
-		} else {
-			configKey = appSignalsConfigKeys[1]
-		}
-		cfg.GRPC.NetAddr.Endpoint = defaultAppSignalsGrpcEndpoint
-		cfg.HTTP.ServerConfig.Endpoint = defaultAppSignalsHttpEndpoint
+func TranslateToEndpointConfig(conf *confmap.Conf, pipelineName string, configKey string, index int) []EndpointConfig {
```
jefchien: nit: This no longer needs to be exported. Consider moving it into `translators.go` since it's only used there.

translator/translate/otel/receiver/otlp/translator.go
```diff
@@ -29,99 +33,141 @@ const (
 type translator struct {
 	common.NameProvider
 	common.IndexProvider
```
jefchien: nit: I don't think we need this anymore. The index and name are handled by the `TranslateToEndpointConfig`.

translator/translate/otel/receiver/otlp/translator.go
```diff
@@ -29,99 +33,141 @@ const (
...
 		opt(t)
 	}
-	if t.Name() == "" && t.signal.String() != "" {
-		t.SetName(t.signal.String())
-		if t.Index() != -1 {
-			t.SetName(t.Name() + "/" + strconv.Itoa(t.Index()))
-		}
-	}
+
+	t.pipelineName = t.Name()
```
jefchien, line 73: Will `t.Name()` ever not be empty? The only place we're calling `otlp.NewTranslator` now just passes in the `EndpointConfig`.

movence: updated the code to pass pipelineName at translator initialization

translator/translate/otel/receiver/otlp/translator.go
```diff
@@ -28,100 +32,144 @@ const (
...
+	// set name as "{type - http or grpc}" then appends an endpoint by replacing special chars with '_' eg. 'http_0_0_0_0_4316'
+	t.SetName(string(epConfig.protocol) + "_" + epRegex.ReplaceAllString(epConfig.endpoint, "_"))
```
sky333999, line 75: nit: Probably fine for now but the hash of the endpoint might be better given some DNS names can get quite long. We have a `HashName` util if we need one.

translator/translate/otel/receiver/otlp/translator.go
```diff
@@ -28,100 +32,144 @@ const (
...
+
+	for cachedConfig := range configCache {
+		if cachedConfig.protocol == t.endpointConfig.protocol && cachedConfig.endpoint == t.endpointConfig.endpoint &&
+			(cachedConfig.certFile != t.endpointConfig.certFile || cachedConfig.keyFile != t.endpointConfig.keyFile) {
+			// ignores (missing) TLS conflict for application signals pipelines when one has TLS and the other doesn't
+			// https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-Agent-Application_Signals.html
+			if t.pipelineName == common.AppSignals {
+				cachedHasTLS := cachedConfig.certFile != "" || cachedConfig.keyFile != ""
+				currentHasTLS := t.endpointConfig.certFile != "" || t.endpointConfig.keyFile != ""
+				if cachedHasTLS != currentHasTLS {
```
sky333999, line 105: Does this always work?
So assume whats in cache has not TLS and current does have TLS. `cachedHasTLS` will be false and `currentHasTLS` will be true -> we will enter this clause but then return what is cached i.e. without TLS. Shouldnt we be updating the cache to the one with TLS and return that?

I think for the flip case where cached has TLS and current does not is fine even per current logic.

movence: Yeah, that is a kind of an edge case with this approach (or previous). Once an endpointConfig is cached that means a receiver (component) config has been already been returned to a pipeline, and there is no way to update it. At the same time, it will be wrong to just update the cached entry when the returned component is clearly different. 

We do have some test cases (Eg https://github.com/aws/amazon-cloudwatch-agent/blob/main/translator/tocwconfig/sampleConfig/appsignals_and_eks_config.json) covering these scenarios. Before some recent commits, the rule was more strict failing these existing cases which seemed wrong for any existing expectations and configs. This is middle ground to support existing cases and use whatever is already in cache.

PR #1827, opened 2025-08-20
Ecs Observer config

agarakan:
# Description of the issue
Updates ECS Observer Config Translator for parity with existing Ecs Service Discovery defaults

# Description of changes
List of modifications
* Dynamically retrieve default value for clusterName and clusterRegion when not provided (make them optional)
* Update refreshDuration default to `1m`
* Update ResultFile default to `/tmp/cwagent_ecs_auto_sd.yaml`
* Update JobNameLabel default to `job`
* Update MetricPathLabel default to `ECS_PROMETHEUS_METRICS_PATH`

# License

# Tests
Updated unit tests & all tests pass
```
make test
make fmt
make fmt-sh
make lint
```

# Requirements
2. Run `make lint`

-------
### Integration Tests

### DIFF
--- a/translator/translate/otel/extension/ecsobserver/translator_test.go
+++ b/translator/translate/otel/extension/ecsobserver/translator_test.go
@@ -371,6 +371,23 @@ func TestTranslator_Translate(t *testing.T) {
 			},
 		},
 		{
+			name: "missing sd_target_cluster uses ECS util",
+			config: map[string]interface{}{
+				"logs": map[string]interface{}{
+					"metrics_collected": map[string]interface{}{
+						"prometheus": map[string]interface{}{
+							"ecs_service_discovery": map[string]interface{}{
+								"sd_frequency":      "1m",
+								"sd_cluster_region": "us-west-2",
+								"sd_result_file":    "/tmp/test.yaml",
+							},
+						},
+					},
+				},
+			},
+			wantErr: true, // ECS util returns empty in test environment
+		},
+		{
 			name: "missing required fields",
 			config: map[string]interface{}{
 				"logs": map[string]interface{}{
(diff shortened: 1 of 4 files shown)


translator/translate/otel/extension/ecsobserver/translator_test.go
```diff
@@ -371,13 +369,15 @@ func TestTranslator_Translate(t *testing.T) {
 			},
 		},
 		{
-			name: "missing required fields",
+			name: "missing sd_target_cluster uses ECS util",
```
dricross: This test is causing it to use ECS util to grab the target cluster, but the expected result is a nil config. That doesn't seem right. Is that just because the unit tests don't have something defined for `ecsutil.GetECSUtilSingleton().Region`? If so , could we set it to something to ensure it gets picked up?

agarakan: Yeah its correct, we dont have a mock set up to return the Region and Cluster Name from ECSUtil. I wasn't able to find guidance on mocks in go. I'll try to set the singleton as you suggested though

agarakan: Great, it worked! Validated tests pass, combined the scenario with an above scenario that tests the default values when none are provided. Resolving thread

translator/util/ecsutil/ecsutil.go
```diff
@@ -37,6 +37,9 @@ var ecsUtilInstance *ecsUtil
 var ecsUtilOnce sync.Once
 
 func GetECSUtilSingleton() *ecsUtil {
+	if ecsUtilInstance != nil {
+		return ecsUtilInstance
+	}
```
dricross: nit: I don't think you want / need this. That's what the `.Do` function does essentially.

agarakan: ack, validated in the docs. Removed and retested.

Paramadon (approved):

PR #1828, opened 2025-08-20
Ecs observer - Add Relabel Config to Prometheus Receiver Translator

agarakan:
# Description of the issue
Add Relabel Config to the Prometheus Receiver Translator when ECS Observer is enabled. The purpose of this change is to modify the Prometheus File-based service discovery labels collected by OTel ECS Observer to maintain parity with the existing  EcsServicediscovery telegraf plugin

# Description of changes
Updates the Prometheus Receiver Translator to append default Relabel Config when ECS Observer is enabled. The config are added to the generated OTEL Yaml file.

# License

# Tests
Added unit tests for added functionality. updated existing tests for latest behavior
```
make test
make fmt
make fmt-sh
make lint
```

See validation of included relabel fields via integration test:
Screenshot 2025-08-27 at 9 29 50 PM

Previous state: 

Screenshot 2025-08-27 at 12 13 01 PM

# Requirements
2. Run `make lint`

-------
### Integration Tests


translator/translate/otel/receiver/prometheus/translator.go
```diff
@@ -160,3 +164,39 @@ func escapeStrings(node any) {
...
+		{Regex: relabel.MustNewRegexp("^__meta_ecs_container_labels_(.+)$"), Action: relabel.LabelMap, Replacement: "$1"},
+	}
+
+	defaultMetricRelabelConfigs := []*relabel.Config{
+		{SourceLabels: model.LabelNames{"source"}, Regex: relabel.MustNewRegexp("^arn:aws:ecs:.*:.*:task.*\\/(.*)$"), Action: relabel.Replace, TargetLabel: "TaskId"},
+	}
+
+	for _, scrapeConfig := range scrapeConfigs {
+		if hasFileServiceDiscovery(scrapeConfig) {
+			scrapeConfig.RelabelConfigs = defaultRelabelConfigs
```
sky333999: We shouldnt overwrite, we should append

translator/translate/otel/receiver/prometheus/translator.go
```diff
@@ -160,3 +164,39 @@ func escapeStrings(node any) {
 		}
 	}
 }
+
+func addDefaultRelabelConfigs(scrapeConfigs []*config.ScrapeConfig) {
+	defaultRelabelConfigs := []*relabel.Config{
```
sky333999: This should all happen only if we are on ECS.

translator/translate/otel/receiver/prometheus/translator.go
```diff
@@ -160,3 +164,39 @@ func escapeStrings(node any) {
...
+		if hasFileServiceDiscovery(scrapeConfig) {
+			scrapeConfig.RelabelConfigs = defaultRelabelConfigs
+			scrapeConfig.MetricRelabelConfigs = defaultMetricRelabelConfigs
+		}
+	}
+}
+
+func hasFileServiceDiscovery(scrapeConfig *config.ScrapeConfig) bool {
+	for _, sdConfig := range scrapeConfig.ServiceDiscoveryConfigs {
+		if sdConfig.Name() == "file" {
```
sky333999: Can we be more specific to target the sd results file that we create?

agarakan: addressed

translator/translate/otel/receiver/prometheus/translator.go
agarakan: addressed

translator/translate/otel/receiver/prometheus/translator.go
agarakan: addressing in follow up CR as we need to consider merge logic

translator/translate/otel/receiver/prometheus/translator.go
```diff
@@ -160,3 +171,58 @@ func escapeStrings(node any) {
...
+	ecsSdResultFileKey := common.ConfigKey(ecsSDKey, ECS_SD_RESULT_FILE)
+	ecsSDFileName := defaultECSSDfileName
+	if conf.IsSet(ecsSdResultFileKey) {
+		if fileName, ok := conf.Get(ecsSdResultFileKey).(string); ok && fileName != "" {
+			ecsSDFileName = fileName
+		}
+	}
+
+	defaultRelabelConfigs := []*relabel.Config{
+		{SourceLabels: model.LabelNames{"__meta_ecs_cluster_name"}, Action: relabel.Replace, TargetLabel: "ClusterName", Regex: relabel.MustNewRegexp("(.*)")},
```
agarakan, line 190: ClusterName relabel config should only be added if not explicitly defined by customer. For now, adding config always. Introduces conditional logic in follow up CR: https://github.com/aws/amazon-cloudwatch-agent/compare/ecs_observer-relabel-config...cluster-name-relabel-conditional

translator/translate/otel/receiver/prometheus/translator.go
agarakan: https://github.com/aws/amazon-cloudwatch-agent/compare/support-labelmap-relabel...merge-prom-scrape-configs

translator/translate/otel/receiver/prometheus/translator.go
sky333999, line 190: Do you need to specify the regex to .*? Isnt that the default behavior?

translator/translate/otel/receiver/prometheus/translator.go
```diff
@@ -160,3 +171,58 @@ func escapeStrings(node any) {
...
+		{SourceLabels: model.LabelNames{"__meta_ecs_task_definition_family"}, Action: relabel.Replace, TargetLabel: "TaskDefinitionFamily", Regex: relabel.MustNewRegexp("(.*)")},
+		{SourceLabels: model.LabelNames{"__meta_ecs_task_definition_revision"}, Action: relabel.Replace, TargetLabel: "TaskRevision", Regex: relabel.MustNewRegexp("(.*)")},
+		{SourceLabels: model.LabelNames{"__meta_ecs_ec2_instance_type"}, Action: relabel.Replace, TargetLabel: "InstanceType", Regex: relabel.MustNewRegexp("(.*)")},
+		{SourceLabels: model.LabelNames{"__meta_ecs_ec2_subnet_id"}, Action: relabel.Replace, TargetLabel: "SubnetId", Regex: relabel.MustNewRegexp("(.*)")},
+		{SourceLabels: model.LabelNames{"__meta_ecs_ec2_vpc_id"}, Action: relabel.Replace, TargetLabel: "VpcId", Regex: relabel.MustNewRegexp("(.*)")},
+		{SourceLabels: model.LabelNames{"__meta_ecs_source"}, Regex: relabel.MustNewRegexp("^arn:aws:ecs:.*:.*:task.*\\/(.*)$"), Action: relabel.Replace, TargetLabel: "TaskId"},
+		{SourceLabels: model.LabelNames{"__meta_ecs_container_labels_app_x"}, Action: relabel.Replace, TargetLabel: "app_x", Regex: relabel.MustNewRegexp("(.*)")},
+	}
+
+	defaultMetricRelabelConfigs := []*relabel.Config{}
```
sky333999: Do we need this if its always just an empty array?

translator/translate/otel/receiver/prometheus/translator.go
sky333999: What is the merge logic needed here?

translator/translate/otel/receiver/prometheus/translator.go
```diff
@@ -160,3 +171,58 @@ func escapeStrings(node any) {
...
+
+	for _, scrapeConfig := range scrapeConfigs {
+		if hasConfiguredServiceDiscoveryResultFile(scrapeConfig, ecsSDFileName) {
+			scrapeConfig.RelabelConfigs = defaultRelabelConfigs
+			scrapeConfig.MetricRelabelConfigs = defaultMetricRelabelConfigs
+		}
+	}
+}
+
+func hasConfiguredServiceDiscoveryResultFile(scrapeConfig *config.ScrapeConfig, ecsSdResultFile string) bool {
```
sky333999: So assuming their scrape config has 10 diff sections in the array, all of them being fileSD configs.
One of the 10 is for the ecs sd file, but the remaining 9 are some other files that the user setup.

Per this func, since the ecs sd file is 1 of the 10, itll return true and the func above is going to append all these relabel configs to all 10 - which isnt right.

So id just merge this logic inline above and only append the relabel configs when there the file matches the ecs sd file.

translator/translate/otel/receiver/prometheus/translator_test.go
```diff
@@ -291,6 +294,187 @@ func TestMetricsEmfTranslator(t *testing.T) {
...
+			},
+		},
+	})
+
+	configKey := "logs.metrics_collected.prometheus"
+
+	addDefaultECSRelabelConfigs(scrapeConfigs, conf, configKey)
+
+	// Should add configs because ecs_service_discovery is explicitly configured
+	assert.Len(t, scrapeConfigWithFileSD.RelabelConfigs, 13, "Should add 14 relabel configs when ecs_service_discovery is explicitly configured")
```
sky333999: The len assert is for 13, but the comment says 14? Probably just drop the 14 from the comment.

translator/translate/otel/receiver/prometheus/translator_test.go
```diff
@@ -291,6 +294,187 @@ func TestMetricsEmfTranslator(t *testing.T) {
...
+	assert.Equal(t, "ClusterName", scrapeConfigWithFileSD.RelabelConfigs[0].TargetLabel)
```
sky333999: Checking just a few based on the index isnt maintainable long term. Since you have the array, just store the expected labels in a set and check if they all exist if you need to.

translator/translate/otel/receiver/prometheus/translator.go
agarakan: Originally I was under the impression we need to replace instances of "originalLabel" with "newLabel", but after digging I found prometheus will handle relabels in order, so as long as I can execute my dynamic relabel config first, I can ensure the customers config is appended after. 

The merge logic is not merge logic anymore, its just appending in order and ensuring the order

translator/translate/otel/receiver/prometheus/translator.go
agarakan: Without defining regex, then when the config is evaluated by the prometheus receiver, it appends an additional field called Separator and Regex, which then append values that are rejected by the Otel yaml validator. I did not document the default values that show up, but ultimately it causes more validation pain to omit the Regex field although it does default to the same value I set it to

I tested it a few times and concluded this is the solution.

translator/translate/otel/receiver/prometheus/translator.go
agarakan: Yeah good find, i actually deleted it in a different commit. Seems I forgot to add that commit

translator/translate/otel/receiver/prometheus/translator_test.go
agarakan: forgot to merge in the fix

translator/translate/otel/receiver/prometheus/translator_test.go
agarakan: ack

translator/translate/otel/receiver/prometheus/translator.go
agarakan: Reviewing logic for the scenario you laid out

agarakan: I see the issue, i added a test for the scenario and modified the logic so it now only adds relabelConfig on the scrapeConfig that has matching resultFile names.

translator/translate/otel/receiver/prometheus/translator.go
```diff
@@ -160,3 +171,47 @@ func escapeStrings(node any) {
...
+		{SourceLabels: model.LabelNames{"__meta_ecs_ec2_vpc_id"}, Action: relabel.Replace, TargetLabel: "VpcId", Regex: relabel.MustNewRegexp("(.*)")},
+		{SourceLabels: model.LabelNames{"__meta_ecs_source"}, Regex: relabel.MustNewRegexp("^arn:aws:ecs:.*:.*:task.*\\/(.*)$"), Action: relabel.Replace, TargetLabel: "TaskId"},
+		{SourceLabels: model.LabelNames{"__meta_ecs_container_labels_app_x"}, Action: relabel.Replace, TargetLabel: "app_x", Regex: relabel.MustNewRegexp("(.*)")},
+	}
+
+	for _, scrapeConfig := range scrapeConfigs {
+		for _, sdConfig := range scrapeConfig.ServiceDiscoveryConfigs {
+			if fileSDConfig, ok := sdConfig.(*file.SDConfig); ok {
+				for _, filePath := range fileSDConfig.Files {
+					if slices.Contains([]string{filePath}, ecsSDFileName) {
```
sky333999, line 209: I dont get this. We seem to be converting a string to an array to only check if another string exists in it?
Why not just compare `filePath` to `ecsSDFileName`

sky333999 (approved):

translator/translate/otel/receiver/prometheus/translator.go
agarakan: discussed offline, seems we can simplify this to use `==` for string equivalence, i was under the impression filepath was an array of strings. Will follow up in next CR

PR #1832, opened 2025-08-21
Adding Nvme Instance Store and EBS Adoption metrics

Paramadon:
# Description of the issue
Currently there is no ebs or instance store specific adoption metrics, so there isn't a easy way to tell how many hosts are scraping ebs or instance store metrics. 

# Description of changes
 This pr adds nvme_ebs/nvme_is to the CloudWatch Agent useragent string if diskio_ebs (for nvme_ebs) or diskio_instance_store(for nvme_is) metric prefix exist in the cloudwatch exporter. If it does, `nvme_ebs/nvme_is` is added to the feature section of the user agent if it exists or creates a feature section if it doesn't exist and adds them.

# Tests
Created a unit test to mock a request and see if the expected outcome was given, also tested manually on a ec2 instance to see if CWA useragent string was populated with the feature flags.

Screenshot 2025-08-21 at 2 08 14 PM


plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -175,17 +207,17 @@ func (c *CloudWatch) pushMetricDatum() {
 			entity, datums := c.BuildMetricDatum(metric)
 			numberOfPartitions := len(datums)
 			/* We currently do not account for entity information as a part of the payload size.
-			This is by design and should be revisited once the SDK protocol changes.
-			In the meantime there has been a payload limit increase applied in the background to accommodate this decision
+			   This is by design and should be revisited once the SDK protocol changes.
```
Paramadon, line 210: auto formats to this for some reason

zhihonl: nit: line 220 is not aligning properly. Let's either remove this modification or fix the alignment.

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -72,6 +82,9 @@ type CloudWatch struct {
 	aggregatorShutdownChan chan struct{}
 	aggregatorWaitGroup    sync.WaitGroup
 	lastRequestBytes       int
+	mu                     sync.RWMutex
```
zhihonl, line 85: Why is mutex needed? Thought this exporter doesn't have concurrencies where we can access `featureList` concurrently but I could be wrong

plugins/outputs/cloudwatch/cloudwatch.go
```diff
@@ -419,6 +451,28 @@ func (c *CloudWatch) WriteToCloudWatch(req interface{}) {
...
+		if _, exists := c.featureList[featureFlagNvmeIS]; !exists && strings.HasPrefix(name, metricPrefixInstanceStore) {
+			c.featureList[featureFlagNvmeIS] = struct{}{}
+			changed = true
+		}
+		if changed {
+			var features []string
+			for f := range c.featureList {
+				features = append(features, f)
+			}
+			sort.Strings(features)
```
zhihonl, line 471: Do we need to check for empty array before setting `c.prebuiltFeature`

plugins/outputs/cloudwatch/cloudwatch.go
Paramadon: We need a Mutex because we are reading and writing to the same map at different points.

plugins/outputs/cloudwatch/cloudwatch.go
Paramadon: array will not be empty since the changed variable is true meaning there is something in the feature list.

zhihonl (approved):

varunch77 (approved):

PR #1848, opened 2025-09-05
Address goroutine leak with dynamically determined log destinations

dricross:
# Description of the issue

## Everliving Destinations
The CloudWatch Agent can publish logs from log files to CloudWatch Log Streams determined by the file name. Each log file the agent is reading from creates a "source" object (specifically `tailerSrc` type) with several running goroutines, and each log stream that the CloudWatch Agent is pushing logs to creates a "destination" object (specifically `cwDest` type) with several running goroutines. The source objects' goroutines are closed when the associated log file is closed, but the destination objects are never subsequently cleaned up. This causes a memory leak as the goroutines are never closed and keep piling up.

Dynamically generated log stream names can be generated when using the `publish_multi_logs` flag. Here's is a sample entry in the collect list:
```json
          {
            "publish_multi_logs": true,
            "file_path": "/tmp/test_logs/publish_multi_logs_*.log",
            "log_group_name": "test-log-rotation",
            "log_stream_name": "rotation-test-stream",
            "timezone": "UTC",
            "timestamp_format": "%Y-%m-%dT%H:%M:%S",
            "multi_line_start_pattern": "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}",
            "retention_in_days": 1
          },
```
For this config, the agent will periodically look for log files that match the `/tmp/test_logs/publish_multi_logs_*.log` glob pattern. For each one that it finds, it will write the contents of that file to `test-log-rotation/rotiatest-test-stream-<logfilename>` loggroup/logstream. So each file it finds will create one new source object and one new destination object.

It's important to note that there may be several source objects referencing one destination. For example, a customer could use the following in their collect list to collect logs from different files but push to the same destination:
```json
          {
            "file_path": "/tmp/test_logs/shared_destination.txt",
            "log_group_name": "test-log-rotation",
            "log_stream_name": "shared-destination-stream",
            "timezone": "UTC",
            "timestamp_format": "%Y-%m-%dT%H:%M:%S",
            "multi_line_start_pattern": "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}",
            "retention_in_days": 1
          },
          {
            "file_path": "/tmp/test_logs/shared_destination_to_close.txt",
            "log_group_name": "test-log-rotation",
            "log_stream_name": "shared-destination-stream",
            "timezone": "UTC",
            "timestamp_format": "%Y-%m-%dT%H:%M:%S",
            "multi_line_start_pattern": "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}",
            "retention_in_days": 1
          }
```

## Duplicate cloudwatch logs clients 
Each time a destination object is created, a new cloudwatch logs client is created. Request/response handlers are injected into the client to collect agent health metrics. These handlers have more underlying clients which have their own goroutines and caches. The underlying cache is a way to associate request data with response data, like Payload size and latency, which are eventually forwarded to the agent health metrics extension. The handlers are created by a middleware configurer. Each time a new cloudwatch logs client is created, a new middleware configurer is created, which creates new request/response handlers and spins up more goroutines. These clients have no way to be closed so the goroutines are never closed and keep piling up.

# Description of changes

There are two sets of changes to address the two issues noted above.

## Reference Counting Destinations
Add reference counting to the destination objects and a notification system for the source objects to tell the destinations it's no longer being used. When the destination object is no longer used, it stops itself and tells the CloudWatchLogs plugin it's no longer usable.

This implementation assumes that nothing other than the source objects are using the destination objects. There is some vestigial code left over from when the CloudWatchLogs telegraf plugin was used to process EMF metrics which could use the destination objects, but that functionality has been moved to the OTel awsemfexporter plugin. So that code path is more or less unreachable.

As mentioned previously, it is important to note that it is possible for multiple source objects to reference one destination. This possibility of sharing means source objects cannot close the destinations directly and a signaling mechanism is necessary.

Using the above JSON as an example, the agent will create two sources for each of the log files and one destination object for the `test-log-rotation/shared-destination-stream` loggroup/logstream. When the file `shared_destination_to_close.txt` is closed, the source object will notify the destination object that the source has closed and it's no longer being referenced. The destination will remain open since it knows the `shared_destination.txt` is still using it. If `shared_destination.txt` is closed, then the destination will then know it's no longer being used and will close itself, terminating the associated goroutines.

## Single Middleware Configurer
Create one middleware configurer (per CloudWatchLogs instance, but it's a singleton, so there's only ever one configurer). These request/response handlers are all identical, so there's really no need to create new ones, nor is there a need to create a new middleware configurer.

The one side effect is that destination objects will share request/response handlers, which means there is only one cache to support all request/responses. There are no concurrency concerns here as the request/response handlers already support concurrent request/responses.

# License

# Tests
Configure agent with the following configuration:
```json
{
  "agent": {"debug":true},
  "logs": {
    "logs_collected": {
      "files": {
        "collect_list": [
          {
            "publish_multi_logs": true,
            "file_path": "/tmp/test_logs/publish_multi_logs_*.log",
            "log_group_name": "test-log-rotation",
            "log_stream_name": "rotation-test-stream",
            "timezone": "UTC",
            "timestamp_format": "%Y-%m-%dT%H:%M:%S",
            "multi_line_start_pattern": "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}",
            "retention_in_days": 1
          },
          {
            "file_path": "/tmp/test_logs/shared_destination.txt",
            "log_group_name": "test-log-rotation",
            "log_stream_name": "shared-destination-stream",
            "timezone": "UTC",
            "timestamp_format": "%Y-%m-%dT%H:%M:%S",
            "multi_line_start_pattern": "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}",
            "retention_in_days": 1
          },
            {
            "file_path": "/tmp/test_logs/shared_destination_to_close.txt",
            "log_group_name": "test-log-rotation",
            "log_stream_name": "shared-destination-stream",
            "timezone": "UTC",
            "timestamp_format": "%Y-%m-%dT%H:%M:%S",
            "multi_line_start_pattern": "^\\d{4}-\\d{2}-\\d{2}T\\d{2}:\\d{2}:\\d{2}",
            "retention_in_days": 1
          }
        ]
      }
    }
  }
```

Use the following python script to generate log files that match the configuration:
```
#!/usr/bin/env python3

import time
import os
from datetime import datetime
from threading import Thread

log_dir = "/tmp/test_logs"

def generate_shared_destination_logs():
    os.makedirs(log_dir, exist_ok=True)
    
    shared_file = open(f"{log_dir}/shared_destination.txt", 'w')
    shared_to_close_file = open(f"{log_dir}/shared_destination_to_close.txt", 'w')
    
    try:
        start_time = time.time()
        line_counter = 1
        
        while True:
            current_time = time.time()
            elapsed = current_time - start_time
            
            log_line = f"{datetime.now().isoformat()} INFO Processing request {line_counter}\n"
            
            # Write to shared_destination throughout
            shared_file.write(log_line)
            shared_file.flush()
            
            # Write to shared_destination_to_close only for first 20 seconds
            if elapsed < 20:
                shared_to_close_file.write(log_line)
                shared_to_close_file.flush()
            elif elapsed >= 20 and shared_to_close_file:
                shared_to_close_file.close()
                shared_to_close_file = None
                print("Closed shared_destination_to_close.txt after 20 seconds")
            
            line_counter += 1
            time.sleep(0.5)
            
    except KeyboardInterrupt:
        print("\nStopping log generation...")
    finally:
        shared_file.close()
        if shared_to_close_file:
            shared_to_close_file.close()


def generate_publish_multi_logs():
    log_dir = "/tmp/test_logs"
    os.makedirs(log_dir, exist_ok=True)

    file_counter = 1
    current_file = None

    try:
        while True:
            # Close previous file and open new one
            if current_file:
               # current_file.close()
                os.remove(current_file.name)

            timestamp = datetime.now().strftime("%Y%m%d_%H%M%S")
            filename = f"{log_dir}/publish_multi_logs_{timestamp}_{file_counter:03d}.log"
            current_file = open(filename, 'w')
            print(f"Created new log file: {filename}")

            # Write logs for 10 seconds
            start_time = time.time()
            line_counter = 1

            while time.time() - start_time < 10:
                log_line = f"{datetime.now().isoformat()} INFO [app_{file_counter:03d}] Processing request {line_counter}\n"
                current_file.write(log_line)
                current_file.flush()
                line_counter += 1
                time.sleep(0.5)  # Write every 500ms

            file_counter += 1

    except KeyboardInterrupt:
        print("\nStopping log generation...")
    finally:
        if current_file:
            current_file.close()

if __name__ == "__main__":
    multi_logs_thread = Thread(target = generate_publish_multi_logs, args = ())
    shared_destination_thread = Thread(target = generate_shared_destination_logs, args = ())
    multi_logs_thread.start()
    shared_destination_thread.start()

    multi_logs_thread.join()
    shared_destination_thread.join()
```

# Requirements
2. Run `make lint`

-------
### Integration Tests


plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -340,12 +367,32 @@ func (cd *cwDest) Publish(events []logs.LogEvent) error {
...
 	cd.retryer.Stop()
+	cd.pusher.Stop()
 	cd.stopped = true
+	if cd.onStopFunc != nil {
+		cd.onStopFunc()
+	}
 }
 
 func (cd *cwDest) AddEvent(e logs.LogEvent) {
+	if cd.stopped {
```
duhminick: questionable nit: possible race condition here? let's say this reads false which can be stale, it'll try to push to a stopped pusher.

though the consequences aren't probably too high (?) I haven't dug into the code so I'm not entirely sure

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -320,11 +335,23 @@ type cwDest struct {
...
+	stopped    bool
+	onStopFunc func()
 }
 
+var _ logs.LogDest = (*cwDest)(nil)
+
 func (cd *cwDest) Publish(events []logs.LogEvent) error {
+	if cd.stopped {
+		return fmt.Errorf("cannot publish events: destination has been stopped")
+	}
```
dricross, line 262: Just noticed there's already a `cd.stopped` check at the end of the function. I don't know why it the check is at the end, I think we want to avoid handling any events if it's already stopped.

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
dricross: That's a good point. We could accidentally write to a closed channel which will panic. We'll need to lock the cwDest for the duration of this call so it doesn't close in the middle.

logs/logs.go
```diff
@@ -168,14 +166,17 @@ func (l *LogAgent) runSrcToDest(src LogSrc, dest LogDest) {
 	for e := range eventsCh {
 		err := dest.Publish([]LogEvent{e})
 		if err == ErrOutputStopped {
-			log.Printf("I! [logagent] Log destination %v has stopped, finalizing %v/%v", l.destNames[dest], src.Group(), src.Stream())
```
dricross, line 171: There was actually a concurrent map write condition here. This non-sync map is written to in the `Run` routine and is potentially read from the `runSrcToDest` goroutine.

plugins/inputs/logfile/logfile.go
```diff
@@ -248,11 +249,6 @@ func (t *LogFile) FindLogSrc() []logs.LogSrc {
 				}
 			}
 
-			destination := fileconfig.Destination
-			if destination == "" {
-				destination = t.Destination
-			}
-
```
dricross, line 258: This variable is unused

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
```diff
@@ -105,33 +108,30 @@ func (q *queue) AddEventNonBlocking(e logs.LogEvent) {
...
+	q.stopped = true
+}
+
 // start is the main loop for processing events and managing the queue.
 func (q *queue) start() {
 	defer q.wg.Done()
 	mergeChan := make(chan logs.LogEvent)
 
-	// Merge events from both blocking and non-blocking channel
-	go func() {
```
dricross, line 114: I moved this to a named function just to make the pprof output a little nicer. Since this is an unnamed function, the name of the function in the output is a bit obfuscated `start.func1+0xcf`:
```
github.com/aws/amazon-cloudwatch-agent/plugins/outputs/cloudwatchlogs/internal/pusher.(*queue).start.func1+0xcf github.com/aws/amazon-cloudwatch-agent/plugins/outputs/cloudwatchlogs/internal/pusher/queue.go:117
```
Still findable from the line number, just not super obvious

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -106,10 +105,8 @@ func (c *CloudWatchLogs) Close() error {
 }
 
 func (c *CloudWatchLogs) Write(metrics []telegraf.Metric) error {
-	for _, m := range metrics {
-		c.writeMetricAsStructuredLog(m)
-	}
-	return nil
+	// we no longer expect this to be used. We now use the OTel awsemfexporter for sending EMF metrics to CloudWatch Logs
+	return fmt.Errorf("unexpected call to Write")
```
dricross, line 109: To make cwDest thread-safe more easily, I removed this functionality. This function is here to adhere to the Telegraf interface, but it is no longer used. This functionality was for pushing EMF metrics to CloudWatch Logs using this telegraf-based output plugin. However, the agent no longer uses this plugin to push EMF metrics. EMF metrics now go through the OTel awsemfexporter plugin.

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -186,166 +199,107 @@ func (c *CloudWatchLogs) createClient(retryer aws.RequestRetryer) *cloudwatchlog
 	return client
 }
 
-func (c *CloudWatchLogs) writeMetricAsStructuredLog(m telegraf.Metric) {
```
dricross, line 189: Only used by `Write` which is no longer used

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -186,166 +199,107 @@ func (c *CloudWatchLogs) createClient(retryer aws.RequestRetryer) *cloudwatchlog
...
-		logStream = c.LogStreamName
-	}
-
-	return pusher.Target{Group: logGroup, Stream: logStream, Class: util.StandardLogGroupClass, Retention: -1}, nil
+// Description returns a one-sentence description on the Output
+func (c *CloudWatchLogs) Description() string {
+	return "Configuration for AWS CloudWatchLogs output."
 }
 
-func (c *CloudWatchLogs) getLogEventFromMetric(metric telegraf.Metric) *structuredLogEvent {
```
dricross, line 229: Only used by `Write` which is no longer used

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -186,166 +199,107 @@ func (c *CloudWatchLogs) createClient(retryer aws.RequestRetryer) *cloudwatchlog
...
-	return &structuredLogEvent{
-		msg: message,
-		t:   metric.Time(),
-	}
-}
+var sampleConfig = `
+  ## Amazon REGION
+  region = "us-east-1"
 
-type structuredLogEvent struct {
```
dricross, line 304: Only used by `Write` which is no longer used

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -366,44 +318,10 @@ func (cd *cwDest) switchToEMF() {
 	}
 }
 
-// Description returns a one-sentence description on the Output
-func (c *CloudWatchLogs) Description() string {
```
dricross, line 370: I moved these two CloudWatchLogs functions to the CloudWatchLogs block (above cwDest). Didn't make much sense to define these these functions separately from the rest.

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -186,166 +199,107 @@ func (c *CloudWatchLogs) createClient(retryer aws.RequestRetryer) *cloudwatchlog
...
 }
 
+func (cd *cwDest) NotifySourceStopped() {
+	cd.Lock()
+	defer cd.Unlock()
+	cd.refCount--
+	if cd.refCount <= 0 {
+		cd.stop()
+	}
+}
```
duhminick, line 286: refCount being negative is probably unexpected, we could add a debug/warning message here

duhminick:
Also, when possible, can you run the integ tests?

dricross:
> Also, when possible, can you run the integ tests?

Was having trouble with the log tailer unit test on Windows again... Eventually worked after a couple of retries. Added the tag and kicked off integ tests.

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
dricross: Good idea

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
jefchien: If we're removing this function, should we also remove `cwDest.switchToEMF`

logs/logs.go
```diff
@@ -168,14 +166,17 @@ func (l *LogAgent) runSrcToDest(src LogSrc, dest LogDest) {
...
 		}
 		if err != nil {
-			log.Printf("E! [logagent] Failed to publish log to %v, error: %v", l.destNames[dest], err)
+			log.Printf("E! [logagent] Failed to publish log to %v, error: %v", src.Destination(), err)
 			return
 		}
 	}
+
+	// eventsCh has been closed meaning the src has been stopped
+	dest.NotifySourceStopped()
```
jefchien: Should this use a defer after the `defer src.Stop()` at the start of the function? What if the `eventsCh` didn't close, but the `runToDest` function exits? Wouldn't that result in an inconsistent `refCount` and prevent it from ever closing the pusher?

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
```diff
@@ -34,21 +35,23 @@ type queue struct {
...
+	stopCh       chan struct{}
+	stopped      bool
 	lastSentTime atomic.Value
 
 	initNonBlockingChOnce sync.Once
 	startNonBlockCh       chan struct{}
 	wg                    *sync.WaitGroup
 }
 
+var _ (Queue) = (*queue)(nil)
```
jefchien, line 47: nit: Not necessary because the constructor should already enforce this.

plugins/outputs/cloudwatchlogs/internal/pusher/pool.go
```diff
@@ -109,6 +109,10 @@ func (s *senderPool) Send(batch *logEventBatch) {
 	})
 }
 
+func (s *senderPool) Stop() {
+	// workerpool is stopped by the plugin
```
jefchien, line 113: Shouldn't this still stop the underlying sender? I think if this doesn't, then nothing else will.

plugins/outputs/cloudwatchlogs/internal/pusher/queue_test.go
```diff
@@ -86,6 +86,10 @@ func (m *mockSender) RetryDuration() time.Duration {
 	return args.Get(0).(time.Duration)
 }
 
+func (m *mockSender) Stop() {
+
```
jefchien: Think it'd be worth asserting this is called when the pusher is stopped in one of the tests.

plugins/outputs/cloudwatchlogs/internal/pusher/sender.go
```diff
@@ -134,6 +138,14 @@ func (s *sender) Send(batch *logEventBatch) {
 	}
 }
 
+func (s *sender) Stop() {
+	if s.stopped {
+		return
+	}
+	close(s.stopCh)
+	s.stopped = true
+}
```
jefchien, line 147: Not thread-safe, but likely not an issue since the sender should only be stopped from a single thread.

logs/logs.go
dricross: You're right we should use a defer here

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
dricross: That's true, it's superfluous. Though I like this to explicitly declare the intention that this type should implement the Queue interface as close to the type definition as possible.

plugins/outputs/cloudwatchlogs/internal/pusher/pool.go
dricross: Yes, it should

plugins/outputs/cloudwatchlogs/internal/pusher/queue_test.go
dricross: Makes sense

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -186,166 +199,111 @@ func (c *CloudWatchLogs) createClient(retryer aws.RequestRetryer) *cloudwatchlog
...
+func (cd *cwDest) NotifySourceStopped() {
+	cd.Lock()
+	defer cd.Unlock()
+	cd.refCount--
+	if cd.refCount <= 0 {
+		cd.stop()
+	}
+
+	if cd.refCount < 0 {
+		fmt.Printf("E! Negative refCount on cwDest detected. refCount: %d, logGroup: %s, logStream: %s", cd.refCount, cd.pusher.Group, cd.pusher.Stream)
```
duhminick, line 284: Would it be possible to pass in a logger instead of using fmt?

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -134,7 +131,13 @@ func (c *CloudWatchLogs) CreateDest(group, stream string, retention int, logGrou
 
 func (c *CloudWatchLogs) getDest(t pusher.Target, logSrc logs.LogSrc) *cwDest {
 	if cwd, ok := c.cwDests.Load(t); ok {
-		return cwd.(*cwDest)
+		d := cwd.(*cwDest)
+		d.Lock()
+		defer d.Unlock()
```
duhminick, line 136: I'm trying to think if there's a possible situation where the destination does get loaded, but some other routine called stop before we get the lock which might lead to a weird refCount. We should be okay here I think considering the tests you did

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
```diff
@@ -105,33 +108,30 @@ func (q *queue) AddEventNonBlocking(e logs.LogEvent) {
...
-
+	go q.merge(mergeChan)
 	go q.manageFlushTimer()
 
 	for {
 		select {
-		case e := <-mergeChan:
+		case e, ok := <-mergeChan:
+			if !ok {
+				q.send()
```
movence, line 132: should we check if there are events to send?

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
```diff
@@ -105,33 +108,30 @@ func (q *queue) AddEventNonBlocking(e logs.LogEvent) {
 	}
 }
 
+// Stop stops all goroutines associated with this queue instance.
+func (q *queue) Stop() {
+	if q.stopped {
+		return
+	}
+	close(q.stopCh)
+	q.stopped = true
```
movence, line 117: it looks like this is safe since only a single thread will be calling this function?

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
movence: oh I see `send()` function already check for events len

logs/logs.go
```diff
@@ -148,8 +146,10 @@ func (l *LogAgent) Run(ctx context.Context) {
 }
 
 func (l *LogAgent) runSrcToDest(src LogSrc, dest LogDest) {
+
 	eventsCh := make(chan LogEvent)
 	defer src.Stop()
+	defer dest.NotifySourceStopped()
```
jefchien, line 152: Does it matter which order this is in? If the source is stopped after the destination reports it, would that be an issue?

jefchien (approved):

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
jefchien: +1. There's an existing logger in the `CloudWatchLogs` that is preferable if we want to log something.

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
dricross: The `Lock()` on line 135 should prevent that situation. If some other routine calls stop on the destination after it was loaded but before the lock is obtained, then this routine should see that reflected after it obtains the lock and checks the `stopped` field. In this case, it'll see that the destination is already stopped and will make a new one instead by continuing through the function.

logs/logs.go
dricross: The order doesn't matter. The two objects are only linked through this function. Maybe the naming isn't optimal here which might cause some confusion. The `NotifySourceStopped` method on `dest` doesn't notify the source, it notifies `dest` that the source is stopped.

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
dricross: `cwDest.switchToEMF` could still be invoked if the agent detects EMF metrics within a log file so I don't think we want to remove that functionality.

PR #1850, opened 2025-09-05
Cluster name relabel conditional

agarakan:
# Description of the issue
When a customer defines clusterName in their provided prometheus config, we should prefer that value and not overwrite it

See original logic: https://github.com/aws/amazon-cloudwatch-agent/blob/75b8144b453928aeced75e39e973d4816d9d9496/plugins/inputs/prometheus/metrics_handler.go#L68-L75 

# Description of changes
This change introduces a check for any customer relabel configs targeting "ClusterName" and prefers that. Otherwise it injects a default ClusterName based on the ecs metadata.

# License

# Tests
Added unit tests, verified all tests pass
E2E testing to verify customer-provided ClusterName takes priority

```
make lint
make fmt
make fmt-sh
make test
```

# Requirements
2. Run `make lint`

-------
### Integration Tests

### DIFF
--- a/translator/tocwconfig/sampleConfig/prometheus_ecs_sd_config_linux.yaml
+++ b/translator/tocwconfig/sampleConfig/prometheus_ecs_sd_config_linux.yaml
@@ -227,9 +227,8 @@ receivers:
                   metrics_path: /metrics
                   relabel_configs:
                     - action: replace
-                      regex: (.*)
-                      source_labels:
-                        - __meta_ecs_cluster_name
+                      regex:
+                      replacement: test-ecs-cluster-0123456789
                       target_label: ClusterName
                     - action: replace
                       regex: (.*)
@@ -330,9 +329,8 @@ receivers:
                   metrics_path: /metrics
                   relabel_configs:
                     - action: replace
-                      regex: (.*)
-                      source_labels:
-                        - __meta_ecs_cluster_name
+                      regex:
+                      replacement: test-ecs-cluster-0123456789
                       target_label: ClusterName
                     - action: replace
                       regex: (.*)
@@ -433,9 +431,8 @@ receivers:
                   metrics_path: /metrics
                   relabel_configs:
                     - action: replace
-                      regex: (.*)
-                      source_labels:
-                        - __meta_ecs_cluster_name
+                      regex: 
+                      replacement: test-ecs-cluster-0123456789
                       target_label: ClusterName
                     - action: replace
                       regex: (.*)
--- a/translator/tocwconfig/sampleConfig/prometheus_ecs_sd_config_linux.yaml
+++ b/translator/tocwconfig/sampleConfig/prometheus_ecs_sd_config_linux.yaml
@@ -227,10 +227,6 @@ receivers:
                   metrics_path: /metrics
                   relabel_configs:
                     - action: replace
-                      regex:
-                      replacement: test-ecs-cluster-0123456789
-                      target_label: ClusterName
-                    - action: replace
                       regex: (.*)
                       source_labels:
                         - __meta_ecs_cluster_name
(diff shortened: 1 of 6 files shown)


translator/translate/otel/receiver/prometheus/translator.go
```diff
@@ -206,11 +205,36 @@ func addDefaultECSRelabelConfigs(scrapeConfigs []*config.ScrapeConfig, conf *con
...
 			}
 		}
 	}
 }
+
+/*
+Adds ClusterName relabelConfig if not already set by customer.
+Customer can specify the clusterName in the scraping job's relabel_config
+CWAgent won't overwrite in this case to support cross-cluster monitoring
+*/
```
dricross: nit: go doc convention is:
```
// <function name> adds ClusterName ...
```

translator/tocwconfig/sampleConfig/prometheus_ecs_sd_config_linux.yaml
```diff
@@ -291,6 +286,10 @@ receivers:
                       source_labels:
                         - __meta_ecs_container_labels_app_x
                       target_label: app_x
+                    - action: replace
+                      regex:
```
dricross, line 290: hmm this looks invalid? should it just be omitted or have an actual value maybe?

translator/translate/otel/receiver/prometheus/translator.go
agarakan: Thanks! I'll update

translator/translate/otel/receiver/prometheus/translator.go
```diff
@@ -206,11 +205,36 @@ func addDefaultECSRelabelConfigs(scrapeConfigs []*config.ScrapeConfig, conf *con
...
+func addClusterNameRelabelConfig(scrapeConfig *config.ScrapeConfig) {
+	for _, relabelConfig := range scrapeConfig.RelabelConfigs {
+		// If ClusterName relabel config is defined by customer, don't override
+		if relabelConfig.TargetLabel == "ClusterName" {
+			return
+		}
+	}
+
+	clusterName := ecsutil.GetECSUtilSingleton().Cluster
```
sky333999, line 233: This is not right. The current logic first checks if an override via the agent config is provided using the `cluster_name` field under the prometheus section.
So we should retain the same logic.

translator/tocwconfig/sampleConfig/prometheus_ecs_sd_config_linux.yaml
agarakan: Its actually valid and is the generated value in the actual result. I'll see if tests pass by omitting it. Else its likely this is just the actual value it needs to be

translator/translate/otel/receiver/prometheus/translator.go
agarakan: Underlying the prometheus section, the cluster_name is populated from the ECS Metadata when the agent is detected to be running on ECS (for example, this specific scenario). I need to check if the prometheus cluster_name is initialized by the time this part of the translation is triggered.

translator/tocwconfig/sampleConfig/prometheus_ecs_sd_config_linux.yaml
agarakan: Yeah verified, thats the valid expectation for adding the clusterName relabel config. No change planned here

translator/translate/otel/receiver/prometheus/translator.go
agarakan: Connected offline, we need to read in clusterName from the CWA json like here: https://github.com/aws/amazon-cloudwatch-agent/blob/main/translator/translate/logs/metrics_collected/prometheus/ruleClusterName.go

PR #1855, opened 2025-09-09
Drop Docker Job label when JobName configured for ecs_sd

agarakan:
# Description of the issue
When a customer configures a JobName or a JobNameLabel, we should use that as the "job" label in the collected metrics. Currently a docker label of "job" will overwrite the configured JobName as its preserved by the labelmap relabel config. We should drop the docker__label of "job" if a JobName or JobNameLabel is explicitly configured. The final "job" label is appended by Prometheus Receiver

# Description of changes
Drop dockerlabel job if JobName or custom JobNameLabel is configured in ecs_service_discovery config.

# License

# Tests
Unit tests passing, and added specific scenarios

```
make fmt
make fmt-sh
make lint
make test
```

Integration tests also pass using sample config, verified custom JobName is captured in the metric (previously not)

<Pending add screenshot of integ test result>

# Requirements
2. Run `make lint`

-------
### Integration Tests

### DIFF
--- a/translator/tocwconfig/sampleConfig/prometheus_ecs_sd_config_linux.yaml
+++ b/translator/tocwconfig/sampleConfig/prometheus_ecs_sd_config_linux.yaml
@@ -286,12 +286,10 @@ receivers:
                       source_labels:
                         - __meta_ecs_source
                       target_label: TaskId
-                    - action: replace
-                      regex: (.*)
-                      replacement: ""
+                    - action: drop
+                      regex: .*
                       source_labels:
                         - __meta_ecs_container_labels_job
-                      target_label: __meta_ecs_container_labels_job
                     - action: labelmap
                       regex: ^__meta_ecs_container_labels_(.+)$
                       replacement: __capture_group_1__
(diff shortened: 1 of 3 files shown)


dricross:
Using the configured JobName (if provided) over the docker label makes sense. When JobNameLabel is configured, though, what will happen if that label is empty? What happens with the service discovery extension today?

My naive expectation would be to fallback to the docker label "job" if both JobName isn't provided and a label with a key matching JobNameLabel doesn't exist (and/or perhaps if the value is empty). But what it should do really only depends on how ecs service discovery works today.

agarakan:
> Using the configured JobName (if provided) over the docker label makes sense. When JobNameLabel is configured, though, what will happen if that label is empty? What happens with the service discovery extension today?
> 
> My naive expectation would be to fallback to the docker label "job" if both JobName isn't provided and a label with a key matching JobNameLabel doesn't exist (and/or perhaps if the value is empty). But what it should do really only depends on how ecs service discovery works today.

Good call out, it seems like theres some funny replacement happening in the Prometheus receiver when the JobNameLabel is empty or invalid, and the behavior differs based on if the Docker label for "job" exists, is valid, invalid or blank too. Tracking a separate item for it in the backlog

PR #1859, opened 2025-09-12
[Fix] Added node_(gpu|efa|neuroncore)_request metric

aadam19:
This PR continues the work of my fellow noble and courageous intern @tudor-manea.
We were scraping already for the `node_(gpu|efa|neuroncore)_request` but somehow forgot to make the agent pick it up.
Link to IntegTests PR that made us realize: https://github.com/aws/amazon-cloudwatch-agent-test/pull/569

### DIFF
--- a/translator/tocwconfig/sampleConfig/emf_and_kubernetes_with_gpu_config.yaml
+++ b/translator/tocwconfig/sampleConfig/emf_and_kubernetes_with_gpu_config.yaml
@@ -190,16 +190,19 @@ exporters:
                 - node_status_condition_unknown
                 - node_status_capacity_pods
                 - node_status_allocatable_pods
+                - node_gpu_request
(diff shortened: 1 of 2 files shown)


translator/tocwconfig/sampleConfig/emf_and_kubernetes_with_gpu_config.yaml
```diff
@@ -190,16 +190,19 @@ exporters:
                 - node_status_condition_unknown
                 - node_status_capacity_pods
                 - node_status_allocatable_pods
+                - node_gpu_request
```
sky333999, line 193: Are we sure we want this? I think this was left out in the past since request & limit are always the same for GPUs.

sky333999 (approved):

PR #1864, opened 2025-09-15
Cloudwatch Agent IPv6 support

Paramadon:
# Description of the issue
Currently the CloudWatch Agent does not support IPv6 environments. This means customer using IPv6 only environments will not be able to use the CWA to get telemetry. 

# Description of changes
This pr allows customers to be able to run the CloudWatch Agent in an IPv6 only environment. Customer simply needs to add `use_dualstack_endpoint=true` in the agent configuration and what this will do is it will have all the agent endpoint calls use aws dualstack endpoints. A dual stack endpoint will work in either IPv4 or IPv6 so enabling dual stack will allow all aws sdk calls to have dual stack endpoints.

### How are we enabling dualstack for all sdk calls?
If the custom sets the `use_dualstack_endpoint=true` in the agent section, the cloudwatch agent will set the `AWS_USE_DUALSTACK_ENDPOINT` to true this environment variable will automatically have all sdk used dual stack endpoints if set (as pointed out in this [aws doc](https://docs.aws.amazon.com/sdkref/latest/guide/feature-endpoints.html)). This is being done in the `toEnvConfig.go` file where we see if `use_dualstack_endpoint` was specified to true and if it was, we'll set the env variable while agent is starting up (following current behavior of setting env variables at agent start up)

There are some extra changes needed to support the CloudWatch Agent fully in IPv6 and those are 1. AMP custom endpoint changes and config-downloader changes.

For the AMP custom endpoint changes, we added a new field in Global_Config of the agent which is currently being used in amp translator. We parse the `use_dualstack_endpoint` in the apply rule and then set the `Global_Config` so that the agent knows that the customer set dual stack to true in amp translation. In the amp translator we check `Global_Config.UseDualStackEndpoint` if set to true, we'll update the amp endpoint to be dualstack endpoint.

##### IPv4 AMP endpoint
`https://aps-workspaces.us-east-1.amazonaws.com/workspaces/ws-12345678-1234-1234-1234-123456789012/api/v1/remote_write`

##### Dualstack AMP endpoint
`https://aps-workspaces.us-east-1.api.aws/workspaces/ws-12345678-1234-1234-1234-123456789012/api/v1/remote_write`

the only difference between the endpoint is `amazonaws.com` is changed to `api.aws`. Which is exactly what we do in the translator to make amp work in an IPv6 only environment

### How about the config downloader IPv6 support?

For the config downloader the agent configuration isn't available as customer uses this to fetch their configuration from ssm parameter store. In this case customer has two options 1. Set the env variable `AWS_USE_DUALSTACK_ENDPOINT` to true and run the config downloader. Or they can pass in `-d` option flag to their `fetch-config`/`append-config` call and the agent will set the env variable for them. This matches the current pattern used in the agent of using short options like for example -s, -h, etc.

### How about cloudwatch agent `use_dualstack_endpoint` vs env var precedence

<html><head></head><body>
<hr>
<h3>CloudWatch Agent Dual-Stack Behavior Matrix (IPv6-only Environment)</h3>

Scenario | AWS_USE_DUALSTACK_ENDPOINT | use_dualstack_endpoint | Metrics Delivered | Notes
--  | -- | -- | -- | --
Default | Not set | Not set | No | Fails in IPv6-only environment
Env only | false | Not set | No | Environment variable disables dual-stack
Env only | true | Not set | Yes | Environment variable enables dual-stack
Agent config | Not set | true | Yes | Agent configuration enables dual-stack
Agent config | false | true | Yes | Agent configuration takes precedence
Agent config | true | false | No | Agent configuration takes precedence (disables dual-stack)
Invalid env | 0 or 1 | Not set | No | Invalid environment variable causes error

</body></html>

# Tests

All testing was done on an IPv6 only environment (IPv6 addresses only no IPv4 address) verifying that we are able to get agent metrics and also verified amp and config-dowloader:

#### Agent json used
```
Agent json used for test:

{
  "agent": {
    "metrics_collection_interval": 15,
    "run_as_user": "root",
    "debug": true,
    "region": "us-west-2",
    "use_dualstack_endpoint": true
  },
  "metrics": {
    "metrics_destinations": {
      "amp": {
        "workspace_id": "ws-xxx"
      }
    },
    "metrics_collected": {
      "cpu": {
        "measurement": [
          {"name": "time_active", "rename":"cpu_time_active_renamed"},
          "time_guest", "time_guest_nice", "time_idle", "time_iowait", "time_irq",
          "time_nice", "time_softirq", "time_steal", "time_system", "time_user",
          "usage_active", "usage_guest", "usage_guest_nice", "usage_idle", "usage_iowait", "usage_irq",
          "usage_nice", "usage_softirq", "usage_steal", "usage_system", "usage_user"
        ],
        "metrics_collection_interval": 10
      }
    }
  }
}

```

#### Successfully querying amp metrics

```
{
  "status": "success",
  "data": {
    "resultType": "vector",
    "result": [
      {
        "metric": {
          "__name__": "cpu_usage_system",
          "aws_StorageResolution": "true",
          "cpu": "cpu-total",
          "host": "i-xxx"
        },
        "value": [
          1757692351,
          "0"
        ]
      }
    ]
  }
}
```

#### CLI Command used:
`sudo /opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent-ctl -d -a fetch-config -m ec2 -c ssm:/config/standard/linux/ec2 -s`
#### CWA json in ssm parameter store
```
{
  "agent": {
    "metrics_collection_interval": 15,
    "run_as_user": "root",
    "debug": true,
    "region": "us-west-2",
    "use_dualstack_endpoint": true
  },
  "metrics": {
    "namespace": "testingIPv6-9/12-config-downloader",
    "metrics_collected": {
      "cpu": {
        "measurement": [
          {"name": "time_active", "rename":"cpu_time_active_renamed"},
          "time_guest", "time_guest_nice", "time_idle", "time_iowait", "time_irq",
          "time_nice", "time_softirq", "time_steal", "time_system", "time_user",
          "usage_active", "usage_guest", "usage_guest_nice", "usage_idle", "usage_iowait", "usage_irq",
          "usage_nice", "usage_softirq", "usage_steal", "usage_system", "usage_user"
        ],
        "metrics_collection_interval": 10
      }
    }
  }
}
```

#### Successfully getting metrics given in the namespace of the agent json stored in the ssm parameter store
Screenshot 2025-09-15 at 5 21 11 PM

### Tested config downloader dual stack changes on Windows 

Agent json used: 
```
{
    "agent": {
      "debug": true,
      "region": "us-west-2",
      "use_dualstack_endpoint": true

    },
    "metrics": {
      "namespace": "Windows-config-downloader-test",
      "metrics_collected": {
        "Network Interface": {
          "resources": [
            "*"
          ],
          "measurement": [
            {
              "name": "Bytes Sent/sec",
              "rename": "Bytes_Sent_Per_Sec"
            },
            {
              "name": "Packets Sent/sec",
              "rename": "Packets_Sent_Per_Sec"
            }
          ],
          "metrics_collection_interval": 1
        },
        "procstat": [
          {
            "exe": "amazon-cloudwatch-agent.exe",
            "measurement": [
              "cpu_usage",
              "memory_rss",
              "memory_vms"
            ],
            "metrics_collection_interval": 1
          }
        ]
      },
      "append_dimensions": {
      },
      "force_flush_interval": 10
    },
	"logs": {
		"logs_collected": {
		  "files": {
			"collect_list": [
			  {
				"file_path": "/tmp/test1.log",
				"log_group_name": "{instance_id}",
				"log_stream_name": "{instance_id}/tmp1",
				"timezone": "UTC"
			  }
			]
		  }
		},
        "force_flush_interval": 60
	}
  }
```

#### Calling control script with -d dual stack option
Screenshot 2025-09-16 at 1 48 27 PM

#### use_dual_stack_endpoint set in toml
Screenshot 2025-09-16 at 1 51 54 PM

#### Getting metrics 
Screenshot 2025-09-16 at 1 53 43 PM

#### Tested Mac config downloader 
Tested by checking that mac control script properly config downloader with `--dualstack` flag.


packaging/dependencies/amazon-cloudwatch-agent-ctl
```diff
@@ -67,6 +70,9 @@ UsageString="
         -s: optionally restart after configuring the agent configuration
             this parameter is used for 'fetch-config', 'append-config', 'remove-config' action only.
 
+        -d: enable dual-stack endpoints for AWS API calls during config download
```
Paramadon, line 73: Adding `-d` to keep option short to add call config downloader with dual stack to match current options in the control script of the agent. Also this is just an extra option being added for customers, they can also simply set AWS_USE_DUALSTACK_ENDPOINT env variable to true which would do the same as passing in -d option. This just gives customers an extra option to enable have config-dowloader work on an ipv6 environment.

dricross:
We also need to update and test the control scripts for Mac (under `packaging/darwin`)  and Windows (under `packaging/windows`)

dricross:
As I understand it, users *could* get IPv6 support today by setting `AWS_USE_DUALSTACK_ENDPOINT` environment variable themselves. Thinking about some use cases here to understand preference between environment variable and agent config.

Just to confirm, what happens in each of these cases:
1. Customer does not set `AWS_USE_DUALSTACK_ENDPOINT` and does not set `use_default_endpoint`?
2. Customer does not set `AWS_USE_DUALSTACK_ENDPOINT` and does set `use_default_endpoint`?
  a. What happens when `use_default_endpoint` is false?
  b. What happens when `use_default_endpoint` is true?
4. Customer does set `AWS_USE_DUALSTACK_ENDPOINT` and does not set `use_default_endpoint`?
  a. What happens when `AWS_USE_DUALSTACK_ENDPOINT` is false?
  b. What happens when `AWS_USE_DUALSTACK_ENDPOINT` is true?
  c. What happens when `AWS_USE_DUALSTACK_ENDPOINT` is some other value (e.g. 0/1)?
6. Customer does set `AWS_USE_DUALSTACK_ENDPOINT` and does set `use_default_endpoint`?
  a. What happens if they differ? Which takes precedence?

A table might be the best way to enumerate these

packaging/dependencies/amazon-cloudwatch-agent-ctl
dricross: I like that you kept it short. Maintains convention and consistency.

translator/translate/agent/use_dualstack_endpoint.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+func (r *UseDualStackEndpoint) ApplyRule(input interface{}) (returnKey string, returnVal interface{}) {
+	agentMap, ok := input.(map[string]interface{})
+	if !ok {
+		returnKey, returnVal = "", translator.ErrorMessages
+		return
+	}
+
+	dualStackValue, exists := agentMap["use_dualstack_endpoint"]
```
dricross: Should `use_dualstack_endpoint` be defined in `cfg/envconfig` package instead?

tool/downloader/flags/flags.go
```diff
@@ -13,4 +13,5 @@ var DownloaderFlags = map[string]cmdwrapper.Flag{
 	"output-dir":      {DefaultValue: "", Description: "Path of output json config directory."},
 	"config":          {DefaultValue: "", Description: "Please provide the common-config file"},
 	"multi-config":    {DefaultValue: "default", Description: "valid values: default, append, remove"},
+	"d":               {DefaultValue: "false", Description: "Use dual-stack endpoints for AWS API calls for config-downloader", IsBool: true},
```
dricross: In this case, since the other flags are more descriptive, we should probably stay consistent and use something more descriptive like `dualstack`.

translator/tocwconfig/toenvconfig/toEnvConfig.go
```diff
@@ -41,6 +42,15 @@ func ToEnvConfig(jsonConfigValue map[string]interface{}) []byte {
...
 		}
+
+		if useDualStack, ok := agentMap[useDualStackEndpointKey].(bool); ok {
+			if useDualStack {
+				envVars["AWS_USE_DUALSTACK_ENDPOINT"] = "true"
+			} else {
+				envVars["AWS_USE_DUALSTACK_ENDPOINT"] = "false"
+			}
+		} else {
+		}
```
dricross: nit: useless `else`. I'm surprised the linter didn't complain about this

translator/translate/agent/use_dualstack_endpoint.go
dricross: Actually, `cfg/envconfig` is for environment variables not agent config fields. Feels like this config field should be defined as a constant somewhere though.

translator/tocwconfig/toenvconfig/toEnvConfig.go
```diff
@@ -41,6 +42,15 @@ func ToEnvConfig(jsonConfigValue map[string]interface{}) []byte {
 		if usageData, ok := agentMap[usageDataKey].(bool); ok && !usageData {
 			envVars[envconfig.CWAGENT_USAGE_DATA] = "FALSE"
 		}
+
+		if useDualStack, ok := agentMap[useDualStackEndpointKey].(bool); ok {
+			if useDualStack {
+				envVars["AWS_USE_DUALSTACK_ENDPOINT"] = "true"
+			} else {
```
dricross, line 48: `AWS_USE_DUALSTACK_ENDPOINT` should probably be defined in `envconfig` like the others

packaging/windows/amazon-cloudwatch-agent-ctl.ps1
```diff
@@ -14,6 +14,8 @@ Param (
     [string]$Mode = 'ec2',
     [Parameter(Mandatory = $false)]
     [string]$LogLevel = '',
+    [Parameter(Mandatory = $false)]
+    [switch]$d = $false,
```
dricross, line 18: nit: for consistency within this file, `d` should probably be `DualStack`

tool/downloader/downloader_test.go
```diff
@@ -0,0 +1,54 @@
...
+				"output-dir":      stringPtr("/tmp/test"),
+				"config":          stringPtr(""),
+				"multi-config":    stringPtr("default"),
+				"dualstack":       stringPtr(tt.dualStackValue),
+			}
+
+			result := *flags["dualstack"] == "true"
+			if result != tt.expectedResult {
+				t.Errorf("Expected dualstack flag conversion to be %v, got %v", tt.expectedResult, result)
+			}
```
dricross: This seems like a useless test. It's setting values in a map and then reading the values from the map. That doesn't really test any of your code. You should also stick to `assert`/`require` since that's our convention.

packaging/darwin/amazon-cloudwatch-agent-ctl
```diff
@@ -69,6 +71,9 @@ UsageString="
         -s: optionally restart after configuring the agent configuration
             this parameter is used for 'fetch-config', 'append-config', 'remove-config' action only.
 
+        -d: enable dual-stack endpoints for AWS API calls during config download
+            this parameter is used for 'fetch-config', 'append-config', 'remove-config' actions only.
```
movence: is this true? I don't think `remove` option would require dualstack?

packaging/dependencies/amazon-cloudwatch-agent-ctl
```diff
@@ -354,6 +366,7 @@ main() {
      cwa_config_location=''
      restart='false'
      mode='ec2'
+     use_dualstack='false'
```
movence, line 369: please remind me what is the concern of enabling this by default? Shouldn't it alway fallback to ipv4 in case ipv6 is not available?

translator/tocwconfig/sampleConfig/dualstack_config.json
```diff
@@ -0,0 +1,65 @@
+{
```
movence, line 1: I think you can trim down this json config when the focus is really just AMP endpoint. We don't really care about extra stuffs here.

translator/translate/agent/use_dualstack_endpoint_test.go
```diff
@@ -0,0 +1,92 @@
...
+		},
+		"DisableDualStack": {
+			input: map[string]interface{}{
+				UseDualStackEndpointKey: false,
+			},
+			expectedKey:    UseDualStackEndpointKey,
+			expectedValue:  false,
+			expectedGlobal: false,
+		},
+		"MissingField": {
```
movence: what are we validating here?

Paramadon: If UseDualStackEndpointKey is not set in the map, we can remove this from the unit test as it might not be necessary

packaging/dependencies/amazon-cloudwatch-agent-ctl
Paramadon: Discussed offline but there exist a case for IPv6 only with a NAT gateway that might cause regions that do not support IPv6 in some sdk clients like China that could break the agent. So as of now we are opting for an opt in experience. It is similar to what AWS CLI is also doing as well as the AWS SSM agents team so following consistent pattern.

packaging/darwin/amazon-cloudwatch-agent-ctl
Paramadon: You can use the remove-config for removing ssm parameter example like:
`amazon-cloudwatch-agent-ctl -a remove-config -c ssm:/config/standard/linux/ec2` this would require dual stack for an ipv6 only environment.

movence (approved):

PR #1875, opened 2025-09-26
Retrieve ECS clusterName

agarakan:
# Description of the issue
When running on ECS, the clusterName should be retrieved from the CWA config or dynamically

# Description of changes
1. Adds back pre-maturely removed prometheus config parser for the logs metrics_collected section

# License

# Tests
Validated end to end
```
make lint
make fmt
make fmt-sh
make test
```

# Requirements
2. Run `make lint`

-------
### Integration Tests


TravisStark (approved):

translator/translate/otel/receiver/prometheus/translator.go
```diff
@@ -172,6 +172,30 @@ func escapeStrings(node any) {
 	}
 }
 
+func getClusterNameFromConfig(conf *confmap.Conf, promConfigKey string) string {
```
Paramadon, line 175: Would prefer to have a unit test for name resolution

translator/translate/otel/receiver/prometheus/translator.go
```diff
@@ -187,7 +211,7 @@ func addDefaultECSRelabelConfigs(scrapeConfigs []*config.ScrapeConfig, conf *con
 	}
 
 	defaultRelabelConfigs := []*relabel.Config{
-		{SourceLabels: model.LabelNames{"__meta_ecs_cluster_name"}, Action: relabel.Replace, TargetLabel: "ClusterName", Regex: relabel.MustNewRegexp("(.*)")},
+		{Replacement: getClusterNameFromConfig(conf, promConfigKey), Action: relabel.Replace, TargetLabel: "ClusterName"},
```
Paramadon, line 214: What if the cluster name is empty should we still set it? Would this cause any issues?

Paramadon: Also could there be any code that previously relied on the old regex-based relabeling and with this it would behave differently. We are fine with this behavior change I'm assuming?

translator/translate/otel/receiver/prometheus/translator.go
agarakan: Itll be part of the next CR

translator/translate/otel/receiver/prometheus/translator.go
agarakan: When blank the relabel config doesnt show up as a dimension on the Metric which is expected

Nothing relied on the regex as that was a placeholder I added before this PR

translator/translate/otel/common/common.go
```diff
@@ -491,7 +492,7 @@ func KueueContainerInsightsEnabled(conf *confmap.Conf) bool {
 }
 
 func GetClusterName(conf *confmap.Conf) string {
-	val, ok := GetString(conf, ConfigKey(LogsKey, MetricsCollectedKey, KubernetesKey, "cluster_name"))
+	val, ok := GetString(conf, ConfigKey(LogsKey, MetricsCollectedKey, KubernetesKey, ClusterNameKey))
```
Paramadon, line 495: So the order of precedence for resolving the cluster name is config > environment variable > ECS metadata > EC2 tagger > ECS fallback? Just want to make sure this is the expected behavior, also a comment for order of precedence here for maintainability purposes.

agarakan: Yes

agarakan: See original order in the old code: https://github.com/aws/amazon-cloudwatch-agent/blob/main/translator/translate/logs/metrics_collected/prometheus/ruleClusterName.go

PR #1876, opened 2025-09-26
Add support for EC2 metadata or tags as dimensions for host metrics

lisguo:
# Description of the issue
Currently if you tried to use `append_dimensions` within a host metrics plugin, it will append the raw string. If you tried to use the `${aws:` formatting to substitute the dimension with EC2 metadata information or tags, it will not work. This is only supported in the global `append_dimensions` section.

```
{
  "metrics": {
    "append_dimensions": {
    },
    "metrics_collected": {
      "cpu": {
        "append_dimensions": {
          "InstanceType": "${aws:InstanceType}",
          "ImageId": "${aws:ImageId}",
          "ServiceName": "MyServiceApplication"
        },
        "measurement": [
          "cpu_usage_idle"
        ]
      }
    }
  }
}
```

The above will create the following dimensions today:

| Key | Value |
|-----|-------|
| InstanceId | i-00000001 |
| InstanceType | ${aws:InstanceType} |
| ImageId | ${aws:ImageId} |
| AutoScalingGroupName | ${aws:AutoScalingGroupName} |
| ServiceName | MyServiceApplication |

# Description of changes
The host plugins are telegraf-based and use TOML. We can query EC2 metadata service and EC2 tags at translation time to populate the correct value into the TOML file.

From the above example:
```
[[inputs.cpu]]
    fieldpass = ["usage_idle"]
    percpu = false
    totalcpu = true
    [inputs.cpu.tags]
      AutoScalingGroupName = "production-web-asg"
      ImageId = "ami-0abcdef1234567890"
      InstanceType = "t3.medium"
      ServiceName = "MyServiceApplication"
```

> Note: This means the global append_dimensions section will use ec2tagger processor which queries the values at runtime, while the inner append_dimensions section for host metrics will query the values at translation time. This can cause a conflict if tags are changed during runtime.

# License

# Tests
Manual testing:
For the following cwagent config
```
{
    "metrics": {
        "append_dimensions": {
        },
        "metrics_collected": {
            "cpu": {
                "append_dimensions": {
                    "InstanceType": "${aws:InstanceType}",
                    "ImageId": "${aws:ImageId}",
                    "AutoScalingGroupName": "${aws:AutoScalingGroupName}"
                },
                "measurement": [
                    "cpu_usage_idle",
                    "cpu_usage_iowait",
                    "cpu_usage_user",
                    "cpu_usage_system"
                ],
                "totalcpu": true
            },
            "disk": {
                "append_dimensions": {
                    "InstanceType": "${aws:InstanceType}",
                    "ImageId": "${aws:ImageId}",
                    "AutoScalingGroupName": "${aws:AutoScalingGroupName}"
                },
                "measurement": [
                    "used_percent",
                    "inodes_free"
                ],
                "resources": [
                    "*"
                ],
                "dimensions": [
                    [
                        "device",
                        "fstype",
                        "path"
                    ]
                ]
            },
            "diskio": {
                "append_dimensions": {
                    "InstanceType": "${aws:InstanceType}",
                    "ImageId": "${aws:ImageId}",
                    "AutoScalingGroupName": "${aws:AutoScalingGroupName}"
                },
                "measurement": [
                    "io_time"
                ],
                "resources": [
                    "*"
                ]
            },
            "mem": {
                "append_dimensions": {
                    "InstanceType": "${aws:InstanceType}",
                    "ImageId": "${aws:ImageId}",
                    "AutoScalingGroupName": "${aws:AutoScalingGroupName}"
                },
                "measurement": [
                    "used_percent"
                ]
            },
            "swap": {
                "append_dimensions": {
                    "InstanceType": "${aws:InstanceType}",
                    "ImageId": "${aws:ImageId}",
                    "AutoScalingGroupName": "${aws:AutoScalingGroupName}"
                },
                "measurement": [
                    "used_percent"
                ]
            }
        }
    }
}
```

Toml:
```
[inputs]

  [[inputs.cpu]]
    fieldpass = ["usage_idle", "usage_iowait", "usage_user", "usage_system"]
    percpu = false
    totalcpu = true
    [inputs.cpu.tags]
      ImageId = "ami-0254b2d5c4c472488"
      InstanceType = "t3.medium"

  [[inputs.disk]]
    fieldpass = ["used_percent", "inodes_free"]
    tagexclude = ["mode"]
    [inputs.disk.tags]
      ImageId = "ami-0254b2d5c4c472488"
      InstanceType = "t3.medium"

  [[inputs.diskio]]
    fieldpass = ["io_time"]
    [inputs.diskio.tags]
      ImageId = "ami-0254b2d5c4c472488"
      InstanceType = "t3.medium"

  [[inputs.mem]]
    fieldpass = ["used_percent"]
    [inputs.mem.tags]
      ImageId = "ami-0254b2d5c4c472488"
      InstanceType = "t3.medium"

  [[inputs.swap]]
    fieldpass = ["used_percent"]
    [inputs.swap.tags]
      ImageId = "ami-0254b2d5c4c472488"
      InstanceType = "t3.medium"
```

EC2 Tagger Yaml Config:
```
    ec2tagger:
        ec2_metadata_tags:
            - InstanceId
        imds_retries: 1
        middleware: agenthealth/statuscode
        refresh_tags_interval: 0s
        refresh_volumes_interval: 0s
```

# Requirements
2. Run `make lint`

-------
### Integration Tests


translator/translate/util/tagsutil.go
```diff
@@ -20,13 +37,107 @@ func AddHighResolutionTag(tags interface{}) {
...
+	if len(result.Tags) > 0 && result.Tags[0].Value != nil {
+		return *result.Tags[0].Value
+	}
+
+	log.Printf("EC2 tag '%s' not found on instance %s", tagKey, instanceID)
+	return ""
+}
+
+// getEC2Metadata retrieves EC2 metadata using the same approach as ec2tagger
+var getEC2Metadata = func() (ec2metadata.EC2InstanceIdentityDocument, error) {
```
dricross: Should we memoize the metadata to reduce API calls? Instance metadata shouldn't change. We could wrap this in a once.Do and set some singleton.

translator/translate/util/tagsutil.go
```diff
@@ -20,13 +37,107 @@ func AddHighResolutionTag(tags interface{}) {
...
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	mdCredentialConfig := &configaws.CredentialConfig{}
+	metadataProvider := ec2metadataprovider.NewMetadataProvider(mdCredentialConfig.Credentials(), 3)
```
dricross: Regarding the value 3 here, we have similar code in the ec2tagger and the entity store extension, all which seem to use a different number of retries. Should we consolidate retryer logic somewhere?

dricross (approved):

plugins/processors/ec2tagger/ec2tagger.go
```diff
@@ -140,27 +140,36 @@ func (t *Tagger) updateOtelAttributes(attributes []pcommon.Map) {
...
+			}
 		}
 		if t.volumeSerialCache != nil {
 			if devName, found := attr.Get(t.DiskDeviceTagKey); found {
 				serial := t.volumeSerialCache.Serial(devName.Str())
 				if serial != "" {
-					attr.PutStr(AttributeVolumeId, serial)
+					if _, exists := attr.Get(AttributeVolumeId); !exists {
+						attr.PutStr(AttributeVolumeId, serial)
+					}
```
jefchien: nit: This could be moved up so we only try to get the serial from the cache or check the device attribute if it isn't set in the attributes.

translator/translate/util/tagsutil.go
```diff
@@ -4,13 +4,30 @@
...
 )
 
 const (
 	High_Resolution_Tag_Key      = "aws:StorageResolution"
 	Aggregation_Interval_Tag_Key = "aws:AggregationInterval"
+
+	// EC2 metadata keys (matching ec2tagger constants)
+	mdKeyInstanceID   = "InstanceId"
+	mdKeyImageID      = "ImageId"
+	mdKeyInstanceType = "InstanceType"
```
jefchien: Would rather not redefine this. Can we export the ec2tagger constants? We're already importing it.

translator/translate/util/tagsutil.go
```diff
@@ -20,13 +37,107 @@ func AddHighResolutionTag(tags interface{}) {
 	tagMap[High_Resolution_Tag_Key] = "true"
 }
 
-// FilterReservedKeys out reserved tag keys
+// FilterReservedKeys out reserved tag keys and resolves AWS metadata variables at translation time
 func FilterReservedKeys(input any) any {
```
jefchien, line 32: I don't think this is the right place for this. Have we considered putting it in `translator/translate/util/placeholderUtil` and using the `ec2tagger.SupportedAppendDimensions` that the `ec2taggerprocessor` translator uses?

translator/translate/util/tagsutil.go
```diff
@@ -20,13 +37,107 @@ func AddHighResolutionTag(tags interface{}) {
...
+			} else {
+				result[k] = v
+			}
 		}
 	}
 	return result
 }
+
+// resolveAWSMetadata resolves AWS metadata variables like ${aws:TagKey} to actual values from EC2 tags
+func resolveAWSMetadata(variable string) string {
```
jefchien: This seems like logic that already exists in the `ec2utils`.

translator/translate/util/tagsutil.go
lisguo: yeah, we can actually the EC2UtilSingleton in the ec2util package to include the instance type and image id. Good call

translator/translate/util/tagsutil.go
lisguo: this is unused. It was for mocking. I just removed it to avoid confusion

translator/translate/util/tagsutil.go
```diff
@@ -20,13 +28,82 @@ func AddHighResolutionTag(tags interface{}) {
...
+			},
+			{
+				Name:   aws.String("key"),
+				Values: []*string{aws.String(tagKey)},
+			},
+		},
+	}
+
+	for {
+		result, err := ec2Client.DescribeTags(input)
```
lisguo, line 87: I tried to ask q if there's any other place in the code that uses DescribeTags that I could borrow this logic from...didn't find anything. Let me know if I can re-use something here

translator/translate/util/tagsutil.go
```diff
@@ -20,13 +28,82 @@ func AddHighResolutionTag(tags interface{}) {
...
+
+var getEC2TagValueFunc = getEC2TagValue
+
+func GetEC2TagValue(tagKey string) string {
+	return getEC2TagValueFunc(tagKey)
+}
+
+func getEC2TagValue(tagKey string) string {
+	ec2Util := ec2util.GetEC2UtilSingleton()
+	if ec2Util.InstanceID == "" || ec2Util.Region == "" {
```
lisguo, line 50: need instance id and region to call describe tags for the current instance

translator/translate/util/placeholderUtil.go
```diff
@@ -87,13 +100,47 @@ func GetMetadataInfo(provider MetadataInfoProvider) map[string]string {
...
+		accountIdPlaceholder:     accountID,
 	}
 
-	return map[string]string{instanceIdPlaceholder: instanceID, hostnamePlaceholder: hostname,
-		localHostnamePlaceholder: localHostname, ipAddressPlaceholder: ipAddress, awsRegionPlaceholder: awsRegion,
-		accountIdPlaceholder: accountID,
+	// Add AWS metadata placeholders
+	metadata[ec2tagger.SupportedAppendDimensions["InstanceId"]] = instanceID
+	metadata[ec2tagger.SupportedAppendDimensions["InstanceType"]] = instanceType
+	metadata[ec2tagger.SupportedAppendDimensions["ImageId"]] = imageID
```
dricross: This doesn't really address jefchien@'s concerns about redefining these constants since these ("InstanceId", "InstanceType", etc) still have to match what's defined in ec2tagger processor.

We to export these and reference them here: https://github.com/aws/amazon-cloudwatch-agent/blob/main/plugins/processors/ec2tagger/constants.go#L64-L67

translator/translate/util/tagsutil.go
dricross: Looks like we use DescribeTags for retrieving the EKS cluster name: https://github.com/aws/amazon-cloudwatch-agent/blob/main/translator/translate/logs/util/get_eks_cluster_name.go#L80. And we have some specialized retry logic with backoff logic, so it might be important to re-use that. That code is specific for retrieving the EKS cluster name though, so some refactoring would be needed to make these two uses mesh.

translator/translate/util/placeholderUtil.go
```diff
@@ -87,13 +100,47 @@ func GetMetadataInfo(provider MetadataInfoProvider) map[string]string {
...
+}
+
+// GetAWSMetadataInfo returns AWS metadata using Ec2MetadataInfoProvider and EC2 Tags
+func GetAWSMetadataInfo() map[string]string {
+	// Start with the existing metadata pattern
+	metadata := GetMetadataInfo(Ec2MetadataInfoProvider)
+
+	// Add EC2 tags that require API calls (like AutoScaling group name)
+	if asgName := GetEC2TagValue(ec2tagger.Ec2InstanceTagKeyASG); asgName != "" {
+		metadata[ec2tagger.SupportedAppendDimensions["AutoScalingGroupName"]] = asgName
```
dricross: Similar comment as above. Use constant defined in ec2tagger instead of string literal: https://github.com/aws/amazon-cloudwatch-agent/blob/main/plugins/processors/ec2tagger/constants.go#L64-L67

translator/translate/util/placeholderUtil.go
lisguo: yeah the `SupportedAppendDimensions` map is good for making sure we are consistent with ec2tagger for what we support for the `aws:` prefix. I can include the exact constants here to replace the dimension names

translator/translate/util/placeholderUtil.go
```diff
@@ -62,38 +75,52 @@ func ResolvePlaceholder(placeholder string, metadata map[string]string) string {
...
+		instanceIdPlaceholder:    instanceID,
+		hostnamePlaceholder:      hostname,
+		localHostnamePlaceholder: localHostname,
+		ipAddressPlaceholder:     ipAddress,
+		awsRegionPlaceholder:     awsRegion,
+		accountIdPlaceholder:     accountID,
+
+		ec2tagger.SupportedAppendDimensions[ec2tagger.MdKeyInstanceID]:   instanceID,
+		ec2tagger.SupportedAppendDimensions[ec2tagger.MdKeyInstanceType]: instanceType,
+		ec2tagger.SupportedAppendDimensions[ec2tagger.MdKeyImageID]:      imageID,
```
jefchien, line 116: Might be worth keeping the AWS metadata info and regular metadata info completely separate. There isn't an overlap in the placeholder keys. Could be confusing if suddenly `{ip_address}` gets replaced in the `append_dimensions` as well.

translator/translate/util/placeholderUtil.go
```diff
@@ -62,38 +75,52 @@ func ResolvePlaceholder(placeholder string, metadata map[string]string) string {
...
+// GetAWSMetadataInfo returns AWS metadata using Ec2MetadataInfoProvider and EC2 Tags
+func GetAWSMetadataInfo() map[string]string {
+	// Start with the existing metadata pattern
+	metadata := GetMetadataInfo(Ec2MetadataInfoProvider)
 
-	accountID := provider().AccountID
-	if accountID == "" {
-		accountID = unknownAccountId
+	// Add EC2 tags that require API calls (like AutoScaling group name)
+	if asgName := GetEC2TagValue(ec2tagger.Ec2InstanceTagKeyASG); asgName != "" {
```
jefchien: Could we change this to only make the describe call if it's been configured in `append_dimensions`?

translator/translate/util/tagsutil.go
jefchien: Instead of doing this as part of the `FilterReservedKeys`, can we just call `ResolveAWSMetadataPlaceholders` after in `commonconfigutil.go`?

translator/translate/util/tagsutil.go
```diff
@@ -20,13 +28,82 @@ func AddHighResolutionTag(tags interface{}) {
...
+				Name:   aws.String("key"),
+				Values: []*string{aws.String(tagKey)},
+			},
+		},
+	}
+
+	for {
+		result, err := ec2Client.DescribeTags(input)
+		if err != nil {
+			log.Printf("Failed to describe EC2 tag '%s': %v", tagKey, err)
```
jefchien: Log lines should have log level prefix (e.g. `E!`).

translator/translate/util/placeholderUtil.go
lisguo: thats a good point. I can revert this to keep them separate

translator/translate/util/tagsutil.go
lisguo: Yeah, I could try to refactor but not sure if I should tackle it in this PR. Can be a follow up

translator/translate/util/placeholderUtil.go
```diff
@@ -62,38 +76,56 @@ func ResolvePlaceholder(placeholder string, metadata map[string]string) string {
...
+		awsRegionPlaceholder:     awsRegion,
+		accountIdPlaceholder:     accountID,
 	}
+}
 
-	hostname := provider().Hostname
-	if hostname == "" {
-		hostname = localHostname
-	}
+func GetAWSMetadataInfo(provider MetadataInfoProvider) map[string]string {
```
jefchien: nit: This isn't used anywhere else but the `GetAWSMetadataPlaceholderInfo`, so could be either not exported or just combined with that function.

translator/translate/util/tagsutil.go
```diff
@@ -30,3 +38,69 @@ func FilterReservedKeys(input any) any {
...
+				Name:   aws.String("key"),
+				Values: []*string{aws.String(tagKey)},
+			},
+		},
+	}
+
+	for {
+		result, err := ec2Client.DescribeTags(input)
+		if err != nil {
+			log.Printf("E! Failed to describe EC2 tag '%s': %v", tagKey, err)
```
jefchien, line 89: nit: Won't this start to show up in every translator log since we don't include `DescribeTags` in the default policy? Are we sure we want to log this?

jefchien (dismissed):

translator/translate/util/tagsutil.go
lisguo: We do this already when trying to get the EKS cluster name: https://github.com/aws/amazon-cloudwatch-agent/blob/main/translator/translate/logs/util/get_eks_cluster_name.go#L80

But this is only invoked if a customer sets their append_dimensions with an aws placeholder. So I think it makes sense to log it because they are trying to use a tag for their dimension, but it will fail.

translator/translate/util/placeholderUtil.go
lisguo: can update if need another revision

translator/translate/util/placeholderUtil.go
```diff
@@ -109,12 +141,36 @@ func getIpAddress() string {
...
+	inputMap := input.(map[string]interface{})
+	result := make(map[string]any, len(inputMap))
+	var awsMetadata map[string]string
+
+	for k, v := range inputMap {
+		if vStr, ok := v.(string); ok && strings.Contains(vStr, awsPlaceholderPrefix) {
+			// Cache AWS metadata on first use
+			if awsMetadata == nil {
+				awsMetadata = GetAWSMetadataPlaceholderInfo()
+			}
```
dricross, line 153: nit: This will cache the metadata for one append_dimensions section, but I think it's going to re-generate it for each append_dimensions section in the agent config, incurring another AWS call to DescribeTags. Hard to know if that's an issue or not. The special retry logic in get_eks_cluster_name.go tells me we've had some issues in the past with the `DescribeTags` call. `GetAWSMetadataPlaceholderInfo` could do the caching instead so that it's shared across the entire invocation of the translator.

jefchien (approved):

dricross (approved):

translator/translate/util/placeholderUtil.go
lisguo: yeah, for ec2 metadata at least we end up calling  the ec2util singleton, so we are kind of double caching here: `ec2 := ec2util.GetEC2UtilSingleton()`. For tags though we want to make sure we aren't making double the calls. Can have a separate singleton for that

PR #1885, opened 2025-10-03
Removing auth-config configmap check for eks detection

Paramadon:
# Description of the issue
Currently we have calls to aws auth config map to determine if our CWA is running in EKS. However, with the launch of EKS access entries, the aws-auth config map is deprecated and is no longer guaranteed to be present. This results in a bunch of noisy 404 not found calls for the config map in the K8s audit logs.

# Description of changes
We will only use the issuer logic to determine if agent is running on eks, this was tested on both the newest and the oldest supported version of kubernetes (1.28). We removed some struct logic as it is not necessary/redundant. 

# Tests

### Issuer debug statement to recongnize eks mode
Screenshot 2025-10-03 at 4 21 18 PM

### Works with the oldest supported kubernetes verison 
Screenshot 2025-10-03 at 12 31 43 PM


translator/util/eksdetector/eksdetector.go
```diff
@@ -4,104 +4,39 @@
...
-				}
-			}
-			isEKSCacheSingleton = IsEKSCache{Value: value, Err: errors}
-		})
-
-		return isEKSCacheSingleton
-	}
+	getInClusterConfig = func() (*rest.Config, error) { return rest.InClusterConfig() }
+	IsEKS              = isEKS
+	NewDetector        = isEKS
```
musa-asad: Do we still need this variable? Looks redundant. If it's used outside of this code, can we refactor?

translator/translate/otel/translate_otel_test.go
```diff
@@ -26,7 +26,7 @@ func TestTranslator(t *testing.T) {
 	testCases := map[string]struct {
 		input           interface{}
 		wantErrContains string
-		detector        func() (eksdetector.Detector, error)
+		detector        func() eksdetector.IsEKSCache
 		isEKSDataStore  func() eksdetector.IsEKSCache
```
musa-asad, line 29: Do we need `detector` and `isEKSDataStore`?

Paramadon: we don't, we have a couple of places this exist, so we'll remove them.

translator/util/eksdetector/eksdetector.go
Paramadon: It is redundant, I'll make it into just one func.

translator/translate/otel/pipeline/applicationsignals/translator_test.go
```diff
@@ -211,7 +204,7 @@ func TestTranslatorMetricsForEC2(t *testing.T) {
 		input      map[string]interface{}
 		want       *want
 		wantErr    error
-		detector   func() (eksdetector.Detector, error)
+		detector   func() eksdetector.IsEKSCache
```
musa-asad: This can be cleaned up too?

translator/util/eksdetector/eksdetector.go
```diff
@@ -122,7 +57,7 @@ func (d *EksDetector) getIssuer() (string, error) {
 		return "", fmt.Errorf("failed to decode token payload: %w", err)
 	}
 
-	var claims map[string]interface{}
+	var claims map[string]any
```
musa-asad: Why was this changed?

musa-asad:
Did we test again after these changes?

musa-asad (approved):

translator/util/eksdetector/eksdetector.go
```diff
@@ -4,104 +4,39 @@
...
+	IsEKS = isEKS
 )
 
-// getConfigMap retrieves the configmap with the provided name in the provided namespace
-func (d *EksDetector) getConfigMap(namespace string, name string) (map[string]string, error) {
-	configMap, err := d.Clientset.CoreV1().ConfigMaps(namespace).Get(context.TODO(), name, metav1.GetOptions{})
+// isEKS checks if the agent is running on EKS by extracting the "iss" field from the service account token and
+// checking if it contains "eks"
+func isEKS() IsEKSCache {
+	issuer, err := getIssuer()
```
movence: This should only happen once, right? There is no need to fetch and check the issuer for every call.

translator/util/eksdetector/eksdetector.go
```diff
@@ -4,104 +4,49 @@
...
-	// NewDetector creates a new singleton detector for EKS
-	NewDetector = func() (Detector, error) {
-		var errors error
-		if clientset, err := getClient(); err != nil {
-			errors = err
-		} else {
-			detector = &EksDetector{Clientset: clientset}
+// isEKS checks if the agent is running on EKS by extracting the "iss" field from the service account token and
+// checking if it contains "eks". The result is cached to avoid repeated expensive operations.
+func isEKS() IsEKSCache {
```
musa-asad, line 33: Why can't we can make this global? Even if it requires refactoring. It would make the code cleaner and avoid `IsEKS = isEKS`.

translator/util/eksdetector/eksdetector.go
```diff
@@ -4,104 +4,49 @@
...
-	detector            Detector
-	isEKSCacheSingleton IsEKSCache
-	once                sync.Once
+	getInClusterConfig = func() (*rest.Config, error) { return rest.InClusterConfig() }
+	// IsEKS is a function variable that can be overridden in tests
+	IsEKS = isEKS
+
+	// Cache for the EKS detection result
+	eksCache *IsEKSCache
+	eksOnce  sync.Once
```
musa-asad: This looks like it can be changed back as a whole.
```suggestion
	isEKSCacheSingleton IsEKSCache
	once                sync.Once
```

translator/util/eksdetector/eksdetector.go
```diff
@@ -4,104 +4,49 @@
...
-	authConfigConfigMap = "aws-auth"
-)
-
-var _ Detector = (*EksDetector)(nil)
-
 var (
-	detector            Detector
-	isEKSCacheSingleton IsEKSCache
-	once                sync.Once
+	getInClusterConfig = func() (*rest.Config, error) { return rest.InClusterConfig() }
```
musa-asad, line 22: Could you explain why we need to do this?

translator/util/eksdetector/eksdetector.go
musa-asad: Okay, looks like it needs to be there for the tests.

translator/util/eksdetector/eksdetector.go
```diff
@@ -135,18 +80,8 @@ func (d *EksDetector) getIssuer() (string, error) {
...
-
-	// Create Clientset using generated configuration
-	clientset, err := getKubernetesClient(confs)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create Clientset for Kubernetes client")
-	}
-
-	return clientset, err
+// resetCacheForTesting resets the EKS detection cache - only used in tests
+func resetCacheForTesting() {
```
musa-asad: nit - If this is only for testing, why does it need to be here?

danburgo:
awesome! thank you

translator/util/eksdetector/eksdetector.go
```diff
@@ -4,104 +4,49 @@
...
-		var errors error
-		if clientset, err := getClient(); err != nil {
-			errors = err
-		} else {
-			detector = &EksDetector{Clientset: clientset}
+// isEKS checks if the agent is running on EKS by extracting the "iss" field from the service account token and
+// checking if it contains "eks". The result is cached to avoid repeated expensive operations.
+func isEKS() IsEKSCache {
+	once.Do(func() {
+		issuer, err := getIssuer()
```
movence, line 37: could be simlified a bit by staying in a similar pattern with the existing approach
```suggestion
    once.Do(func() {
		var err error
		var value bool
        issuer, err := getIssuer()
        if len(issuer) > 0 && err == nil {
            value := strings.Contains(strings.ToLower(issuer), "eks")    
        }
        isEKSCacheSingleton = IsEKSCache{Value: value, Err: err}
	})
```

movence (approved):

PR #1886, opened 2025-10-06
Error for missing clustername on container

agarakan:
# Description of the issue
Pulling in missing error logic when clusterName is missing on a container. This is an edge case that we should ideally not see but should still log errors for. 

# Description of changes
Add error check and test

# License

# Tests
Added and ran unit tests
```
make lint
make fmt
make fmt-sh
make test
```
# Requirements
2. Run `make lint`

-------
### Integration Tests


translator/translate/otel/receiver/prometheus/translator.go
```diff
@@ -238,6 +240,11 @@ func getClusterNameFromConfig(conf *confmap.Conf, promConfigKey string) string {
 		clusterName = util.GetECSClusterName(clusterNameSectionKey, prometheusConfigInput)
 	}
 
+	// Cluster Name is mandatory for Containerized Workloads
+	if context.CurrentContext().RunInContainer() && clusterName == "" {
+		translator.AddErrorMessages(promConfigKey, "ClusterName is not defined")
+	}
+
```
musa-asad, line 247: Not sure if this is viable, but is there any way we can consolidate the logic here in the `ApplyRule` and call that?

agarakan: actually this branch removes the applyRule methods to just process the config linearly.

musa-asad (approved):

TravisStark (approved):

PR #1892, opened 2025-10-13
Support AutoScalingGroup as an append_dimension for host metrics

lisguo:
# Description of the issue
Currently, the append_dimensions section for host metrics can only support raw key/value pairs. The only way to support having a dynamic AutoScalingGroup dimension is to put it in the global metrics_collected section which uses the EC2 resource tag to populate it.

At translation time, we will call EC2 DescribeTags to populate placeholder values in the agent json configuration.

Example Agent JSON:
```
{
  "metrics": {
    "append_dimensions": {
    },
    "metrics_collected": {
      "cpu": {
        "append_dimensions": {
          "InstanceType": "${aws:InstanceType}",
          "ImageId": "${aws:ImageId}",
          "AutoScalingGroupName": "${aws:AutoScalingGroupName}", // <--- This will trigger EC2 DescribeTags
          "ServiceName": "MyServiceApplication"
        },
        "measurement": [
          "cpu_usage_idle"
        ]
      }
    }
  }
}
```

Output TOML for the cpu section:
```
[inputs]

  [[inputs.cpu]]
    fieldpass = ["usage_idle"]
    percpu = false
    totalcpu = true
    [inputs.cpu.tags]
      AutoScalingGroupName = "production-web-asg" // <--- populated from ASG tag "aws:autoscaling:groupName" value
      ImageId = "ami-0abcdef1234567890"
      InstanceType = "t3.medium"
      ServiceName = "MyServiceApplication"
```

# Description of changes
- Added singleton for storing tags on translation time in `tagsutil.go` to be referenced in `placeholderUtil.go` and `get_eks_cluster_name.go`

- Since we are calling EC2 DescribeTag for both getting the AutoScalingGroup name and the EKS cluster name, we can use a singleton that caches the tags for both use cases. If the EKS cluster name does not exist in the cache initially, it will use the existing retry and backoff logic to refresh the cache.

- Create GetAutoScalingGroupName and GetEKSClusterName functions in tagutil package

# License

# Tests
- Added unit test for e2e translation
- Tested on EC2 instance with tag:
```
aws:autoscaling:groupName test-asg
```

```
  [[inputs.cpu]]
    fieldpass = ["usage_idle"]
    percpu = false
    totalcpu = true
    [inputs.cpu.tags]
      AutoScalingGroupName = "test-asg"
      ImageId = "ami-0c503d55a14d7a5f0"
      InstanceType = "m6i.xlarge"
      ServiceName = "MyServiceApplication"
```

# Requirements
2. Run `make lint`

-------
### Integration Tests


translator/translate/logs/util/get_eks_cluster_name.go
```diff
@@ -4,26 +4,8 @@
...
-
-	"github.com/aws/aws-sdk-go/aws"
-	"github.com/aws/aws-sdk-go/aws/session"
-	"github.com/aws/aws-sdk-go/service/ec2"
-
-	configaws "github.com/aws/amazon-cloudwatch-agent/cfg/aws"
 	"github.com/aws/amazon-cloudwatch-agent/translator/util/ec2util"
-)
-
-const (
```
lisguo, line 19: moved this logic to `tagutil.go`

translator/translate/util/placeholderUtil.go
```diff
@@ -13,6 +13,27 @@ import (
 	"github.com/aws/amazon-cloudwatch-agent/plugins/processors/ec2tagger"
 	"github.com/aws/amazon-cloudwatch-agent/translator/translate/agent"
 	"github.com/aws/amazon-cloudwatch-agent/translator/util/ec2util"
+	"github.com/aws/amazon-cloudwatch-agent/translator/util/tagutil"
+)
+
+const (
```
lisguo, line 19: moved this up to the top of the file

translator/translate/util/placeholderUtil.go
```diff
@@ -26,13 +47,7 @@ type Metadata struct {
 
 type MetadataInfoProvider func() *Metadata
 
-var ec2MetadataInfoProviderFunc = ec2MetadataInfoProvider
```
lisguo, line 29: simplified this by just creating an anon function that directly calls `GetEC2UtilSingleton()`

translator/translate/util/placeholderUtil.go
```diff
@@ -140,26 +157,48 @@ func getIpAddress() string {
...
+	return metadata
+}
+
 func ResolveAWSMetadataPlaceholders(input any) any {
 	inputMap := input.(map[string]interface{})
 	result := make(map[string]any, len(inputMap))
-	var awsMetadata map[string]string
 
-	for k, v := range inputMap {
+	hasAWSPlaceholders := false
```
lisguo, line 187: flags to determine if we need to query EC2 metadata or EC2 tags

translator/translate/util/tagsutil.go
```diff
@@ -1,106 +0,0 @@
-// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
lisguo, line 1: moved all this into the `util/tagutil` package as this caused circular dependency issues with get_eks_cluster_name

translator/translate/util/placeholderUtil.go
```diff
@@ -117,6 +118,28 @@ func getAWSMetadataInfo(provider MetadataInfoProvider) map[string]string {
 	}
 }
 
+var tagMetadataProvider func() map[string]string
+
+func getTagMetadata() map[string]string {
```
lisguo, line 123: this function is used to determine what tag metadata we will support in append_dimensions. currently it is only auto scaling group

translator/util/tagutil/tagutil.go
```diff
@@ -0,0 +1,220 @@
...
+			LogLevel: configaws.SDKLogLevel(),
+			Logger:   configaws.SDKLogger{},
+		})
+}
+
+var ec2APIProvider EC2APIProvider = defaultEC2APIProvider
+
+type TagsCache struct {
+	instanceID string
+	tags       sync.Map
```
dricross: Caching all of the tags makes sense. Call it once and be done.

nit: I think the translator is all single threaded, so a regular map should be fine. I doubt that using a sync.Map will slow it down noticeably though.

translator/util/tagutil/tagutil.go
```diff
@@ -0,0 +1,220 @@
...
+var defaultEC2APIProvider = func() EC2TagsClient {
+	ec2CredentialConfig := &configaws.CredentialConfig{
+		Region: agent.Global_Config.Region,
+	}
+	return ec2.New(
+		ec2CredentialConfig.Credentials(),
+		&aws.Config{
+			LogLevel: configaws.SDKLogLevel(),
+			Logger:   configaws.SDKLogger{},
+		})
```
dricross: I noticed we're changing up how we create the ec2 session for the DescribeTags calls. Instead of using the default credentials provider, we're using our custom one. The `Region` also comes from the agent configuration instead of from `ec2util.GetEC2UtilSingleton().Region`. I don't know what the potential side effects of those changes are, but are we certain this is the right way to construct the client?

translator/util/tagutil/tagutil.go
```diff
@@ -0,0 +1,220 @@
...
+
+func (tc *TagsCache) loadAllTags() {
+	tc.once.Do(func() {
+		ec2API := ec2APIProvider()
+		if ec2API == nil {
+			log.Printf("W! loadAllTags: EC2 API client not available for instance %s", tc.instanceID)
+			return
+		}
+
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
dricross: Makes sense to add a timeout here

translator/util/tagutil/tagutil.go
```diff
@@ -0,0 +1,220 @@
...
+					Values: aws.StringSlice([]string{tc.instanceID}),
+				},
+			},
+		}
+
+		result, err := ec2API.DescribeTagsWithContext(ctx, input)
+		if err != nil {
+			log.Printf("W! loadAllTags: Failed to describe tags for instance %s: %v", tc.instanceID, err)
+			return
+		}
```
dricross: With the current changes, the retry logic only applies to the DescribeTag failures when trying to poll for the cluster name. Would it be better to have the retry logic on any DescribeTags call instead? Should we be using `result, err := callFuncWithRetries(ec2.DescribeTagsWithContext, input, "Describe EC2 Tag Fail.")` here instead to maintain functionality?

translator/util/tagutil/tagutil.go
```diff
@@ -0,0 +1,220 @@
...
+					Values: aws.StringSlice([]string{"instance"}),
+				},
+				{
+					Name:   aws.String("resource-id"),
+					Values: aws.StringSlice([]string{tc.instanceID}),
+				},
+			},
+		}
+
+		result, err := ec2API.DescribeTagsWithContext(ctx, input)
```
dricross: We should make sure we get all pages of the API call i.e. check if result.NextToken exists. If so set input.SetNextToken(*result.NextToken), and call DescribeTags again.

translator/util/tagutil/tagutil.go
lisguo: Yeah I wasn't sure if we will have simultaneous reads on the maps for each translator component (ie look for ASG at the same time as EKS Cluster name).

If we are confident that all the translators run sequentially then I can revert this back to a map, or use a R mutex if it's simpler.

translator/util/tagutil/tagutil.go
lisguo: You're right. It should be getting the region from the singleton. I'll fix on next rev.

I noticed the singleton retrieves the Region from IMDS, but I wonder what our global config is used for then?

translator/util/tagutil/tagutil.go
lisguo: Yeah this was intentional to keep the existing behavior with EKS cluster name retrieval, without also adding the retry logic on the ASG. 

I wasn't sure if we wanted the retry logic in all areas of the translator that will call DescribeTags moving forward

translator/util/tagutil/tagutil.go
lisguo: I reverted this to keep the old logic which just called DescribeTags instead of DescribeTagsWithContext

dricross (approved):

JayPolanco (approved):

PR #1903, opened 2025-10-15
Add workload discovery

jefchien:
# Description of changes
- Adds a new binary (`workload-discovery`), which scans running processes and devices to detect supported workloads (e.g. JVM, Tomcat, Kafka, NVIDIA GPU).
  - 2.8MB
  - Each detector has a README with more details on the behavior.
- Updates build scripts and packaging to build and install the `workload-discovery` binary alongside existing binaries for all distributions.
- Adds a new `status-with-workloads` action to the `amazon-cloudwatch-agent-ctl` script, which will run the `workload-discovery` binary and return its output.

# License

# Tests
Adds unit tests.

Manual testing on instance with sample apps, Tomcat, and Kafka.
```
[
  {
    "categories": [
      "JVM"
    ],
    "name": "org.apache.zookeeper.server.quorum.QuorumPeerMain",
    "status": "NEEDS_SETUP/JMX_PORT"
  },
  {
    "categories": [
      "JVM",
      "KAFKA/BROKER"
    ],
    "name": "Kafka Broker",
    "status": "NEEDS_SETUP/JMX_PORT"
  },
  {
    "categories": [
      "JVM"
    ],
    "name": "SampleApp",
    "telemetry_port": 9999,
    "status": "READY"
  },
  {
    "categories": [
      "JVM",
      "TOMCAT"
    ],
    "name": "/opt/tomcat/latest",
    "telemetry_port": 1080,
    "status": "READY"
  },
  {
    "categories": [
      "JVM"
    ],
    "name": "io.opentelemetry.contrib.jmxmetrics.JmxMetrics",
    "status": "NEEDS_SETUP/JMX_PORT"
  },
  {
    "categories": [
      "JVM"
    ],
    "name": "com.example.demo.DemoApplication",
    "telemetry_port": 2030,
    "status": "READY"
  }
]
```

# Requirements
2. Run `make lint`

-------
### Integration Tests

### DIFF
--- a/cmd/workload-discovery/discovery.go
+++ b/cmd/workload-discovery/discovery.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"errors"
 	"log/slog"
+	"math"
 	"os"
 	"runtime"
 	"sync"
@@ -18,6 +19,7 @@ import (
 	"github.com/aws/amazon-cloudwatch-agent/internal/detector"
 	"github.com/aws/amazon-cloudwatch-agent/internal/detector/filter"
 	"github.com/aws/amazon-cloudwatch-agent/internal/detector/java"
+	"github.com/aws/amazon-cloudwatch-agent/internal/detector/nvidia"
 	"github.com/aws/amazon-cloudwatch-agent/internal/detector/util"
 	"github.com/aws/amazon-cloudwatch-agent/tool/paths"
 )
@@ -33,6 +35,7 @@ type Discoverer struct {
 	cfg              Config
 	logger           *slog.Logger
 	processDetectors []detector.ProcessDetector
+	deviceDetectors  []detector.DeviceDetector
 	filters          filter.Filters
 }
 
@@ -44,25 +47,51 @@ func NewDiscoverer(cfg Config, logger *slog.Logger) *Discoverer {
 		processDetectors: []detector.ProcessDetector{
 			java.NewDetector(logger, filters.Process.Name),
 		},
+		deviceDetectors: []detector.DeviceDetector{
+			nvidia.NewDetector(logger),
+		},
 		filters: filters,
 	}
 }
 
 func (d *Discoverer) Discover(ctx context.Context) error {
 	start := time.Now()
-	processes, err := process.Processes()
-	if err != nil {
-		return err
+
+	var wg sync.WaitGroup
+	var processMd, deviceMd detector.MetadataSlice
+	var processErr error
+
+	// Run process detection
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		processes, err := process.Processes()
+		if err != nil {
+			processErr = err
+			return
+		}
+		processMd, processErr = d.detectMetadataFromProcesses(ctx, processes)
+	}()
+
+	// Run device detection
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		deviceMd = d.detectMetadataFromDevices()
+	}()
+
+	wg.Wait()
+
+	if processErr != nil {
+		return processErr
 	}
-	md, err := d.detectMetadataFromProcesses(ctx, processes)
-	if err != nil {
-		return err
-	}
+
+	md := append(processMd, deviceMd...)
 	d.logger.Debug("Discovered metadata", "elapsed", time.Since(start))
 	if len(md) > 0 {
 		encoder := json.NewEncoder(os.Stdout)
 		encoder.SetIndent("", "  ")
-		if err = encoder.Encode(md); err != nil {
+		if err := encoder.Encode(md); err != nil {
 			return err
 		}
 	}
@@ -95,18 +124,16 @@ func (d *Discoverer) detectMetadataFromProcesses(ctx context.Context, processes 
 		}
 	}()
 
-	selfPID := int32(os.Getpid()) // nolint:gosec
+	selfPID, err := convertToPid32(os.Getpid())
 	for _, p := range processes {
 		// skip the workload discovery process
-		if p.Pid == selfPID {
+		if err == nil && p.Pid == selfPID {
 			continue
 		}
 		select {
 		case jobs <- p:
 		case <-ctx.Done():
-			close(jobs)
-			workerWg.Wait()
-			return nil, ctx.Err()
+			break
 		}
 	}
 	close(jobs)
@@ -114,6 +141,9 @@ func (d *Discoverer) detectMetadataFromProcesses(ctx context.Context, processes 
 	close(results)
 	collectorWg.Wait()
 
+	if err = ctx.Err(); err != nil {
+		return nil, err
+	}
 	return mds, nil
 }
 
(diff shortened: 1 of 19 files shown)


Tools/src/create_deb.sh
```diff
@@ -60,6 +61,7 @@ cp ${BUILD_SPACE}/packaging/debian/debian-binary ${BUILD_ROOT}/
 chmod ug+rx ${BUILD_ROOT}/opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent
 chmod ug+rx ${BUILD_ROOT}/opt/aws/amazon-cloudwatch-agent/bin/amazon-cloudwatch-agent-ctl
 chmod ug+rx ${BUILD_ROOT}/opt/aws/amazon-cloudwatch-agent/bin/start-amazon-cloudwatch-agent
+chmod 700 ${BUILD_ROOT}/opt/aws/amazon-cloudwatch-agent/bin/workload-discovery
```
the-mann, line 64: should this be 700 or ug+rx? does the owner of the file need to be able to write to the workload discovery file?

lisguo (approved):

cmd/workload-discovery/discovery.go
```diff
@@ -0,0 +1,233 @@
...
+	var mds detector.MetadataSlice
+	collectorWg.Add(1)
+	go func() {
+		defer collectorWg.Done()
+		for r := range results {
+			mds = append(mds, r)
+		}
+	}()
+
+	selfPID := int32(os.Getpid()) // nolint:gosec
```
okankoAMZ: why do we have nolint flag here?

cmd/workload-discovery/discovery.go
```diff
@@ -0,0 +1,233 @@
...
+		}
+	}
+	return nil
+}
+
+func (d *Discoverer) detectMetadataFromProcesses(ctx context.Context, processes []*process.Process) (detector.MetadataSlice, error) {
+	d.logger.Debug(
+		"Starting discovery",
+		"num_process", len(processes),
+		"num_worker", d.cfg.Concurrency,
```
okankoAMZ, line 105: can `d.cfg.concurrency` be 0? if so should we validate that it is NOT 0 before proceeding in this function?

cmd/workload-discovery/discovery.go
```diff
@@ -0,0 +1,233 @@
...
+	selfPID := int32(os.Getpid()) // nolint:gosec
+	for _, p := range processes {
+		// skip the workload discovery process
+		if p.Pid == selfPID {
+			continue
+		}
+		select {
+		case jobs <- p:
+		case <-ctx.Done():
+			close(jobs)
```
okankoAMZ: we not closing results here

Tools/src/create_deb.sh
jefchien: Yes, the owner is root. The package managers need to be able to write to the file to be able to replace it during an update.
The `ug+rx` gives
```
-rwxr-xr-x 1 root root 183086664 Oct 15 23:50 amazon-cloudwatch-agent
```
`700` gives
```
-rwx------ 1 root root   2801848 Oct 16 15:29 workload-discovery
```

cmd/workload-discovery/discovery.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+	handler := slog.NewTextHandler(os.Stderr, opts)
+	return slog.New(handler)
+}
+
+func main() {
+	cfg := Config{
+		Concurrency: runtime.NumCPU(),
+		LogLevel:    slog.LevelDebug,
+		Timeout:     500 * time.Millisecond,
```
okankoAMZ, line 224: how did we determine these timings? Does the ratio between MinUpTime and Timeout matter?

internal/detector/kafka/kafka.go
```diff
@@ -0,0 +1,87 @@
...
+
+// Detect identifies Kafka broker processes by looking at the command-line arguments to find the Kafka main class
+// (`kafka.Kafka`). Returns metadata with the KAFKA/BROKER category and Kafka Broker as the name if detected.
+func (d *kafkaDetector) Detect(ctx context.Context, process detector.Process) (*detector.Metadata, error) {
+	var md *detector.Metadata
+	for _, sd := range d.subDetectors {
+		var err error
+		md, err = sd.Detect(ctx, process)
+		if err != nil {
+			continue
```
okankoAMZ: Nit: should we break when there is an error to make it easier to read?

internal/detector/nvidia/nvidia_linux.go
```diff
@@ -0,0 +1,57 @@
...
+	return &checker{
+		devPath:     "/dev",
+		driverPaths: []string{defaultNvidiaSMIPath},
+	}
+}
+
+// hasNvidiaDevice checks for NVIDIA GPU devices on Linux by looking for device files.
+// NVIDIA GPUs create device files in /dev/nvidia[0-9]+ when the kernel module is loaded.
+func (c *checker) hasNvidiaDevice() bool {
+	entries, err := os.ReadDir(c.devPath)
```
okankoAMZ, line 32: shouldn't we check the `c.devPath` with `os.Stat` before we try to read the directory.

packaging/windows/amazon-cloudwatch-agent-ctl.ps1
```diff
@@ -229,7 +236,24 @@ Function StatusAll() {
...
+        $workload_discovery_path = $WORKLOAD_DISCOVERY
+        if (Test-Path -LiteralPath $workload_discovery_path) {
+            try {
+                $workloads_json = cmd /c "`"$workload_discovery_path`" 2>null"
+                if (-not $workloads_json) { $workloads_json = "[]" }
+            } catch {
+                $workloads_json = "[]"
+            }
+        }
+        Write-Output "  `"version`": `"${version}`","
```
okankoAMZ, line 251: why are we moving `Write-Output " `"version`": `"${version}`""` should this if statement just print workloads or not, I don't this statement should effect the version

cmd/workload-discovery/discovery.go
jefchien: It's currently not configurable.

cmd/workload-discovery/discovery.go
jefchien: We had some internal discussions about the timeout limits. The ratio does not matter. The `MinUpTime` is for the scanned processes., while the `Timeout` is for the workload discovery run.

internal/detector/nvidia/nvidia_linux.go
jefchien: It'll fail the same way if the `devPath` doesn't exist with a `os.PathError`.

the-mann: similar note, what about SELinux? will accessing the device work with the existing SELinux policies Param wrote?

cmd/workload-discovery/discovery.go
jefchien: You're right.

packaging/windows/amazon-cloudwatch-agent-ctl.ps1
jefchien: The `status` action and `status-with-workloads` action is using a combined function, so that's why it was moved.

cmd/workload-discovery/discovery.go
jefchien: gopsutil process pids are int32, so we need to cast the os pid.

okankoAMZ (approved):

PR #1915, opened 2025-10-20
Fallback to iterative DescribeLogGroups if batch not supported

duhminick:
# Description of the issue
In #1717, we implemented batching for `DescribeLogGroups` (DLG) calls using the `LogGroupIdentifiers` parameter to reduce API throttling when customers have many log groups. However, this batched DLG API is not supported in all cases. When unsupported, the agent fails with an `InvalidParameterException` error: `"Input filter on Log group identifiers is not supported."`, preventing log group retention policies from being checked and updated.

# Description of changes
This PR adds a fallback mechanism to handle regions where the batched DLG API is not available:

  1. **Primary path**: Attempts to use the batched `DescribeLogGroups` API with `LogGroupIdentifiers` (introduced in #1717)

  2. **Fallback path**: If the batch call fails with the specific unsupported error, falls back to individual `DescribeLogGroups` calls using `LogGroupNamePrefix` for each log group. This is the same logic that was in the agent before.

 **Implementation details:**
  - Refactored `updateTargetBatch()` → `updateTargets()` which orchestrates the batch-first approach
  - Added `updateTargetsBatch()` for the batched API call (returns error if not supported)
  - Added `updateTargetsIteratively()` for the fallback individual API calls
  - Added back `getRetention()` helper to retrieve retention using prefix-based lookup
  - Error detection checks for both `InvalidParameterException` error code AND exact error message to avoid false positives
  - Logs at DEBUG level when fallback occurs

# License

# Tests
1. Unit tests
2. Integration tests

# Requirements
3. Run `make lint`

-------
### Integration Tests


plugins/outputs/cloudwatchlogs/internal/pusher/target.go
```diff
@@ -225,6 +241,49 @@ func (m *targetManager) updateTargetBatch(targets map[string]Target) {
...
+	for _, target := range targets {
+		for attempt := 0; attempt < numBackoffRetries; attempt++ {
+			currentRetention, err := m.getRetention(target)
+			if err != nil {
+				m.logger.Errorf("failed to describe log group retention for target %v: %v", target, err)
+				time.Sleep(m.calculateBackoff(attempt))
+				continue
+			}
+
+			if currentRetention != target.Retention && target.Retention > 0 {
```
dricross, line 258: We should skip the target if the target.Retention <= 0. no point in the making the DescribeLogGroups call if we're not going to use it, and we should minimize AWS calls where we can.

duhminick: https://github.com/aws/amazon-cloudwatch-agent/blob/583aeed4f557826216cd9f12ea6627c767938ef5/plugins/outputs/cloudwatchlogs/internal/pusher/target.go#L93-L99

The target map will only have targets with retention greater than 0

dricross: Ok great. So the `target.Retention > 0` check isn't needed but good to have a double check.

PR #1930, opened 2025-10-30
Adding ipv6 to useragent

Paramadon:
# Description of the issue
Adding ipv6 to customer useragent. The agent set's `AWS_USE_DUALSTACK_ENDPOINT` to true if `use_dualstack_endpoint` is passed into the agent configuration and set to true. Having AWS_USE_DUALSTACK_ENDPOINT to true makes it so that all sdk calls make dualstack endpoint calls which makes the agent work on ipv6 environments. 

# Description of changes
Adding a check to see if the AWS_USE_DUALSTACK_ENDPOINT exist and if so we'll add ipv6 as part of feature flag for it to get parsed.

### Tested change - Useragent string has ipv6
Screenshot 2025-10-30 at 4 45 45 PM

### DIFF
--- a/extension/agenthealth/handler/useragent/useragent.go
+++ b/extension/agenthealth/handler/useragent/useragent.go
@@ -141,9 +141,15 @@ func (ua *userAgent) SetComponents(otelCfg *otelcol.Config, telegrafCfg *telegra
 		ua.inputs.Add(flagRunAsUser)
 	}
 
+	// Add ipv6 feature flag if dualstack endpoint is enabled
+	if os.Getenv(envconfig.AWS_USE_DUALSTACK_ENDPOINT) == "true" {
+		ua.feature.Add("ipv6")
+	}
+
 	ua.inputsStr.Store(componentsStr(typeInputs, ua.inputs))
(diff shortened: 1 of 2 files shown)


extension/agenthealth/handler/useragent/useragent.go
```diff
@@ -141,9 +141,15 @@ func (ua *userAgent) SetComponents(otelCfg *otelcol.Config, telegrafCfg *telegra
 		ua.inputs.Add(flagRunAsUser)
 	}
 
+	// Add ipv6 feature flag if dualstack endpoint is enabled
+	if os.Getenv(envconfig.AWS_USE_DUALSTACK_ENDPOINT) == "true" {
```
varunch77, line 145: are these agent health metrics just for EC2 instances? I'm assuming AWS_USE_DUALSTACK_ENDPOINT isn't available everywhere so are we worried about that?

Paramadon: The test was done on eks, so it does work both on eks and ec2

TravisStark:
Build windows-2022 seems to be failing for some reason unrelated to this change

TravisStark (approved):

PR #1931, opened 2025-10-31
Improving windows build flakiness

Paramadon:
# Description of the issue
Improving flaky windows build failures. Currently the windows build is flaky and is causing merging delays as tests needs to be rerun after failure repeatedly until they work - here's an example of a pr with 4 windows test failures where each run takes 20-25+ min... - https://github.com/aws/amazon-cloudwatch-agent/actions/runs/18954707387/job/54203959187?pr=1930. 

It seems like the issue is that we are setting up watcher when follow (follow true just means basically like tail -f vs false is like a cat) is false which should not happen and the watch is deadlocking. And having a unique prefix for temp to avoid matching temp files which seems to make test less flaky.

Test is going from 

# Description of changes
Make sure to only set up watcher if tail.Follow is true and unique test name

# Tests
Test goes from 25% pass rate to 96% passing As shown in this test: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/19050471827/job/54409684451


plugins/inputs/logfile/tail/tail.go
```diff
@@ -398,9 +398,12 @@ func (tail *Tail) tailFileSync() {
 		}
 	}
 
-	if err := tail.watchChanges(); err != nil {
-		tail.Killf("Error watching for changes on %s: %s", tail.Filename, err)
-		return
+	// Only set up file watchers if we're following the file
+	if tail.Follow {
```
jefchien, line 402: `Follow` is always set to true. https://github.com/aws/amazon-cloudwatch-agent/blob/main/plugins/inputs/logfile/logfile.go#L222

agarakan (approved):

PR #2014, opened 2026-02-11
Fix poison pill deadlock: Make retry heap unbounded

the-mann:
## Summary

Fix deadlock when failing log groups exceed concurrency limit by making the retry heap unbounded.

## Problem

With bounded retry heap (size = concurrency):
- Retry heap size = concurrency (e.g., 2)
- When failing log groups (10) > heap size (2)
- Heap fills with failed batches
- Workers block trying to push more failed batches
- System deadlocks - no progress possible
- Allowed log groups get starved

## Solution

Make retry heap unbounded:
- Remove maxSize constraint and semaphore
- Push() is now non-blocking
- Failed batches queue up without blocking workers
- Allowed log groups continue publishing normally

## Changes

- Remove `maxSize` and `semaphore` from retryHeap struct
- Make `Push()` non-blocking (no semaphore wait)
- Remove semaphore release from `PopReady()`
- Replace `sync.Cond` halt/resume with channel-based approach + mutex to prevent shutdown deadlock
- Add `TestQueueStopWhileHalted` to verify no shutdown deadlock
- Add state callback tests for retry, expiry, and shutdown scenarios
- Add `poison_pill_test.go` with comprehensive poison pill scenario tests
- Clean up test assertions and remove redundant tests

## Test Results

**Before Fix**: Test deadlocked after 30s with all goroutines blocked

**After Fix**: ✅ Test PASSES
```
Allowed success=5 (all batches published)
Denied attempts=110
Heap size=28 (grew beyond concurrency limit of 2)
No deadlock or starvation
```

## Integration Test

Integration test run: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/21973935546

New integration test added in amazon-cloudwatch-agent-test:
- Test directory: `test/cloudwatchlogs_concurrency`
- Config: concurrency=2, force_flush_interval=5s
- Validates: 1 allowed + 10 denied log groups
- Expected: Allowed log group continues publishing despite denied groups

### 1. Memory Stabilizes, Not Growing Unbounded ✅
- **Initial spike**: +1.6 MB in first 3 minutes (161s)
- **Stabilization**: Memory drops to +1.1 MB and remains stable
- **Final 2 minutes**: No growth despite 5,000 more events

### 2. Garbage Collection Active
Memory actually decreased from 125,244 KB (161s) to 124,816 KB (247s), showing Go's garbage collector is reclaiming memory from processed batches.

### 3. Retry Heap Bounded by Target Count
With 10 denied log groups:
- Each has 1 queue → max 1 batch in retry heap per target
- Circuit breaker halts queues after failure → prevents new batch creation
- Retry heap stabilizes at ~10 batches (one per denied log group)
- Memory growth is from batch metadata, not unbounded accumulation

### 4. No Memory Leak
- **0.9% growth** over 5 minutes with 29,000 events
- **Stable memory** in final 2 minutes despite continued event ingestion
- **Well below threshold**: <1% growth vs 50% leak threshold

## Conclusion

✅ No memory leak: 0.9% growth over 5 minutes with 29,000 events
✅ Memory stabilizes: Peaks at 1.6 MB, drops to 1.1 MB, remains stable
✅ Garbage collection works: Memory decreases after peak
✅ Retry heap bounded: Limited by number of failing targets (~10 batches)
✅ Production ready: Safe for long-running deployments with persistent failures

## Related PRs

- #1988 - Original PR (DRAFT, targeting sender-block-on-failure)
- #2012 - Duplicate targeting sender-block-on-failure
- #2023 - Draft PR targeting main (full feature set)

## Graceful Shutdown Test

Manually tested shutdown behavior with a loaded retry heap on EC2 (c7i.xlarge, AL2023).

**Setup**: concurrency=2, 7 log groups (2 allowed, 5 denied via IAM explicit deny), ~10 lines/sec per file for 4 minutes. Retry heap actively cycling through 5 denied groups with `AccessDeniedException` at time of shutdown.

**Stop method**: `amazon-cloudwatch-agent-ctl -a stop`

| Metric | Result |
|--------|--------|
| Shutdown time | **85ms** |
| Process exit | Clean — no residual process |
| systemd | `Deactivated successfully`, no timeout/forced kill |
| Allowed groups (2) | Final batches flushed before exit, 2000+ events each in CloudWatch |
| Denied groups (5) | Never created in CloudWatch, did not block shutdown |
| Memory at shutdown | 112 MB RSS, 10 threads |
| dmesg/OOM | Clean |

**Shutdown sequence** (all within 1 second):
```
Pusher published 1 log events to cwa-debug-allowed-1 (final flush)
Pusher published 1 log events to cwa-debug-allowed-2 (final flush)
Profiler is stopped during shutdown
Received signal from OS signal=terminated
Starting shutdown...
Stopping extensions...
Shutdown complete.
```

✅ No deadlock, no starvation, no goroutine leaks. Retry heap abandoned cleanly without blocking the shutdown path.


plugins/outputs/cloudwatchlogs/internal/pusher/queue_test.go
```diff
@@ -814,3 +790,66 @@ func TestQueueCallbackRegistration(t *testing.T) {
...
+	mockSender.AssertExpectations(t)
+}
+
+// TestQueueResumeOnBatchExpiry verifies that when a batch expires after 14 days of retrying,
+// the circuit breaker resumes the queue to allow new batches to be processed.
+// This prevents the target from being permanently blocked when a bad batch is eventually dropped.
+//
+// Scenario from PR comment: "Say a bad batch from a target caused this to halt. Now that bad batch
+// is re-tried for 14 days and eventually dropped - but this never gets resumed in that case right?
+// So this target is blocked forever in that scenario?"
```
jefchien: What is this?

plugins/outputs/cloudwatchlogs/internal/pusher/retryheap.go
```diff
@@ -41,51 +42,44 @@ func (h *retryHeapImpl) Pop() interface{} {
...
+	stopCh  chan struct{}
+	stopped bool
+	logger  telegraf.Logger
 }
 
 var _ RetryHeap = (*retryHeap)(nil)
 
-// NewRetryHeap creates a new retry heap with the specified maximum size
+// NewRetryHeap creates a new retry heap (unbounded)
 func NewRetryHeap(maxSize int, logger telegraf.Logger) RetryHeap {
```
jefchien: nit: `maxSize` is no longer used.

plugins/outputs/cloudwatchlogs/internal/pusher/pool_test.go
```diff
@@ -135,22 +131,20 @@ func TestSenderPool(t *testing.T) {
 	assert.Equal(t, int32(200), completed.Load())
 }
 
-func TestSenderPoolRetryHeap(t *testing.T) {
-	assert.NotPanics(t, func() {
-		logger := testutil.NewNopLogger()
-		mockService := new(mockLogsService)
-		mockService.On("PutLogEvents", mock.Anything).Return(&cloudwatchlogs.PutLogEventsOutput{}, nil)
+func TestSenderPoolRetryHeap(_ *testing.T) {
```
jefchien: nit: Still don't see the point of this test. It's not asserting anything.

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
```diff
@@ -245,3 +257,34 @@ func hasValidTime(e logs.LogEvent) bool {
...
 }
+
+// waitIfHalted blocks until the queue is unhalted (circuit breaker functionality)
+func (q *queue) waitIfHalted() {
+	q.haltCond.L.Lock()
+	for q.halted {
+		q.haltCond.Wait()
+	}
+	q.haltCond.L.Unlock()
+}
```
jefchien, line 310: Is there a potential for deadlocking here? If `Wait` is blocking the queue's goroutine and nothing calls `resume`, will it prevent the shutdown from happening?

https://pkg.go.dev/sync#Cond.Wait
> Wait atomically unlocks c.L and suspends execution of the calling goroutine. After later resuming execution, Wait locks c.L before returning. Unlike in other systems, Wait cannot return unless awoken by [Cond.Broadcast](https://pkg.go.dev/sync#Cond.Broadcast) or [Cond.Signal](https://pkg.go.dev/sync#Cond.Signal).

Makes it sound like it will block. Would it make sense to use a channel to block on the halt? That way we could do something like
```go
select {
case <-ch:
case <-q.stopCh:
}
```
which would allow it to unblock during a shutdown. 

Or we could add `q.resume()` to the `Stop` function. Maybe that would work.

plugins/outputs/cloudwatchlogs/internal/pusher/retryheap.go
```diff
@@ -201,9 +196,10 @@ func (p *RetryHeapProcessor) processReadyMessages() {
 
 	for _, batch := range readyBatches {
 		// Check if batch has expired
-		if batch.isExpired(p.maxRetryDuration) {
+		if batch.isExpired() {
 			p.logger.Errorf("Dropping expired batch for %v/%v", batch.Group, batch.Stream)
 			batch.updateState()
+			batch.done() // Resume circuit breaker to allow target to process new batches
```
jefchien: nit: `batch.done()` calls `batch.updateState()`, so we can remove that.

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
```diff
@@ -175,6 +181,11 @@ func (q *queue) merge(mergeChan chan logs.LogEvent) {
 func (q *queue) send() {
 	if len(q.batch.events) > 0 {
 		q.batch.addDoneCallback(q.onSuccessCallback(q.batch.bufferedSize))
+		q.batch.addFailCallback(q.onFailCallback())
```
jefchien: nit: Could be simplified to

```suggestion
		q.batch.addFailCallback(q.halt)
```

plugins/outputs/cloudwatchlogs/internal/pusher/queue_test.go
```diff
@@ -814,3 +790,66 @@ func TestQueueCallbackRegistration(t *testing.T) {
...
+	// Trigger flush to force send of second batch - this should block until resumed
+	done := make(chan bool)
+	go func() {
+		time.Sleep(100 * time.Millisecond) // Wait a bit
+		// Manually resume by calling success callback on a dummy batch
+		dummyBatch := newLogEventBatch(Target{"G", "S", util.StandardLogGroupClass, -1}, nil)
+		dummyBatch.addDoneCallback(func() {
+			// This simulates a successful send that should resume the queue
+		})
+		dummyBatch.done()
```
jefchien: nit: Don't understand what this is for. It's created outside of the queue, so the halt/resume functions aren't added to the batch. The batch isn't sent either.

plugins/outputs/cloudwatchlogs/internal/pusher/queue_test.go
```diff
@@ -814,3 +790,66 @@ func TestQueueCallbackRegistration(t *testing.T) {
 		mockSender.AssertExpectations(t)
 	})
 }
+func TestQueueHaltResume(t *testing.T) {
```
jefchien, line 894: Not convinced this is testing halt/resume. Where is the assertion for the resume? I see the assertion for halt.

plugins/outputs/cloudwatchlogs/internal/pusher/circuitbreaker_test.go
```diff
@@ -0,0 +1,100 @@
...
+		failingPusher.AddEvent(newStubLogEvent("fail-more", now))
+		healthyPusher.AddEvent(newStubLogEvent("ok-more", now))
+	}
+
+	time.Sleep(500 * time.Millisecond)
+
+	// Circuit breaker assertion: after the first failure, the failing target should
+	// NOT have sent additional batches. Only 1 send attempt should have been made
+	// before the circuit breaker blocks it.
+	assert.LessOrEqual(t, failingTargetSendCount.Load(), int32(1),
```
jefchien: nit: Why LessOrEqual? Shouldn't this always be 1?

plugins/outputs/cloudwatchlogs/internal/pusher/retryheap_expiry_test.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/aws/amazon-cloudwatch-agent/sdk/service/cloudwatchlogs"
+	"github.com/aws/amazon-cloudwatch-agent/tool/testutil"
+)
+
+// TestRetryHeapProcessorExpiredBatchShouldResume demonstrates the bug where
+// expired batches don't resume the circuit breaker, leaving the target permanently blocked.
+//
+// From PR comment: "Say a bad batch from a target caused this to halt. Now that bad batch
+// is re-tried for 14 days and eventually dropped - but this never gets resumed in that case right?
+// So this target is blocked forever in that scenario?"
```
jefchien: nit: Was this intentionally included?

plugins/outputs/cloudwatchlogs/internal/pusher/retryheap_expiry_test.go
```diff
@@ -0,0 +1,104 @@
...
+	// Create a batch that will expire
+	batch := newLogEventBatch(target, nil)
+	batch.append(newLogEvent(time.Now(), "test message", nil))
+
+	// Set up callbacks to track circuit breaker state
+	var circuitBreakerHalted atomic.Bool
+	var circuitBreakerResumed atomic.Bool
+
+	batch.addFailCallback(func() {
+		circuitBreakerHalted.Store(true)
```
jefchien: nit: What is the point of this if it's never checked?

plugins/outputs/cloudwatchlogs/internal/pusher/queue_test.go
the-mann: Kiro slop, i can delete the comment. the test itself I think is valuable, it checks to make sure that if a batch expires, it will be dropped and the circuit breaker will be reopened. The idea here being we want to recreate the synchronous behavior as closely as possible for dropping after 14 days

plugins/outputs/cloudwatchlogs/internal/pusher/retryheap.go
the-mann: will clean that up, thanks

plugins/outputs/cloudwatchlogs/internal/pusher/pool_test.go
the-mann: yeah not sure how this slipped through, i think the original PR owner said they would remove it but never got around to it

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
the-mann: discussed offline, will implement stopCh approach instead of trying q.resume()

plugins/outputs/cloudwatchlogs/internal/pusher/queue_test.go
the-mann: Removed in [`2d07b385`](https://github.com/aws/amazon-cloudwatch-agent/commit/2d07b385).

plugins/outputs/cloudwatchlogs/internal/pusher/retryheap.go
the-mann: Removed in [`fd2ea56a`](https://github.com/aws/amazon-cloudwatch-agent/commit/fd2ea56a).

plugins/outputs/cloudwatchlogs/internal/pusher/pool_test.go
the-mann: Deleted in [`3e1ed82c`](https://github.com/aws/amazon-cloudwatch-agent/commit/3e1ed82c).

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
the-mann: Replaced `sync.Cond` with channel-based `select` on `haltCh` and `stopCh` in [`3d9d7791`](https://github.com/aws/amazon-cloudwatch-agent/commit/3d9d7791). Added `TestQueueStopWhileHalted` to verify shutdown doesn't deadlock.

plugins/outputs/cloudwatchlogs/internal/pusher/retryheap.go
the-mann: Removed in [`38afc5f6`](https://github.com/aws/amazon-cloudwatch-agent/commit/38afc5f6).

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
the-mann: Done in [`b6f3b3ed`](https://github.com/aws/amazon-cloudwatch-agent/commit/b6f3b3ed). Deleted `onFailCallback` wrapper.

plugins/outputs/cloudwatchlogs/internal/pusher/queue_test.go
the-mann: Removed in [`eb589127`](https://github.com/aws/amazon-cloudwatch-agent/commit/eb589127).

plugins/outputs/cloudwatchlogs/internal/pusher/queue_test.go
the-mann: Added resume assertions and third event verification in [`3d9d7791`](https://github.com/aws/amazon-cloudwatch-agent/commit/3d9d7791).

plugins/outputs/cloudwatchlogs/internal/pusher/circuitbreaker_test.go
the-mann: Changed to `assert.Equal` in [`f640a193`](https://github.com/aws/amazon-cloudwatch-agent/commit/f640a193).

plugins/outputs/cloudwatchlogs/internal/pusher/retryheap_expiry_test.go
the-mann: Removed in [`c739bb97`](https://github.com/aws/amazon-cloudwatch-agent/commit/c739bb97).

plugins/outputs/cloudwatchlogs/internal/pusher/retryheap_expiry_test.go
the-mann: Removed in [`4e7393af`](https://github.com/aws/amazon-cloudwatch-agent/commit/4e7393af).

plugins/outputs/cloudwatchlogs/internal/pusher/queue_test.go
```diff
@@ -814,3 +789,94 @@ func TestQueueCallbackRegistration(t *testing.T) {
...
+	queueImpl.haltMu.Unlock()
+
+	// Add second event - should be queued but not sent due to halt
+	q.AddEvent(newStubLogEvent("second message", time.Now()))
+
+	// Verify only one send happened (queue is halted)
+	assert.Equal(t, int32(1), sendCount.Load(), "Should have only one send due to halt")
+
+	// Trigger resume by calling the success callback directly
+	queueImpl.resume()
```
jefchien, line 930: nit: Comment isn't quite right, but it accomplishes the same thing.

plugins/outputs/cloudwatchlogs/internal/pusher/queue_test.go
```diff
@@ -814,3 +789,94 @@ func TestQueueCallbackRegistration(t *testing.T) {
...
+	queueImpl := q.(*queue)
+	queueImpl.haltMu.Lock()
+	assert.True(t, queueImpl.halted, "Queue should be halted after failure")
+	queueImpl.haltMu.Unlock()
+
+	// Add second event - should be queued but not sent due to halt
+	q.AddEvent(newStubLogEvent("second message", time.Now()))
+
+	// Verify only one send happened (queue is halted)
+	assert.Equal(t, int32(1), sendCount.Load(), "Should have only one send due to halt")
```
jefchien: This doesn't wait long enough for the second batch to have been created and attempted to flush, so even without the halting logic this could be true. The flush interval is 10ms, so we should wait at least that amount of time before checking the sendCount and resuming.

plugins/outputs/cloudwatchlogs/internal/pusher/retryheap_recovery_test.go
```diff
@@ -0,0 +1,279 @@
...
+	mockService.AssertExpectations(t)
+}
+
+func stringPtr(s string) *string {
+	return &s
+}
+
+func int64Ptr(i int64) *int64 {
+	return &i
+}
```
jefchien: nit: If we need these functions, we typically use the functions provided by AWS SDK (e.g. `aws.String("string")`). Can see them in other tests/code in the repo.

plugins/outputs/cloudwatchlogs/internal/pusher/retryheap.go
```diff
@@ -139,7 +134,7 @@ type RetryHeapProcessor struct {
 func NewRetryHeapProcessor(retryHeap RetryHeap, workerPool WorkerPool, service cloudWatchLogsService, targetManager TargetManager, logger telegraf.Logger, maxRetryDuration time.Duration, retryer *retryer.LogThrottleRetryer) *RetryHeapProcessor {
```
jefchien: `maxRetryDuration` was removed everywhere else. This is no longer used.

plugins/outputs/cloudwatchlogs/internal/pusher/retryheap_recovery_test.go
```diff
@@ -0,0 +1,279 @@
...
+	// Heap should be empty (batch successfully sent)
+	assert.Equal(t, 0, heap.Size(), "Heap should be empty after successful retry")
+
+	// Verify both PutLogEvents calls were made
+	mockService.AssertExpectations(t)
+}
+
+// TestRecoveryAfterSystemRestart validates that when the system restarts with
+// retry ongoing, it resumes correctly by loading state and continuing retries.
+func TestRecoveryAfterSystemRestart(t *testing.T) {
```
jefchien: I'm not seeing the restart or anything to do with the state file.

plugins/outputs/cloudwatchlogs/internal/pusher/poison_pill_test.go
```diff
@@ -0,0 +1,225 @@
...
+		defer ticker.Stop()
+		for {
+			select {
+			case <-processorDone:
+				return
+			case <-ticker.C:
+				processor.processReadyMessages()
+			}
+		}
+	}()
```
jefchien, line 139: nit: Isn't this what the retry heap processor is for? Is there a reason we aren't using `processor.Start()`?

plugins/outputs/cloudwatchlogs/internal/pusher/poison_pill_test.go
```diff
@@ -0,0 +1,225 @@
...
+	"github.com/stretchr/testify/mock"
+
+	"github.com/aws/amazon-cloudwatch-agent/internal/retryer"
+	"github.com/aws/amazon-cloudwatch-agent/sdk/service/cloudwatchlogs"
+)
+
+// TestPoisonPillScenario validates that when 10 denied + 1 allowed log groups
+// share a worker pool with concurrency=2, the allowed log group continues
+// publishing without being starved by failed retries.
+func TestPoisonPillScenario(t *testing.T) {
```
jefchien: So this only tests the retry heap? Is that intentional?

plugins/outputs/cloudwatchlogs/internal/pusher/retryheap_recovery_test.go
```diff
@@ -0,0 +1,279 @@
...
+	// Verify retry metadata was preserved
+	assert.Equal(t, 2, batch.retryCountShort, "Retry count should be preserved across restart")
+	assert.False(t, batch.startTime.IsZero(), "Start time should be preserved across restart")
+
+	mockService.AssertExpectations(t)
+}
+
+// TestRecoveryWithMultipleTargets validates that when one target has permission
+// issues, other healthy targets continue publishing successfully.
+func TestRecoveryWithMultipleTargets(t *testing.T) {
```
jefchien: What is this testing differently than `TestSingleDeniedLogGroup`? Don't see which part of this is recovery.

plugins/outputs/cloudwatchlogs/internal/pusher/state_callback_test.go
```diff
@@ -0,0 +1,188 @@
...
+		r:     state.NewRange(0, 100),
+		queue: queue,
+	})
+	batch.append(evt)
+	return batch
+}
+
+// TestRetryHeapSuccessCallsStateCallback verifies that when a batch succeeds
+// on retry through the heap, state callbacks fire to persist file offsets.
+func TestRetryHeapSuccessCallsStateCallback(t *testing.T) {
```
jefchien: This one and `TestRetryHeapProcessorSendsBatch` are pretty similar and a bit redundant.

plugins/outputs/cloudwatchlogs/internal/pusher/retryheap_expiry_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/aws/amazon-cloudwatch-agent/sdk/service/cloudwatchlogs"
+	"github.com/aws/amazon-cloudwatch-agent/tool/testutil"
+)
+
+// TestRetryHeapProcessorExpiredBatchShouldResume verifies that expired batches
+// resume the circuit breaker, preventing the target from being permanently blocked.
+func TestRetryHeapProcessorExpiredBatchShouldResume(t *testing.T) {
```
jefchien: This, `TestRetryHeapProcessorExpiredBatch`, and `TestRetryHeapExpiryCallsStateCallback` are pretty similar as well. Could have all of the assertions in a single test. Would help improve maintainability and test run time.

jefchien (approved):
I think the tests could be improved, but functionally, it looks good to me.

PR #2023, opened 2026-02-13
feat(pusher): Add concurrent log publishing with circuit breaker and retry heap

the-mann:
## Summary

Concurrent CloudWatch Logs publishing: shared worker pool + per-target circuit breaker + unbounded retry heap, so backoff never holds a worker.

**Problem.** Retry backoff sleeps in the send path. Since #1499 made the pool *shared*, a sleeping sender holds a worker — N failing targets starve healthy groups. Opt-in (`concurrency > 1`, default 0).

```
Queue → SenderPool → WorkerPool (N) → Sender → PutLogEvents
                                        ↓ fail
                            RetryHeap → RetryHeapProcessor → WorkerPool
```

The heap is unbounded deliberately: a bounded one (size = concurrency) deadlocked once failing groups outnumbered slots, as workers blocked in `Push()`.

**Batch finalizers.** Every exit path must clear the breaker *and* be explicit about delivery, or a target latches off / offsets advance past unsent events:

| | Breaker | Offsets | Delivered |
|---|---|---|---|
| `done()` — success | resume | advance | yes |
| `drop()` — terminal error / 14-day expiry | resume | advance | **no** |
| `abandon()` — heap closed at shutdown | resume | untouched | no; re-read after restart |

## Fixed in review

- **Data loss** — a panic mid-send and a task submitted just after `Stop()` left batches unfinalized (breaker halted, offsets frozen); both finalize.
- **EMF header dropped on retry** (raised by @lorespiz; severity corrected by @jefchien) — `x-amzn-logs-format` is per-client state, so a batch retried through the processor's client went out shaped unlike its origin. Fix pins the originating client, correct for any per-destination client state, the header included.
- `waitIfHalted` now loops, re-checking the halted flag each wakeup, so a halt re-engaged after `resume()` recreates the halt channel no longer lets a send onto a halted queue (@jefchien; mutation-verified regression test).
- Retry-heap processor reuses the shared client instead of a self-built client and `LogThrottleRetryer` — reachable only as the `batch.service == nil` fallback, which never fires since heap batches pin to their origin before the first failing send; also drops an idle retryer goroutine (@jefchien).
- **Deadlocks** — four real hangs: `Publish` holding the destination lock across a blocked `AddEvent`, the same on the EMF path, `Submit` holding a read lock against `Stop`, and `RetryHeapProcessor.Stop()` parked in `Submit`; plus a leaked flush-timer goroutine.
- **Correctness** — breaker latching on terminal errors; false delivery on drop paths; a callback panic skipping `resume()`; a worker panic killing the agent; a heap-drain panic losing queued batches.
- **Deliberately unbounded**: `waitIfHalted`, `workerPool.Stop`, `pusherWaitGroup.Wait`. Bounding was tried and reverted — timing out `waitIfHalted` removes the backpressure capping a failing target at ~1 batch in flight. Liveness comes from `stopCh`; documented at each site.

## Testing

- `-race` green on both packages; `golangci-lint --new-from-rev` clean; tests grouped by behaviour and round-tripped against a broken variant.
- Real AL2023 host: 90 PASS / 0 FAIL / 0 DATA RACE in pusher, new fixes mutation-verified.
- **Live against real CloudWatch Logs** — identical config and load, one group denied `PutLogEvents` plus one healthy, concurrency 8:

| Build | Healthy-group delivery |
|---|---|
| `main` | ~7.6% of generated events |
| this PR | ~99.99% |

- Also live-verified: the bounded-heap deadlock, 10-file fan-out (no loss, bounded memory), mid-run permission restore, synchronous default (0/1), and EMF A/B (no leak onto a plain group).
- Throttling fix in `main` re-verified; a rebase-exposed nil panic fixed with a test.
- **Gaps:** the flush-timer fix has no test — mine passed with the bug reinstated, so I removed it rather than imply coverage; and `workerPool.Stop` awaiting in-flight sends can't be asserted without the bounding rejected above.

## Known limitations

- The breaker is a single `bool`, so `halt`/`resume` aren't 1:1 — the first success resumes a target whose other in-flight batches are still failing. A counter is a redesign, not a bug fix.
- Heap unbounded under mass failure; memory bounded only indirectly, by the breaker.
- Batches for an already-stopped destination linger to the 14-day expiry.

Related: #1975, #1977, #2022.


the-mann:
Incorporated review feedback. Integration tests passing:
- Full suite: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/22458358617
- Concurrency tests (recovery-test branch): https://github.com/aws/amazon-cloudwatch-agent/actions/runs/22458610567

the-mann:
Updated PR with review feedback and merged latest from main. Commit to review: 9aabc042

Integration tests passing:
- Full suite: https://github.com/aws/amazon-cloudwatch-agent/actions/runs/22458358617
- Concurrency tests (recovery-test branch): https://github.com/aws/amazon-cloudwatch-agent/actions/runs/22458610567

plugins/outputs/cloudwatchlogs/internal/pusher/retryheap.go
```diff
@@ -0,0 +1,222 @@
...
+// processLoop runs the main processing loop
+func (p *RetryHeapProcessor) processLoop() {
+	defer p.wg.Done()
+	ticker := time.NewTicker(100 * time.Millisecond)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			p.processReadyMessages()
```
olowosulu, line 219: If this panics, the goroutine dies.
Can we wrap it with a deferred recover that logs and continues?

plugins/outputs/cloudwatchlogs/internal/pusher/batch.go
```diff
@@ -196,6 +218,49 @@ func (b *logEventBatch) updateState() {
...
+		callback := b.resumeCallbacks[i]
+		callback()
+	}
+}
+
+// drop finalizes a batch that will never be delivered after a permanent failure. It
+// persists state so the events are not reprocessed on restart and clears the circuit
+// breaker, but deliberately does NOT run doneCallbacks: those signal successful
+// delivery (per-event LogEvent.Done plus the queue's success bookkeeping).
+func (b *logEventBatch) drop() {
```
olowosulu, line 257: Can we add an atomic counter that increments inside `drop()` and expose it as a named metric?
This would allow users build alarms on data loss rate without needing to parse logs.

olowosulu (dismissed):

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -158,15 +174,23 @@ func (c *CloudWatchLogs) getDest(t pusher.Target, logSrc logs.LogSrc) *cwDest {
...
 		c.sharedRetryer = retryer.NewLogThrottleRetryer(c.Log)
 		c.sharedClient = c.createClient(c.sharedRetryer)
 		c.targetManager = pusher.NewTargetManager(c.Log, c.sharedClient)
+
+		if c.Concurrency > 1 {
+			c.workerPool = pusher.NewWorkerPool(c.Concurrency)
+			c.retryHeap = pusher.NewRetryHeap(c.Log)
+
+			retryHeapProcessorRetryer := retryer.NewLogThrottleRetryer(c.Log)
+			retryHeapProcessorClient := c.createClient(retryHeapProcessorRetryer)
```
lorespiz: Retried batches go out through this client, but EMF handling is client state: switchToEMF() pushes the `x-amzn-logs-format: json/emf` header onto the originating dest's client only. So an EMF batch that fails and gets retried via `RetryHeapProcessor` is re-sent without the EMF header, and CloudWatch Logs will ingest it as plain log events with no metric extraction

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
```diff
@@ -245,3 +264,36 @@ func hasValidTime(e logs.LogEvent) bool {
...
+		q.haltMu.Unlock()
+		return
+	}
+	ch := q.haltCh
+	q.haltMu.Unlock()
+	select {
+	case <-ch:
+	case <-q.stopCh:
+	}
+}
```
jefchien, line 310: nit: Should this be in a for loop so if `q.halted` is re-engaged after the resume, it will recheck and catch it?
```suggestion
func (q *queue) waitIfHalted() {
    for {
  		q.haltMu.Lock()
  		if !q.halted {
  			q.haltMu.Unlock()
  			return
  		}
  		ch := q.haltCh
  		q.haltMu.Unlock()
  		select {
  		case <-ch:
  		case <-q.stopCh:
  			return
  		}
	}
}
```

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
movence: Good catch! Batches now pin the client that first sent them (batch.service), so retries reuse the originating destination's handlers; `TestEMFHeaderPreservedOnRetriedBatch` and `TestEMFHeaderNotLeakedToOtherDestinationsOnRetry` cover both directions.

jefchien: nit: `x-amzn-logs-format: json/emf` hasn't been required for a while now, so it's not as critical of an issue as no metric extraction, but it's still a fair callout. A request going out from the retry heap would have a different shape than the original request.

https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch_Embedded_Metric_Format_Generation_PutLogEvents.html
> When calling PutLogEvents, you have the option to include the following HTTP header, which tells CloudWatch Logs the metrics should be extracted, but it's not required. 
> ```
> x-amzn-logs-format: json/emf
> ```

jefchien: If it's pinning the client, then what is this client used for?

lorespiz (dismissed):

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
movence: Nothing. it only reached the `batch.service == nil` fallback, which never fires since heap batches are pinned before their first failing send. Removed it; the processor reuses `c.sharedClient`, which also drops a LogThrottleRetryer goroutine started for nothing.

plugins/outputs/cloudwatchlogs/internal/pusher/batch.go
```diff
@@ -5,19 +5,35 @@ package pusher
...
+// droppedLogEvents is the cumulative count of log events dropped without delivery after a
+// terminal send failure or retry-timeout expiry. It is incremented in drop() and can be
+// read via DroppedLogEvents() so data loss can be tracked without scraping per-batch logs.
+var droppedLogEvents atomic.Int64
+
+// DroppedLogEvents returns the cumulative number of log events permanently dropped
+// (never delivered) across all targets since process start.
+func DroppedLogEvents() int64 {
+	return droppedLogEvents.Load()
+}
```
jefchien: nit: It doesn't look like this is used outside of tests within this package. The `profiler` is already used to track and log this. Do we need it?

plugins/outputs/cloudwatchlogs/internal/pusher/batch.go
```diff
@@ -226,3 +317,44 @@ func (t byTimestamp) Swap(i, j int) {
 func (t byTimestamp) Less(i, j int) bool {
 	return *t[i].Timestamp < *t[j].Timestamp
 }
+
+// initializeStartTime sets the start time and expiration time if not already set.
+func (b *logEventBatch) initializeStartTime() {
+	if b.startTime.IsZero() {
+		b.startTime = time.Now()
+		b.expireAfter = b.startTime.Add(maxRetryTimeout)
```
jefchien, line 311: If the goal is to drop expired events before making PLE calls, shouldn't `expireAfter` be based on the oldest timestamp in the batch instead of the current time?

plugins/outputs/cloudwatchlogs/internal/pusher/batch.go
```diff
@@ -226,3 +317,44 @@ func (t byTimestamp) Swap(i, j int) {
...
+	return !b.expireAfter.IsZero() && time.Now().After(b.expireAfter)
+}
+
+// isReadyForRetry checks if enough time has passed since the last failure to retry this batch.
+func (b *logEventBatch) isReadyForRetry() bool {
+	if b.nextRetryTime.IsZero() {
+		return true
+	}
+	return time.Now().After(b.nextRetryTime)
+}
```
jefchien: nit: Seems like `isReadyForRetry` isn't used anywhere but in testing. Consider either moving it into the test file or removing it and inlining the check.

plugins/outputs/cloudwatchlogs/internal/pusher/batch.go
```diff
@@ -226,3 +317,44 @@ func (t byTimestamp) Swap(i, j int) {
...
+		b.startTime = time.Now()
+		b.expireAfter = b.startTime.Add(maxRetryTimeout)
+	}
+}
+
+// updateRetryMetadata updates the retry metadata after a failed send attempt.
+// It increments the appropriate retry counter based on the error type and calculates the next retry time.
+func (b *logEventBatch) updateRetryMetadata(err error) {
+	// Store the error
+	b.lastError = err
```
jefchien: nit: We store this error, but the only place it's referenced is a single test. Do we really need it?

plugins/outputs/cloudwatchlogs/internal/pusher/batch_test.go
```diff
@@ -404,3 +407,185 @@ func TestValidateAndTruncateMessage(t *testing.T) {
...
+	assert.Equal(t, firstStartTime, batch.startTime, "startTime should not change on third call")
+	assert.Equal(t, firstExpireAfter, batch.expireAfter, "expireAfter should not change on third call")
+}
+
+// A batch dropped because the retry heap is already stopped (shutdown in
+// progress) must not fire done callbacks. done() runs updateState() AND every
+// doneCallback -- per-event LogEvent.Done() plus the queue's onSuccessCallback -- which
+// marks never-delivered events as delivered and advances file offsets past them, so the
+// events are neither shipped nor re-read after restart.
+func TestHeapStoppedDropDoesNotSignalSuccess(t *testing.T) {
```
jefchien: nit: This seems like it should be in `retryheap_test.go` instead of `batch_test.go`.

plugins/outputs/cloudwatchlogs/internal/pusher/batch_test.go
```diff
@@ -404,3 +407,185 @@ func TestValidateAndTruncateMessage(t *testing.T) {
...
+	assert.Zero(t, delivered.Load(),
+		"undelivered batch dropped at shutdown fired done callbacks: events are reported delivered")
+	assert.Zero(t, offsetsAdvanced.Load(),
+		"undelivered batch dropped at shutdown advanced file offsets: events are lost across restart")
+}
+
+// Second site: the same false-success signal on the expired-batch drop
+// inside flushReadyBatches. Expiry is a permanent give-up, so offsets SHOULD advance --
+// but done callbacks must not fire, because nothing was delivered.
+func TestExpiredBatchDropDoesNotSignalSuccess(t *testing.T) {
```
jefchien: nit: This seems like it should be in `retryheap_test.go` instead `batch_test.go`.

plugins/outputs/cloudwatchlogs/internal/pusher/batch_test.go
```diff
@@ -404,3 +407,185 @@ func TestValidateAndTruncateMessage(t *testing.T) {
...
+	b := newLogEventBatch(Target{Group: "g", Stream: "s"}, nil)
+	b.append(newLogEvent(time.Now(), "a", func() {}))
+	b.append(newLogEvent(time.Now(), "b", func() {}))
+	b.drop()
+
+	require.Equal(t, before+2, DroppedLogEvents(),
+		"drop() must add every undelivered event in the batch to the dropped counter")
+}
+
+// N5: a permanently invalid batch must advance offsets so it is not re-read forever (its own
```
jefchien: nit: What is N5? I see it and other labels like N3, N4, P8, etc. in comments throughout the PR.

plugins/outputs/cloudwatchlogs/internal/pusher/circuitbreaker_test.go
```diff
@@ -0,0 +1,287 @@
...
+		healthyPusher.AddEvent(newStubLogEvent("ok-more", now))
+	}
+
+	time.Sleep(500 * time.Millisecond)
+
+	// Circuit breaker assertion: after the first failure, the failing target should
+	// NOT have sent additional batches. Only 1 send attempt should have been made
+	// before the circuit breaker blocks it.
+	assert.Equal(t, int32(1), failingTargetSendCount.Load(),
+		"Circuit breaker should block failing target after exactly 1 send attempt")
```
jefchien: nit: Instead of a `time.Sleep`, consider using `assert.Never` or `require.Never`. Can lead to a less flaky test when dealing with clocks.

```suggestion
	// Circuit breaker assertion: after the first failure, the failing target should
	// NOT have sent additional batches. Only 1 send attempt should have been made
	// before the circuit breaker blocks it.
	require.Never(t, func() bool {
         return failingTargetSendCount.Load() > 1
	}, 500*time.Millisecond, 50*time.Millisecond, "Circuit breaker should block failing target after exactly 1 send attempt")
```

plugins/outputs/cloudwatchlogs/internal/pusher/poison_pill_test.go
```diff
@@ -0,0 +1,315 @@
...
+	err = heap.Push(deniedBatch)
+	assert.NoError(t, err)
+
+	processor.processReadyMessages()
+	time.Sleep(100 * time.Millisecond)
+
+	// Verify allowed log group received events
+	assert.Greater(t, allowedGroupSuccessCount.Load(), int32(0),
+		"Allowed log group must receive events with single denied log group")
+}
```
jefchien: nit: Consider using `assert.Eventually` instead of `time.Sleep`.

plugins/outputs/cloudwatchlogs/internal/pusher/poison_pill_test.go
```diff
@@ -0,0 +1,315 @@
...
+	for _, ps := range pushers {
+		for i := 0; i < 10; i++ {
+			ps.AddEvent(newStubLogEvent("fail", time.Now()))
+		}
+	}
+	require.Eventually(t, func() bool { return calls.Load() > 0 }, 10*time.Second, 50*time.Millisecond)
+	time.Sleep(2 * time.Second)
+
+	// breaker caps each failing target at roughly one in-flight batch
+	assert.LessOrEqual(t, retryHeap.Size(), 24, "heap grew beyond one batch per target: %d", retryHeap.Size())
```
jefchien: nit: Why is this 24? 4 workers x 6 pushers?

plugins/outputs/cloudwatchlogs/internal/pusher/pool.go
```diff
@@ -47,24 +52,53 @@ func (p *workerPool) worker() {
...
+		}
 	}
 }
 
+// runTask isolates a panic to the one task that raised it. Sends run here, so without
+// this an unrecovered panic in a worker goroutine terminates the whole agent process.
+func runTask(task func()) {
+	defer func() {
+		if r := recover(); r != nil {
+			log.Printf("E! [cloudwatchlogs] Recovered from panic in worker task: %v", r)
```
jefchien: nit: Noticed that this is the only place that uses `log.Printf` instead of the logger. Technically, this will still work since the default log gets set up to write to the same file/sink, but it stands out a bit.

plugins/outputs/cloudwatchlogs/internal/pusher/pool.go
```diff
@@ -47,24 +52,53 @@ func (p *workerPool) worker() {
...
+// Stop no longer closes p.tasks, only stopCh.
+func (p *workerPool) Submit(task func()) bool {
+	// Check stopping first: select is random when both cases are ready, so without this a
+	// task could be enqueued after Stop and never run, leaving its batch unfinalized.
+	if p.stopping.Load() {
+		return false
+	}
 	select {
+	case p.tasks <- task:
+		return true
```
jefchien, line 100: Do we need to check `p.stopping.Load()` again here?
```suggestion
        if p.stopping.Load() {
            select {
            case <-p.tasks: // remove the task from the queue since we're stopping
            default:
            }
            return false
        }
        return true
```

plugins/outputs/cloudwatchlogs/internal/pusher/pool.go
```diff
@@ -74,52 +108,66 @@ func (p *workerPool) WorkerCount() int32 {
 }
 
 // Stop closes the channels and waits for the workers to stop.
+// Stop signals the workers and waits for them to drain and exit. p.tasks is deliberately
```
jefchien, line 111: nit: Looks like this was meant to replace the previous comment, but it appended to it instead.

plugins/outputs/cloudwatchlogs/internal/pusher/pool_test.go
```diff
@@ -134,3 +132,141 @@ func TestSenderPool(t *testing.T) {
...
+	// A later task must still execute: the worker has to survive the panic above.
+	pool.Submit(func() { ran.Add(1) })
+
+	deadline := time.Now().Add(10 * time.Second)
+	for time.Now().Before(deadline) && ran.Load() == 0 {
+		time.Sleep(20 * time.Millisecond)
+	}
+	require.Equal(t, 1, int(ran.Load()),
+		"a panicking task must not kill the worker; without a recover in worker() the "+
+			"panic is unrecovered and the agent process exits")
```
jefchien: nit: Consider using `require.Eventually` here as well.

plugins/outputs/cloudwatchlogs/internal/pusher/pool_test.go
```diff
@@ -134,3 +132,141 @@ func TestSenderPool(t *testing.T) {
 	s.Stop()
 	assert.Equal(t, int32(200), completed.Load())
 }
+
+// N4: the worker pool must behave across the concurrency values operators actually set,
+// including more workers than CPUs.
+func TestWorkerPoolConcurrencyMatrix(t *testing.T) {
+	for _, size := range []int{1, 2, 8, runtime.NumCPU() * 4} {
```
jefchien, line 140: nit: If this is a matrix test, consider using `t.Run` within the loop so they all run even if one fails.

plugins/outputs/cloudwatchlogs/internal/pusher/pool_test.go
```diff
@@ -134,3 +132,141 @@ func TestSenderPool(t *testing.T) {
...
+	}
+	require.Equal(t, 1, int(resumed.Load()),
+		"a panic mid-send must still clear the target's circuit breaker via abandon()")
+	require.Zero(t, stateRuns.Load(),
+		"abandon() must NOT persist offsets: the events were never delivered")
+}
+
+// Submit holds stopLock.RLock while blocking on the task channel, and
+// Stop needs stopLock.Lock to close stopCh. The write lock waits for that reader, so Stop
+// can never release the blocked Submit -- the pool deadlocks on a saturated shutdown.
```
jefchien: nit: This seems outdated. `stopLock` was removed.

plugins/outputs/cloudwatchlogs/internal/pusher/pusher_test.go
```diff
@@ -124,3 +124,38 @@ func setupPusher(t *testing.T, workerPool WorkerPool, wg *sync.WaitGroup) *Pushe
...
+		time.Second,
+		&wg,
+		retryHeap,
+	)
+
+	assert.NotNil(t, pusher)
+	assert.Equal(t, target, pusher.Target)
+
+	// Verify pusher has retryHeap when concurrency enabled
+	// (RetryHeap is now passed to the underlying sender, not senderPool)
```
jefchien: There's no assertion that does what this is claiming.

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
```diff
@@ -109,12 +128,13 @@ func (q *queue) AddEventNonBlocking(e logs.LogEvent) {
...
+// cwDest.signalStop) and again on the normal Stop path.
 func (q *queue) Stop() {
-	if q.stopped {
-		return
-	}
-	close(q.stopCh)
-	q.stopped = true
+	q.stopOnce.Do(func() {
+		close(q.stopCh)
+		q.stopped = true
```
jefchien: It doesn't look like this is used anywhere. Do we still need to set this or have this field?

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
```diff
@@ -175,6 +195,17 @@ func (q *queue) merge(mergeChan chan logs.LogEvent) {
...
 	if len(q.batch.events) > 0 {
 		q.batch.addDoneCallback(q.onSuccessCallback(q.batch.bufferedSize))
+		q.batch.addFailCallback(q.halt)
+		// Lets a permanently-failed or abandoned batch clear the breaker without being
+		// reported as delivered. Without this, halt() latches until the next done().
+		q.batch.addResumeCallback(q.resume)
+
+		// In synchronous mode (no retryHeap), halt() is never called because
+		// sender only calls batch.fail() when retryHeap != nil. So waitIfHalted
+		// is a no-op. The lock acquisition is negligible overhead (~20ns) on
```
jefchien: nit: How did we measure the overhead?

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
```diff
@@ -93,7 +106,13 @@ func (q *queue) AddEventNonBlocking(e logs.LogEvent) {
...
+		// Stop-aware: startNonBlockCh is unbuffered and its only receiver is the merge loop,
+		// which exits on stopCh. Without this, the first EMF event after shutdown blocks
+		// forever while cwDest.Publish holds the destination lock.
+		select {
+		case q.startNonBlockCh <- struct{}{}: // let the select loop pick up the merged channel
+		case <-q.stopCh:
+		}
 	})
 
 	// Drain the channel until new event can be added
```
jefchien, line 117: Do we need to have a stopCh check here as well?

plugins/outputs/cloudwatchlogs/internal/pusher/queue_test.go
```diff
@@ -912,3 +887,119 @@ func TestForceFlushTimer_SingleLowVolumeEventFlushes(t *testing.T) {
...
+// Without the stopCh select in waitIfHalted, this would deadlock.
+func TestQueueStopWhileHalted(t *testing.T) {
+	logger := testutil.NewNopLogger()
+
+	mockSender := &mockSender{}
+	mockSender.On("Send", mock.Anything).Run(func(args mock.Arguments) {
+		batch := args.Get(0).(*logEventBatch)
+		batch.fail() // Halt the queue
+	}).Return()
+	mockSender.On("Stop").Return()
```
jefchien: nit: This doesn't do anything. `queue.Stop` doesn't call `sender.Stop`.

plugins/outputs/cloudwatchlogs/internal/pusher/retryheap.go
```diff
@@ -0,0 +1,286 @@
...
+
+// Stop stops the retry heap
+func (rh *retryHeap) Stop() {
+	rh.mutex.Lock()
+	defer rh.mutex.Unlock()
+
+	if rh.stopped {
+		return
+	}
+	close(rh.stopCh)
```
jefchien: What is the point of this `stopCh`? It looks like it's just initialized and then closed. Is it used anywhere?

plugins/outputs/cloudwatchlogs/internal/pusher/retryheap.go
```diff
@@ -0,0 +1,286 @@
...
+type RetryHeap interface {
+	Push(batch *logEventBatch) error
+	PopReady() []*logEventBatch
+	Size() int
+	Stop()
+}
+
+type retryHeap struct {
+	heap    retryHeapImpl
+	mutex   sync.RWMutex
```
jefchien: nit: Doesn't need to be a `sync.RWMutex`. A `sync.Mutex` would work just fine. It seems like the only place the read lock is used is the `Size()` method, which is only called by tests?

plugins/outputs/cloudwatchlogs/internal/pusher/retryheap.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+	close(rh.stopCh)
+	rh.stopped = true
+}
+
+// RetryHeapProcessor manages the retry heap and moves ready batches back to sender queue
+type RetryHeapProcessor struct {
+	retryHeap  RetryHeap
+	senderPool Sender
+	retryer    *retryer.LogThrottleRetryer
```
jefchien: Do we still need to pass in a retryer? It seems like in the only cases where we call `NewRetryHeapProcessor` (in non-test code), it's nil.

plugins/outputs/cloudwatchlogs/internal/pusher/retryheap.go
```diff
@@ -0,0 +1,286 @@
...
+	old[n-1] = nil // don't stop the GC from reclaiming the item eventually
+	*h = old[0 : n-1]
+	return item
+}
+
+// stopFlushTimeout bounds the shutdown flush so a saturated worker pool cannot hang Stop.
+const stopFlushTimeout = 5 * time.Second
+
+// stopWaitTimeout bounds the wait for the process loop, which may be parked in Submit.
+const stopWaitTimeout = 5 * time.Second
```
jefchien, line 46: nit: Any reason why they're both 5s, but defined separately? Would we tune them separately?

plugins/outputs/cloudwatchlogs/internal/pusher/retryheap.go
```diff
@@ -0,0 +1,286 @@
...
+	retryer    *retryer.LogThrottleRetryer
+	stopCh     chan struct{}
+	logger     telegraf.Logger
+	stopped    atomic.Bool
+	stopOnce   sync.Once
+	wg         sync.WaitGroup
+}
+
+// NewRetryHeapProcessor creates a new retry heap processor
+func NewRetryHeapProcessor(retryHeap RetryHeap, workerPool WorkerPool, service cloudWatchLogsService, targetManager TargetManager, logger telegraf.Logger, retryer *retryer.LogThrottleRetryer) *RetryHeapProcessor {
```
jefchien: nit: Does this constructor or the struct need to be exported? It uses unexported structs in its methods, so it won't work if used outside of this package.

plugins/outputs/cloudwatchlogs/internal/pusher/retryheap.go
```diff
@@ -0,0 +1,286 @@
...
+	p.stopOnce.Do(func() {
+		// Best-effort flush of remaining ready batches, but bounded: the flush submits into
+		// the worker pool, which blocks while the pool is saturated. The pool is stopped
+		// after this processor, and Submit falls through on the pool's stopCh, so a flush
+		// still in flight unblocks there instead of hanging shutdown here.
+		flushed := make(chan struct{})
+		go func() {
+			defer close(flushed)
+			p.flushReadyBatches()
+		}()
```
jefchien, line 182: Should we be tracking this goroutine with the wait group as well?

plugins/outputs/cloudwatchlogs/internal/pusher/retryheap.go
```diff
@@ -0,0 +1,286 @@
...
+
+// Size returns the current number of batches in the heap
+func (rh *retryHeap) Size() int {
+	rh.mutex.RLock()
+	defer rh.mutex.RUnlock()
+	return len(rh.heap)
+}
+
+// Stop stops the retry heap
+func (rh *retryHeap) Stop() {
```
jefchien: nit: Consider renaming this method. `Stop` implies that there's a running goroutine or something to shutdown. The retryHeap doesn't have anything running in the background. This is more of a `Close` to me.

plugins/outputs/cloudwatchlogs/internal/pusher/retryheap_expiry_test.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/aws/amazon-cloudwatch-agent/sdk/service/cloudwatchlogs"
+	"github.com/aws/amazon-cloudwatch-agent/tool/testutil"
+)
+
+// TestRetryHeapProcessorExpiredBatchShouldResume verifies that expired batches
+// resume the circuit breaker, preventing the target from being permanently blocked.
+func TestRetryHeapProcessorExpiredBatchShouldResume(t *testing.T) {
+	logger := testutil.NewNopLogger()
+
+	var sendAttempts atomic.Int32
```
jefchien: nit: This is unused.

plugins/outputs/cloudwatchlogs/internal/pusher/retryheap_expiry_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/aws/amazon-cloudwatch-agent/sdk/service/cloudwatchlogs"
+	"github.com/aws/amazon-cloudwatch-agent/tool/testutil"
+)
+
+// TestRetryHeapProcessorExpiredBatchShouldResume verifies that expired batches
+// resume the circuit breaker, preventing the target from being permanently blocked.
+func TestRetryHeapProcessorExpiredBatchShouldResume(t *testing.T) {
```
jefchien: nit: A lot of overlap with `TestRetryHeapProcessorExpiredBatch`. Consider combining.

plugins/outputs/cloudwatchlogs/internal/pusher/retryheap_test.go
```diff
@@ -0,0 +1,512 @@
...
+	batch3 := newLogEventBatch(target, nil)
+	batch3.nextRetryTime = now.Add(20 * time.Millisecond)
+
+	// Push in random order
+	heap.Push(batch1)
+	heap.Push(batch2)
+	heap.Push(batch3)
+
+	// Wait for all to be ready
+	time.Sleep(100 * time.Millisecond)
```
jefchien: nit: Can't we just set the `nextRetryTime` in the past instead of sleeping?

plugins/outputs/cloudwatchlogs/internal/pusher/retryheap_test.go
```diff
@@ -0,0 +1,512 @@
...
+	})
+
+	// Verify the stopped processor didn't process the new batch
+	assert.Equal(t, 1, heap.Size())
+
+	// Verify processor is marked as stopped
+	assert.True(t, processor.stopped.Load())
+}
+
+// RetryHeapProcessor.Stop() takes stopMu with a deferred Unlock and holds it
```
jefchien: nit: `stopMu` no longer exists.

plugins/outputs/cloudwatchlogs/internal/pusher/sender.go
```diff
@@ -104,54 +110,63 @@ func (s *sender) Send(batch *logEventBatch) {
...
-func (s *sender) SetRetryDuration(retryDuration time.Duration) {
-	s.retryDuration.Store(retryDuration)
-}
-
-// RetryDuration returns the current maximum retry duration.
-func (s *sender) RetryDuration() time.Duration {
-	return s.retryDuration.Load().(time.Duration)
+	s.stopOnce.Do(func() {
+		close(s.stopCh)
+		s.stopped = true
```
jefchien: It doesn't look like this is used. Can we get rid of it?

plugins/outputs/cloudwatchlogs/internal/pusher/sender.go
```diff
@@ -104,54 +110,63 @@ func (s *sender) Send(batch *logEventBatch) {
...
+			wait = 0
+		}
+
+		s.logger.Warnf("Retried %v time, going to sleep %v before retrying.", totalRetries, wait)
 
 		select {
 		case <-s.stopCh:
-			s.logger.Errorf("Stop requested after %v retries to %v/%v failed for PutLogEvents, request dropped.", retryCountShort+retryCountLong-1, batch.Group, batch.Stream)
-			batch.updateState()
+			// drop(), not abandon(): #1789 deliberately persists state here so the batch is
```
jefchien: What is `#1789`? The PR? Would prefer if we didn't reference PRs and kept the context in the comment.

plugins/outputs/cloudwatchlogs/internal/pusher/sender.go
jefchien: nit: Outdated. `RetryDuration` doesn't exist anymore.

plugins/outputs/cloudwatchlogs/internal/pusher/state_callback_test.go
```diff
@@ -0,0 +1,188 @@
...
+func newStatefulBatch(target Target, queue *mockFileRangeQueue) *logEventBatch {
+	batch := newLogEventBatch(target, nil)
+	now := time.Now()
+	evt := newStatefulLogEvent(now, "test", nil, &logEventState{
+		r:     state.NewRange(0, 100),
+		queue: queue,
+	})
+	batch.append(evt)
+	return batch
+}
```
jefchien: nit: Should these get moved to `helpers_test.go`?

plugins/outputs/cloudwatchlogs/internal/pusher/state_callback_test.go
jefchien, line 1: nit: These are retryheap tests. Consider moving them into `retryheap_test.go`.

plugins/outputs/cloudwatchlogs/internal/pusher/retryheap.go
movence: They bound different waits: `stopFlushTimeout` caps the best-effort flush goroutine, `stopWaitTimeout` caps the join on `processLoop`. Same value today but independently tunable, so I would rather keep them separate.

plugins/outputs/cloudwatchlogs/internal/pusher/retryheap.go
movence: Deliberate. The flush submits into the worker pool, which only stops after `Stop` returns, so joining it to the waitgroup could hang shutdown; it is bounded by `stopFlushTimeout` and unblocks via the pool's `stopCh` instead.

plugins/outputs/cloudwatchlogs/internal/pusher/batch.go
movence: `expireAfter` is a retry budget measured from the first send attempt (`startTime + maxRetryTimeout`), not an event-age filter, so anchoring on the oldest event is not needed. Event age is already rejected by `hasValidTime` and by CloudWatch Logs itself.

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
movence: This one cannot block: the channel is buffered and the `default` drains one then retries, so it always makes progress. The unbuffered `startNonBlockCh` send just above does need the `stopCh` case, since its receiver can exit.

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
movence: Fair, that was not measured. I have removed the claim; the comment now just says it is a no-op guard.

lorespiz (approved):

lorespiz (approved):

plugins/outputs/cloudwatchlogs/internal/pusher/sender.go
```diff
@@ -108,54 +112,62 @@ func (s *sender) Send(batch *logEventBatch) {
...
+		if s.retryHeap != nil {
+			if err := s.retryHeap.Push(batch); err != nil {
+				// Heap closed because shutdown is in progress. This is transient, so
+				// state is NOT persisted: the events are re-read after restart rather
+				// than being reported as delivered and silently lost.
+				s.logger.Warnf("RetryHeap stopped, abandoning batch for %v/%v (will be re-read after restart): %v", batch.Group, batch.Stream, err)
+				batch.abandon()
+				return
+			}
+			batch.fail()
```
jefchien: nit: Should `fail` happen before the retryHeap push so the halt is in place before the batch is handed over to the retry heap?
```suggestion
			batch.fail()
			if err := s.retryHeap.Push(batch); err != nil {
				// Heap closed because shutdown is in progress. This is transient, so
				// state is NOT persisted: the events are re-read after restart rather
				// than being reported as delivered and silently lost.
				s.logger.Warnf("RetryHeap stopped, abandoning batch for %v/%v (will be re-read after restart): %v", batch.Group, batch.Stream, err)
				batch.abandon()
				return
			}
```

jefchien (approved):

PR #2032, opened 2026-02-18
Support metadata resolution for multi-cloud envs

movence:
# Description of changes
Adds multi-cloud support for metric decoration and placeholder resolution. Azure (and future cloud providers) 
use OTel's resourcedetectionprocessor for metadata, while EC2 continues using ec2tagger. Disk/volume tagging 
is extracted into a standalone `disktagger` processor that supports both AWS and Azure.

Also include test fixes from https://github.com/aws/amazon-cloudwatch-agent/pull/2010 and minor update to remove an unnecessary wrapper function ([ref](https://github.com/aws/amazon-cloudwatch-agent/pull/2010#discussion_r2825076228)). 

**Cloud metadata (internal/cloudmetadata/)**
- Cloud-agnostic Provider interface with AWS and Azure implementations
- Wraps re-exported metadataproviders from otel-contrib (no custom IMDS code)
- Singleton with auto-detection (AWS ? Azure), used by config-downloader and config-translator
- Replaces ec2util.GetEC2UtilSingleton() in all callers

**Cloud provider enum (internal/cloudprovider/)**
- Shared CloudProvider type (Unknown, AWS, Azure)

**Re-export wrappers (otel-contrib)**
- resourcedetectionprocessor/metadataproviders/{ec2,azure}/ exposes internal packages for external use

**resourcedetection processor translator**
- Added for non-EC2 pipelines (Azure, GCP)
- Reads append_dimensions and enables only requested attributes
- Supports OTel-style placeholders (${host.id}, ${azure.vm.size})

**ec2tagger changes**
- Removed all volume/disk tagging code (moved to disktagger)
- Supports OTel placeholders (${host.id}) alongside legacy (${aws:InstanceId})

**disktagger processor (plugins/processors/disktagger/)**
- Extracted from ec2tagger as standalone processor
- AWS: reuses existing volume package (DescribeVolumes + NVMe serial)
- Azure: IMDS storageProfile (OS + data disks) + /dev/disk/azure/ symlinks with sysfs SCSI fallback
- Prefix matching for partition names (e.g. sda1 matches sda)

**Pipeline translator**
- EC2 mode ? ec2tagger (unchanged)
- Non-EC2 ? resourcedetectionprocessor
- disktagger added when disk VolumeId/DiskId dimension is configured

**Placeholder resolution**
- Runtime placeholders (${aws:VolumeId}, ${disk.id}) stripped during translation
- cloudmetadata.GetProvider() replaces ec2util for metadata

### Config examples

EC2 (backward compatible):
```
{"InstanceId": "${aws:InstanceId}", "VolumeId": "${aws:VolumeId}"}
```

EC2 (OTel style):
```
{"InstanceId": "${host.id}", "VolumeId": "${aws:VolumeId}"}
```

Azure:
```
{"InstanceId": "${host.id}", "Region": "${cloud.region}", "VolumeId": "${disk.id}"}
```

**Known gap**

Attribute renaming for non-EC2 clouds: `resourcedetectionprocessor` adds OTel semantic convention attributes (
e.g. `host.id`, `cloud.region`, `azure.vm.size`) to the Resource. These are not automatically renamed to CloudWatch 
dimension names (e.g. `InstanceId`, `Region`, `VMSize`). The `resource_to_telemetry_conversion` setting copies them to metrics, but with OTel names for CW exporters.

# License

# Tests
Tested on Ubuntu hosts in AWS and Azure by emitting CPU/MEM/DISK metrics
- AWS
Screenshot 2026-02-18 at 4 24 58 PM

- Azure
Screenshot 2026-02-18 at 4 25 05 PM

# Requirements
2. Run `make lint`

-------
### Integration Tests


internal/cloudmetadata/factory.go
```diff
@@ -0,0 +1,64 @@
...
+
+const detectTimeout = 3 * time.Second
+
+// GetProvider returns the singleton cloud metadata provider.
+// It auto-detects the cloud by trying each provider in order.
+// Detection uses a short timeout to avoid blocking agent startup
+// on hosts where IMDS is unreachable.
+func GetProvider() Provider {
+	once.Do(func() {
+		ctx, cancel := context.WithTimeout(context.Background(), detectTimeout)
```
jefchien: The same context is used for both the `aws.NewProvider` and the `azure.NewProvider`. If `aws.NewProvider` takes 3 seconds and times out, then the `azure.NewProvider` will never get run, since the context will already have been cancelled by the time it gets there.

They cannot share contexts.

plugins/processors/disktagger/factory.go
```diff
@@ -0,0 +1,83 @@
...
+		return awsprovider.NewProvider(ec2.NewFromConfig(awsCfg), p.InstanceID())
+	case cloudprovider.Azure:
+		set.Logger.Info("disktagger: using Azure managed disk provider")
+		ap := azureprovider.NewProvider()
+		return newMapProvider(ap.DeviceToDiskID)
+	default:
+		set.Logger.Warn("Unsupported cloud provider for disktagger")
+		return nil
+	}
+}
```
jefchien: It looks like `cloudmetadata.GetProvider()` is only being used here to determine which cloud the agent is running in. Isn't this something that could be determined at translation time and configured on this component?

It looks like all other `cloudmetadata.GetProvider()` usages are at translation time. Don't think it's worth calling and caching the IMDS info just for this switch case at runtime. We would also be able to add a validation check for the config.

plugins/processors/disktagger/factory.go
```diff
@@ -0,0 +1,83 @@
...
+
+func createMetricsProcessor(
+	ctx context.Context,
+	set processor.Settings,
+	cfg component.Config,
+	next consumer.Metrics,
+) (processor.Metrics, error) {
+	c := cfg.(*Config)
+	provider := createDiskProvider(ctx, set)
```
jefchien: nit: This should be created as part of the processor's constructor.

plugins/processors/disktagger/config.go
```diff
@@ -0,0 +1,15 @@
...
+import "time"
+
+const (
+	AttributeDiskID = "VolumeId"
+)
+
+type Config struct {
+	RefreshInterval  time.Duration `mapstructure:"refresh_interval"`
+	DiskDeviceTagKey string        `mapstructure:"disk_device_tag_key"`
+}
```
jefchien, line 29: Can we add godocs for these fields?

plugins/processors/disktagger/internal/volume/volume_test.go
```diff
@@ -36,11 +37,11 @@ func TestCache(t *testing.T) {
 	c.fetchBlockName = func(s string) string {
 		return ""
 	}
-	assert.ErrorIs(t, c.Refresh(), errNoProviders)
+	assert.ErrorIs(t, c.Refresh(context.Background()), errNoProviders)
```
jefchien: nit: Could use `t.Context()`

plugins/processors/disktagger/aws/provider.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"context"
+
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+
+	"github.com/aws/amazon-cloudwatch-agent/plugins/processors/disktagger/internal/volume"
+)
+
+// Provider maps device names to EBS volume IDs.
+type Provider struct {
```
jefchien: nit: This doesn't add anything. It's just a wrapper around the `volume.Cache`. Could extract the `volume.Provider` and rename the package to `aws`.

plugins/processors/disktagger/provider.go
```diff
@@ -0,0 +1,57 @@
...
+package disktagger
+
+import (
+	"context"
+	"strings"
+	"sync"
+)
+
+// DiskProvider maps device names to disk identifiers (volume IDs, managed disk names, etc.)
+type DiskProvider interface {
```
jefchien: nit: This is basically re-defining the `volume.Cache` interface. Could this be simplified?

plugins/processors/disktagger/provider.go
```diff
@@ -0,0 +1,57 @@
...
+	// The context allows cancellation during shutdown.
+	Refresh(ctx context.Context) error
+
+	// Serial returns the disk identifier for a device name.
+	// Supports prefix matching (e.g. "nvme0n1p1" matches "nvme0n1").
+	Serial(devName string) string
+}
+
+// mapProvider wraps a simple map for providers that don't need prefix matching (e.g. Azure).
+type mapProvider struct {
```
jefchien: nit: This is also re-defining the implementation of `volume.cache`. I think these could be simplified. If the `azure.Provider` matched the `volume.Provider` interface, then it could be used directly with the `volume.Cache`.

```go
volume.NewCache(azure.NewProvider())
```

plugins/processors/disktagger/disktagger.go
```diff
@@ -0,0 +1,118 @@
...
+		t.logger.Warn("disktagger: no provider, disk tagging disabled")
+		return nil
+	}
+
+	if err := t.provider.Refresh(ctx); err != nil {
+		t.logger.Warn("Initial disk refresh failed, will retry", zap.Error(err))
+	}
+
+	if t.config.RefreshInterval > 0 {
+		t.ctx, t.cancel = context.WithCancel(context.Background())
```
jefchien: It's a bit strange to store the context on the `Tagger`. Consider just passing it into `t.refreshLoop` instead.

```go
func (t *Tagger) Start(ctx context.Context, _ component.Host) error {
    ...
    if t.config.RefreshInterval > 0 {
        ctx, t.cancel = context.WithCancel(context.Background())
        go t.refreshLoop(ctx)
    }
    return nil
}

func (t *Tagger) refreshLoop(ctx context.Context) {
    ticker := time.NewTicker(t.config.RefreshInterval)
    defer ticker.Stop()
    for {
        select {
        case <-ticker.C:
            if err := t.provider.Refresh(ctx); err != nil {
                t.logger.Warn("Disk refresh failed", zap.Error(err))
            }
        case <-ctx.Done():
            return
        }
    }
}
```

translator/tocwconfig/tocwconfig_test.go
```diff
@@ -899,6 +900,18 @@ func TestIgnoreInvalidAppendDimensions(t *testing.T) {
 	checkTranslation(t, "ignore_append_dimensions", "linux", expectedEnvVars, "")
 }
 
+func TestEC2DisktaggerConfig(t *testing.T) {
```
jefchien: nit: typo
```suggestion
func TestEC2DiskTaggerConfig(t *testing.T) {
```

translator/util/tagutil/tagutil.go
```diff
@@ -82,7 +82,10 @@ func callFuncWithRetries(fn func() (*ec2.DescribeTagsOutput, error), errorMsg st
 var ec2ClientFactory = createDescribeTagsClient
 
 func createDescribeTagsClient(ctx context.Context) (ec2.DescribeTagsAPIClient, error) {
-	region := ec2util.GetEC2UtilSingleton().Region
+	region := ""
+	if p := cloudmetadata.GetProvider(); p != nil {
+		region = p.Region()
```
jefchien, line 88: What if the region we get back is an azure one (like `eastus`)? This isn't going to map to an AWS region name.

translator/translate/otel/processor/disktaggerprocessor/translator.go
jefchien, line 1: Missing unit tests.

translator/util/sdkutil_test.go
```diff
@@ -9,28 +9,46 @@ import (
...
+		"WithEC2Region": {provider: &mockProvider{region: "us-east-1", cloudProvider: cloudprovider.AWS}, wantMode: config.ModeEC2},
+		"WithAzure":     {provider: &mockProvider{region: "westus2", cloudProvider: cloudprovider.Azure}, wantMode: config.ModeOnPrem},
 		"WithECSRegion": {ecsRegion: "us-east-1", wantMode: config.ModeEC2},
 		"WithNone":      {wantMode: config.ModeOnPrem},
 	}
 	for name, testCase := range testCases {
 		t.Run(name, func(t *testing.T) {
 			runInAws = testCase.runInAws
-			DefaultEC2Region = func() string { return testCase.ec2Region }
+			cloudmetadata.SetForTest(testCase.provider)
```
jefchien, line 48: nit: Missing defer `ResetForTest`.

I think we'll have problems if any of these tests run in parallel. Is there a better way to inject the global provider for testing?

translator/translate/otel/processor/disktaggerprocessor/translator.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+func (t *translator) Translate(conf *confmap.Conf) (component.Config, error) {
+	if !IsSet(conf) {
+		return nil, &common.MissingKeyError{ID: t.ID(), JsonKey: diskVolumeKey}
+	}
+
+	cfg := t.factory.CreateDefaultConfig().(*disktagger.Config)
+	cfg.RefreshInterval = 5 * time.Minute
+	cfg.DiskDeviceTagKey = "device"
```
jefchien: Aren't these already the defaults? Why are we setting them again?

translator/translate/otel/processor/resourcedetectionprocessor/translator.go
```diff
@@ -0,0 +1,128 @@
...
+
+var appendDimensionsKey = common.ConfigKey(common.MetricsKey, common.AppendDimensionsKey)
+
+// legacyToOTel maps legacy ${aws:*} placeholders to OTel resource attribute names.
+// Used for backward compatibility when EC2 eventually cuts over to resourcedetection.
+var legacyToOTel = map[string]string{
+	"${aws:InstanceId}":   "host.id",
+	"${aws:InstanceType}": "host.type",
+	"${aws:ImageId}":      "host.image.id",
+	"${aws:Region}":       "cloud.region",
```
jefchien: Is `${aws:Region}` something we currently support?

translator/translate/otel/processor/resourcedetectionprocessor/translator.go
```diff
@@ -0,0 +1,128 @@
...
+	ra.CloudPlatform.Enabled = requested["cloud.platform"]
+	ra.CloudProvider.Enabled = requested["cloud.provider"]
+	ra.CloudRegion.Enabled = requested["cloud.region"]
+	ra.HostID.Enabled = requested["host.id"]
+	ra.HostImageID.Enabled = requested["host.image.id"]
+	ra.HostName.Enabled = requested["host.name"]
+	ra.HostType.Enabled = requested["host.type"]
+}
+
+func configureAzureAttributes(cfg *resourcedetectionprocessor.Config, requested map[string]bool) {
```
jefchien, line 118: Shouldn't this check to see if azure is included in the detectors?

translator/translate/otel/pipeline/host/translator.go
```diff
@@ -91,17 +93,31 @@ func (t translator) Translate(conf *confmap.Conf) (*common.ComponentTranslators,
...
+				translators.Processors.Set(ec2taggerprocessor.NewTranslator())
+				ec2TaggerEnabled = true
+			}
+		} else {
+			// Non-EC2 clouds always need resourcedetection for cloud metadata
+			// (instance ID, type, etc.) used by entity store and resource attribution.
+			// Unlike EC2 where ec2tagger is optional, this is the only source of
+			// cloud metadata on Azure/GCP.
+			log.Printf("D! resourcedetection processor required for cloud metadata")
+			translators.Processors.Set(resourcedetectionprocessor.NewTranslator([]string{"env", "azure", "gcp", "system"}))
```
jefchien, line 108: Should we be including `gcp`? We don't support it anywhere else.

translator/translate/otel/processor/resourcedetectionprocessor/translator.go
```diff
@@ -0,0 +1,128 @@
...
+// It reads append_dimensions from the JSON config and enables only the
+// referenced resource attributes in each detector.
+func (t *translator) Translate(conf *confmap.Conf) (component.Config, error) {
+	cfg := t.factory.CreateDefaultConfig().(*resourcedetectionprocessor.Config)
+	cfg.Detectors = t.detectors
+	cfg.Timeout = 5 * time.Second
+	cfg.Override = false
+
+	requested := collectRequestedAttributes(conf)
+	configureEC2Attributes(cfg, requested)
```
jefchien, line 63: Is this too pre-emptive? We haven't switched the EC2 metadata over to the resourcedetection processor yet.

plugins/processors/disktagger/factory.go
jefchien, line 1: Missing unit tests for factory functions.

plugins/processors/disktagger/azure/provider.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+func (p *Provider) DeviceToDiskID(ctx context.Context) (map[string]string, error) {
+	profile, err := p.fetchStorageProfile(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// Determine which device resolution method to use.
+	useSymlinks := symlinkAvailable()
```
jefchien: Does this need to happen on each refresh?

plugins/processors/disktagger/azure/provider.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+func NewProvider() *Provider {
+	return &Provider{
+		client:   &http.Client{Timeout: 5 * time.Second},
+		endpoint: storageProfileEndpoint,
+	}
+}
+
+func (p *Provider) DeviceToDiskID(ctx context.Context) (map[string]string, error) {
```
jefchien: Missing unit testing. Consider following something similar to `internal/volume/host_linux.go` which allows overrides for the `os` functions.

plugins/processors/disktagger/azure/provider.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+	return false
+}
+
+// resolveSymlink reads a /dev/disk/azure/* symlink and returns the base
+// device name (without partition suffix).
+//
+// Example: /dev/disk/azure/root → ../../sda → "sda"
+func resolveSymlink(path string) string {
+	for _, prefix := range []string{"", "/rootfs"} {
```
jefchien: Can we determine the prefix once (in the constructor perhaps) and store it on the provider? Seems like a waste to try `""` every time if it's never going to be right. Other option is to make it configurable.

PR #2086, opened 2026-04-16
Improve otel-fork-replace workflow: dynamic component discovery, changelog in PR

sky333999:
# Description of changes
## Changes to otel-fork-replace workflow
- Parse go.mod dynamically instead of hardcoding fork components, so the workflow never goes stale when components are added/removed
- Fix shell syntax bug in commit SHA conditional (missing spaces)
- Add changelog table to PR description with commit and PR links
- Run go mod tidy once instead of after every go mod edit
- Add set -euo pipefail for fail-fast error handling
- Use repo-local git config instead of global with fragile cleanup
- Adds a safeguard that aborts if fork components point to multiple distinct SHAs (indicating an intentional pin).
## Changes to PR-test workflow
- Allow a "skip testing" label to be set on the PR to bypass testing and still satisfy GH checks. Should only be used for PRs that do no impact functionality of the agent such as doc updates, workflow file updates etc.

# License

# Tests
Tested the chunks locally.

# Requirements
2. Run `make lint`

-------
### Integration Tests

### DIFF
--- a/.github/workflows/otel-fork-replace.yml
+++ b/.github/workflows/otel-fork-replace.yml
@@ -38,12 +38,17 @@ jobs:
       - name: Get old commit sha from go.mod
         id: get-old-commit
         run: |
+          set -euo pipefail
           # Extract all distinct 12-char hash suffixes from fork pseudo-versions in go.mod.
           # Pseudo-versions look like: v0.0.0-20260407222105-72a988178008
           shas=$(grep 'amazon-contributing/opentelemetry-collector-contrib/' go.mod \
             | grep -oE 'v0\.0\.0-[0-9]+-[0-9a-f]{12}' \
             | awk -F'-' '{print $NF}' \
             | sort -u)
+          if [ -z "$shas" ]; then
+            echo "::error::No fork components found in go.mod. Nothing to update."
+            exit 1
+          fi
           count=$(echo "$shas" | wc -l | tr -d ' ')
           if [ "$count" -ne 1 ]; then
             echo "::error::Found $count distinct fork SHAs in go.mod. A component may be intentionally pinned. Aborting."
--- a/.github/workflows/otel-fork-replace.yml
+++ b/.github/workflows/otel-fork-replace.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Get latest commit sha
         id: get-latest-commit
         run: |
-          if ["${{ inputs.CommitSha }}" == ""]; then
+          if [ "${{ inputs.CommitSha }}" == "" ]; then
             echo "sha=$(git ls-remote ${{ env.UPSTREAM }} ${{ env.UPSTREAM_BRANCH }} | awk '{print $1;}')" >> $GITHUB_OUTPUT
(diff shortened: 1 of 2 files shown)


TravisStark (approved):

.github/workflows/otel-fork-replace.yml
```diff
@@ -35,48 +35,90 @@ jobs:
...
+        env:
+          NEW_SHA: ${{ steps.get-latest-commit.outputs.sha }}
+          OLD_SHA: ${{ steps.get-old-commit.outputs.sha }}
+        run: |
+          set -euo pipefail
+          REPO_URL="https://github.com/amazon-contributing/opentelemetry-collector-contrib"
+          git clone --bare ${{ env.UPSTREAM }} /tmp/otel-fork
+          # old_sha is a 12-char prefix, git log handles prefix matching
+          raw=$(git -C /tmp/otel-fork log --oneline "${OLD_SHA}..${NEW_SHA}" 2>/dev/null || true)
+          rm -rf /tmp/otel-fork
```
jefchien, line 70: nit: Could use `gh` CLI instead of the full clone or use `@action/checkout`.



```suggestion          
          REPO="amazon-contributing/opentelemetry-collector-contrib"
          REPO_URL="https://github.com/${REPO}"

          # GitHub compare API returns commits between two refs
          raw=$(gh api "repos/${REPO}/compare/${OLD_SHA}...${NEW_SHA}" \
             --jq '.commits[] | "\(.sha[:7]) \(.commit.message | split("\n")[0])"' 2>/dev/null || true)
```

.github/workflows/otel-fork-replace.yml
```diff
@@ -35,48 +35,90 @@ jobs:
...
+          {
+            echo "log<<CHANGELOG_EOF"
+            if [ -z "$raw" ]; then
+              echo "Could not generate changelog (old SHA ${OLD_SHA} not found in history)"
+            else
+              echo "| Commit | Description | PR |"
+              echo "| ------ | ----------- | -- |"
+              echo "$raw" | while IFS= read -r line; do
+                sha=$(echo "$line" | awk '{print $1}')
+                msg=$(echo "$line" | cut -d' ' -f2-)
```
jefchien, line 80: nit: Might want to escape `|` in the message so it doesn't break formatting.


```suggestion
                msg=$(echo "$line" | cut -d' ' -f2-)
                msg=$(echo "$msg" | sed 's/|/\\|/g')
```

jefchien (approved):

.github/workflows/otel-fork-replace.yml
sky333999: Fixed in https://github.com/aws/amazon-cloudwatch-agent/pull/2088

.github/workflows/otel-fork-replace.yml
sky333999: Fixed in https://github.com/aws/amazon-cloudwatch-agent/pull/2088

PR #2093, opened 2026-04-17
Add support for configuring EFA receiver on EC2 via JSON

mitali-salvi:
# Description of the issue

Add support for configuring the awsefareceiver via the CWA JSON configuration. Customers will be able to configure EFA (Elastic Fabric Adapter) metrics collection under `metrics.metrics_collected.efa`.

EFA is a distinct hardware type (Elastic Fabric Adapter) that exposes 22 cumulative monotonic sum metrics via `/sys/class/infiniband/` sysfs. The receiver is a native OTel receiver (not adapter-based) that runs in the `hostDeltaMetrics` pipeline with cumulative-to-delta processing.

## Configuration Examples

```json
{
  "metrics": {
    "metrics_collected": {
      "efa": {
        "measurement": ["efa_tx_bytes", "efa_rx_bytes", "efa_rdma_read_bytes"],
        "metrics_collection_interval": 30
      }
    }
  }
}

{
  "metrics": {
    "metrics_collected": {
      "efa": {
        "measurement": ["tx_bytes", "rx_dropped"]
      }
    }
  }
}
```

```json
{
  "metrics": {
    "aggregation_dimensions": [["aws.efa.device"]],
    "metrics_collected": {
      "efa": {
        "measurement": ["efa_tx_bytes", "efa_rx_bytes"],
        "drop_original_metrics": ["efa_tx_bytes", "efa_rx_bytes"]
      }
    }
  }
}

```

# License

# Tests
- Unit tests: 9 test cases all passing (including wildcard and measurement-required error cases)
- Manual: deployed CWA to EC2 instance with EFA devices, generated traffic between two EFA-enabled instances, confirmed agent starts and collects metrics

Screenshot 2026-04-17 at 14 51 41
Screenshot 2026-05-01 at 13 57 06

# Requirements

*Before committing your code, please do the following steps.*
2. Run `make lint`

### Integration Tests


JayPolanco (dismissed):

translator/config/schema.json
```diff
@@ -358,6 +361,20 @@
             }
           ]
         },
+        "efaDefinitions": {
+          "type": "object",
+          "properties": {
+            "metrics_collection_interval": {
+              "$ref": "#/definitions/timeIntervalDefinition"
+            },
+            "append_dimensions": {
```
sky333999: I dont see where this is implemented by the current code. Did we test if this is working?

translator/translate/otel/receiver/awsefa/testdata/agentInterval.json
```diff
@@ -0,0 +1,10 @@
+{
+  "agent": {
+    "metrics_collection_interval": 120
+  },
+  "metrics": {
+    "metrics_collected": {
+      "efa": {}
```
sky333999, line 7: We should be consistent with our existing plugins. For all these host plugins, `measurements` is a required field and we only collect what is specified.

sky333999: Along similar lines, we emit a default `host` dimension` today if InstanceId is not specified so things dont get aggregated up to the account level - I dont see that being implemented here.

translator/translate/otel/receiver/awsefa/translator.go
```diff
@@ -0,0 +1,123 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
sky333999, line 1: Need to add goldel yaml tests

translator/translate/otel/receiver/awsefa/translator.go
```diff
@@ -0,0 +1,123 @@
...
+			return nil, fmt.Errorf("unable to unmarshal efa receiver (%s): %w", t.ID(), err)
+		}
+	}
+
+	return cfg, nil
+}
+
+// allEfaMetrics is the complete list of EFA metrics for disabling unselected ones.
+// Keep in sync with awsefareceiver/internal/metadata.MetricsConfig.
+var allEfaMetrics = map[string]bool{
```
sky333999: This will go stale over time.. can we not hardcode and use reflect or something instead?

translator/translate/otel/receiver/awsefa/translator.go
```diff
@@ -0,0 +1,123 @@
...
+	"efa_impaired_remote_conn_events": true,
```
sky333999: Please double check if the unit being defined for these in the efareceiver is accurately being reflected in CW.

translator/config/schema.json
```diff
@@ -358,6 +361,20 @@
             }
           ]
         },
+        "efaDefinitions": {
+          "type": "object",
+          "properties": {
```
sky333999, line 366: No device/interface filters like with `resources`? Check what our other plugins have to be consistent.

translator/translate/otel/receiver/awsefa/translator.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+func getEnabledMeasurements(measurements []string) map[string]any {
+	// Disable all metrics first.
+	metrics := map[string]any{}
+	for m := range allEfaMetrics {
+		metrics[m] = map[string]any{"enabled": false}
+	}
+	// Enable only the selected ones that are valid EFA metrics.
+	for _, m := range measurements {
```
sky333999, line 105: Does this work if user specifies `*`?

sky333999:
Mind posting a screenshot of what the actual metrics look like in the CW metrics console? Do they have device and interfaceID as dimensions? Without host/InstanceId, are these currently getting aggregated to the account level?

translator/config/schema.json
mitali-salvi: Fixed — removed `append_dimensions` from the schema entirely since it was declared but never implemented in the translator. Per-plugin `append_dimensions` would require explicit translator support for native OTel receivers (adapter-based plugins get it for free via ec2tagger). The host/InstanceId dimension is handled at the pipeline level by the ec2tagger processor via the top-level `metrics.append_dimensions` config, which EFA participates in as part of the `hostDeltaMetrics` pipeline. See commit d3f4db3.

translator/translate/otel/receiver/awsefa/testdata/agentInterval.json
mitali-salvi: The default host/InstanceId dimension is handled at the pipeline level by the `ec2tagger` processor, not per-receiver. Since EFA is registered in the `hostDeltaMetrics` pipeline, it automatically gets the ec2tagger processor applied — same as diskio and net. When users configure `metrics.append_dimensions.InstanceId: \${aws:InstanceId}`, the ec2tagger adds it to all metrics in the pipeline including EFA. No per-receiver change needed here.

translator/translate/otel/receiver/awsefa/translator.go
mitali-salvi: The existing tests already use the golden JSON→YAML pattern (testdata/*.json input → testdata/*.yaml expected output). We now have 9 test cases covering: selective metrics, non-prefixed names, empty measurements, wildcard, measurement-required errors, and interval overrides. Added `wildcardConfig.json/yaml` golden pair. See commit 4e79df7.

translator/translate/otel/receiver/awsefa/translator.go
mitali-salvi: Fixed — replaced the hardcoded `allEfaMetrics` map with `getAllEfaMetrics()` which uses `reflect.TypeOf(cfg.Metrics)` to extract metric names from the `mapstructure` struct tags on `MetricsConfig`. The metric list is now derived from the receiver's own config struct at init time, so it stays in sync automatically when the receiver adds or removes metrics. See commit 4e79df7.

translator/translate/otel/receiver/awsefa/translator.go
mitali-salvi: Verified — the awsefareceiver defines byte metrics with unit `By` and count metrics with unit `1`. The CloudWatch exporter maps OTel unit `By` → CloudWatch `Bytes` and `1` → `None`. Confirmed in manual testing that metrics appear with correct units in the CW console.

translator/config/schema.json
mitali-salvi: Acknowledged — the EFA receiver currently collects from all EFA devices on the instance. Most EFA deployments have a single device, so device filtering is low priority for v1. The `resources` pattern from diskio/net is adapter-specific (`basicResourcesDefinition`). For the native OTel receiver, device filtering would need to be implemented in the awsefareceiver itself.

translator/translate/otel/receiver/awsefa/translator.go
mitali-salvi: No other plugin (diskio, net, nvidia_gpu, cpu) supports ["*"] in the measurement array,
so I've removed it to stay consistent with existing behavior. Customers will need to list metrics explicitly

JayPolanco (approved):

translator/translate/otel/receiver/awsefa/translator.go
sky333999: By the golden yaml tests, I mean the ones in `translator/tocwconfig/sampleConfig/` that test the end to end pipelines and not just the individual components. You could probably add this to the complete linux one.

sky333999 (approved):

PR #2095, opened 2026-04-21
feat(nodemetadataenricher): Add log signal support

miconeilaws:
Extend the nodemetadataenricher processor to handle logs in addition to metrics. The processLogs method uses the same logic as processMetrics: looks up k8s.node.name in resource attributes and stamps host.id, host.name, host.type, host.image.id, and cloud.availability_zone from the Lease-based node metadata cache.

This enables the cluster-scraper's K8s events pipeline to enrich event logs with per-node IMDS metadata.


sky333999 (dismissed):

plugins/processors/nodemetadataenricher/processor.go
```diff
@@ -72,3 +73,38 @@ func (p *nodeMetadataEnricherProcessor) processMetrics(_ context.Context, md pme
 
 	return md, nil
 }
+
+func (p *nodeMetadataEnricherProcessor) processLogs(_ context.Context, ld plog.Logs) (plog.Logs, error) {
```
sky333999, line 94: Can we update processor_test.go to keep the tests in sync as well?

sky333999:
```
  Can we add parallel tests in processor_test.go. Minimum set:

  import "go.opentelemetry.io/collector/pdata/plog"

  func createTestLogs(nodeName string, existingAttrs map[string]string) plog.Logs {
      ld := plog.NewLogs()
      rl := ld.ResourceLogs().AppendEmpty()
      if nodeName != "" {
          rl.Resource().Attributes().PutStr("k8s.node.name", nodeName)
      }
      for k, v := range existingAttrs {
          rl.Resource().Attributes().PutStr(k, v)
      }
      sl := rl.ScopeLogs().AppendEmpty()
      sl.LogRecords().AppendEmpty().Body().SetStr("test event")
      return ld
  }

  func TestEnrichLogsWithCacheHit(t *testing.T) { /* mirror TestEnrichmentWithCacheHit */ }
  func TestPassThroughLogsWithCacheMiss(t *testing.T) { /* mirror */ }
  func TestPassThroughLogsWithoutNodeName(t *testing.T) { /* mirror */ }
  func TestPassThroughLogsWithEmptyNodeName(t *testing.T) { /* mirror */ }
  func TestLogCountPreservation(t *testing.T) { /* mirror TestMetricCountPreservation */ }
  func TestLogsCloudAZOverwrite(t *testing.T) { /* mirror */ }
  func TestMixedLogs(t *testing.T) { /* mirror TestMixedMetrics */ }

  You should also add at least one test that covers the lazy-init path on the logs side (cache nil at construction, set later), since that's the only branch whose behavior diverges across calls.
```

louisall: Done

louisall:
4 failing checks are unrelated:

sles-15:ca_bundle_test fails on both amd64 and arm64 (showing as 2 reds). The matrix hardcodes caCertPath: /etc/ssl/certs/ca-bundle.crt, which doesn't exist on SLES 15 — SUSE puts it at /etc/ssl/ca-bundle.pem. Fails on every PR
  that runs the SLES tests.

  rocky-linux-9:ssm_document_test is a 5-minute SSM-ready timeout — the EC2 instance didn't register with Systems Manager in time. Same flake has hit several recent unrelated PRs.

  Verify All PR Test Jobs is the workflow aggregator and just rolls up the 3 above.

movence (dismissed):

plugins/processors/nodemetadataenricher/factory.go
```diff
@@ -56,3 +57,26 @@ func createMetricsProcessor(
 		processorhelper.WithCapabilities(processorCapabilities),
 	)
 }
+
+func createLogsProcessor(
```
mitali-salvi, line 61: missing unit tests for this function and maybe this file as well

plugins/processors/nodemetadataenricher/processor.go
mitali-salvi, line 94: seems like this is an almost duplicate of processMetrics, can we extract the common functionality into a util function and use it inside both processing functions ?

PR #2096, opened 2026-04-21
feat: Register k8seventsreceiver in default components

miconeilaws:
Add the k8s_events receiver to the CW Agent's OTEL collector build so the cluster-scraper can watch Kubernetes events.


go.mod
```diff
@@ -496,6 +496,7 @@ require (
 	github.com/open-telemetry/opentelemetry-collector-contrib/pkg/translator/prometheus v0.124.1 // indirect
 	github.com/open-telemetry/opentelemetry-collector-contrib/pkg/translator/prometheusremotewrite v0.124.1 // indirect
 	github.com/open-telemetry/opentelemetry-collector-contrib/pkg/translator/zipkin v0.124.1 // indirect
+	github.com/open-telemetry/opentelemetry-collector-contrib/receiver/k8seventsreceiver v0.124.1
```
sky333999: Mind moving this up to the actual dependencies section and not have it in this indirect list

louisall: Done

miconeilaws:
Approved

louisall:
All 4 failing checks are environmental or pre-existing flakes, not caused by this PR. 

  - rocky-linux-9:ssm_document_test: EC2 instance failed to register with SSM within the 5-minute timeout. Same flake has hit multiple recent unrelated PRs (e.g. #2095, kevwar/syslog-receiver) — AWS infra registration latency.
  - sles-15:ca_bundle_test: pre-existing matrix bug. The test hardcodes caCertPath: /etc/ssl/certs/ca-bundle.crt, which doesn't exist on SLES 15 (SUSE puts it at /etc/ssl/ca-bundle.pem). Fails on every PR that runs the SLES tests.
  - sles-15:cloudwatchlogs_test (one arch): TestRotatingLogsDoesNotSkipLines flaked with actual log events count (2) does not match expected (3) — agent missed one line in the inotify window during a rotation. The test was added in
  2022 (commit 4072cc94, predates this work by ~3.5 years) and exercises the legacy Telegraf-based logs plugin. The parallel arch's job for this same SHA passed, and the test passed on every recent run before this one.
  - Verify All PR Test Jobs: aggregator — fails because of the three above.

movence (approved):

PR #2099, opened 2026-04-23
Add helm_chart_branch support to EKS daemon tests

spanaik:
Add helm_chart_branch support to EKS daemon tests

  Pass helm_chart_branch to Terraform modules that declare it (currently
  the OTEL MetricsV2 test modules). Auto-detects the latest release-*
  branch from aws-observability/helm-charts via the GitHub refs API.
  Engineers can override via the helm_chart_branch workflow input to
  test against their own branch.

  Priority: manual input → auto-lookup latest release → fallback to main.
  Existing EKS tests without a helm_chart_branch variable are unaffected.
  
  
  image

### DIFF
--- a/.github/workflows/test-artifacts.yml
+++ b/.github/workflows/test-artifacts.yml
@@ -1213,7 +1213,8 @@ jobs:
               HELM_BRANCH="${HELM_CHART_BRANCH_INPUT}"
               if [ -z "$HELM_BRANCH" ]; then
                 HELM_BRANCH=$(curl -sL "https://api.github.com/repos/aws-observability/helm-charts/git/matching-refs/heads/release-" \
-                  | jq -r '[.[].ref | ltrimstr("refs/heads/")] | sort | last // empty')
+                  | jq -r '.[].ref | ltrimstr("refs/heads/")' \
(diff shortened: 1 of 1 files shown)


.github/workflows/test-artifacts.yml
```diff
@@ -1198,13 +1208,31 @@ jobs:
             fi
 
             terraform init
+            HELM_BRANCH=""
+            if grep -q 'helm_chart_branch' *.tf 2>/dev/null; then
+              HELM_BRANCH="${HELM_CHART_BRANCH_INPUT}"
+              if [ -z "$HELM_BRANCH" ]; then
+                HELM_BRANCH=$(curl -sL "https://api.github.com/repos/aws-observability/helm-charts/git/matching-refs/heads/release-" \
+                  | jq -r '[.[].ref | ltrimstr("refs/heads/")] | sort | last // empty')
```
movence: think it's an edge case but `sort | last' would return `4.9` when there is `4.10`

spanaik: Nice catch, updated.

sky333999 (approved):

movence (approved):

PR #2111, opened 2026-05-08
Add Application Signals logs pipeline translator and allow metric routing to otlp endpoint

jj22ee:
# Description of the issue
Application Signals currently collects metrics and traces. This PR updates and refactors the `applicationsignals` translator to also generate OTel logs and metrics pipelines that route OTLP data to CloudWatch via the CW OTLP endpoints, with automatic per-service log group creation and selective metrics routing.

Customers configure `logs.logs_collected.application_signals` with optional `log_group_name` (supporting placeholders like `{service.name}`) and `log_stream_name`. The provisioner extension creates log groups/streams on first encounter.

# Description of changes

Update the existing `translator/translate/otel/pipeline/applicationsignals/` translator with routing connectors for both logs and metrics signals. Each signal uses a 3-pipeline topology (receive → route → export) connected via a routing connector.

**Metrics (3 pipelines via routing connector):**
- `metrics/application_signals_metrics_route`: OTLP receiver → routing connector
- `metrics/application_signals_metrics_logs_destination` (default): routing → existing AppSignals processors → EMF exporter
- `metrics/application_signals_metrics_otlp_destination`: routing → batch → otlphttp to CW OTLP monitoring endpoint
  - Routes by: `attributes["Telemetry.Source"] == "ServiceEvents"` (datapoint context)

**Logs (3 pipelines via routing connector):**
- `logs/application_signals_logs_route`: OTLP receiver → transform → attributestocontext → transform → routing connector
  - Routes by: `attributes["event.name"] == "aws.service_events.aggregate_profile"` (log context)
  - If there are no placeholders, `transform` and `attributestocontext` are not used
  - temporary resource attributes (prefixed with `temporary_key.`) are used to help build the `aws.log.group.name/aws.log.stream.name` attributes in the first transform, and will be discarded in the second transform. 
- `logs/application_signals_logs_batch` (default): routing → batch → otlphttp to CW OTLP logs endpoint
- `logs/application_signals_logs_nobatch`: routing → otlphttp (no batch, for large payloads)
  - Auth chain: `otlphttp → headers_setter → awscloudwatchlogsprovisioner → sigv4auth`

**Traces (unchanged):**
- Single pipeline with existing processors and X-Ray exporter

Notable behaviors:
- Logs auto-opt-in: logs pipeline activates when `metrics_collected.application_signals` is configured. Explicitly disable via: `logs.logs_collected.application_signals.disabled: true`
- Multiple placeholders supported in both group and stream (e.g. `/{deployment.environment}/prefix/{service.name}`) (Default log group: `/aws/service-events/{service.name}`)
- Customer-set resource attributes for `aws.log.group.name/aws.log.stream.name` are ignored

New/Updated shared translator packages:
- `extension/awscloudwatchlogsprovisioner` — new
- `extension/headerssetter` — new
- `processor/attributestocontext` — new
- `connector/routing` — new
- `exporter/otlphttp` — new (generic, signal-aware `EndpointConfig`)
- `extension/sigv4auth` — Updated with `WithService` option
- `processor/batchprocessor` — Updated with `WithMetadataKeys` option
- `processor/transformprocessor` — Updated with `NewTranslatorWithLogStatements`

# License

# Tests
- Unit tests
- Integration tests in `amazon-cloudwatch-agent-test`: <TODO: Make PR>

# Requirements
1. [x] Run `make fmt` and `make fmt-sh`
2. [x] Run `make lint`

-------
### Integration Tests


go.mod
jj22ee, line 1: `go.mod` will be updated since this PR depends on:
- [pull/521](https://github.com/amazon-contributing/opentelemetry-collector-contrib/pull/521) - Cherry-picks for extension/headersetter
- [pull/510](https://github.com/amazon-contributing/opentelemetry-collector-contrib/pull/510) - attributestocontext
- [pull/511](https://github.com/amazon-contributing/opentelemetry-collector-contrib/pull/511) - awscloudwatchlogsprovisioner
- [pull/506](https://github.com/amazon-contributing/opentelemetry-collector-contrib/pull/506)- EMF Exporter enhancement

translator/translate/otel/pipeline/applicationsignalslogs/translator.go
```diff
@@ -0,0 +1,299 @@
...
+	batchproc "github.com/aws/amazon-cloudwatch-agent/translator/translate/otel/processor/batchprocessor"
+	transformproc "github.com/aws/amazon-cloudwatch-agent/translator/translate/otel/processor/transformprocessor"
+	"github.com/aws/amazon-cloudwatch-agent/translator/translate/otel/receiver/otlp"
+)
+
+const (
+	pipelineName = "application_signals_logs"
+
+	// TODO: Update default log group prefix before PR is merged.
+	defaultLogGroupPrefix = "/aws/telemetry/"
```
jj22ee: Still pending and not final.

translator/translate/otel/pipeline/applicationsignalslogs/translator.go
```diff
@@ -0,0 +1,299 @@
...
+	defaultLogStreamName  = "default"
+
+	metadataKeyLogGroup  = "aws.cloudwatch.log_group.destination"
+	metadataKeyLogStream = "aws.cloudwatch.log_stream.destination"
```
jj22ee: These key values are not following any otel spec, they're just arbitrary attributes created by Transform Processor which are used by AttributesToContext Processor, so we don't expect customers to know about these keys or even set them in their exported logs.

However, if customer does know and set these these attributes in their logs, we'll respect it and won't override them.

translator/translate/otel/translate_otel.go
```diff
@@ -52,6 +53,10 @@ func Translate(jsonConfig interface{}, os string) (*otelcol.Config, error) {
 	if !ok {
 		return nil, errors.New("invalid json config")
 	}
+	// Auto-enable logs.logs_collected.application_signals when metrics is configured
+	// Must happen before creating confmap since confmap may copy the map
+	applicationsignalslogs.AutoEnableIfNeeded(m)
```
sky333999: You dont need to mutate the json for this.. just rely on the existing AppSignals enabled checks in the pipeline translator.

translator/translate/otel/translate_otel.go
```diff
@@ -73,6 +78,7 @@ func Translate(jsonConfig interface{}, os string) (*otelcol.Config, error) {
 	translators.Merge(containerInsightsTranslators)
 	translators.Set(applicationsignals.NewTranslator(pipeline.SignalTraces))
 	translators.Set(applicationsignals.NewTranslator(pipeline.SignalMetrics))
+	translators.Set(applicationsignalslogs.NewTranslator())
```
sky333999: Does this really need to be a new translator? We've been trying to keep one translator per "solution" that can internally create multiple pipelines.

translator/translate/otel/receiver/adapter/translators.go
```diff
@@ -32,7 +32,7 @@ const (
 var (
 	logKey           = common.ConfigKey(common.LogsKey, common.LogsCollectedKey)
 	metricKey        = common.ConfigKey(common.MetricsKey, common.MetricsCollectedKey)
-	skipInputSet     = collections.NewSet[string](files.SectionKey, windowsevents.SectionKey)
+	skipInputSet     = collections.NewSet[string](files.SectionKey, windowsevents.SectionKey, common.AppSignals, common.AppSignalsFallback)
```
sky333999, line 35: What in your changes forced you to add this?
i.e. what complained when you didnt have this?

translator/translate/otel/processor/transformprocessor/translate.go
```diff
@@ -22,14 +23,21 @@ var transformJmxConfig string
...
 var _ common.ComponentTranslator = (*translator)(nil)
 
 func NewTranslatorWithName(name string) common.ComponentTranslator {
-	return &translator{name, transformprocessor.NewFactory()}
+	return &translator{name: name, factory: transformprocessor.NewFactory()}
+}
+
+// NewTranslatorWithLogStatements creates a transform processor translator that
+// executes the given OTTL statements in the "resource" context for logs.
+func NewTranslatorWithLogStatements(name string, statements []string) common.ComponentTranslator {
```
sky333999: nit: Can we follow the same pattern you did with `WithMetadataKeys` in the batch processor translator instead of creating this?

translator/translate/otel/processor/attributestocontext/translator.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+var _ common.ComponentTranslator = (*translator)(nil)
+
+func NewTranslator(actions []ActionMapping) common.ComponentTranslator {
+	return &translator{actions: actions}
+}
+
+func (t *translator) ID() component.ID {
+	return component.MustNewID("attributestocontext")
```
sky333999: nit: Can we just do `return component.NewIDWithName(t.factory.Type(), t.name)`?

translator/translate/otel/extension/headerssetter/translator.go
```diff
@@ -0,0 +1,54 @@
...
+func NewTranslatorWithName(name string, additionalAuth component.ID, headers []HeaderMapping) common.ComponentTranslator {
+	return &translator{
+		name:           name,
+		additionalAuth: additionalAuth,
+		headers:        headers,
+	}
+}
+
+func (t *translator) ID() component.ID {
+	return component.NewIDWithName(component.MustNewType("headers_setter"), t.name)
```
sky333999: nit: Same here too and for the other components - can we just do `return component.NewIDWithName(t.factory.Type(), t.name)`.

translator/translate/otel/extension/headerssetter/translator.go
```diff
@@ -0,0 +1,54 @@
...
+
+type translator struct {
+	name           string
+	additionalAuth component.ID
+	headers        []HeaderMapping
+}
+
+var _ common.ComponentTranslator = (*translator)(nil)
+
+func NewTranslatorWithName(name string, additionalAuth component.ID, headers []HeaderMapping) common.ComponentTranslator {
```
sky333999: nit: The `With` might be better suited unless we are saying these are all mandatory.

translator/translate/otel/receiver/adapter/translators.go
jj22ee: Before my change, `application_signals` only existed under `logs.metrics_collected` (for metrics pipeline) and `traces.traces_collected` (for traces pipeline). The adapter never saw it because it only scans `logs_collected`.

Now that we added `logs.logs_collected.application_signals`, the adapter scans it and tries to create a `telegraf_application_signals` receiver which doesn't exist and is why I try to skip it here.

translator/translate/otel/translate_otel.go
jj22ee: Not needed, moved into the existing `applicationsignals` translator.

translator/translate/otel/pipeline/applicationsignalslogs/translator.go
jj22ee: Will now be `/aws/service-events/{service.name}`.

translator/tocwconfig/sampleConfig/appsignals_and_eks_config.yaml
```diff
@@ -1,3 +1,24 @@
...
+              context: log
+              pipelines:
+                - logs/application_signals_logs_nobatch
+              statement: ""
+    routing/application_signals_metrics:
+        default_pipelines:
+            - metrics/application_signals_metrics_logs_destination
+        error_mode: ignore
+        table:
+            - condition: attributes["Telemetry.Source"] == "Telemend"
```
mxiamxia: "Telemend" is internal code that we're going to remove it from public visible places, including datamodel

translator/tocwconfig/sampleConfig/appsignals_and_eks_config.yaml
```diff
@@ -1,3 +1,24 @@
...
+              context: datapoint
+              pipelines:
+                - metrics/application_signals_metrics_otlp_destination
+              statement: ""
```
mxiamxia, line 21: is empty of `statement` required?

translator/tocwconfig/sampleConfig/appsignals_and_eks_config.yaml
```diff
@@ -1,3 +1,24 @@
...
+              pipelines:
+                - logs/application_signals_logs_nobatch
+              statement: ""
+    routing/application_signals_metrics:
+        default_pipelines:
+            - metrics/application_signals_metrics_logs_destination
+        error_mode: ignore
+        table:
+            - condition: attributes["Telemetry.Source"] == "Telemend"
+              context: datapoint
```
mxiamxia, line 18: why this is named `datapoint`

translator/tocwconfig/sampleConfig/appsignals_and_eks_config.yaml
jj22ee: This is now changed to be `ServiceEvents`.

translator/tocwconfig/sampleConfig/appsignals_and_eks_config.yaml
jj22ee: Since our SDK will set Telemetry.Source as an attribute on `counter.add(count, {"Telemetry.Source": "ServiceEvents", ...})`, it's a datapoint attribute, so `context: datapoint` is required to match it.

translator/tocwconfig/sampleConfig/appsignals_and_eks_config.yaml
jj22ee: `statement: ""` is just the default serialization from the routing connector's RoutingTableItem struct.
We only need `condition` for routing, no need `statement` which can be used to modify data as well.

translator/config/schema.json
```diff
@@ -658,6 +658,34 @@
...
+                  "minLength": 1
+                },
+                "log_stream_name": {
+                  "type": "string",
+                  "minLength": 1
+                }
+              },
+              "additionalProperties": true
+            },
+            "app_signals": {
```
jefchien: Why are we keeping `app_signals` for this new logs section? I get that metrics/traces needed it for backwards compatibility, but this is entirely new and all the current documentation uses `application_signals`.

translator/translate/otel/pipeline/applicationsignals/translator.go
```diff
@@ -78,16 +216,289 @@ func (t *translator) Translate(conf *confmap.Conf) (*common.ComponentTranslators
...
+	}
+
+	translators.Receivers.Set(connectorTranslator)
+
+	translators.Processors.Set(batchproc.NewTranslator(common.WithName(metricsVariantOtlpDest), batchproc.WithTelemetrySection(common.LogsKey)))
+	translators.Exporters.Set(otlphttp.NewTranslatorWithName(metricsVariantOtlpDest, metricsEndpoint,
+		otlphttp.WithAuthenticator(sigv4ID),
+	))
+	translators.Extensions.Set(sigv4auth.NewTranslatorWithName(metricsComponentName, sigv4auth.WithService("monitoring")))
+	translators.Extensions.Set(agenthealth.NewTranslator(agenthealth.LogsName, []string{agenthealth.OperationPutLogEvents}))
```
okankoAMZ: Why is translate Metric Route sending agenthealth metrics to PLE?

translator/translate/otel/pipeline/applicationsignals/translator.go
```diff
@@ -78,16 +216,289 @@ func (t *translator) Translate(conf *confmap.Conf) (*common.ComponentTranslators
...
+			return true
+		}
+	}
+	return false
+}
+
+func templateToLiteral(segments []templateSegment) string {
+	var sb strings.Builder
+	for _, seg := range segments {
+		sb.WriteString(seg.literal)
```
okankoAMZ, line 464: are intentionally dropping the `seg.attributes` ?

translator/tocwconfig/sampleConfig/appsignals_and_ecs_config.yaml
```diff
@@ -1327,6 +1443,16 @@ processors:
...
                     enabled: false
         timeout: 2s
+    transform/application_signals_logs:
+        error_mode: propagate
+        flatten_data: false
+        log_statements:
+            - context: resource
+              error_mode: propagate
+              statements:
+                - set(resource.attributes["aws.cloudwatch.log_group.destination"], Concat(["/aws/service-events/", resource.attributes["service.name"]], "")) where resource.attributes["aws.cloudwatch.log_group.destination"] == nil
```
jefchien: nit: Should we try to stick with a resource attribute closer to [semconv](https://opentelemetry.io/docs/specs/semconv/resource/cloud-provider/aws/logs/) (e.g. `aws.log.group.name`)?

translator/translate/otel/processor/attributestocontext/translator.go
```diff
@@ -0,0 +1,54 @@
...
+func NewTranslator(actions []ActionMapping) common.ComponentTranslator {
+	return &translator{factory: attributestocontextprocessor.NewFactory(), actions: actions}
+}
+
+func (t *translator) ID() component.ID {
+	return component.NewIDWithName(t.factory.Type(), "")
+}
+
+func (t *translator) Translate(_ *confmap.Conf) (component.Config, error) {
+	cfg := attributestocontextprocessor.NewFactory().CreateDefaultConfig()
```
jefchien: Doesn't use the factory instance in the translator.


```suggestion
	cfg := t.factory.CreateDefaultConfig()
```

translator/translate/otel/processor/attributestocontext/translator.go
```diff
@@ -0,0 +1,54 @@
...
+	"go.opentelemetry.io/collector/processor"
+
+	"github.com/aws/amazon-cloudwatch-agent/translator/translate/otel/common"
+)
+
+// ActionMapping defines a resource attribute to copy to client metadata.
+type ActionMapping struct {
+	Key                   string
+	FromResourceAttribute string
+}
```
jefchien: Would rather we re-export the `actions.KeyValue` (https://github.com/amazon-contributing/opentelemetry-collector-contrib/blob/aws-cwa-dev/processor/attributestocontextprocessor/config.go#L17) instead of redefining it here. Makes this fragile to changes in the processor.

translator/translate/otel/exporter/otlphttp/translator.go
```diff
@@ -0,0 +1,97 @@
...
+var _ common.ComponentTranslator = (*translator)(nil)
+
+// NewTranslatorWithName creates an otlphttp exporter translator with the given
+// endpoint configuration.
+func NewTranslatorWithName(name string, endpoint EndpointConfig, opts ...Option) common.ComponentTranslator {
+	t := &translator{
+		name:          name,
+		factory:       otlphttpexporter.NewFactory(),
+		endpoint:      endpoint,
+		authenticator: component.NewID(component.MustNewType(common.SigV4Auth)),
```
jefchien: Is there ever a case where we have a non-namespaced `sigv4auth` extension?

translator/translate/otel/pipeline/applicationsignals/translators.go
```diff
@@ -0,0 +1,42 @@
...
+	case pipeline.SignalTraces:
+		translators.Set(NewTranslator(signal))
+	case pipeline.SignalMetrics:
+		translators.Set(NewTranslator(signal, SetVariant(metricsVariantRoute)))
+		translators.Set(NewTranslator(signal, SetVariant(metricsVariantLogDest)))
+		translators.Set(NewTranslator(signal, SetVariant(metricsVariantOtlpDest)))
+	case pipeline.SignalLogs:
+		translators.Set(NewTranslator(signal, SetVariant(logsVariantRoute)))
+		translators.Set(NewTranslator(signal, SetVariant(logsVariantBatch)))
+		translators.Set(NewTranslator(signal, SetVariant(logsVariantNoBatch)))
```
jefchien: Why not check `isLogsDisabled` at this point?

translator/translate/otel/pipeline/applicationsignals/translators.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"go.opentelemetry.io/collector/confmap"
+	"go.opentelemetry.io/collector/pipeline"
+
+	"github.com/aws/amazon-cloudwatch-agent/translator/translate/otel/common"
+)
+
+func isLogsDisabled(conf *confmap.Conf, configKeys []string) bool {
+	for _, key := range configKeys {
+		if v, ok := common.GetBool(conf, common.ConfigKey(key, "disable")); ok && v {
```
jefchien: Please try to be consistent with the configuration. The `logs.metrics_collected.application_signals.limiter` has a `disabled` field.

translator/translate/otel/pipeline/applicationsignals/translators.go
```diff
@@ -0,0 +1,42 @@
...
+package applicationsignals
+
+import (
+	"go.opentelemetry.io/collector/confmap"
+	"go.opentelemetry.io/collector/pipeline"
+
+	"github.com/aws/amazon-cloudwatch-agent/translator/translate/otel/common"
+)
+
+func isLogsDisabled(conf *confmap.Conf, configKeys []string) bool {
```
jefchien: nit: Why do we need `configKeys`? Isn't this a fixed path?

translator/translate/otel/extension/headerssetter/translator.go
```diff
@@ -0,0 +1,81 @@
...
+type translator struct {
+	name           string
+	factory        extension.Factory
+	additionalAuth *component.ID
+	headers        []HeaderMapping
+}
+
+var _ common.ComponentTranslator = (*translator)(nil)
+
+func NewTranslatorWithName(name string, opts ...Option) common.ComponentTranslator {
```
jefchien, line 45: nit: How dynamic does this need to be? Isn't the `headersetter` a fixed configuration? Could we use an embedded YAML?

translator/translate/otel/exporter/otlphttp/translator.go
```diff
@@ -0,0 +1,97 @@
...
+)
+
+// EndpointConfig specifies the base endpoint and signal-specific endpoint for
+// the otlphttp exporter.
+type EndpointConfig struct {
+	BaseEndpoint    string // e.g. "https://logs.us-east-1.amazonaws.com"
+	LogsEndpoint    string // e.g. "https://logs.us-east-1.amazonaws.com/v1/logs"
+	MetricsEndpoint string
+	TracesEndpoint  string
+}
```
jefchien: When would we ever need anything beyond the base endpoint?

translator/translate/otel/exporter/otlphttp/translator.go
```diff
@@ -0,0 +1,97 @@
...
+		t.authenticator = id
+	}
+}
+
+// WithHeaders sets static HTTP headers on the exporter.
+func WithHeaders(headers map[string]string) Option {
+	return func(t *translator) {
+		t.headers = headers
+	}
+}
```
jefchien: Is this ever used? Aren't we always using the `headers_setter` extension?

translator/translate/otel/pipeline/applicationsignals/translator.go
```diff
@@ -78,16 +216,289 @@ func (t *translator) Translate(conf *confmap.Conf) (*common.ComponentTranslators
...
+		}
+	}
+
+	if !hasAttributes {
+		var literal string
+		for _, seg := range segments {
+			literal += seg.literal
+		}
+		return fmt.Sprintf(`set(resource.attributes["%s"], "%s")`, metadataKey, literal) + whereGuard
+	}
```
jefchien: Is this ever used? The two places where `buildOTTLSetStatement` is called are within `hasPlaceholder` checks, so `hasAttributes` will always be true.

go.mod
```diff
@@ -255,7 +258,16 @@ require (
...
+require (
+	github.com/open-telemetry/opentelemetry-collector-contrib/extension/awscloudwatchlogsprovisionerextension v0.0.0-00010101000000-000000000000
+	github.com/open-telemetry/opentelemetry-collector-contrib/pkg/ottl v0.124.1
+	github.com/open-telemetry/opentelemetry-collector-contrib/processor/attributestocontextprocessor v0.0.0-00010101000000-000000000000
+	github.com/safchain/ethtool v0.0.0-20210803160452-9aa261dae9b1
+)
+
+require github.com/aws/aws-sdk-go-v2/service/cloudwatchlogs v1.46.0 // indirect
+
+require github.com/aws/aws-sdk-go-v2/aws/protocol/eventstream v1.6.10 // indirect
```
jefchien: nit: Please group these in their respective existing require blocks.

translator/translate/otel/extension/sigv4auth/translator.go
```diff
@@ -13,19 +13,33 @@ import (
...
 
-func NewTranslatorWithName(name string) common.ComponentTranslator {
-	return &translator{name, sigv4authextension.NewFactory()}
+func NewTranslatorWithName(name string, opts ...Option) common.ComponentTranslator {
+	t := &translator{name: name, factory: sigv4authextension.NewFactory()}
+	for _, opt := range opts {
+		opt(t)
+	}
+	return t
 }
```
jefchien: I'm wondering if there's ever a situation where we'd want to have two separate sigv4auth extensions for the same service. Could we simplify this by just requiring a `service` in `NewTranslator` and using that for the `ID`?

jefchien (commented):
Lots of changes. Please make sure it is thoroughly tested.

translator/translate/otel/pipeline/applicationsignals/translator.go
```diff
@@ -78,16 +216,289 @@ func (t *translator) Translate(conf *confmap.Conf) (*common.ComponentTranslators
...
+func (t *translator) translateMetricsRouteToOtlp(_ *confmap.Conf) (*common.ComponentTranslators, error) {
+	region := agent.Global_Config.Region
+	if region == "" {
+		return nil, fmt.Errorf("region is required for %s OTLP metrics pipeline", common.AppSignals)
+	}
+
+	connectorTranslator := newMetricsRoutingConnectorTranslator()
+	sigv4ID := component.NewIDWithName(component.MustNewType("sigv4auth"), metricsComponentName)
+	metricsEndpoint := otlphttp.EndpointConfig{
+		MetricsEndpoint: fmt.Sprintf("https://monitoring.%s.amazonaws.com/v1/metrics", region),
```
jefchien: Don't hard code this. Get this from the AWS SDK. This won't be the same for all partitions.

translator/translate/otel/pipeline/applicationsignals/translator.go
```diff
@@ -78,16 +216,289 @@ func (t *translator) Translate(conf *confmap.Conf) (*common.ComponentTranslators
...
+			Pipelines: []pipeline.ID{noBatchPipelineID},
+		}),
+	)
+}
+
+func (t *translator) translateLogsReceiveToRoute(conf *confmap.Conf) (*common.ComponentTranslators, error) {
+	region := agent.Global_Config.Region
+	if region == "" {
+		return nil, fmt.Errorf("region is required for %s logs pipeline", common.AppSignals)
+	}
```
jefchien: Why is this a check in this function? It doesn't even use the region. If we want to guard against a missing region, it should be done earlier at a higher level.

translator/translate/otel/pipeline/applicationsignals/translator.go
```diff
@@ -78,16 +216,289 @@ func (t *translator) Translate(conf *confmap.Conf) (*common.ComponentTranslators
...
+func newLogsRoutingConnectorTranslator() common.ComponentTranslator {
+	batchPipelineID := pipeline.NewIDWithName(pipeline.SignalLogs, logsVariantBatch)
+	noBatchPipelineID := pipeline.NewIDWithName(pipeline.SignalLogs, logsVariantNoBatch)
+
+	return routing.NewTranslator(common.AppSignals+"_logs",
+		routing.WithErrorMode(ottl.IgnoreError),
+		routing.WithDefaultPipelines(batchPipelineID),
+		routing.WithTable(routingconnector.RoutingTableItem{
+			Context:   "log",
+			Condition: `attributes["event.name"] == "aws.telemend.aggregate_profile"`,
```
jefchien: This is outdated and should be replaced.

translator/translate/otel/pipeline/applicationsignals/translator.go
```diff
@@ -78,16 +216,289 @@ func (t *translator) Translate(conf *confmap.Conf) (*common.ComponentTranslators
...
+				common.WithName(logsComponentName),
+				batchproc.WithMetadataKeys(metadataKeys),
+				batchproc.WithTelemetrySection(common.LogsKey),
+			))
+		} else {
+			translators.Processors.Set(batchproc.NewTranslator(
+				common.WithName(logsComponentName),
+				batchproc.WithTelemetrySection(common.LogsKey),
+			))
+		}
```
jefchien: nit: This could just be 

```
			translators.Processors.Set(batchproc.NewTranslator(
				common.WithName(logsComponentName),
				batchproc.WithMetadataKeys(metadataKeys),
				batchproc.WithTelemetrySection(common.LogsKey),
			))
```

`metadataKeys` is only set if there are placeholders.

translator/tocwconfig/sampleConfig/base_appsignals_fallback_config.yaml
```diff
@@ -1363,16 +1489,55 @@ service:
...
+                - otlphttp/application_signals_logs
+            processors: []
+            receivers:
+                - routing/application_signals_logs
+        logs/application_signals_logs_route:
+            exporters:
+                - routing/application_signals_logs
+            processors:
+                - transform/application_signals_logs
+                - attributestocontext
```
jefchien, line 1529: Does the context get carried over to the routing pipeline? I'd imagine they have separate contexts.

translator/tocwconfig/sampleConfig/base_appsignals_fallback_config.yaml
```diff
@@ -1363,16 +1489,55 @@ service:
...
+            exporters:
+                - otlphttp/application_signals_logs
+            processors:
+                - batch/application_signals_logs
+            receivers:
+                - routing/application_signals_logs
+        logs/application_signals_logs_nobatch:
+            exporters:
+                - otlphttp/application_signals_logs
+            processors: []
```
jefchien, line 1521: Have we tested this and determined that we won't cause throttle limits to get hit without the batching?

translator/translate/otel/pipeline/applicationsignals/translator.go
```diff
@@ -78,16 +216,289 @@ func (t *translator) Translate(conf *confmap.Conf) (*common.ComponentTranslators
...
+	translators := &common.ComponentTranslators{
+		Receivers:  common.NewTranslatorMap[component.Config, component.ID](),
+		Processors: common.NewTranslatorMap[component.Config, component.ID](),
+		Exporters:  common.NewTranslatorMap[component.Config, component.ID](),
+		Extensions: common.NewTranslatorMap[component.Config, component.ID](),
+	}
+
+	translators.Receivers.Set(connectorTranslator)
+
+	translators.Processors.Set(batchproc.NewTranslator(common.WithName(metricsVariantOtlpDest), batchproc.WithTelemetrySection(common.LogsKey)))
```
jefchien: It's strange that the flush interval in the logs section would apply to a metrics pipeline. Are we sure we want this behavior?

jj22ee: It's a mistake, changed to `common.MetricsKey`

translator/tocwconfig/sampleConfig/base_appsignals_fallback_config.yaml
jj22ee: We determined it won't cause throttle limits since an instance of our SDK would export an `aws.service_events.aggregate_profile` log once every 60s, where [OTLP Logs TPS limit](https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-OTLPEndpoint.html) is be 5000/s per region. We haven't tried to test getting close to the limit.

translator/tocwconfig/sampleConfig/base_appsignals_fallback_config.yaml
jj22ee: Yes, routing connect does propagate context. Routing & LogGroup placeholder works in our sample apps.

translator/translate/otel/pipeline/applicationsignals/translator.go
jj22ee: Updated to `aws.service_events.aggregate_profile`.

translator/translate/otel/pipeline/applicationsignals/translator.go
jj22ee: Done, removed the check here.

translator/translate/otel/pipeline/applicationsignals/translator.go
jj22ee: Made a fix in https://github.com/aws/amazon-cloudwatch-agent/pull/2111/changes/7f773e1c38f6ef85d186a5bbff708747f0cb2925

translator/config/schema.json
jj22ee: Initially it was for consistency, but I agree that the new logs section should not need it - updated.

translator/translate/otel/processor/attributestocontext/translator.go
jj22ee: Export Type PR: https://github.com/amazon-contributing/opentelemetry-collector-contrib/pull/543

translator/translate/otel/pipeline/applicationsignals/translators.go
jj22ee: Updated to `disabled`.

translator/translate/otel/pipeline/applicationsignals/translator.go
jj22ee: It was some leftover from when we were using emf exporter instead of otlp exporter. Removed this.

translator/translate/otel/pipeline/applicationsignals/translator.go
jj22ee: `templateToLiteral` is only called when the template has no placeholders (!hasPlaceholders), so every segment's attribute field should empty so we can ignore it.

translator/translate/otel/pipeline/applicationsignals/translators.go
jj22ee: Makes sense to have it here, updated. Initial idea was to group it together with the existing `conf.IsSet(...)` checks

translator/translate/otel/pipeline/applicationsignals/translator.go
jj22ee: Not used, removed. Initially had it before I added the `hasPlaceholder` checks.

translator/translate/otel/exporter/otlphttp/translator.go
jj22ee: I was thinking someone may want to just have 1 otlphttp exporter in the future for logs/metrics/traces, in which they don't share the same BaseEndpoint so they'll need to configure all the signal-specific endpoints. For now I've removed `BaseEndpoint` since I'm not making use of it. Do you suggest I just stick with only using `BaseEndpoint` instead?

translator/translate/otel/exporter/otlphttp/translator.go
jj22ee: Not used anymore, removed. I used to use it in the scenario where there are no placeholders, but ended up with using `headers_setter` in that scenario as well.

translator/translate/otel/pipeline/applicationsignals/translators.go
jj22ee: You're right, it's not needed. Simplified `isLogsDisabled` logic.

translator/translate/otel/extension/sigv4auth/translator.go
jj22ee: Probably no such situation. Did the simplification suggestion - renamed to `NewTranslatorWithService(service string)` which uses the service as both the ID name and the signing service.

I kept `NewTranslator()` unchanged if someone else wants to leverage sigv4auth's ability to auto-detect the service from endpoint.

translator/translate/otel/exporter/otlphttp/translator.go
jj22ee: Maybe in a scenario where someone wants to leverage the service auto-detection feature of SigV4Auth. But I've removed this default value since I always use `WithAuthenticator` for metrics & logs pipelines.

translator/translate/otel/extension/headerssetter/translator.go
jj22ee: There's some dynamic settings:
- gets value `from_context` when the log group/stream has placeholders, or a static `value` when there are no placeholders
- `additional_auth` value

go.mod
jj22ee, line 1: Now depends on:
- https://github.com/amazon-contributing/opentelemetry-collector-contrib/pull/541
- https://github.com/amazon-contributing/opentelemetry-collector-contrib/pull/543

translator/translate/otel/extension/sigv4auth/translator.go
```diff
@@ -15,17 +15,18 @@ import (
...
 
 func NewTranslator() common.ComponentTranslator {
-	return NewTranslatorWithName("")
+	return &translator{factory: sigv4authextension.NewFactory()}
 }
 
-func NewTranslatorWithName(name string) common.ComponentTranslator {
-	return &translator{name, sigv4authextension.NewFactory()}
+func NewTranslatorWithService(service string) common.ComponentTranslator {
+	return &translator{name: service, service: service, factory: sigv4authextension.NewFactory()}
```
jefchien: nit: Don't need a separate field for name since it's always the same as service.

translator/translate/otel/pipeline/applicationsignals/translators.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"go.opentelemetry.io/collector/confmap"
+	"go.opentelemetry.io/collector/pipeline"
+
+	"github.com/aws/amazon-cloudwatch-agent/translator/translate/otel/common"
+)
+
+func isLogsDisabled(conf *confmap.Conf) bool {
+	v, ok := common.GetBool(conf, common.ConfigKey(common.AppSignalsLogs, "disabled"))
```
okankoAMZ, line 14: The PR description says to disable via "disable": true, but code checks "disabled":

translator/translate/otel/common/common.go
```diff
@@ -137,12 +137,15 @@ const (
 var (
 	AppSignalsTraces          = ConfigKey(TracesKey, TracesCollectedKey, AppSignals)
 	AppSignalsMetrics         = ConfigKey(LogsKey, MetricsCollectedKey, AppSignals)
+	AppSignalsLogs            = ConfigKey(LogsKey, LogsCollectedKey, AppSignals)
 	AppSignalsTracesFallback  = ConfigKey(TracesKey, TracesCollectedKey, AppSignalsFallback)
 	AppSignalsMetricsFallback = ConfigKey(LogsKey, MetricsCollectedKey, AppSignalsFallback)
+	AppSignalsLogsFallback    = ConfigKey(LogsKey, LogsCollectedKey, AppSignalsFallback)
```
jefchien: nit: Seems unused

translator/translate/otel/pipeline/applicationsignals/translators.go
```diff
@@ -0,0 +1,41 @@
...
+
+	switch signal {
+	case pipeline.SignalTraces:
+		translators.Set(NewTranslator(signal))
+	case pipeline.SignalMetrics:
+		translators.Set(NewTranslator(signal, SetVariant(metricsVariantRoute)))
+		translators.Set(NewTranslator(signal, SetVariant(metricsVariantLogDest)))
+		translators.Set(NewTranslator(signal, SetVariant(metricsVariantOtlpDest)))
+	case pipeline.SignalLogs:
+		if isLogsDisabled(conf) {
```
jefchien: nit: Should have a nil check for the conf.

translator/config/schema.json
```diff
@@ -658,6 +658,20 @@
...
+            "application_signals": {
+              "type": "object",
+              "properties": {
+                "log_group_name": {
+                  "type": "string",
+                  "minLength": 1
+                },
+                "log_stream_name": {
+                  "type": "string",
+                  "minLength": 1
```
jefchien: nit: `maxLength` could be set to 512 to match https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_PutLogEvents.html although I'm sure the placeholders will allow it to go past that.

translator/tocwconfig/sampleConfig/appsignals_and_ecs_config.yaml
```diff
@@ -1,3 +1,24 @@
...
+              pipelines:
+                - logs/application_signals_logs_nobatch
+              statement: ""
+    routing/application_signals_metrics:
+        default_pipelines:
+            - metrics/application_signals_metrics_logs_destination
+        error_mode: ignore
+        table:
+            - condition: attributes["Telemetry.Source"] == "ServiceEvents"
+              context: datapoint
```
jefchien, line 18: Is there a reason this is on the datapoint attributes and not the resource attributes?

translator/tocwconfig/sampleConfig/appsignals_and_ecs_config.yaml
```diff
@@ -1327,6 +1442,17 @@ processors:
...
                     enabled: false
         timeout: 2s
+    transform/application_signals_logs:
+        error_mode: propagate
+        flatten_data: false
+        log_statements:
+            - context: resource
+              error_mode: propagate
+              statements:
+                - set(resource.attributes["aws.cloudwatch.log_group.destination"], Concat(["/aws/service-events/", resource.attributes["service.name"]], "")) where resource.attributes["aws.cloudwatch.log_group.destination"] == nil
```
jefchien: Do we want to support existing `aws.cloudwatch.log_group.destination` resource attributes or should we always replace it with the pattern? If we need the `service.name` for the pattern to work, what do we do when the `service.name` doesn't exist?

If we want it to be "undefined", consider using the `Coalesce` function (https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/pkg/ottl/ottlfuncs#coalesce).

```
set(resource.attributes["aws.cloudwatch.log_group.destination"], 
    Concat(["/aws/service-events/", Coalesce([resource.attributes["service.name"], "undefined"])], ""))
    where resource.attributes["aws.cloudwatch.log_group.destination"] == nil
```

translator/translate/otel/exporter/otlphttp/translator.go
jefchien: Yes, but I'm fine with what you have as well.

translator/tocwconfig/sampleConfig/appsignals_and_ecs_config.yaml
```diff
@@ -1327,6 +1442,17 @@ processors:
...
+                - replace_pattern(resource.attributes["aws.cloudwatch.log_group.destination"], "<nil>", "undefined")
```
jefchien: Should we remove this `aws.cloudwatch.log_group.destination` resource attribute after `attributestocontext` sets it in the context? If we don't it's going to be included in every single log we send. Could add a transformprocessor for to clean it up before routing or in the export pipeline.

translator/translate/otel/pipeline/applicationsignals/translator.go
```diff
@@ -5,40 +5,86 @@ package applicationsignals
...
+	logsVariantBatch   = "application_signals_logs_batch"
+	logsVariantNoBatch = "application_signals_logs_nobatch"
+)
+
+const (
+	defaultLogGroupName  = "/aws/service-events/{service.name}"
+	defaultLogStreamName = "default"
+
+	metadataKeyLogGroup  = "aws.cloudwatch.log_group.destination"
+	metadataKeyLogStream = "aws.cloudwatch.log_stream.destination"
```
jefchien: nit: Consider making the metadata key consistent with the header key.

translator/translate/otel/pipeline/applicationsignals/translator.go
```diff
@@ -5,40 +5,86 @@ package applicationsignals
...
+	signal  pipeline.Signal
+	variant string
 }
 
 var _ common.PipelineTranslator = (*translator)(nil)
 
-func NewTranslator(signal pipeline.Signal) common.PipelineTranslator {
-	return &translator{
-		signal,
+func NewTranslator(signal pipeline.Signal, opts ...common.TranslatorOption) common.PipelineTranslator {
```
jefchien: nit: Don't export this function or `SetVariant` so it's clear that the pipelines should only be created using the `NewTranslators` function.

translator/translate/otel/pipeline/applicationsignals/translator.go
```diff
@@ -78,16 +214,276 @@ func (t *translator) Translate(conf *confmap.Conf) (*common.ComponentTranslators
...
+}
+
+func (t *translator) translateMetricsRouteToOtlp(_ *confmap.Conf) (*common.ComponentTranslators, error) {
+	region := agent.Global_Config.Region
+	if region == "" {
+		return nil, fmt.Errorf("region is required for %s OTLP metrics pipeline", common.AppSignals)
+	}
+
+	connectorTranslator := newMetricsRoutingConnectorTranslator()
+	sigv4ID := component.NewIDWithName(component.MustNewType("sigv4auth"), "monitoring")
```
jefchien: nit: Consider getting the ID from the translator.

```go
sigExt := sigv4auth.NewTranslatorWithService("monitoring")
...
otlphttp.WithAuthenticator(sigExt.ID())
```

jj22ee: Fixed in https://github.com/aws/amazon-cloudwatch-agent/pull/2111/commits/64ce8ce54a347942607ed9014a268e61c7b14bb7

translator/translate/otel/pipeline/applicationsignals/translator.go
jj22ee: Fixed in https://github.com/aws/amazon-cloudwatch-agent/pull/2111/commits/1b2f6a277d4bbe33744ba95525f23fc3d5021a4b

translator/translate/otel/pipeline/applicationsignals/translator.go
jj22ee: Do you mean to use the naming `aws.log.group.name / aws.log.stream.name` instead of `aws.cloudwatch.log_...`?

Done in https://github.com/aws/amazon-cloudwatch-agent/pull/2111/commits/cb34b0aed144522a35db83628d91e7ed54af45fe

translator/tocwconfig/sampleConfig/appsignals_and_ecs_config.yaml
jj22ee: As discussed, added a transform processor to cleanup: https://github.com/aws/amazon-cloudwatch-agent/pull/2111/commits/5cec42274fd9dfeec9fa910283c1df084cc1fb08

translator/tocwconfig/sampleConfig/appsignals_and_ecs_config.yaml
jj22ee: In our SDK, it will set the attribute in on `counter.add()` attributes, so it's a datapoint attribute. SDK can send datapoints with different kinds of `Telemetry.Source` so we can't have it on resource.attributes.

translator/config/schema.json
jj22ee: Done in https://github.com/aws/amazon-cloudwatch-agent/pull/2111/commits/9763ae62526dd085628dbdf7bdb73eaf516daa2e.

translator/translate/otel/extension/sigv4auth/translator.go
jj22ee: Done in https://github.com/aws/amazon-cloudwatch-agent/pull/2111/commits/6b3465cb51e4df3457951574e9586901e167d919

translator/translate/otel/common/common.go
jj22ee: Done in https://github.com/aws/amazon-cloudwatch-agent/pull/2111/commits/61ff1d1eda84da6130e013173aab528e5e48d185

translator/translate/otel/pipeline/applicationsignals/translators.go
jj22ee: Done in https://github.com/aws/amazon-cloudwatch-agent/pull/2111/commits/62723cfa458e1fdec837f070701b5df079b341fa

translator/tocwconfig/sampleConfig/appsignals_and_ecs_config.yaml
jj22ee: Updated how we address missing attributes. Before we scan for `"<nil>"` and replace with `"undefined"`.

Now, instead I make use of temporary resource attributes of the placeholders that will also be cleaned up in the end, so I avoid modifying original attributes.
- https://github.com/aws/amazon-cloudwatch-agent/pull/2111/commits/d3c5ded653601decbed3c8d10e356d2b0c5fe211

This way I can also add this enhancement:
- service.name: `nil → "unknown_service"` & `"unknown_service:*" → "unknown_service"`
  - (java could set `unknown_service:java`)
- all other attributes: `nil → "unknown"`

translator/translate/otel/pipeline/applicationsignals/translator.go
jefchien: Meant `x-aws-log-group`/`x-aws-log-stream` like with https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/CWL_HTTP_Endpoints_OTLP.html so it's clear what they're for, but I think it's fine either way.

translator/config/schema.json
```diff
@@ -658,6 +658,22 @@
             },
             "windows_events": {
               "$ref": "#/definitions/logsDefinition/definitions/logsWindowsEventsDefinition"
+            },
+            "application_signals": {
```
jefchien, line 662: nit: Should we declare `disabled` in the schema as well?

okankoAMZ (dismissed):

jefchien (dismissed):

go.mod
```diff
@@ -647,6 +647,6 @@ require (
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
 
-replace github.com/open-telemetry/opentelemetry-collector-contrib/extension/awscloudwatchlogsprovisionerextension => github.com/amazon-contributing/opentelemetry-collector-contrib/extension/awscloudwatchlogsprovisionerextension v0.0.0-20260515160158-c86549cf773d
+replace github.com/open-telemetry/opentelemetry-collector-contrib/extension/awscloudwatchlogsprovisionerextension => github.com/amazon-contributing/opentelemetry-collector-contrib/extension/awscloudwatchlogsprovisionerextension v0.0.0-20260521173229-510795fdd8b4
 
-replace github.com/open-telemetry/opentelemetry-collector-contrib/processor/attributestocontextprocessor => github.com/amazon-contributing/opentelemetry-collector-contrib/processor/attributestocontextprocessor v0.0.0-20260515160158-c86549cf773d
+replace github.com/open-telemetry/opentelemetry-collector-contrib/processor/attributestocontextprocessor => github.com/amazon-contributing/opentelemetry-collector-contrib/processor/attributestocontextprocessor v0.0.0-20260521173229-510795fdd8b4
```
jefchien: Bring this up into the respective replace group to make it easier to maintain.

jefchien (approved):

sky333999 (approved):

PR #2119, opened 2026-05-13
ci: declare permissions on the PR-build workflow

arpitjain099:
`PR-build.yml` runs Go lint + tests across the matrix with `actions/cache@v3` for the Go module cache. It is the one workflow in `.github/workflows/` without an explicit `permissions:` block.

Scopes:
- `contents: read` — `actions/checkout`
- `actions: write` — `actions/cache@v3` save path. The action restores AND saves; once explicit permissions kick in (instead of repo default), the save call needs `actions: write`.

YAML validates.

### DIFF
--- a/.github/workflows/PR-build.yml
+++ b/.github/workflows/PR-build.yml
@@ -19,6 +19,10 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref_name }}
   cancel-in-progress: true
 
+permissions:
+  contents: read
+  actions: write   # actions/cache save path on cache miss
+
 jobs:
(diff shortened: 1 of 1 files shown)


arpitjain099:
Hi @okankoAMZ, gentle ping on this. PR has been open for 4 days without review. I noticed you've been on the recent-merger side of recent merges in this repo. When you have a moment, would you mind giving it a quick look? No urgency. Happy to address any feedback.

arpitjain099:
Hi, Wanted to check whether this is still on the radar. Happy to revise. Thanks!

PR #2121, opened 2026-05-13
Add journald log collection support for CloudWatch Agent

Paamicky:
# Description of the issue
Modern Linux distributions (AL2023) use systemd-journald as the sole logging system, dropping rsyslog by default. Customers on AL2023 have no /var/log/messages to tail. This feature adds native systemd-journald log collection to the CloudWatch Agent by integrating the upstream OpenTelemetry journald receiver.  This fills that gap.

# Description of changes
   - Config translator - Maps customer JSON config (units, priority, matches, filters) to OTel journald receiver config. Follows the same pattern as existing CWAgent translators (windows event logs, log files).
  - Pipeline translator  - Creates one OTel pipeline per collect_list entry with independent receiver, filter processor, batch processor, and exporter. Each entry routes to its own CloudWatch log group/stream.
  - Filter processor - Implements regex-based include/exclude filtering using the OTel filter processor with OTTL expressions.
  - File storage extension - Persists journald cursor to /opt/aws/amazon-cloudwatch-agent/logs/state/ for state management across agent restarts. No log duplication or loss on restart.
  - JSON schema validation - Adds units, priority, matches, filters field definitions with type/length constraints and additionalProperties: false.
  - Agent health - Adds journald feature flag to the user-agent string for adoption metrics tracking.
  - Config wizard- Interactive journald configuration with prompts for units, priority, matches, and regex filters.
  - Sample configs and translation tests — journaldlogs_units, journaldlogs_priority, journaldlogs_matches, journaldlogs_filters with generated .conf and .yaml snapshots.

### Config Example:
```{
    "agent": {
      "debug": true
    },
    "logs": {
      "logs_collected": {
        "journald": {
          "collect_list": [
	    {
              "log_group_name": "journald-all-logs",
              "log_stream_name": "{instance_id}",
              "retention_in_days": 1
            },
            {
              "log_group_name": "journald-units-test",
              "log_stream_name": "{instance_id}",
              "units": ["sshd"],
              "retention_in_days": 1
            },
	    {
              "log_group_name": "journald-priority-info",
              "log_stream_name": "{instance_id}",
              "priority": "info",
              "retention_in_days": 1
            },
            {
              "log_group_name": "journald-priority-err",
              "log_stream_name": "{instance_id}",
              "priority": "err",
              "retention_in_days": 1
            },
            {
              "log_group_name": "journald-matches-test",
              "log_stream_name": "{instance_id}",
              "matches": [{"_UID": "0"}],
              "retention_in_days": 1
            },
            {
              "log_group_name": "journald-filter-test",
              "log_stream_name": "{instance_id}",
              "filters": [
                {"type": "include", "expression": ".*CODE_FILE.*"}
              ],
              "retention_in_days": 1
            }
          ]
        }
      }
    }
  }
```

# License

# Tests
  - TestJournaldUnitsLogs - validates only entries from configured unit appear in CloudWatch
  - TestJournaldPriorityLogs - validates only entries at err priority and above appear
  - TestJournaldMatchesLogs - validates only entries matching _UID=0 appear
  - TestJournaldRegexLogs - validates only entries matching include regex appear
  - TestJournaldLogState - validates no duplicates or missing logs across agent restart (cursor persistence)
  - All tests run on AL2 and AL2023
  - Integration test: **feature_linux_fournald_log_test:** https://github.com/aws/amazon-cloudwatch-agent/actions/runs/26290578857/job/78162376974
  
EKS and some other flakey test are failing on main, team is try to resolve it.

# Requirements
Integration test and SElinux Policy update PR's needs to be merged before this PR
Integration test: https://github.com/aws/amazon-cloudwatch-agent-test/pull/695
SElinux Policy Update: https://github.com/aws/amazon-cloudwatch-agent-selinux/pull/9

-------
### Integration Tests


internal/merge/confmap/confmap.go
```diff
@@ -10,7 +10,7 @@ import (
 	otelconfmap "go.opentelemetry.io/collector/confmap"
 )
 
-const (
+var (
 	KeyDelimiter = otelconfmap.KeyDelimiter
```
dricross, line 14: Do we need to change this from const to var?

Paamicky: > Do we need to change this from const to var?

I needed to change that to match the upstream. There is actually a commit for it:   https://github.com/aws/amazon-cloudwatch-agent/commit/17ab20db I will talk to kaushik if I need to revert the change.

Paamicky: The original commit above will roll in a couple of weeks. I will revert to main.

translator/translate/otel/extension/filestorage/translator.go
```diff
@@ -0,0 +1,47 @@
...
+
+type translator struct {
+	name    string
+	factory extension.Factory
+}
+
+var _ common.ComponentTranslator = (*translator)(nil)
+
+// StorageID is the component.ID for the file_storage/journald extension, exported for use by the receiver.
+var StorageID component.ID
```
okankoAMZ: The receiver reads `filestorage.StorageID` (a global var set as a side effect of `NewTranslator()`). If `build()` order ever changes or the receiver is tested in isolation, this silently becomes `component.ID{}`. No other extension/receiver pair communicates via a global var. Can we expose a pure function instead?

```go
func StorageComponentID() component.ID {
    return component.NewIDWithName(filestorage.NewFactory().Type(), "journald")
}
```

translator/config/schema.json
```diff
@@ -1148,6 +1151,75 @@
...
+                    "items": {
+                      "$ref": "#/definitions/logsDefinition/definitions/filterDefinition"
+                    }
+                  },
+                  "units": {
+                    "type": "array",
+                    "description": "Array of systemd units to read from",
+                    "items": {
+                      "type": "string",
+                      "maxLength": 256
```
okankoAMZ, line 1205: `minItems` and `uniqueItems` are array keywords, they're silently ignored on `"type": "string"`. This means no validation on priority values. Can we use an `"enum"` like `event_levels` in `logsWindowsEventsDefinition` does?

```json
"enum": ["emerg", "alert", "crit", "err", "warning", "notice", "info", "debug", "0", "1", "2", "3", "4", "5", "6", "7"]
```

translator/translate/otel/exporter/journald/translator.go
```diff
@@ -0,0 +1,119 @@
...
+	cfg.MiddlewareID = &agenthealth.LogsID
+
+	// Configure from the specific collect_list entry
+	if t.collectConfig != nil {
+		if err := t.setJournaldFieldsFromConfig(cfg); err != nil {
+			return nil, err
+		}
+	}
+
+	// Set AWS session configuration
```
okankoAMZ, line 58: This sets `cfg.Region`, `cfg.Profile`, etc. directly, but the existing `awscloudwatchlogs` exporter (same config type) uses `cfg.AWSSessionSettings.*`. Could this cause different credential resolution behavior? Should match `translator/translate/otel/exporter/awscloudwatchlogs/translator.go`.

translator/config/schema.json
```diff
@@ -1148,6 +1151,75 @@
             "collect_list"
           ]
         },
+        "logsJournaldDefinition": {
```
okankoAMZ, line 1173: `log_group_name` isn't required here, but both `logsWindowsEventsDefinition` and `logsFilesDefinition` require their key field. A customer who omits it silently gets all logs routed to a shared `"journald-logs"` group. Should we add `"required": ["log_group_name"]`?

translator/translate/otel/extension/filestorage/translator.go
Paamicky: Good catch! replaced with a pure function StorageComponentID() that returns the ID. No global state dependency.

translator/config/schema.json
Paamicky: Added enum validation for priority values matching both string names and numeric equivalents.

translator/translate/otel/exporter/journald/translator.go
Paamicky: I think both resolve identically at runtime. AWSSessionSettings is an embedded struct so cfg.Region and cfg.AWSSessionSettings.Region are the same field. The lint rule enforced that I remove the embedded prefix so i kept the direct path for consistency with lint. Should I add a comment in the code for clarity?

translator/config/schema.json
Paamicky: Yes and this is fixed, thanks!

tool/data/config/logs/journald.go
```diff
@@ -0,0 +1,41 @@
...
+package logs
+
+import "github.com/aws/amazon-cloudwatch-agent/tool/runtime"
+
+type Journald struct {
+	JournaldConfigs []*JournaldConfig
+}
+
+func (config *Journald) ToMap(ctx *runtime.Context) (string, map[string]interface{}) {
+	resultMap := make(map[string]interface{})
```
okankoAMZ: Since this is calling it recursively, what is our base case here? shouldn't we check to avoid a recursive overload and stack overflow

tool/data/config/logs/journald.go
```diff
@@ -0,0 +1,41 @@
...
+
+	collectList := []map[string]interface{}{}
```
okankoAMZ: Double allocation, feels a bit excessive we can do something like:
```go
resultMap := map[string]interface{}{
    "collect_list": make([]map[string]interface{}, 0, len(config.JournaldConfigs)),
}
```

tool/data/config/logs/journaldConfig.go
```diff
@@ -0,0 +1,57 @@
...
+			}
+		}
+		resultMap["filters"] = filters
+	}
+
+	if config.RetentionInDays != 0 {
+		resultMap["retention_in_days"] = config.RetentionInDays
+	}
+
+	return "", resultMap
```
okankoAMZ, line 56: what is this string for?

tool/data/config/logs/journald.go
Paamicky: I don't think it's recursive since `Journald.ToMap` calls `JournaldConfig.ToMap `on a different type.

tool/data/config/logs/journaldConfig.go
Paamicky: The empty string is intentional because `ToMap` returns (key, map) where the key is the config section. For leaf types like `JournaldConfig` , embedded inside `collect_list`, the key is unused by the caller (discarded with _ ). This is also used by other config types in the package.

tool/data/config/logs/journald.go
Paamicky: makes sense, fixed!

translator/translate/otel/exporter/journald/translator.go
```diff
@@ -0,0 +1,119 @@
...
+	cfg.CertificateFilePath = os.Getenv(envconfig.AWS_CA_BUNDLE)
+	if endpoint, ok := common.GetString(c, endpointOverrideKey); ok {
+		cfg.Endpoint = endpoint
+		cfg.AWSSessionSettings.Endpoint = endpoint
+	}
+	cfg.IMDSRetries = retryer.GetDefaultRetryNumber()
+	if profileKey, ok := agent.Global_Config.Credentials[agent.Profile_Key]; ok {
+		cfg.Profile = fmt.Sprintf("%v", profileKey)
+	}
```
okankoAMZ, line 69: Using `fmt.Sprintf("%v", profileKey)` won’t panic but can produce undesirable values like `"<nil>"` or a non-string representation if `profileKey` isn’t actually a string; prefer an explicit type assertion and validation: 
```suggestion
if profileKey, ok := agent.Global_Config.Credentials[agent.Profile_Key]; ok {
	if s, ok := profileKey.(string); ok && s != "" {
		cfg.Profile = s
	}
}

File: translator/translate/otel/processor/journaldfilter/translator.go [hunk 10]
```diff
@@ -0,0 +1,101 @@
...
+	var logRecordFilters []string
+	for _, filter := range t.filters {
+		var ottlExpr string
+		switch filter.Type {
+		case "exclude":
+			ottlExpr = fmt.Sprintf(`IsMatch(body, "%s")`, filter.Expression)
+		case "include":
+			// The filter processor drops records matching the condition.
+			// To "include" only matching records, we drop records that do NOT match.
+			ottlExpr = fmt.Sprintf(`not IsMatch(body, "%s")`, filter.Expression)
```
okankoAMZ: What if I want to use 2 includes, like include warn, include error, wouldn't this drop both of them?

translator/translate/otel/pipeline/journald/translator.go
```diff
@@ -0,0 +1,189 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
okankoAMZ, line 1: nit: should we make these only go:build for linux releases?

translator/translate/otel/extension/filestorage/translator.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/open-telemetry/opentelemetry-collector-contrib/extension/storage/filestorage"
+	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/confmap"
+	"go.opentelemetry.io/collector/extension"
+
+	"github.com/aws/amazon-cloudwatch-agent/translator/translate/otel/common"
+)
+
+const (
+	directory = "/opt/aws/amazon-cloudwatch-agent/logs/state"
```
okankoAMZ: This path is a Linux absolute path hardcoded in a package that has no build tag. Even setting aside the portability concern, CWA has a config.GetProgramDataPath() / translator/config package for resolving canonical agent paths. Using it here would be consistent with how other translators resolve paths and avoids a second place to update when the path changes.

tool/data/config/logs.go
```diff
@@ -47,3 +47,10 @@ func (config *Logs) AddWindowsEvent(eventName, logGroupName, logStream, eventFor
 	}
 	config.LogsCollect.AddWindowsEvent(eventName, logGroupName, logStream, eventFormat, eventLevels, eventIDs, filters, retention, logGroupClass)
 }
+
+func (config *Logs) AddJournald(logGroupName, logStream string, units []string, priority string, matches []map[string]string, filters []*logs.JournaldFilter, retention int) {
+	if config.LogsCollect == nil {
+		config.LogsCollect = &logs.Collection{}
+	}
+	config.LogsCollect.AddJournald(units, priority, matches, logGroupName, logStream, filters, retention)
```
okankoAMZ: Why is the argument order here different from top level, feel inconsistent?

translator/translate/otel/processor/journaldfilter/translator.go
Paamicky: True, multiple include filters now combined to one expression. Test passes

translator/translate/otel/pipeline/journald/translator.go
Paamicky: So the upstream `journaldreceiver` and `journaldinput` packages already have the tags `(//go:build !linux)`, which compiles on all platforms. Adding build constraints here creates test complications. I just tested and failed it failed on `macos-15`. I will leave as is and and keep it consistent with other pipeline translators

translator/translate/otel/extension/filestorage/translator.go
Paamicky: fixed! Now consistent with other linux resolves paths: `cfg.Directory = filepath.Join(paths.AgentDir, "logs", "state")`

tool/data/config/logs.go
Paamicky: Fixed it and now consistent across necessary files

translator/translate/otel/exporter/journald/translator.go
Paamicky: fixed, now uses explicit type assertion

plugins/inputs/logfile/logfile_test.go
```diff
@@ -1145,211 +1144,211 @@ func TestLogsPartialLineReading(t *testing.T) {
...
-			}
-			defer lsrc.Stop()
-		}
-		time.Sleep(1 * time.Second)
-		c++
-	}
-	wg.Wait()
-	tt.Stop()
-}
+// func TestLogFileMultiLogsReading(t *testing.T) {
```
okankoAMZ: Why are we commenting this out?

Paamicky: error during rebase fixed.

translator/translate/otel/pipeline/journald/translator.go
```diff
@@ -0,0 +1,189 @@
...
+		Exporters:  common.NewTranslatorMap[component.Config, component.ID](),
+		Extensions: common.NewTranslatorMap[component.Config, component.ID](),
+	}
+
+	// Create a unique name suffix based on index
+	suffix := fmt.Sprintf("_%d", index)
+
+	// Add journald receiver for this entry
+	receiverName := "journald" + suffix
+	units, _ := entryConfig["units"].([]interface{})
```
okankoAMZ: why are we dropping the error check here? How do we make sure the type casting succeeds

translator/translate/otel/pipeline/journald/translator.go
```diff
@@ -0,0 +1,189 @@
...
+	if t.Index() != -1 {
+		t.name += "/" + strconv.Itoa(t.Index())
+	}
+	return t
+}
+
+func NewTranslators(conf *confmap.Conf) common.TranslatorMap[*common.ComponentTranslators, pipeline.ID] {
+	translators := common.NewTranslatorMap[*common.ComponentTranslators, pipeline.ID]()
+
+	journaldKey := common.ConfigKey(common.LogsKey, "logs_collected", "journald")
```
okankoAMZ: can we use the constant here instead of hardcoding the strings, just like the LogsKey

translator/translate/otel/exporter/journald/translator.go
```diff
@@ -0,0 +1,114 @@
...
+	// Set log stream name with placeholder resolution
+	if logStreamName, ok := t.collectConfig[common.LogStreamName].(string); ok && logStreamName != "" {
+		cfg.LogStreamName = logsutil.ResolvePlaceholder(logStreamName, globallogs.GlobalLogConfig.MetadataInfo)
+	} else {
+		// Default log stream name if not specified
+		cfg.LogStreamName = logsutil.ResolvePlaceholder("{instance_id}", globallogs.GlobalLogConfig.MetadataInfo)
+	}
+
+	// Set retention in days if available
+	if retentionInDays, ok := t.collectConfig["retention_in_days"].(float64); ok {
```
okankoAMZ, line 112: why is the days a float and not an int?

translator/translate/otel/exporter/journald/translator.go
```diff
@@ -0,0 +1,114 @@
...
+		// Default log group name if not specified
+		cfg.LogGroupName = "journald-logs"
+	}
+
+	// Set log stream name with placeholder resolution
+	if logStreamName, ok := t.collectConfig[common.LogStreamName].(string); ok && logStreamName != "" {
+		cfg.LogStreamName = logsutil.ResolvePlaceholder(logStreamName, globallogs.GlobalLogConfig.MetadataInfo)
+	} else {
+		// Default log stream name if not specified
+		cfg.LogStreamName = logsutil.ResolvePlaceholder("{instance_id}", globallogs.GlobalLogConfig.MetadataInfo)
```
okankoAMZ: Check `context.CurrentContext().Mode()` at translation time and use `{hostname}` for on-prem modes, matching the wizard's behavior. Alternatively, enforce `log_stream_name` as required and remove the fallback entirely.

translator/translate/otel/exporter/journald/translator.go
```diff
@@ -0,0 +1,114 @@
...
+func (t *translator) setJournaldFieldsFromConfig(cfg *awscloudwatchlogsexporter.Config) error {
+	if t.collectConfig == nil {
+		return nil
+	}
+
+	// Set log group name with placeholder resolution
+	if logGroupName, ok := t.collectConfig[common.LogGroupName].(string); ok && logGroupName != "" {
+		cfg.LogGroupName = logsutil.ResolvePlaceholder(logGroupName, globallogs.GlobalLogConfig.MetadataInfo)
+	} else {
+		// Default log group name if not specified
```
okankoAMZ, line 96: Should we log that we are using the default, could be useful in debugging

translator/translate/otel/pipeline/journald/translator.go
Paamicky: fixed!

translator/translate/otel/exporter/journald/translator.go
Paamicky: JSON numbers are decoded as float64 in go, so we need to assert float and convert to int

translator/translate/otel/exporter/journald/translator.go
Paamicky: fixed. uses host name for on-prem modes

translator/translate/otel/pipeline/journald/translator.go
Paamicky: Schema validation ensures that units is an array of strings. Also missing field will be and empty slice and the agent collects all units.

translator/translate/otel/exporter/journald/translator.go
Paamicky: no OTel translator currently logs, I don't know if we should for journald

translator/translate/otel/pipeline/journald/translator.go
Paamicky: I added an error check for all fields assuming customer use direct yaml

translator/translate/otel/exporter/journald/translator.go
```diff
@@ -0,0 +1,117 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
okankoAMZ, line 1: MIssing  `log_group_class` when passing the Config

translator/translate/otel/exporter/journald/translator.go
```diff
@@ -0,0 +1,117 @@
...
+func NewTranslatorWithConfig(name string, collectConfig map[string]interface{}) common.ComponentTranslator {
+	return &translator{name, awscloudwatchlogsexporter.NewFactory(), collectConfig}
+}
+
+func (t *translator) ID() component.ID {
+	return component.NewIDWithName(t.factory.Type(), t.name)
+}
+
+// Translate creates an awscloudwatchlogsexporter config specifically configured for journald logs
+func (t *translator) Translate(c *confmap.Conf) (component.Config, error) {
```
okankoAMZ, line 47: nit: we should probably make logs-exporter translator module so we can reuse it between our logs

okankoAMZ (changes requested):

translator/translate/otel/exporter/journald/translator.go
Paamicky: The upstream `awscloudwatchlogsexporter` doesn't have the`LogGroupClass` field in its Config. This would require changes to the [upstream fork](https://github.com/amazon-contributing/opentelemetry-collector-contrib/blob/aws-cwa-dev/exporter/awscloudwatchlogsexporter/config.go) first

okankoAMZ (approved):

dricross (approved):

extension/agenthealth/handler/useragent/useragent_linux.go
```diff
@@ -0,0 +1,24 @@
...
+
+package useragent
+
+import (
+	"github.com/open-telemetry/opentelemetry-collector-contrib/receiver/journaldreceiver"
+	"go.opentelemetry.io/collector/otelcol"
+)
+
+const (
+	flagJournaldEnabled = "jd_enabled"
```
sky333999, line 14: Why do we need this? This seems redundant to the presence of the actual receiver itself.

PR #2131, opened 2026-05-28
Retain custom environment variables in env-config.json between translation

jefchien:
# Description of the issue
When the config translator runs (e.g. on `fetch-config`), it overwrites `env-config.json` entirely. Any values written to that file outside of translation are lost. This means custom environment variables are not retained.

# Description of changes
Adds file-level helpers to the existing `cfg/envconfig` package (`ReadFile`, `LoadFile`, `MergeFile`) that centralize all `env-config.json` I/O.

The key behavioral change is in `TranslateJSONMapToEnvConfigFile`. Instead of overwriting `env-config.json`, it now calls `MergeFile` with `toenvconfig.ManagedKeys`, the explicit set of keys that translation owns. On each translation, managed keys are reset to reflect the current JSON config (clearing stale values), while unmanaged keys are preserved.

Since the agent is trying to retain custom environment variables now, an incorrectly formatted `env-config.json` will prevent the agent from starting rather than wiping the values.

# License

# Tests
Added unit tests. Built and tested on EC2 instance.

# Requirements
2. Run `make lint`

-------
### Integration Tests


JayPolanco (dismissed):

cmd/amazon-cloudwatch-agent/amazon-cloudwatch-agent.go
```diff
@@ -201,20 +200,10 @@ func reloadLoop(
...
-	envVars := map[string]string{}
-	err = json.Unmarshal(bytes, &envVars)
+	envVars, err := envconfig.ReadEnvConfigFile(path)
 	if err != nil {
-		return fmt.Errorf("cannot create env config due to: %s", err.Error())
+		return fmt.Errorf("cannot read env config file %s: %w", path, err)
 	}
-
 	for key, val := range envVars {
 		os.Setenv(key, val)
```
sky333999: nit: Can we use `LoadEnvConfigFile` here instead of this and the Read above?

cfg/envconfig/file.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	for k, v := range envVars {
+		os.Setenv(k, v)
+	}
+	return nil
+}
+
+// MergeEnvConfigFile merges the given values into the env-config.json at path.
+// Existing values are retained; the provided values take precedence.
+func MergeEnvConfigFile(path string, values map[string]string) error {
```
movence: nit: is this just to clearly distinguish between merge and replace? seems unnecessary wrapper when the end result is always replace.

cfg/envconfig/file.go
```diff
@@ -0,0 +1,74 @@
...
+		delete(merged, k)
+	}
+	for k, v := range values {
+		merged[k] = v
+	}
+	data, err := json.MarshalIndent(merged, "", "\t")
+	if err != nil {
+		return err
+	}
+	return os.WriteFile(path, data, 0644) //nolint:gosec // G306: 0644 is intentional for env-config.json
```
okankoAMZ: why 0644?

cfg/envconfig/file.go
```diff
@@ -0,0 +1,74 @@
...
+// LoadEnvConfigFile loads environment variables from the given env-config.json into the process environment.
+// If visitor is non-nil, it is called for each key-value pair after setting.
+func LoadEnvConfigFile(path string, visitor func(key, value string)) error {
+	envVars, err := ReadEnvConfigFile(path)
+	if err != nil {
+		return err
+	}
+	for k, v := range envVars {
+		if os.Setenv(k, v) == nil && visitor != nil {
+			visitor(k, v)
```
okankoAMZ: Should we log out if we fail to setEnv ?

cfg/envconfig/file.go
```diff
@@ -0,0 +1,74 @@
...
+
+// MergeEnvConfigFile merges the given values into the env-config.json at path.
+// Existing values are retained; the provided values take precedence.
+func MergeEnvConfigFile(path string, values map[string]string) error {
+	return ReplaceEnvConfigFile(path, values, nil)
+}
+
+// ReplaceEnvConfigFile merges the given values into the env-config.json at path,
+// first removing keysToRemove so stale values don't persist.
+func ReplaceEnvConfigFile(path string, values map[string]string, keysToRemove []string) error {
```
okankoAMZ: I know that this function is NOT atomic so it can technically corrupt the env.json via a race condition. I think it should be okay since we have the `0644`; however, if the user sets to run CWAgent as X user, is it possible for that user to right to the `env-config.json` and cause the race condition?

cfg/envconfig/file.go
```diff
@@ -0,0 +1,74 @@
...
+	for k, v := range envVars {
+		if os.Setenv(k, v) == nil && visitor != nil {
+			visitor(k, v)
+		}
+	}
+	return nil
+}
+
+// MergeEnvConfigFile merges the given values into the env-config.json at path.
+// Existing values are retained; the provided values take precedence.
```
okankoAMZ: nit: I would say overwrite rather than precedence

cfg/envconfig/file.go
jefchien: This is the permissions that the `env-config.json` currently has. Don't want to introduce a potential regression by changing it.

cfg/envconfig/file.go
jefchien: Simplified into a single function.

cfg/envconfig/file.go
jefchien: I suppose if they're writing to it at the exact same time as the translation. I think that's difficult to avoid.

mitali-salvi (approved):

okankoAMZ (approved):

PR #2135, opened 2026-05-29
Add OpenTelemetry component translators for postgres database monitoring

JayPolanco:
# Description of the issue

Add component translators for PostgreSQL database monitoring via the OpenTelemetry pipeline. These are the building blocks (receivers, connectors, processors) that will be wired into pipelines in a follow-up PR.

Part 1 of 2.

# Description of changes

Adds standalone component translators following the same pattern as existing translators (JMX, Container Insights):

- **Receivers**: postgresql (metrics + events modes with TLS, query sampling, top query collection) and filelog (generic file tailing)
- **Connectors**: count (active session metrics) and signaltometrics (query statistics metrics from pg_stat_statements)
- **Processors**: transform/dbi_db_attrs (sets resource attributes), transform/dbi_log_destination (sets log group/stream),
transform/dbi_fix_start_time (fixes zero timestamps + fills empty attributes), filter/dbi_drop_monitor (excludes monitoring
user)

Also registers the new components (postgresqlreceiver, countconnector, signaltometricsconnector) and adds JSON schema
validation for the database_insights config section.

These components are not yet wired into any pipeline — no existing translation paths are affected.

# License
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the
terms of your choice.

# Tests

- Unit tests for every translator (ID generation, config output verification with exact assertions)
- Schema validation tests (valid + invalid sample configs)
- make fmt applied
- All existing tests continue to pass

# Requirements

1. [x] Run make fmt and make fmt-sh
2. [ X] Run make lint


go.mod
```diff
@@ -268,6 +275,45 @@ require (
 	github.com/safchain/ethtool v0.0.0-20210803160452-9aa261dae9b1
 )
 
+require (
```
movence, line 278: it seems pretty heavy lift by introducing indirect deps

JayPolanco: The vast majority of them come from the addition of the postgressql receiver. 

```
postgresqlreceiver → internal/sqlquery → snowflakedb/gosnowflake.
```

Basically all sql receivers use `internal/sqlquery` which pulls in these dependencies. SO mysql and the others shouldn't pull any extra stuff when we implement them

movence (approved):

translator/translate/otel/receiver/postgresql/translator.go
```diff
@@ -0,0 +1,99 @@
...
+	username            string
+	passfile            string
+	caFile              string
+	isLocalhost         bool
+	index               int
+	querySampleInterval time.Duration
+	maxRowsPerQuery     int64
+}
+
+func WithName(name string) Option   { return func(t *translator) { t.name = name } }
```
mitali-salvi, line 33: seems like these functions are only used in unit testing

mitali-salvi (approved):

PR #2141, opened 2026-06-03
Add OpenTelemetry container_insights pipeline translator

mitali-salvi:
## Description

Adds config translation for `opentelemetry.collect.container_insights` that generates OTel collector pipelines matching the `otelContainerInsights` YAML configuration.

### Config schema

```json
{
  "opentelemetry": {
    "collect": {
      "container_insights": {
        "cluster_name": "my-cluster",
        "collection_interval": 30,
        "mode": "node",
        "logs": { "enabled": true }
      }
    }
  }
}
```

| Field | Required | Default | Description |
|-------|----------|---------|-------------|
| cluster_name | Yes | — | EKS cluster name |
| collection_interval | No | 30 | Scrape interval in seconds |
| mode | No | all | `node` (daemonset) or `cluster` (deployment) pipelines |
| logs.enabled | No | false | Enable filelog pipelines |

### Pipelines generated

**Daemonset (mode=node):** kubeletstats, cadvisor, node-exporter, dcgm, neuron, efa, ebs-csi, lis-csi, filelog-app, filelog-node

**Deployment (mode=cluster):** apiserver, kube-state-metrics

### Architecture

Uses embedded YAML templates (`//go:embed`) with Go template placeholders for dynamic values. Each YAML contains the full pipeline definition (receivers, processors, exporters, extensions) and is unmarshaled at translation time using a factory registry.

### Known discrepancies vs live addon

1. **scope.version omitted** for dcgm/neuron/node-exporter/kube-state-metrics (avoids stale hardcoded versions)
2. **Logs use headerssetter pattern** (3 extra processors) instead of raw exporter headers — functionally equivalent
3. **OTel default values serialized** in output — no behavioral difference

## Tests

- `TestContainerInsightsConfig` golden file test (all 12 pipelines)
- Parity tested against live EKS addon v6.2.0 config — all pipeline structures match
- `make fmt` ✓ | `make lint` ✓ (only pre-existing plugins/plugins.go impi issue) | `go build` ✓


translator/tocwconfig/generate_golden_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+	"github.com/aws/amazon-cloudwatch-agent/translator"
+	"github.com/aws/amazon-cloudwatch-agent/translator/cmdutil"
+	"github.com/aws/amazon-cloudwatch-agent/translator/config"
+	"github.com/aws/amazon-cloudwatch-agent/translator/context"
+	"github.com/aws/amazon-cloudwatch-agent/translator/tocwconfig/toyamlconfig"
+	"github.com/aws/amazon-cloudwatch-agent/translator/translate/agent"
+)
+
+func TestGenerateContainerInsightsGolden(t *testing.T) {
```
JayPolanco: Is this useful? I don't see it called anywhere else and it looks like it just updates the yaml, but it doesn't actually test it

translator/cmdutil/translatorutil.go
```diff
@@ -254,6 +254,16 @@ func TranslateJsonMapToYamlConfig(jsonConfigValue interface{}) (interface{}, err
 	return mapstructure.Marshal(cfg)
 }
 
+// TranslateJsonMapToYamlConfigNoValidation is like TranslateJsonMapToYamlConfig but skips
+// config validation. Use in tests where filesystem-dependent validation would fail.
+func TranslateJsonMapToYamlConfigNoValidation(jsonConfigValue interface{}) (interface{}, error) { //nolint:revive
```
JayPolanco: should this live in a test file since it's only used for testing?

translator/config/schema.json
```diff
@@ -1772,6 +1772,37 @@
...
               "additionalProperties": false
+            },
+            "container_insights": {
+              "type": "object",
+              "properties": {
+                "cluster_name": {
+                  "description": "The name of the EKS cluster",
+                  "type": "string"
+                },
+                "collection_interval": {
```
JayPolanco, line 1783: Probably worth just using the existing definition: `#/definitions/timeIntervalDefinition`

translator/translate/otel/pipeline/opentelemetry/containerinsights/translator.go
```diff
@@ -0,0 +1,278 @@
...
+//   - omitted: all pipelines
+func NewTranslators(conf *confmap.Conf) common.PipelineTranslatorMap {
+	translators := common.NewTranslatorMap[*common.ComponentTranslators, pipeline.ID]()
+	mode := getMode(conf)
+
+	// Daemonset metrics pipelines
+	if mode == "" || mode == "node" {
+		translators.Set(newYAMLPipeline("kubeletstats", pipeline.SignalMetrics, kubeletstatsYAML))
+		translators.Set(newYAMLPipeline("cadvisor", pipeline.SignalMetrics, cadvisorYAML))
+		translators.Set(newYAMLPipeline("node_exporter", pipeline.SignalMetrics, nodeExporterYAML))
```
JayPolanco, line 70: Was checking the helm chart to compare, and i see that some of these are gated by condition checks (https://github.com/aws-observability/helm-charts/blob/release-6.2.0/charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl#L16), but in the json we set everything regardless. This intentional?

translator/tocwconfig/generate_golden_test.go
mitali-salvi: fixed

translator/cmdutil/translatorutil.go
mitali-salvi: fixed

translator/translate/otel/pipeline/opentelemetry/containerinsights/translator.go
mitali-salvi: Yes, intentional. In the helm chart, those conditions (nodeExporter.enabled, dcgmExporter.enabled, etc.) gate whether the exporter pods get deployed — not the agent's scrape config. When
  otelContainerInsights is enabled, the agent config always includes all pipeline definitions regardless.

  The reasoning: if the scrape target doesn't exist (e.g., no GPUs → no dcgm-exporter-service), the prometheus receiver simply fails DNS resolution and produces no data. No harm, no cost — the pipeline is
  effectively a no-op. This avoids coupling the agent config to knowledge of what's deployed alongside it.

For the JSON translation path, the helm chart / operator decides whether to include container_insights in the config at all. Once it's present, everything is on.

translator/config/schema.json
mitali-salvi: fixed

JayPolanco (approved):

movence (approved):

PR #2152, opened 2026-06-11
Add opentelemetry root level config section with host_metrics, prometheus, OTLP, container_insights, and database_insights pipelines

mitali-salvi:
## Summary

Adds the `opentelemetry.collect` config translation pipeline, enabling users to configure OTel pipelines via JSON config instead of raw YAML. This is the unified feature branch containing all v2 OTel collect sub-features.

## Features

- **host_metrics** — hostmetrics receiver (cpu, disk, filesystem, load, memory, network, processes) with configurable collection interval
- **container_insights** — Full CI metrics pipeline (kubeletstats, cadvisor, node_exporter, dcgm, neuron, ebs-csi, lis-csi, kube-state-metrics, apiserver) with per-source processor chains, k8sattributes enrichment, and OTTL transforms
- **container_insights logs** — Application and node log pipelines (filelog receiver) with dedicated exporters (compression: none for FluentBit migration parity)
- **prometheus** — User-provided prometheus scrape config (config_path) with optional cluster_name enrichment
- **otlp** — OTLP receivers (gRPC + HTTP) for metrics, logs, and traces with configurable endpoints
- **Shared metrics/logs/traces export pipelines** — resourcedetection, batch, otlphttp export to CloudWatch OTLP endpoint with sigv4 auth and agenthealth
- **Identity and log routing processors** — Platform-aware (host vs K8s) service.name, cloud.resource_id, deployment.environment derivation; log group/stream routing from aws.log.source

## Architecture

```
Source Pipelines (per-feature):          Destination Pipelines (shared):
┌─────────────────────────┐             ┌──────────────────────────┐
│ host_metrics            │             │ metrics/opentelemetry    │
│ container_insights (x8) │──forward──▶│  resourcedetection       │
│ prometheus              │             │  identity transform      │
│ otlp (metrics)          │             │  batch → otlphttp        │
├─────────────────────────┤             ├──────────────────────────┤
│ otlp (logs)             │──forward──▶│ logs/opentelemetry       │
│ container_insights logs │             │  resourcedetection       │
├─────────────────────────┤             │  identity + routing      │
│ otlp (traces)           │──forward──▶│  batch → otlphttp        │
└─────────────────────────┘             └──────────────────────────┘
```

## Key Design Decisions

- **rawMapConfig passthrough** for CI templates — avoids factory struct round-trip bugs ($1 expansion, zero-value ErrorMode, filelog operator marshaling)
- **Per-source pipelines** matching helm chart architecture (separate prometheus receivers, not consolidated)
- **forward connector** fan-in to shared export pipelines
- **Explicit batch sizes** from CW OTLP endpoint limits (1000 metrics, 10000 logs/traces per request; see https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-OTLPEndpoint.html)
- **escapeDollarDigit** scoped to CI translator only for k8sattributes regex backreferences

## Testing

- Full translator unit tests with golden YAML comparison
- Schema validation tests
- Verified on EKS cluster (cwa-otel-test) with all CI pipelines producing metrics
- Lint passes (`make lint` clean)


go.mod
```diff
@@ -9,7 +9,9 @@ replace collectd.org v0.4.0 => github.com/collectd/go-collectd v0.4.0
 // Replace with https://github.com/amazon-contributing/opentelemetry-collector-contrib, there are no requirements for all receivers/processors/exporters
 // to be all replaced since there are some changes that will always be from upstream
 replace (
+	github.com/open-telemetry/opentelemetry-collector-contrib/connector/countconnector => github.com/amazon-contributing/opentelemetry-collector-contrib/connector/countconnector v0.0.0-20260522182944-6434433557a0
```
JayPolanco: These connectors look like there on an older commit. Think I forgot to update the dependency to the latest. Should be the same as postgresreceiver

translator/translate/otel/processor/transformprocessor/translate.go
```diff
@@ -57,35 +89,64 @@ func (t *translator) ID() component.ID {
...
 		}
 		cfgMap := map[string]interface{}{
-			"log_statements": []interface{}{
-				map[string]interface{}{
-					"context":    "resource",
-					"error_mode": "propagate",
-					"statements": stmts,
-				},
-			},
+			"error_mode": errorMode,
```
movence, line 107: should the `errorMode` be applied to metric/logs statements level as well? Currently, `t.errorMode` only gets applied to the top level which might be different to individual statement level `error_mode`. For example, `translator/tocwconfig/sampleConfig/opentelemetry/combined_v1_v2_ec2_config.yaml` has
```
    transform/logs_cleanup:
        error_mode: ignore
        flatten_data: false
        log_statements:
            - context: resource
              error_mode: propagate
```

translator/config/schema.json
```diff
@@ -1796,6 +1799,115 @@
...
+                      }
+                    },
+                    "required": ["endpoint", "instance_name", "username", "password_file"],
+                    "additionalProperties": false
+                  }
+                }
+              },
+              "additionalProperties": false
+            },
+            "host_metrics": {
```
movence, line 1837: can you update PR description with the latest change?

translator/translate/otel/pipeline/opentelemetry/hostmetrics/translator.go
```diff
@@ -0,0 +1,66 @@
...
+func NewTranslator() common.PipelineTranslator {
+	return &hostMetricsTranslator{}
+}
+
+func (t *hostMetricsTranslator) ID() pipeline.ID {
+	return pipeline.NewIDWithName(pipeline.SignalMetrics, pipelineNameHostMetrics)
+}
+
+func (t *hostMetricsTranslator) Translate(conf *confmap.Conf) (*common.ComponentTranslators, error) {
+	if conf == nil || (!conf.IsSet(hostMetricsKey) && !conf.IsSet(common.DatabaseInsightsConfigKey)) {
```
movence, line 34: I think it's better to check whether `host_metrics` is configured from DBI translator since that's the one requiring it.

translator/config/schema.json
mitali-salvi: done

translator/translate/otel/processor/transformprocessor/translate.go
mitali-salvi: Fixed. Removed the WithErrorMode("ignore") from otlp_log_source and logs_cleanup since their statements (set with where guard, delete_key) can't produce errors — propagate is the correct default for both

translator/translate/otel/pipeline/opentelemetry/hostmetrics/translator.go
mitali-salvi: a user can enable it without DBI (e.g., they just want host metrics). If DBI also needs hostmetrics, they share the same receiver instance since they use the same component ID (hostmetrics/opentelemetry). Moving the gate into DBI would prevent standalone usage.

translator/translate/otel/pipeline/opentelemetry/translators.go
```diff
@@ -0,0 +1,29 @@
...
+package opentelemetry
+
+import (
+	"go.opentelemetry.io/collector/confmap"
+	"go.opentelemetry.io/collector/pipeline"
+
+	"github.com/aws/amazon-cloudwatch-agent/translator/translate/otel/common"
+	ci "github.com/aws/amazon-cloudwatch-agent/translator/translate/otel/pipeline/opentelemetry/containerinsights"
+	dbi "github.com/aws/amazon-cloudwatch-agent/translator/translate/otel/pipeline/opentelemetry/databaseinsights"
+	hi "github.com/aws/amazon-cloudwatch-agent/translator/translate/otel/pipeline/opentelemetry/hostmetrics"
```
sky333999, line 13: nit: hi -> hm

translator/translate/otel/pipeline/opentelemetry/containerinsights/translator.go
```diff
@@ -0,0 +1,217 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
sky333999, line 1: nit: this file should be called `translators.go` to be consistent with the others.

translator/translate/otel/pipeline/opentelemetry/hostmetrics/translator.go
movence: how is it preventing standalone usage? 

```
# for hostmetrics we can drop DBI config 
if conf == nil || !conf.IsSet(hostMetricsKey) {}

# for DBI, we check both
if conf == nil || !conf.IsSet(common.DatabaseInsightsConfigKey) || !conf.IsSet(hostMetricsKey) {
```

translator/translate/otel/pipeline/opentelemetry/containerinsights/translator.go
```diff
@@ -0,0 +1,217 @@
...
+//go:embed ebs_csi.yaml
+var ebsCsiYAML string
+
+//go:embed lis_csi.yaml
+var lisCsiYAML string
+
+// CI logs pipelines are self-contained with dedicated exporters (compression: none)
+// to match the helm chart behavior for FluentBit migration parity. They cannot share
+// the base logs/opentelemetry exporter which uses gzip compression. See:
+// https://github.com/aws-observability/helm-charts/blob/main/charts/amazon-cloudwatch-observability/templates/linux/_otel-container-insights-config.tpl
```
sky333999, line 62: nit: This reference in the helm chart is going to disappear from `main` pretty soon, either pin it to the commit or copy over the actual reasoning from the helm chart in here - like to prevent a cpu spike etc.

translator/translate/otel/pipeline/opentelemetry/containerinsights/translator.go
```diff
@@ -0,0 +1,217 @@
...
+var kubeStateMetricsYAML string
+
+// NewTranslators returns all container insights pipeline translators.
+// The pipelines generated depend on the "mode" config field:
+//   - "node": daemonset pipelines (per-node metrics + logs)
+//   - "cluster": deployment pipelines (cluster-wide metrics)
+//   - omitted: all pipelines
+func NewTranslators(conf *confmap.Conf) common.PipelineTranslatorMap {
+	translators := common.NewTranslatorMap[*common.ComponentTranslators, pipeline.ID]()
+	mode := getMode(conf)
```
sky333999, line 83: As discussed, we should consider aligning this to the CollectionRole we have today. It will allow us to use the same default config in our helm and tweak the behavior with an env var.

We could continue to do this flag in addition to the env var and have a fallthrough ordering, but lets make the naming consistent i.e. instead of `mode`, might wanna call this `role` or something.

translator/translate/otel/pipeline/opentelemetry/containerinsights/translator.go
```diff
@@ -0,0 +1,217 @@
...
+
+		// Daemonset logs pipelines (gated by logs.enabled)
+		if logsEnabled(conf) {
+			translators.Set(newYAMLPipeline("app", pipeline.SignalLogs, filelogAppYAML))
+			translators.Set(newYAMLPipeline("node", pipeline.SignalLogs, filelogNodeYAML))
+		}
+	}
+
+	// Deployment metrics pipelines
+	if mode == "" || mode == "cluster" {
```
sky333999: If mode is empty, itll end up doing the work for both cluster and node.. do we want that?
Might be safer to just do node level work if empty.

translator/translate/otel/pipeline/opentelemetry/containerinsights/translator.go
```diff
@@ -0,0 +1,217 @@
...
+	if collectionInterval < scrapeTimeout {
+		scrapeTimeout = collectionInterval
+	}
+
+	data := templateData{
+		ClusterName:        clusterName,
+		Region:             agent.Global_Config.Region,
+		CollectionInterval: collectionInterval.String(),
+		ScrapeTimeout:      scrapeTimeout.String(),
+		NodeName:           envOrPlaceholder("K8S_NODE_NAME"),
```
sky333999, line 150: nit: While we set both, `HOST_NAME` is what we've historically used in the context of CI, so Id stick to that.

translator/translate/otel/pipeline/opentelemetry/hostmetrics/translator.go
mitali-salvi: For DatabaseInsights to "require" host_metrics, it would need a mechanism to pull in another pipeline during registration, which would be a larger architectural change. Keeping the check here means the host_metrics pipeline auto-activates when DBI is configured (since DBI depends on host metrics data).

movence (dismissed):

translator/translate/otel/pipeline/opentelemetry/containerinsights/common.go
```diff
@@ -0,0 +1,235 @@
...
+
+	// Determine order: use pipeline spec order if provided, otherwise use map keys
+	var keys []string
+	if len(order) > 0 {
+		keys = order
+	} else {
+		keys = make([]string, 0, len(sectionMap))
+		for k := range sectionMap {
+			keys = append(keys, k)
+		}
```
sky333999, line 206:
```suggestion
		slices.Collect(maps.Keys(sectionMap))
```

translator/tocwconfig/sampleConfig/opentelemetry/otlp_otel_config.yaml
```diff
@@ -0,0 +1,644 @@
...
+                faas.instance:
+                    enabled: true
+                faas.max_memory:
+                    enabled: true
+                faas.name:
+                    enabled: true
+                faas.version:
+                    enabled: true
+        max_idle_conns: 100
+        middleware: agenthealth/statuscode
```
jefchien: This references `agenthealth/statuscode`, but that component isn't declared in `extension` or `service.extensions`.

translator/translate/otel/receiver/hostmetrics/translator.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+//go:embed scrapers_linux.yaml
+var scrapersLinuxConfig []byte
+
+//go:embed scrapers_windows.yaml
+var scrapersWindowsConfig []byte
+
+// Config is a serializable representation of
+// hostmetricsreceiver.Config. The upstream type uses mapstructure:"-" on its
```
sky333999, line 30: Can be a follow up but we should ideally update this in our fork instead of the below.

translator/tocwconfig/sampleConfig/opentelemetry/otlp_otel_config.yaml
mitali-salvi: thanks for pointing that out, update resourcedetection processor to not add middleware if the component comes from the opentelemetry pipelines

translator/translate/otel/pipeline/opentelemetry/prometheus/translator.go
```diff
@@ -0,0 +1,130 @@
...
+
+const (
+	pipelineName = "otel_prometheus"
+)
+
+var clusterNameRegex = regexp.MustCompile(`^[a-zA-Z0-9._-]+$`)
+
+var prometheusKey = common.ConfigKey(common.OpenTelemetryKey, common.CollectKey, common.PrometheusKey)
+var configPathKey = common.ConfigKey(prometheusKey, "config_path")
+var clusterNameKey = common.ConfigKey(prometheusKey, "cluster_name")
```
sky333999, line 31: Can be a follow up, but some of these we should standerdize in this new schema. Like if cluster_name can be set here, why not for `otlp`.

translator/translate/otel/processor/k8sattributesprocessor/translator.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+func (t *translator) Translate(_ *confmap.Conf) (component.Config, error) {
+	cfg := t.factory.CreateDefaultConfig()
+
+	cfgMap := map[string]interface{}{
+		"auth_type":   "serviceAccount",
+		"passthrough": false,
+		"filter": map[string]interface{}{
+			"node_from_env_var": "K8S_NODE_NAME",
```
sky333999, line 45: related to the other comment if we do end up changing that to HOST_NAME at some point

translator/translate/otel/common/endpoints.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/aws/aws-sdk-go/aws/endpoints"
+)
+
+// CloudWatch OTLP endpoint batch limits.
+// See: https://docs.aws.amazon.com/AmazonCloudWatch/latest/monitoring/CloudWatch-OTLPEndpoint.html
+const (
+	MaxMetricsPerRequest = 1000
+	MaxLogsPerRequest    = 10000
+	MaxSpansPerRequest   = 10000
+	BatchTimeout         = 15 * time.Second
```
sky333999: Should this be 15 or 30?

translator/translate/otel/pipeline/opentelemetry/databaseinsights/translators.go
```diff
@@ -0,0 +1,96 @@
...
+		return nil
+	}
+	if !ottlSafeRegex.MatchString(value) {
+		return fmt.Errorf("database_insights %s contains invalid characters: %q (must match %s)", field, value, ottlSafeRegex.String())
+	}
+	return nil
+}
+
+func isLocalhostEndpoint(endpoint string) bool {
+	return strings.HasPrefix(endpoint, "localhost") ||
```
sky333999, line 93: nit: This seems a little fragile... consider parsing the host (net.SplitHostPort) and compare exactly.

translator/config/schema.json
```diff
@@ -1796,6 +1799,115 @@
...
+              "type": "object",
+              "properties": {
+                "metrics_collection_interval": {
```
sky333999, line 1840: nit: We should probably standardize and call it just `collection_interval`

sky333999 (approved):

jefchien (approved):

musa-asad (commented):
.

PR #2166, opened 2026-06-18
cloudwatchlogs: buffer resetTimerCh to prevent lost wakeup in force-flush timer

musa-asad:
## Summary

A one-line concurrency fix that prevents the force-flush timer from permanently stalling on low-volume log streams.

## Background

The CloudWatch Agent reads log files and publishes events to CloudWatch Logs. Each log stream has a force-flush timer (default 5 seconds) that sends accumulated events even when the batch is not full. This keeps delivery timely for low-volume streams.

The flush timer is one-shot: after it fires, `resetFlushTimer()` must be called to start it again. If the reset is dropped, the timer stays dead and events only leave when the batch fills up, which for a quiet stream may take minutes or longer.

## Problem

`resetFlushTimer()` uses a non-blocking send:

```go
select {
case q.resetTimerCh <- struct{}{}:
default:
}
```

When `manageFlushTimer` is blocked sending on `flushCh` (waiting for the main loop to accept a flush signal), it cannot receive from `resetTimerCh`. With an unbuffered channel, the rearm hits the `default` branch and is silently dropped. After a size-based send, this can leave the force-flush timer permanently un-rearmed. A low-volume stream's trailing batch then sits unsent with no error or retry logged.

## Fix

```go
// Before
resetTimerCh: make(chan struct{}),

// After
resetTimerCh: make(chan struct{}, 1),
```

A buffer of 1 parks the rearm signal so `manageFlushTimer` picks it up on its next loop iteration. The signal is idempotent (presence means "please reset"), so a buffer of 1 is sufficient.

## Tests

Three new tests exercise the fix:

| Test | What it proves |
|------|---------------|
| `TestResetTimerChIsBuffered` | `resetTimerCh` has buffer capacity 1; cap 0 reintroduces the bug. |
| `TestForceFlushTimerRearmsAfterHeldSend` | A trailing event is flushed within 3× `flushTimeout` even when a size-based send held the flush path. |
| `TestForceFlushTimer_SingleLowVolumeEventFlushes` | A single event on a quiet stream is flushed by the timer alone. |

```
$ go test -count=1 -race -run "TestResetTimerCh|TestForceFlush" -timeout 30s ./plugins/outputs/cloudwatchlogs/internal/pusher/
ok  	github.com/aws/amazon-cloudwatch-agent/plugins/outputs/cloudwatchlogs/internal/pusher	4.060s
```


plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
```diff
@@ -62,12 +63,19 @@ func newQueue(
...
 		stopCh:          make(chan struct{}),
 		startNonBlockCh: make(chan struct{}),
 		wg:              wg,
 	}
+	if !q.watchdog.Stop() {
+		select {
+		case <-q.watchdog.C:
+		default:
+		}
+	}
```
jefchien: Isn't this the same function as `q.stopWatchDog`? Why are we stopping the timer at this point?

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
```diff
@@ -62,12 +63,19 @@ func newQueue(
 		sender:          sender,
 		eventsCh:        make(chan logs.LogEvent, 100),
 		flushCh:         make(chan struct{}),
-		resetTimerCh:    make(chan struct{}),
+		resetTimerCh:    make(chan struct{}, 1),
 		flushTimer:      time.NewTimer(flushTimeout),
+		watchdog:        time.NewTimer(flushTimeout),
```
jefchien: Why is this the same timeout as the `flushTimer`?

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
```diff
@@ -34,6 +34,7 @@ type queue struct {
 	flushCh      chan struct{}
 	resetTimerCh chan struct{}
 	flushTimer   *time.Timer
+	watchdog     *time.Timer
```
jefchien: `watchdog` is too generic of a name. It isn't immediately clear what this is for.

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
```diff
@@ -224,6 +239,15 @@ func (q *queue) stopFlushTimer() {
...
 
+func (q *queue) stopWatchdog() {
+	if !q.watchdog.Stop() {
+		select {
+		case <-q.watchdog.C:
+		default:
+		}
+	}
+}
+
```
jefchien: nit: Could be a shared function with `stopFlushTimer`. Like a `stopTimer` function that takes in a timer.

plugins/outputs/cloudwatchlogs/internal/pusher/lostwakeup_test.go
```diff
@@ -0,0 +1,125 @@
...
+	held := secondSendAt.Load().(time.Time).Sub(tRelease)
+	require.Lessf(t, held, flushBound,
+		"trailing batch flushed %v after the held send returned; expected < %v (timer re-armed)",
+		held, flushBound)
+
+	t.Logf("REARM-RESULT trailing batch flushed %v after release (flushTimeout=%v sends=%d)\n",
+		held.Round(time.Millisecond), flushTimeout, sendCount.Load())
+}
+
+func TestForceFlushTimer_SingleLowVolumeEventFlushes(t *testing.T) {
```
jefchien: What does this have to do with `lostwakeup`?

plugins/outputs/cloudwatchlogs/internal/pusher/lostwakeup_test.go
jefchien, line 1: nit: Does this one test case need to be split out into a separate test file?

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
```diff
@@ -202,12 +210,19 @@ func (q *queue) manageFlushTimer() {
 		select {
 		case <-q.flushTimer.C:
 			q.flushCh <- struct{}{}
+			if flushTimeout, ok := q.flushTimeout.Load().(time.Duration); ok {
+				q.watchdog.Reset(2 * flushTimeout)
+			}
```
jefchien: Is there a reason we reset it during the flush timer trigger and not during the reset?

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
```diff
@@ -202,12 +210,19 @@ func (q *queue) manageFlushTimer() {
...
 		case <-q.resetTimerCh:
+			q.stopWatchdog()
 			q.stopFlushTimer()
 			if flushTimeout, ok := q.flushTimeout.Load().(time.Duration); ok {
 				q.flushTimer.Reset(flushTimeout)
 			}
+		case <-q.watchdog.C:
+			q.logger.Warnf("Force-flush timer for (%v/%v) did not re-arm after a flush; log delivery may be stalled.", q.target.Group, q.target.Stream)
```
jefchien: Is stalled the right wording for this warning?

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
musa-asad: Reworked the watchdog into its own independent goroutine (`flush_watchdog.go`). It no longer shares timers with `manageFlushTimer`, so that stop logic is gone. The watchdog arms when `send()` is called and disarms when delivery completes (success or terminal failure).

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
musa-asad: Reworked. The watchdog now uses `2*flushTimeout` as its initial deadline (not the same timer). It runs in a separate goroutine, arms only when a send starts, and backs off exponentially (10s, 20s, 40s... cap 24h) if delivery stalls.

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
musa-asad: Named `flushWatchdog` in the current code. It lives in its own file (`flush_watchdog.go`) with a dedicated goroutine for the arm/disarm/warn lifecycle.

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
musa-asad: The watchdog no longer uses `time.Timer` stop/drain from outside. It manages its own timer internally in `flush_watchdog.go`. `stopFlushTimer` remains in `queue.go` for the flush timer only.

plugins/outputs/cloudwatchlogs/internal/pusher/lostwakeup_test.go
musa-asad: Consolidated all tests into queue_test.go and removed lostwakeup_test.go entirely.

plugins/outputs/cloudwatchlogs/internal/pusher/lostwakeup_test.go
musa-asad: Consolidated all tests into queue_test.go and removed the separate file.

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
musa-asad: Reworked. The watchdog no longer couples to the flush timer at all. It arms when `send()` is called (delivery is now expected) and disarms when delivery completes (success or terminal failure). This keeps it independent of the flush timer lifecycle.

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
musa-asad: Current warning text:

```
Force-flush for (%v/%v) has not completed a successful delivery in %s; the flush path may be stalled.
```

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
okankoAMZ: should we log here if flush it NOT ok, what does it mean if it is NOT ok?

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
```diff
@@ -202,28 +205,45 @@ func (q *queue) manageFlushTimer() {
...
+				q.flushTimerWatchdog.Reset(2 * flushTimeout)
+			}
 		case <-q.resetTimerCh:
+			q.stopFlushTimerWatchdog()
 			q.stopFlushTimer()
 			if flushTimeout, ok := q.flushTimeout.Load().(time.Duration); ok {
 				q.flushTimer.Reset(flushTimeout)
 			}
+		case <-q.flushTimerWatchdog.C:
+			q.logger.Warnf("Force-flush timer for (%v/%v) was not re-armed within the expected window after a flush; the send may be slow/retrying, or the re-arm signal may have been lost.", q.target.Group, q.target.Stream)
```
okankoAMZ: shouldn't we rearm watchdog after it fires?

musa-asad: It does. The watchdog re-arms itself with exponential backoff after each warning (10s, 20s, 40s, ... cap 24h). It disarms when delivery completes, whether the send succeeds or exhausts all retries. A completed delivery also resets the backoff to the initial 10s.

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
musa-asad: That branch guards the flush timer reset in `manageFlushTimer`. The `!ok` path is unreachable because `flushTimeout` is stored in `newQueue` before the goroutine launches. If it ever did fail, the only effect is skipping the flush timer reset. The watchdog is fully independent now and not affected by this branch.

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
```diff
@@ -62,14 +63,16 @@ func newQueue(
...
+		resetTimerCh:    make(chan struct{}, 1),
 		flushTimer:      time.NewTimer(flushTimeout),
+		flushWatchdog:   newFlushWatchdog(target, logger, flushTimeout, wg),
 		stopCh:          make(chan struct{}),
 		startNonBlockCh: make(chan struct{}),
 		wg:              wg,
 	}
 	q.flushTimeout.Store(flushTimeout)
 	q.wg.Add(1)
+	q.wg.Add(1)
```
okankoAMZ: why are we adding this separately?

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
```diff
@@ -175,12 +180,15 @@ func (q *queue) merge(mergeChan chan logs.LogEvent) {
...
+		q.batch.addStateCallback(func() {
+			q.flushWatchdog.drain()
+		})
+		q.flushWatchdog.arm()
 		q.sender.Send(q.batch)
 		q.batch = newLogEventBatch(q.target, q.entityProvider)
 	}
 }
 
-// onSuccessCallback returns a callback function to be executed after a successful send.
```
okankoAMZ, line 183: why are we deleting this comment?

plugins/outputs/cloudwatchlogs/internal/pusher/queue_test.go
```diff
@@ -663,9 +663,15 @@ func TestResendWouldStopAfterExhaustedRetries(t *testing.T) {
 	time.Sleep(2 * time.Second)
 
 	logLines := logSink.Lines()
-	lastLine := logLines[len(logLines)-1]
 	expected := fmt.Sprintf("All %v retries to G/S failed for PutLogEvents, request dropped.", cnt.Load()-1)
-	require.True(t, strings.HasSuffix(lastLine, expected), fmt.Sprintf("Expecting error log to end with request dropped, but received '%s' in the log", logSink.String()))
```
okankoAMZ, line 668: why are we no longer checking for last line? why do we check ever line instead?

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
```diff
@@ -114,6 +117,7 @@ func (q *queue) Stop() {
 		return
 	}
 	close(q.stopCh)
+	close(q.flushWatchdog.stopCh)
```
okankoAMZ: 1. Does the order matter? 2. if it does matter what happens if we close the channels in wrong order due to a race condition?

musa-asad: Order doesn't matter here.

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
musa-asad: Fixed.

plugins/outputs/cloudwatchlogs/internal/pusher/queue.go
musa-asad: Fixed.

plugins/outputs/cloudwatchlogs/internal/pusher/queue_test.go
musa-asad: So we can find the drop message regardless of its position.

okankoAMZ (approved):

jefchien (approved):

PR #2175, opened 2026-06-23
fix: exclude telegraf_journald adapter and useragent

Paamicky:
## Description of the issue
The agent crashes when the config includes both `journald` and `metrics`, the translator creates a `telegraf_journald` receiver via the telegraf adapter. This receiver type doesn't exist in the agen. Journald is handled by the OTEL journald receiver.  Also, the redundant `setJournaldFeatureFlags` is removed. The receiver's presence already tracks journald adoption.

## Description of changes

  - Add `common.JournaldKey` to `skipInputSet` to prevent `telegraf_journald` creation
  - Remove `setJournaldFeatureFlags` from useragent
  - Delete `useragent_linux.go`, `useragent_notlinux.go`, and `useragent_linux_test.go` as they are no longer needed
  - Add test case to adapter translators test
  - Regenerate `complete_linux_config.yaml`  for the expected output (removes `telegraf_journald`)

 ## License

## Tests

- Added unit test to make sure the `telegraf_journald` is not created
- Regenerated the `comple_linux_config` file to ensure the expected outcome

## Requirements

  2. Run make lint


translator/translate/otel/receiver/adapter/translators.go
```diff
@@ -32,7 +32,7 @@ const (
 var (
 	logKey           = common.ConfigKey(common.LogsKey, common.LogsCollectedKey)
 	metricKey        = common.ConfigKey(common.MetricsKey, common.MetricsCollectedKey)
-	skipInputSet     = collections.NewSet[string](files.SectionKey, windowsevents.SectionKey, common.AppSignals, common.AppSignalsFallback)
+	skipInputSet     = collections.NewSet[string](files.SectionKey, windowsevents.SectionKey, common.AppSignals, common.AppSignalsFallback, common.JournaldKey)
```
jefchien, line 35: Is there a unit test we can update to try to catch this better?

Paamicky: I added a unit test to make sure the receiver adapter doesn't generate `telegraf_journald`

jefchien (approved):

sky333999 (approved):

PR #2177, opened 2026-06-25
Add MySQL workload detection

Klarsen04:
### Description of the issue
Adds workload detection support for MySQL database servers as part of the workload-discovery feature.

### Description of changes
Implements a MySQL process detector that:

- Identifies MySQL server instances by checking for the mysqld executable
- Extracts the listening port from process arguments (--port/-P flag) or the MYSQL_TCP_PORT environment variable, defaulting to 3306 if neither is set. This is consistent with the upstream OTel MySQL receiver default
- Always returns READY status with the resolved port

**New files:**

- mysql.go - Main detector implementation
- port.go - Port extraction logic
Unit tests for both components

### License

### Tests
Added unit tests for MySQL detector (mysql_test.go)
Added unit tests for port extractor (port_test.go)
make fmt and make lint pass

### Test on AL2:
```

$ sudo /tmp/workload-discovery -debug
time=2026-06-25T13:48:04.298Z level=DEBUG source=internal/detector/nvidia/nvidia.go:36 msg="Starting NVIDIA GPU detection"
time=2026-06-25T13:48:04.299Z level=DEBUG source=internal/detector/nvidia/nvidia.go:39 msg="No NVIDIA GPU devices found"
time=2026-06-25T13:48:04.347Z level=DEBUG source=cmd/workload-discovery/discovery.go:105 msg="Starting discovery" num_process=271 num_worker=8
time=2026-06-25T13:48:04.348Z level=DEBUG source=internal/detector/mysql/mysql.go:46 msg="MySQL process detected" pid=19573
time=2026-06-25T13:48:04.348Z level=DEBUG source=cmd/workload-discovery/discovery.go:201 msg="Detected supported workload(s) for process" pid=19573 categories=[MYSQL]
time=2026-06-25T13:48:04.348Z level=DEBUG source=cmd/workload-discovery/discovery.go:93 msg="Discovered metadata" elapsed=49.58251ms
[
  {
    "categories": [
      "MYSQL"
    ],
    "name": "mysql",
    "telemetry_port": 3306,
    "status": "READY"
  }
]
```

### Requirements
 Run make fmt and make fmt-sh
 Run make lint


internal/detector/mysql/extract/port.go
```diff
@@ -0,0 +1,109 @@
...
+	args, err := process.CmdlineSliceWithContext(ctx)
+	if err != nil {
+		return 0, err
+	}
+
+	for i, arg := range args {
+		// --port=3307 or --port 3307
+		if arg == portFlagLong && i+1 < len(args) {
+			port, err := strconv.Atoi(args[i+1])
+			if err == nil && util.IsValidPort(port) {
```
musa-asad: `util.IsValidPort` accepts 0, so `mysqld --port=0` is extracted as port 0. The mysqld reference treats 0 as "use the default value", and since the metadata tag is `json:"telemetry_port,omitempty"`, a 0 port drops the field from the JSON entirely, so we'd advertise a READY MySQL workload with no port field.

Should we require `port > 0` for explicit values? Rejecting 0 also lands on the right answer, since the fallback returns 3306, which is what mysqld actually listens on when given port 0.

internal/detector/mysql/extract/port.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+func (e *portExtractor) Extract(ctx context.Context, process detector.Process) (int, error) {
+	for _, sub := range e.subExtractors {
+		port, err := sub.Extract(ctx, process)
+		if err == nil {
+			return port, nil
+		}
+	}
+	return defaultMySQLPort, nil
```
musa-asad, line 45: Every sub-extractor error lands here as a successful 3306: a permission failure or process-exit race while reading the cmdline or environment becomes READY metadata for a port we never observed, and `--skip-networking` gets 3306 with no TCP endpoint at all. The description scopes the default to "if neither is set", so the on-error and invalid-value fallbacks are wider than documented.

If defaulting on error is intentional (it does match the "Always returns READY" note), can we state that in the description and README? Otherwise the errors should propagate like the Java detector's do.

internal/detector/mysql/extract/port.go
```diff
@@ -0,0 +1,109 @@
...
+		}
+		if strings.HasPrefix(arg, portFlagShort) && len(arg) > len(portFlagShort) {
+			port, err := strconv.Atoi(arg[len(portFlagShort):])
+			if err == nil && util.IsValidPort(port) {
+				return port, nil
+			}
+		}
+	}
+
+	return 0, detector.ErrExtractPort
```
musa-asad, line 102: An explicit value that is present but unusable takes the same path as no value at all: `--port 99999`, `--port abc`, or the near-miss `-P=3307` (the prefix branch parses `=3307`, `Atoi` fails) all fall through to `ErrExtractPort`, which the aggregate converts into a silent 3306.

`internal/detector/extractor.go` already declares `ErrInvalidPort` for the found-but-unusable case and the Java extractor propagates it. Should we distinguish the two here? Worth adding malformed-value cases to the table test either way.

internal/detector/mysql/extract/port_test.go
```diff
@@ -0,0 +1,89 @@
...
+			wantPort: 3310,
+		},
+		"Success/PortFromEnv": {
+			setup: func(mp *detectortest.MockProcess) {
+				mp.On("CmdlineSliceWithContext", ctx).Return([]string{"mysqld"}, nil)
+				mp.On("EnvironWithContext", ctx).Return([]string{"PATH=/usr/bin", "MYSQL_TCP_PORT=3311"}, nil)
+			},
+			wantPort: 3311,
+		},
+		"Success/CmdlineTakesPrecedence": {
```
musa-asad, line 55: This case stubs only the command line, so it exercises the same path as `Success/PortFromLongFlag` and nothing observes cmdline-over-env precedence. Could you add a competing `MYSQL_TCP_PORT=3311` to the env stub and keep `wantPort: 3307`? The env stub needs `.Maybe()`, since the env extractor is never reached when cmdline wins and `AssertExpectations` fails on an uncalled expectation.

internal/detector/mysql/mysql_test.go
```diff
@@ -0,0 +1,142 @@
...
+			},
+			wantErr: detector.ErrIncompatibleDetector,
+		},
+		"Error/ExeWithContext": {
+			setup: func(mp *detectortest.MockProcess) {
+				mp.On("ExeWithContext", ctx).Return("", assert.AnError)
+			},
+			wantErr: assert.AnError,
+		},
+		"Error/CmdlineSliceWithContext": {
```
musa-asad: The case name says Error but it asserts a successful READY at 3306, and the PostgreSQL test of the same name expects the stub error to surface. If swallowing extractor errors is the intended design here, should we rename it to something like `Success/DefaultPortOnCmdlineError` so the name documents the behavior instead of contradicting it?

internal/detector/mysql/README.md
```diff
@@ -0,0 +1,24 @@
...
+## Status Results
+- `READY`: MySQL process detected with a port (explicit via `--port`/`-P` flag or `MYSQL_TCP_PORT`, otherwise defaults to 3306).
+
+## Sample Metadata Result
+```json
+{
+  "categories": ["MYSQL"],
+  "name": "mysql",
+  "status": "READY",
+  "telemetryPort": 3306
```
musa-asad: nit: the wire format is snake_case, the struct tag is `json:"telemetry_port,omitempty"` and the AL2 output in the description prints `"telemetry_port": 3306`. The Java README uses the correct spelling. The PostgreSQL README has the same typo if you want to fix both in one pass.

internal/detector/mysql/extract/port.go
Klarsen04: The current behavior (defaulting to 3306) is intentional and matches PostgreSQL's approach. The port extractor follows a fallback chain:

1. Try cmdline flags (--port/-P)
2. Try environment variable (MYSQL_TCP_PORT)
Fall back to MySQL default (3306)
3. Fall back to MySQL default (3306)
This ensures MySQL processes are always detected even if port extraction encounters permission errors or process state issues. I've updated the README to clearly document this behavior and added explicit handling for invalid ports (port 0, out-of-range values, malformed values) which now return ErrInvalidPort before falling back to the default.

internal/detector/mysql/extract/port.go
Klarsen04: Fixed. Port 0 is now explicitly rejected and treated as invalid. When --port=0 is found, it returns ErrInvalidPort and falls back to the default 3306. This matches MySQL's behavior where port 0 means "use default socket" which isn't appropriate for explicit port flags.

internal/detector/mysql/extract/port.go
Klarsen04: Fixed. The extractor now properly distinguishes:

ErrInvalidPort: Port found but invalid (malformed values like --port=abc, --port=99999, or --port=0)
ErrExtractPort: Port not found in cmdline/env

internal/detector/mysql/extract/port_test.go
Klarsen04: Fixed. Updated the test to include a competing environment variable with .Maybe() to actually verify that cmdline takes precedence

internal/detector/mysql/mysql_test.go
Klarsen04: Fixed. Renamed test from Error/CmdlineSliceWithContext to Success/DefaultPortOnCmdlineError to accurately reflect MySQL's behavior.
Note: This intentionally differs from PostgreSQL. PostgreSQL requires cmdline access to filter worker processes (checkpointer, autovacuum, etc.) and fails detection if cmdline is unavailable. MySQL has no worker processes to filter, so it succeeds with the default port 3306 when cmdline is unavailable.
Also PostgreSQL calls CmdlineSliceWithContext() first (to filter worker processes) and fails detection if that call errors. MySQL only calls CmdlineSliceWithContext() inside the port extractor, which catches errors and falls back to 3306, so detection succeeds.

internal/detector/mysql/README.md
Klarsen04: Fixed

internal/detector/mysql/README.md
```diff
@@ -0,0 +1,33 @@
...
+The detector examines the executable path of each process and checks if the base name is `mysqld`.
+
+## Port Detection
+
+The detector attempts to find the MySQL port using the following priority:
+1. Command-line flags: `--port` or `-P`
+2. Environment variable: `MYSQL_TCP_PORT`
+3. Default fallback: 3306
+
+Invalid or malformed port values (e.g., `--port=0`, `--port=99999`, `--port=abc`) are rejected and the detector falls back to the default port.
```
musa-asad: This sentence contradicts the priority list two lines above it and the code: on an invalid cmdline value the extractor advances to the next source, not to the default. `mysqld --port=abc` with `MYSQL_TCP_PORT=3308` reports 3308, not 3306.

Suggest: "...are rejected and the detector falls back to the next source in the priority list (e.g., `MYSQL_TCP_PORT`, then the default 3306)."

internal/detector/mysql/extract/port.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil {
+		return 0, err
+	}
+
+	for i, arg := range args {
+		// --port=3307 or --port 3307
+		if arg == portFlagLong && i+1 < len(args) {
+			port, err := strconv.Atoi(args[i+1])
+			if err != nil {
+				return 0, detector.ErrInvalidPort
```
musa-asad: The reject arms are now unconditional returns inside the loop, so an invalid occurrence masks a later valid one: `mysqld --port=abc --port=3307` returns `ErrInvalidPort` here and ends up on the env var or the default instead of 3307. The previous revision fell through and picked up 3307, the sibling extractor keeps that fall-through shape, and the MySQL manual says the last occurrence of a repeated option takes precedence, so the real server would run on the later value.

Should we go back to continuing the loop on an invalid value and keep the reject only when the scan finds nothing usable? A two-flag case in the table test would pin whichever behavior is intended.

internal/detector/mysql/extract/port_test.go
```diff
@@ -0,0 +1,118 @@
...
+			wantPort: 3306,
+		},
+		"Success/DefaultOnMalformedPort": {
+			setup: func(mp *detectortest.MockProcess) {
+				mp.On("CmdlineSliceWithContext", ctx).Return([]string{"mysqld", "--port", "abc"}, nil)
+				mp.On("EnvironWithContext", ctx).Return([]string{"PATH=/usr/bin"}, nil)
+			},
+			wantPort: 3306,
+		},
+		"Success/DefaultOnPort0": {
```
musa-asad, line 90: Each flag branch now carries its own reject arms, but 0 and invalid values are only tested through the space-separated forms. The equals form (`--port=0`), the attached short form (`-P0`), and an invalid `MYSQL_TCP_PORT` each take a different validation path and none has a case. Could you add one case per form so a regression in a single branch gets caught?

internal/detector/mysql/README.md
Klarsen04: Updated it to reflect the priority chain order

internal/detector/mysql/extract/port.go
Klarsen04: I switched to using continue instead of return allows the loop to scan all occurrences and return the last valid one.

internal/detector/mysql/extract/port_test.go
Klarsen04: Fixed. Added comprehensive test coverage for all flag formats and error paths:

Equals format (--port=X):
DefaultOnEqualsInvalidPort (99999)
DefaultOnEqualsMalformed (abc)
DefaultOnEqualsPort0 (0)

Short flag attached (-PX):
DefaultOnShortFlagAttachedInvalid (99999)
DefaultOnShortFlagAttachedMalformed (abc)
DefaultOnShortFlagAttachedPort0 (0)

Environment variable:
InvalidEnvFallsBackToDefault (abc)
Env0FallsBackToDefault (0)
EnvInvalidPortFallsBackToDefault (99999)

musa-asad (approved):

Paamicky (approved):

PR #2182, opened 2026-06-30
Fix OTTL scope attribute warnings and simplify configuration

jefchien:
# Description of changes
Fixes OTTL scope transform statements to use fully-qualified `scope.attributes` instead of bare `attributes`, eliminating the "one or more paths were modified to include their context prefix, please rewrite them accordingly" warnings at agent startup.

Refactors the transform processor translator with
- Per signal scope options  (`WithMetricScopeStatements`, `WithLogScopeStatements`) so metrics-only and logs-only pipelines don't emit dead statements for other signals
- Extracts `common.AgentScopeStatements(solution)` helper to eliminate repeated scope statement construction
- Changes default `error_mode` from `"propagate"` to `"ignore"` since all dynamic OTTL statements are either static string sets or guarded with `where` clauses that can't fail
- DBI scope transform shared as `transform/dbi_scope` (no index suffix needed since config is identical across instances)

# License

# Tests
Updated unit tests.

# Requirements
2. Run `make lint`

-------
### Integration Tests


translator/tocwconfig/sampleConfig/appsignals_and_eks_config.conf
```diff
@@ -14,14 +14,12 @@
   quiet = false
   round_interval = false
 
-[inputs]
```
mitali-salvi, line 17: why is this being updated ?

translator/translate/otel/processor/transformprocessor/translator.go
```diff
@@ -71,13 +71,29 @@ func WithScopeStatements(statements []string) Option {
 	}
 }
 
+// WithLogScopeStatements sets OTTL statements to execute in the "scope" context for logs only.
```
mitali-salvi, line 74: nit - missing unit testing

mitali-salvi (dismissed):

translator/tocwconfig/sampleConfig/appsignals_and_eks_config.conf
jefchien: Was regenerated

mitali-salvi (approved):

JayPolanco (approved):

PR #2183, opened 2026-07-01
Add Azure platform detection to the translator

movence:
## Summary

Adds automatic Azure platform detection to the CloudWatch Agent translator so Azure VM and AKS environments generate the correct OTEL configuration with no customer input, without affecting existing AWS (EC2/ECS/EKS/on-prem) detection.

### revision

Fixes two Azure credential-path regressions surfaced on real hardware:

- Azure VM with common-config credentials: detecting AzureVM mode no longer forces the web-identity chain. If the operator set a profile/shared_credential_file in `common-config.toml`, those win (plain AssumeRole); `oidctoken/web_identity_token_file` are skipped. Web-identity applies only when no shared creds are configured (role_arn still required there). Gated via new agent.HasSharedCredentials().

- AKS: on `KubernetesMode()==ModeAKS` the translator no longer requires `credentials.role_arn` and does not emit `oidctoken`. AKS authenticates via the Helm chart's projected service-account token + `AWS_ROLE_ARN/AWS_WEB_IDENTITY_TOKEN_FILE` env vars through the AWS SDK default chain (bare sigv4auth). No chart change needed. Plain Azure VM keeps the IMDS + oidctoken path.

- Testing

  - New golden TestAzureVMHostMetricsSharedCredsConfig; regenerated host_metrics_aks_config (no oidctoken, bare sigv4auth). All EC2/AzureVM/AKS goldens pass.
  - build, gofmt, golangci-lint, -race all clean.
  - Real EC2: 5 scenarios (creds±role_arn, EC2) all correct; EC2 untouched.
  - Real Azure VM (--mode auto, auto-detected): creds → plain AssumeRole; no creds → web-identity intact.
  - Real AKS: agent went from CrashLoopBackOff (749 restarts, credentials.role_arn is required) to 1/1 Running; config translates as mode=AKS, collector starts, and SigV4 export authenticates via the projected SA token with no STS/403 errors.

## What changed

- **New `translator/util/azuredetector` package**
  - `IsAzureVM()` detects an Azure VM by querying the Azure Instance Metadata Service through the contrib `metadataproviders/azure` provider (`azuremeta.NewProvider().Metadata()`). The IMDS HTTP specifics (endpoint, `Metadata: true` header, response parsing) are owned by the contrib provider; this package wraps that call with a short per-attempt timeout, a bounded retry, and result caching — added here because the contrib client ships none of those. Any probe error is treated as retryable and a definitive (non-error) result is cached.
  - `IsAKS()` reads the `RUN_IN_AKS` environment variable set by the AKS Helm chart.
- **Mode resolution (`translator/util/sdkutil.go`)** — AWS-first: Azure is only considered after every AWS signal (`RUN_IN_AWS`, IRSA, EC2, ECS) is exhausted; EKS keeps precedence over AKS. AKS nodes resolve to the `AzureVM` host mode. Detection runs once at startup.
- **Platform modes** — `ModeAzureVM` / `ModeAKS` (+ short modes) added to the context `SetMode`/`SetKubernetesMode`.
- **Generated config driven by platform** — `resourcedetection` emits `[env, azure]` for Azure VM and `[env, aks, azure]` for AKS; the `oidctoken` extension is emitted on Azure VM (`provider: azure`) so telemetry egress can later obtain a bearer token from Azure IMDS. Wiring `oidctoken` as the exporter authenticator (replacing sigv4auth) is follow-up egress work and is intentionally out of scope for platform detection — the extension is emit-only at this stage.

## Testing

- Detector unit tests use a fake `azuremeta.Provider` stub covering detected, not-Azure, unreachable, transient-error-not-cached, and result-cached cases, plus `RUN_IN_AKS` env stubs — all pass with `-race`.
- Mode-resolution table tests assert AWS (EC2/ECS/IRSA) and EKS results are unchanged even when the Azure signal is forced true.
- Golden `tocwconfig` fixtures for Azure VM and AKS validate the full JSON→YAML translation (including `resourcedetection` detectors and the `oidctoken` extension) through `xconfmap.Validate`.
- `go build ./...`, `go vet`, `gofmt`, and the repo-pinned `golangci-lint` are clean.

### Validated on a real Azure VM

Ran the built `config-translator` on a live Azure VM (`Standard_D2s_v3`) and diffed the generated OTEL YAML against the golden fixtures:

- **Azure VM** (`-mode auto`, live IMDS): detected AzureVM; output byte-identical to `host_metrics_azurevm_config.yaml` — detectors `[env, azure]` + `oidctoken` (`provider: azure`).
- **AKS** (`RUN_IN_AKS=True` + `RUN_IN_CONTAINER=True`): detected AzureVM host + AKS; byte-identical to `host_metrics_aks_config.yaml` — detectors `[env, aks, azure]`.
- **AWS unaffected**: `-mode ec2` → `[eks, env, ec2]`, no Azure detectors; `RUN_IN_AWS=True` on the Azure VM still resolves to EC2 (AWS wins over Azure).


translator/translate/otel/extension/oidctoken/translator.go
```diff
@@ -0,0 +1,47 @@
...
+
+func (t *translator) ID() component.ID {
+	return component.NewID(t.factory.Type())
+}
+
+func (t *translator) Translate(_ *confmap.Conf) (component.Config, error) {
+	cfg := t.factory.CreateDefaultConfig().(*oidctokenextension.Config)
+	// Azure VM is the only platform this extension is wired on today, so pin the
+	// provider explicitly rather than relying on "auto" environment detection.
+	cfg.Provider = oidctokenextension.ProviderAzure
```
jefchien, line 33: Shouldn't this be based on the mode detection?

translator/util/azuredetector/azuredetector.go
```diff
@@ -0,0 +1,222 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+// Package azuredetector detects whether the CloudWatch agent is running on an
+// Azure platform (Azure VM or Azure Kubernetes Service). It mirrors the
+// structure of the eksdetector package: detection results are cached in
+// singletons and the underlying primitives are exposed as overridable function
+// variables so they can be stubbed in unit tests.
+package azuredetector
```
jefchien, line 5: Is there a reason we didn't end up using https://github.com/amazon-contributing/opentelemetry-collector-contrib/blob/aws-cwa-dev/processor/resourcedetectionprocessor/metadataproviders/azure/azure.go#L17?

go.mod
```diff
@@ -20,6 +20,7 @@ replace (
 	github.com/open-telemetry/opentelemetry-collector-contrib/extension/awscloudwatchlogsprovisionerextension => github.com/amazon-contributing/opentelemetry-collector-contrib/extension/awscloudwatchlogsprovisionerextension v0.0.0-20260617204523-70530ca095f9
 	github.com/open-telemetry/opentelemetry-collector-contrib/extension/awsproxy => github.com/amazon-contributing/opentelemetry-collector-contrib/extension/awsproxy v0.0.0-20260617204523-70530ca095f9
 	github.com/open-telemetry/opentelemetry-collector-contrib/extension/headerssetterextension => github.com/amazon-contributing/opentelemetry-collector-contrib/extension/headerssetterextension v0.0.0-20260617204523-70530ca095f9
+	github.com/open-telemetry/opentelemetry-collector-contrib/extension/oidctokenextension => github.com/amazon-contributing/opentelemetry-collector-contrib/extension/oidctokenextension v0.0.0-20260629153525-3b0a252e5908
```
jefchien, line 23: Update all contrib versions so they stay in sync.

translator/util/sdkutil_test.go
```diff
@@ -14,44 +14,80 @@ import (
...
+	origRunInAws := runInAws
+	origEC2 := DefaultEC2Region
+	origECS := DefaultECSRegion
+	defer func() {
+		IsAzureVM = origIsAzureVM
+		IsAKS = origIsAKS
+		runInAws = origRunInAws
+		DefaultEC2Region = origEC2
+		DefaultECSRegion = origECS
+	}()
```
jefchien: nit: Consider using `t.Cleanup()` instead of `defer func()` and a helper function that will do this for you. That way it can be reused.

translator/translate/otel/extension/oidctoken/translator.go
jefchien, line 1: Doesn't look like this is used yet. Are we planning on hooking it up in the next PR?

translator/translate/otel/processor/resourcedetection/configs/aks_config.yaml
```diff
@@ -0,0 +1,3 @@
+detectors: [aks, env]
```
jefchien: Should this be `[env, aks, azure]`?

translator/context/context.go
```diff
@@ -129,8 +129,11 @@ func (ctx *Context) SetMode(mode string) {
 	case config.ModeWithIRSA:
 		ctx.mode = config.ModeWithIRSA
 		ctx.shortMode = config.ShortModeWithIRSA
+	case config.ModeAzureVM:
```
jefchien, line 132: Have we checked to make sure the introduction of a new mode doesn't break something else that's expecting only the existing modes?

movence: onprem checks treat AzureVM as a cloud host which is correct, and appsignal's exhaustive switch sends unknown modes to NewGenericResolver (safe fallback). No existing path breaks

translator/translate/otel/extension/oidctoken/translator.go
movence: `translate_otel.go` sets it on `ModeAzureVM` and it shows up in the `host_metrics_azurevm` extensions list

translator/util/azuredetector/azuredetector.go
movence: good call. updated

translator/translate/otel/processor/transformprocessor/transform_identity_host.yaml
jefchien, line 1: We'll want to have a separate transform configuration for Azure, but that can be done in a separate PR if you want.

translator/translate/otel/processor/transformprocessor/transform_identity_host.yaml
```diff
@@ -4,7 +4,7 @@ trace_statements:
     context: resource
     statements: &identity_statements
       - set(resource.attributes["service.name"], "unknown_service") where resource.attributes["service.name"] == nil
-      - set(resource.attributes["cloud.resource_id"], Concat(["arn:aws:ec2", resource.attributes["cloud.region"], resource.attributes["cloud.account.id"], Concat(["instance", resource.attributes["host.id"]], "/")], ":")) where resource.attributes["cloud.resource_id"] == nil and resource.attributes["host.id"] != nil
+      - set(resource.attributes["cloud.resource_id"], Concat(["arn:aws:ec2", resource.attributes["cloud.region"], resource.attributes["cloud.account.id"], Concat(["instance", resource.attributes["host.id"]], "/")], ":")) where resource.attributes["cloud.resource_id"] == nil and resource.attributes["host.id"] != nil and resource.attributes["cloud.provider"] == "aws"
```
jefchien: nit: `cloud.platform` == `aws_ec2`

translator/translate/otel/translate_otel.go
```diff
@@ -113,6 +115,14 @@ func translateInternal(jsonConfig interface{}, os string, validate bool) (*otelc
 		pipelines.Translators.Extensions.Set(server.NewTranslator())
 	}
 
+	// Azure VM has no AWS credentials, so it needs oidctoken to mint a bearer
+	// token for egress auth. The extension is emitted here; wiring it as the
+	// exporter authenticator (replacing sigv4auth) is handled by the Azure
+	// egress work and is intentionally not part of platform detection.
```
jefchien: This isn't correct. It has nothing to do with bearer tokens and is used as part of sigv4auth.

translator/translate/otel/extension/oidctoken/translator.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/aws/amazon-cloudwatch-agent/translator/config"
+	"github.com/aws/amazon-cloudwatch-agent/translator/context"
+	"github.com/aws/amazon-cloudwatch-agent/translator/translate/otel/common"
+	"github.com/aws/amazon-cloudwatch-agent/translator/util"
+)
+
+// linuxOutputTokenFile is where the extension writes the OIDC token on
+// Linux/Darwin, under the agent dir so a sibling auth component (e.g. sigv4auth)
+// reads the same path. Windows builds the equivalent path under ProgramData.
+const linuxOutputTokenFile = "/opt/aws/amazon-cloudwatch-agent/var/run/oidc-token"
```
jefchien: Why did we choose `<agent-dir>/var/run`? What do we use this directory for now?

translator/translate/otel/extension/oidctoken/translator.go
jefchien: Doesn't the `sigv4auth` extension need to know about the token file that's getting created?

translator/tocwconfig/tocwconfig_test.go
```diff
@@ -132,6 +132,25 @@ func TestAppSignalsAndEKSConfig(t *testing.T) {
 	checkTranslation(t, "appsignals_and_eks_config", "windows", expectedEnvVars, "")
 }
 
+func TestAzureVMHostMetricsConfig(t *testing.T) {
+	resetContext(t)
```
jefchien: nit: Would recommend having this test be for the default_otel_config test case because that's the main one we care about.

translator/util/oidctokenpath.go
```diff
@@ -0,0 +1,19 @@
...
+	"github.com/aws/amazon-cloudwatch-agent/translator/config"
+)
+
+// linuxOIDCTokenFile is the OIDC token path shared by oidctoken (writer) and sigv4auth (reader) on Linux/Darwin (etc dir).
+const linuxOIDCTokenFile = "/opt/aws/amazon-cloudwatch-agent/etc/.oidc-token"
+
+// OIDCTokenFilePath returns the OIDC token file path for the target platform; single source of truth for oidctoken and sigv4auth.
+func OIDCTokenFilePath(targetPlatform string) string {
+	if targetPlatform == config.OS_TYPE_WINDOWS {
+		return GetWindowsProgramDataPath() + "\\Amazon\\AmazonCloudWatchAgent\\.oidc-token"
```
jefchien: Linter is failing. Instead of this function, define this path in `tool/paths`.

translator/util/azuredetector/azuredetector.go
```diff
@@ -0,0 +1,81 @@
...
+		}
+		return result
+	}
+	return azureVMCache
+}
+
+// defaultProbeAzureIMDS reports whether the contrib provider returns a compute doc with a VM ID; any error is retryable and uncached.
+func defaultProbeAzureIMDS() (bool, error) {
+	var lastErr error
+	for attempt := 0; attempt < azureIMDSMaxAttempts; attempt++ {
```
jefchien: Should this be determined by the IMDS retries variable as well?

nit: also since we're not using `attempt`
```go
	for range azureIMDSMaxAttempts {
```

translator/util/azuredetector/azuredetector.go
```diff
@@ -0,0 +1,81 @@
...
+// IsAzureVMCache holds the cached result of the Azure VM detection probe.
+type IsAzureVMCache struct {
+	Value bool
+	Err   error
+}
+
+var (
+	// metadataProvider is the Azure IMDS provider; overridable in tests.
+	metadataProvider = azuremeta.NewProvider()
+	probeAzureIMDS   = defaultProbeAzureIMDS
```
jefchien: Doesn't look like we're overriding this anymore and just replacing the metadataProvider. Do we still need it?

translator/util/azuredetector/azuredetector.go
```diff
@@ -0,0 +1,81 @@
...
+	// Azure VM detection cache; mutex (not sync.Once) so tests can reset it safely.
+	azureVMMu       sync.Mutex
+	azureVMResolved bool
+	azureVMCache    IsAzureVMCache
+)
+
+// isAKS reports whether RUN_IN_AKS is set (by the AKS Helm chart); no I/O.
+func isAKS() bool {
+	return isRunningInAKS()
+}
```
jefchien: Is this function really necessary? Couldn't we set `IsAKS = envconfig.IsRunningInAKS` directly? Why do we need this wrapper?

translator/util/azuredetector/azuredetector.go
```diff
@@ -0,0 +1,81 @@
...
+		result := IsAzureVMCache{Value: value, Err: err}
+		// Cache only a definitive answer so a boot-time transient error can re-probe.
+		if err == nil {
+			azureVMCache = result
+			azureVMResolved = true
+		}
+		return result
+	}
+	return azureVMCache
+}
```
jefchien: There's has to be a simpler way to do this. It looks like we never store the Error in the `azureVMCache`, so there's not really a point in including it in the struct or having the struct. We also don't use the error outside of this package. This really could just be `isAzureVM() bool`. Have we considered using an `atomic.Pointer[bool]` as well to avoid the mutex?

```suggestion
var isAzureVMCache atomic.Pointer[bool]

func isAzureVM() bool {
  ok, _ := loadAzureVM()
  return ok
}

func detectAzureVM() (bool, error) { // for unit tests
  if c := isAzureVMCache.Load(); c != nil {
    return *c, nil
  }
  value, err := probeAzureIMDS()
  if err == nil {
    isAzureVMCache.CompareAndSwap(nil, &value)
  }
  return value, err
}
```

translator/util/sdkutil.go
```diff
@@ -122,6 +147,16 @@ func isEKS() eksdetector.IsEKSCache {
...
 
+// isAKS dispatches to azuredetector.IsAKS dynamically (symmetric with isAzureVM).
+func isAKS() bool {
+	return azuredetector.IsAKS()
+}
+
+// isAzureVM reports Azure VM detection; an unreachable IMDS reads as not-Azure.
+func isAzureVM() bool {
+	return azuredetector.IsAzureVM().Value
+}
```
jefchien: What is the point of these wrapper functions? `isAKS` is already the same as `azuredetector.IsAKS()` so the indirection seems unnecessary.

translator/translate/otel/translate_otel.go
```diff
@@ -113,6 +116,14 @@ func translateInternal(jsonConfig interface{}, os string, validate bool) (*otelc
 		pipelines.Translators.Extensions.Set(server.NewTranslator())
 	}
 
+	// Azure VM/AKS (AKS resolves to ModeAzureVM) reaches AWS only via oidctoken->sigv4 web identity, which needs role_arn.
+	if context.CurrentContext().Mode() == config.ModeAzureVM {
+		if agent.Global_Config.Role_arn == "" {
+			return nil, errors.New("credentials.role_arn is required on Azure VM/AKS for the oidctoken web-identity credential chain")
+		}
+		pipelines.Translators.Extensions.Set(oidctoken.NewTranslator())
+	}
```
jefchien: Feel like this should be defined in a pipeline that actually uses it. Either the `opentelemetry` one or some other one. For example, if I run this on Azure with something that doesn't use the `sigv4auth` extension, the `oidctoken` extension isn't actually being used.

translator/util/sdkutil.go
movence: removed

tool/paths/paths.go
```diff
@@ -11,6 +11,10 @@ const (
 	ENV            = "env-config.json"
 	AGENT_LOG_FILE = "amazon-cloudwatch-agent.log"
 	JMXJarName     = "opentelemetry-jmx-metrics.jar"
+	// OIDCTokenFile is the file the oidctoken extension writes and sigv4auth reads for Azure web-identity auth.
+	OIDCTokenFile = ".oidc-token" // #nosec G101 -- filename, not a credential
+	// LinuxOIDCTokenPath is the Linux/Darwin OIDC token path; literal (not AgentDir-derived) to stay linux-valued when targeting linux on a Windows build host.
+	LinuxOIDCTokenPath = "/opt/aws/amazon-cloudwatch-agent/etc/" + OIDCTokenFile // #nosec G101 -- filesystem path, not a credential
```
jefchien: There shouldn't be OS specific paths in the base `tool/paths/paths.go`. The whole point of it is to use build tags to split up the values for each of the OSes we support in (`tool/paths/paths_unix.go` and `tool/paths/paths_windows.go`.). Take a look at the other paths that are defined.

translator/util/oidctokenpath.go
jefchien, line 1: Is there a reason we still need this file? What can't be handled by `tool/paths`?

translator/tocwconfig/sampleConfig/opentelemetry/host_metrics_aks_config.yaml
```diff
@@ -0,0 +1,611 @@
...
+        is_usage_data_enabled: true
+        stats:
+            operations:
+                - '*'
+            usage_flags:
+                mode: AKS
+                region_type: ACJ
+    oidctoken:
+        output_token_file: /opt/aws/amazon-cloudwatch-agent/etc/.oidc-token
+        provider: azure
```
jefchien: I don't think we need the `oidctoken` extension on AKS, but I'm fine with cleaning this up after.

translator/util/sdkutil.go
```diff
@@ -57,19 +62,37 @@ func DetectAgentMode(configuredMode string) string {
...
+	if isEKS.Err == nil && isEKS.Value {
+		return config.ModeEKS
 	}
 
-	if isEKS.Value {
-		return config.ModeEKS
+	// AKS is an explicit env signal, so it applies even if the EKS probe errored.
+	if IsAKS() {
+		return config.ModeAKS
+	}
```
jefchien: Should we even do the EKS check if AKS is explicitly set as an environment variable?

translator/translate/otel/translate_otel.go
```diff
@@ -140,6 +151,16 @@ func translateInternal(jsonConfig interface{}, os string, validate bool) (*otelc
 	return cfg, nil
 }
 
+// hasSigv4auth reports whether a sigv4auth extension is present, i.e. some pipeline will consume the oidctoken file.
+func hasSigv4auth(extensions common.ComponentTranslatorMap) bool {
+	for _, id := range extensions.Keys() {
+		if id.Type().String() == "sigv4auth" {
```
okankoAMZ: Can we use `sigv4authextension.NewFactory().Type()` instead of hardcoding?

jefchien (dismissed):

translator/translate/agent/ruleCommonCreds.go
```diff
@@ -54,6 +54,13 @@ func (c *CommonCreds) ApplyRule(input interface{}) (returnKey string, returnVal
 	return
 }
 
+// HasSharedCredentials reports whether common-config.toml supplied a profile or credentials file.
+func HasSharedCredentials() bool {
+	_, hasProfile := Global_Config.Credentials[Profile_Key]
+	_, hasFile := Global_Config.Credentials[CredentialsFile_Key]
+	return hasProfile || hasFile
+}
+
```
jefchien, line 64: nit: I would expect this defined as a method on the `Global_Config` struct.

translator/translate/otel/extension/sigv4auth/translator.go
```diff
@@ -56,9 +61,9 @@ func (t *translator) Translate(_ *confmap.Conf) (component.Config, error) {
 	cfg.IMDSRetries = retryer.GetDefaultRetryNumber()
 	if agent.Global_Config.Role_arn != "" {
 		cfg.AssumeRole = sigv4authextension.AssumeRole{ARN: agent.Global_Config.Role_arn, STSRegion: agent.Global_Config.Region}
-		// Azure VM: feed the oidctoken file into AssumeRoleWithWebIdentity for AWS creds.
-		if context.CurrentContext().Mode() == config.ModeAzureVM {
-			cfg.AssumeRole.WebIdentityTokenFile = oidctoken.OutputTokenFile(context.CurrentContext().Os())
+		// Azure VM (non-AKS) without common-config creds: feed the oidctoken IMDS file into AssumeRoleWithWebIdentity. AKS instead uses the chart's projected SA token via the SDK env chain, so it must not point at the unwritten oidctoken file.
+		if context.CurrentContext().Mode() == config.ModeAzureVM && context.CurrentContext().KubernetesMode() != config.ModeAKS && !agent.HasSharedCredentials() {
```
jefchien: nit: This seems like a shared check that we'll need in other places. Consider moving it into a shared package.

jefchien (dismissed):

lisguo:
FYI if you run into issues cleaning AKS clusters via terraform due to PodDisruptionBudget conflicts you might need to use cli to force delete nodepools :https://github.com/hashicorp/terraform-provider-azurerm/issues/30294#issuecomment-3745732745

jefchien (approved):

okankoAMZ (approved):

PR #2185, opened 2026-07-06
Support CWAGENT_ROLE env var for Container Insights mode in JSON v2 config

Paamicky:
## Description of the issue

The OTel Container Insights pipeline had no defined priority order for resolving the pipeline mode. The getMode function returned "" when no mode was configured, leaving callers to implicitly treat an empty string as "node". There was also no way for the Helm chart to set the mode without an explicit JSON config change

## Description of changes

Introduces a defined priority order for resolving the Container Insights pipeline mode:

  - JSON config field (mode) takes highest precedence
  - CWAGENT_ROLE environment variable: used by the Helm chart to indicate whether the agent runs as a DaemonSet (NODE) or Deployment (LEADER)
  - Default: "node":  if neither is set, the agent defaults to DaemonSet/node mode

The getMode function is updated to implement this priority order explicitly. The NewTranslators function is updated to use named constants (modeNode, modeCluster) instead of bare string comparisons, and a nil/missing-key guard is added to return early with zero translators if the container_insights config key is not present.

## License

## Tests

  Added unit tests covering all priority order cases:

  - TestGetMode_JSONConfig:  mode read from JSON config
  - TestGetMode_EnvVarFallback: env var used when JSON config has no mode set
  - TestGetMode_DefaultsToNode: defaults to "node" when neither config nor env var is set
  - TestGetMode_JSONOverridesEnv: JSON config wins when both are present
  - TestNewTranslators_MissingKey: returns zero translators for nil/empty config

 

## Requirements

  2. Run make lint


translator/translate/otel/pipeline/opentelemetry/containerinsights/common.go
```diff
@@ -117,16 +129,27 @@ func logsEnabled(conf *confmap.Conf) bool {
...
+//  2. Environment variable
+//  3. Default: "node" (DaemonSet)
 func getMode(conf *confmap.Conf) string {
-	if conf == nil {
-		return ""
+	if conf != nil {
+		key := common.ConfigKey(ciConfigKey, "mode")
+		if v, ok := common.GetString(conf, key); ok && v != "" {
+			return v
+		}
```
jefchien, line 136: Is there a problem with this returning any arbitrary string? How do we handle an invalid mode?

translator/translate/otel/pipeline/opentelemetry/containerinsights/common.go
```diff
@@ -17,6 +19,16 @@ import (
 const (
 	ciPrefix                  = "cw_k8s_ci_v0"
 	defaultCollectionInterval = 30 * time.Second
+
+	// envCWAgentRole is the environment variable used by the helm chart to indicate
+	// whether the agent runs as a DaemonSet (node-level) or Deployment (cluster-level).
+	envCWAgentRole = "CWAGENT_ROLE"
```
jefchien: Define this with the other environment variables that are supported in `envconfig`.

mitali-salvi: Should this variable be declared and retrieved as part of https://github.com/aws/amazon-cloudwatch-agent/blob/main/cfg/envconfig/envconfig.go ?

sky333999: These are already defined in envconfig.. so we shouldnt duplicate here again

mitali-salvi: actually seems like its already defined here - https://github.com/aws/amazon-cloudwatch-agent/blob/main/cfg/envconfig/envconfig.go#L23

translator/translate/otel/pipeline/opentelemetry/containerinsights/common.go
Paamicky: The schema enum at `translator/config/schema.json` handles this. Invalid mode values are rejected
at config parse time.
```
"mode": {
                  "description": "Pipeline mode: node (daemonset), cluster (deployment), or omit for all",
                  "type": "string",
                  "enum": ["node", "cluster"]
```

translator/translate/otel/pipeline/opentelemetry/containerinsights/common.go
Paamicky: Fixed, removed the local duplicates and switched to the existing env var constants in the `envconfig` package

translator/translate/otel/pipeline/opentelemetry/containerinsights/common_test.go
```diff
@@ -35,3 +42,100 @@ func TestEscapeDollarDigit(t *testing.T) {
...
+	})
+
+	t.Setenv(envconfig.CWAGENT_ROLE, envconfig.LEADER)
+	assert.Equal(t, modeCluster, getMode(cfg))
+
+	t.Setenv(envconfig.CWAGENT_ROLE, envconfig.NODE)
+	assert.Equal(t, modeNode, getMode(cfg))
+}
+
+func TestGetMode_DefaultsToNode(t *testing.T) {
```
jefchien, line 75: nit: Consider clearing the environment variable before the assertion to ensure it isn't set. Also in `TestNewTranslators_DefaultMode`

```suggestion
func TestGetMode_DefaultsToNode(t *testing.T) {
  t.Setenv(envconfig.CWAGENT_ROLE, "")
```

translator/translate/otel/pipeline/opentelemetry/containerinsights/common_test.go
jefchien, line 1: nit: Missing a unit test to cover the case where the role is not leader/node.

```go
func TestGetMode_UnrecognizedRole(t *testing.T) {
    cfg := confmap.NewFromStringMap(map[string]interface{}{
        "opentelemetry": map[string]interface{}{
            "collect": map[string]interface{}{
                "container_insights": map[string]interface{}{},
            },
        },
    })
    t.Setenv(envconfig.CWAGENT_ROLE, "WORKER")
    assert.Equal(t, modeNode, getMode(cfg))
}
```

jefchien (approved):

mitali-salvi (approved):

PR #2190, opened 2026-07-09
fix: prevent throttling deadlock in cloudwatchlogs output

lorespiz:
# Description of the issue
On a ThrottlingException for CreateLogStream/CreateLogGroup, log-stream creation could wedge for the entire output plugin with nothing logged.

LogThrottleRetryer.ShouldRetry performed an unconditional blocking send to a capacity-1 throttle-event channel drained by a single consumer goroutine. The shared TargetManager was bound (via sync.Once) to the first destination's client and retryer; when that destination stopped (e.g. log rotation with auto_removal), its retryer's consumer goroutine exited. The next throttle then filled the buffer and blocked the send forever, and because TargetManager.InitTarget holds a mutex across the create call, all other targets waiting on that mutex stalled as well.

# Description of changes

Two changes, either of which breaks the deadlock; both applied:

- `logthrottle.go`: make the throttle-event send non-blocking (select with a default) so a full channel or stopped consumer can never block ShouldRetry. The retry decision is unchanged.
- `cloudwatchlogs.go`: give the shared TargetManager its own dedicated retryer and client owned by the plugin, stopped only in Close(), instead of reusing the first destination's client. This decouples the TargetManager lifecycle from any single destination.

# License

# Tests
- `TestShouldRetryDoesNotBlockAfterStop`: ShouldRetry returns after the retryer is stopped (fails via timeout on the pre-fix code).
- `TestSharedRetryerLifecycle`: stopping the first destination does not prevent creating additional destinations.
- T`estInitTargetNoDeadlockUnderThrottling`: end-to-end regression driving the real SDK retry loop against an always-throttling endpoint through InitTarget with a stopped retryer consumer; deadlocks on the pre-fix code, passes with the fix.

Manually validated agent behaviour before and after the fix using a local HTTP server mocking `CreateLogGroup` / `CreateLogStream` throttled requests
# Requirements
2. Run `make lint`

-------
### Integration Tests


plugins/outputs/cloudwatchlogs/internal/pusher/target_deadlock_test.go
```diff
@@ -0,0 +1,98 @@
...
+
+	go func() {
+		wg.Wait()
+		close(done)
+	}()
+
+	select {
+	case <-done:
+		// Both InitTarget calls returned: no deadlock.
+	case <-time.After(60 * time.Second):
```
jefchien, line 95: nit: A minute is pretty long to wait for a unit test to fail. Consider reducing to 5-10 seconds.

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -151,7 +161,10 @@ func (c *CloudWatchLogs) getDest(t pusher.Target, logSrc logs.LogSrc) *cwDest {
 		if c.Concurrency > 1 {
 			c.workerPool = pusher.NewWorkerPool(c.Concurrency)
 		}
-		c.targetManager = pusher.NewTargetManager(c.Log, client)
+		// Dedicated retryer/client so the TargetManager isn't tied to the first dest.
+		c.sharedRetryer = retryer.NewLogThrottleRetryer(c.Log)
+		c.sharedClient = c.createClient(c.sharedRetryer)
```
jefchien, line 166: nit: `sharedClient` isn't used anywhere but here. No need to store a reference to it on `CloudWatchLogs`.

jefchien (approved):

plugins/outputs/cloudwatchlogs/cloudwatchlogs.go
```diff
@@ -101,6 +106,11 @@ func (c *CloudWatchLogs) Close() error {
 		c.workerPool.Stop()
 	}
 
+	// Stop the shared retryer last, after all pushers have drained.
+	if c.sharedRetryer != nil {
+		c.sharedRetryer.Stop()
```
sky333999, line 111: nit: either do this with a `sync.Once` or set `c.sharedRetryer` to nil after the Stop() here such that this cant be triggered twice even with future refactors and cause a panic.

sky333999 (approved):

PR #2207, opened 2026-07-16
Add resource_attributes support to opentelemetry section

movence:
## Description

Adds support for setting additional **resource attributes** at the top level of the `opentelemetry` section. When a customer sets `opentelemetry.resource_attributes`, a `resource` processor is prepended to the **front** of each OTEL export pipeline (metrics, logs, traces) and `upsert`s the configured key/value pairs onto every record.

### Example config
```json
{
  "opentelemetry": {
    "resource_attributes": {
      "team": "cloudwatch",
      "deployment.environment": "prod"
    },
    "collect": {
      "otlp": { "grpc_endpoint": "127.0.0.1:4317", "http_endpoint": "127.0.0.1:4318" }
    }
  }
}
```

### Generated pipeline (metrics/logs/traces)
```
processors:
  - resource/opentelemetry        # <- new, first
  - resourcedetection/opentelemetry
  - transform/identity
  - ... (logs adds routing/cleanup)
  - batch/opentelemetry_*
```

## Implementation
- New JSON schema key `opentelemetry.resource_attributes` (string map, `maxProperties: 30`, per-value `minLength: 1`/`maxLength: 1024`), mirroring the existing metrics `append_dimensions` definition.
- `common.OtelResourceAttributesKey` + a `common.GetStringMap` helper.
- Reuses the existing contrib `resourceprocessor` and its `WithAttributes` option (emits `upsert` actions) — no new dependencies.
- Prepends the processor in the three base pipeline translators, only when the key is set (existing configs are unchanged).

## Design note (precedence)
The processor is placed **first**, i.e. **before** `resourcedetection`. Since `resourcedetection` runs with `override: true`, an agent-detected attribute (e.g. `cloud.region`, `host.id`) will win over a same-named customer value. If we want the customer value to always be authoritative for detected keys, the processor should instead run **after** `resourcedetection` (but still before the transforms). Raising this for a second opinion.

The field name is centralized in a single constant, so renaming (e.g. `append_resource_attributes`) is a one-line change if preferred.

## Testing
- Unit tests: `GetStringMap`; processor is `Processors.Keys()[0]` in all three base pipelines.
- Schema validation samples: valid map passes; non-string value fails with 1 `invalid_type` error.
- `tocwconfig` golden round-trip `otlp_otel_resource_attributes_config` (json input → yaml/toml goldens). TOML is byte-identical to the baseline (OTEL-only feature); YAML diff is exactly the new processor block + its prepend to the 3 pipelines.
- `go test` + `go vet` clean across all touched packages.


translator/tocwconfig/sampleConfig/opentelemetry/otlp_otel_resource_attributes_config.json
jefchien, line 1: nit: Can we just add this to an existing one? We're starting to have to manage and regenerate too many sample configs.

translator/translate/otel/processor/resourceprocessor/translator.go
```diff
@@ -27,11 +28,23 @@ func WithAttributes(attrs map[string]string) common.TranslatorOption {
...
+}
+
 type translator struct {
 	common.NameProvider
 	common.IndexProvider
-	factory    processor.Factory
-	attributes map[string]string
+	factory      processor.Factory
+	attributes   map[string]string
+	reservedKeys []string
```
jefchien: nit: Should we just store this as a `collection.Set[string]`?

translator/translate/otel/processor/resourceprocessor/translator.go
```diff
@@ -60,13 +73,30 @@ func (t *translator) Translate(conf *confmap.Conf) (component.Config, error) {
...
+	for k := range t.attributes {
+		if strings.TrimSpace(k) == "" {
+			return nil, fmt.Errorf("%s: resource attribute keys must not be empty", t.ID())
+		}
+		if _, ok := reserved[k]; ok {
+			return nil, fmt.Errorf("%s: resource attribute key %q is reserved and cannot be overridden", t.ID(), k)
+		}
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
```
jefchien, line 96: Should we use `errors.Join` so all errors are shown at once instead of just the first?

translator/translate/otel/pipeline/opentelemetry/resource_attributes.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/aws/amazon-cloudwatch-agent/translator/translate/otel/processor/resourceprocessor"
+)
+
+// reservedResourceAttributeKeys are attributes the agent manages internally for
+// log routing; customers must not override them via resource_attributes.
+var reservedResourceAttributeKeys = []string{
+	"aws.log.group.name",
+	"aws.log.stream.name",
+	"aws.log.source",
+}
```
jefchien, line 19: This makes sense in practice, but I think we'll likely want to rename all of our internally temporary resource attributes that we use and remove. That can be done as a follow up.

translator/config/schema.json
```diff
@@ -1806,6 +1806,16 @@
     "opentelemetryDefinition": {
       "type": "object",
       "properties": {
+        "resource_attributes": {
+          "type": "object",
+          "description": "Additional resource attributes to add to every metric, log, and trace exported by the opentelemetry section. Applied via a resource processor at the front of each export pipeline. Attributes are added before resource detection, so for keys the agent auto-detects (e.g. cloud.region, host.id, cloud.provider) the detected value takes precedence over the configured one; use this for attributes the agent does not detect. Keys managed internally for log routing (aws.log.group.name, aws.log.stream.name, aws.log.source) are reserved and rejected.",
```
jefchien, line 1815: nit: I think this is too detailed for a schema description.

jefchien (dismissed):

okankoAMZ (dismissed):

jefchien (approved):

sky333999 (approved):

PR #2211, opened 2026-07-20
Add set-env ctl action to persist agent environment variables

sky333999:
# Description of the issue
Setup/enablement scripts need to persist environment variables such as `CWAGENT_ROLE_ARN` and `AWS_REGION` into `env-config.json`. Today the only way to do that is invoking the agent binary's internal `-setenv`/`-envconfig` flags directly and hardcoding the `env-config.json` path, bypassing the public ctl interface. The ctl script already wraps this exact mechanism for one specific key via `set-log-level`.

# Description of changes
Adds a generic `set-env` action to the Linux and Windows ctl scripts, following the existing `set-log-level` pattern:

- `amazon-cloudwatch-agent-ctl -a set-env -e KEY=VALUE` (Linux)
- `amazon-cloudwatch-agent-ctl.ps1 -a set-env -e KEY=VALUE` (Windows)

The action validates the KEY=VALUE format (rejects missing `=` or empty key) and delegates to the agent binary's `-setenv`, which merges the key into `env-config.json`. Keys set this way are retained across config translations per #2131, unless they are translator-managed keys.

# License

# Tests
Tested on EC2 with agent 1.300069.0b1529 plus this change, on both Amazon Linux 2023 (x86_64) and Windows Server 2022. On both platforms:

- `set-env` creates `env-config.json` when missing, merges additional keys, overwrites repeated keys
- Rejects `NOEQUALS`, `=value`, and missing `-e` with exit 1 and usage
- Custom keys survive `fetch-config` and full agent restart; translator-managed `CWAGENT_LOG_LEVEL` is wiped/regenerated from the JSON config as expected
- Agent process loads the vars at startup (verified via `I! MY_CUSTOM_VAR is set to ...` log lines) and the 30s env watcher hot-reloads changes
- `set-log-level` regression: still works end to end, including hot reload

### DIFF
--- a/packaging/dependencies/amazon-cloudwatch-agent-ctl
+++ b/packaging/dependencies/amazon-cloudwatch-agent-ctl
@@ -82,6 +82,9 @@ UsageString="
 
         -e: environment variable to set for the agent process, in KEY=VALUE format
             this parameter is used for 'set-env' only.
+            note: values that are also produced by agent config translation (such as
+            proxy, CA bundle, and log level settings) are overwritten from the agent
+            configuration on each fetch-config/append-config or agent restart.
 
 "
 
@@ -393,8 +396,12 @@ set_env() {
           ;;
      esac
 
-     runEnvConfigCommand=$("${CMDDIR}/amazon-cloudwatch-agent" -setenv "${env_kv}" -envconfig "${ENV_CONFIG}")
-     echo "${runEnvConfigCommand}" || return
+     if ! runEnvConfigCommand=$("${CMDDIR}/amazon-cloudwatch-agent" -setenv "${env_kv}" -envconfig "${ENV_CONFIG}" 2>&1); then
+          echo "${runEnvConfigCommand}" >&2
+          echo "Failed to set environment variable ${env_kv%%=*}" >&2
+          exit 1
+     fi
+     echo "${runEnvConfigCommand}"
      echo "Set ${env_kv%%=*}"
 }
 
--- a/packaging/dependencies/amazon-cloudwatch-agent-ctl
+++ b/packaging/dependencies/amazon-cloudwatch-agent-ctl
@@ -31,12 +31,13 @@ UsageString="
 
 
         usage:  amazon-cloudwatch-agent-ctl -a
-                stop|start|status|status-with-workloads|fetch-config|append-config|remove-config|set-log-level
+                stop|start|status|status-with-workloads|fetch-config|append-config|remove-config|set-log-level|set-env
                 [-m ec2|onPremise|onPrem|auto]
                 [-c default|all|ssm:<parameter-store-name>|file:<file-path>]
                 [-s]
                 [-d]
                 [-l INFO|DEBUG|WARN|ERROR|OFF]
+                [-e KEY=VALUE]
 
         e.g.
         1. apply a SSM parameter store config on EC2 instance and restart the agent afterwards:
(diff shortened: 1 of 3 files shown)


sky333999:
Addressed review findings in 742c8872:

1. **Windows value handling (critical):** `SetEnv` now invokes the agent binary directly instead of through `cmd /c`, so values containing spaces or shell metacharacters are passed as a single argument and never interpreted by cmd.exe. Verified on Windows Server 2022: `KEY=hello world with spaces` stored intact, `KEY=val & echo pwned > C:\pwned.txt` stored verbatim with no command execution, `%PATH%` not expanded.
2. **Linux error handling:** `set_env` now checks the agent binary exit code explicitly and prints a failure message with exit 1 instead of relying implicitly on `set -e`. Verified on AL2023 with an unwritable env-config.json.
3. **Usage text:** added a note (both platforms) that translation-produced values (proxy, CA bundle, log level, etc.) are overwritten on fetch-config/append-config or restart.

Additional re-test coverage on both platforms: `KEY=A=B=C` (splits on first `=` only), values with double quotes, failure path leaves env-config.json unmodified, and existing basic/validation cases regression-tested.

Deferred (pre-existing or out of scope, tracked separately): `-setenv` without `-envconfig` silently exits 0 in the agent binary; `set_log_level_all` shares the old exit-code pattern; darwin ctl lacks both set-log-level and set-env; strict key-name validation.

sky333999:
Added darwin platform parity in a8b95865: the darwin ctl now supports both `set-env` and `set-log-level` (it never received set-log-level when that action was added to Linux/Windows). Both use the same explicit exit-code handling introduced for the Linux `set_env` in 742c8872.

Tested on an EC2 mac2.metal (macOS Sonoma, arm64) with agent 1.300069.0b1529 plus this branch's binaries and ctl:
- set-env: creates/merges env-config.json, values with spaces/metacharacters stored intact, `KEY=A=B` splits on first `=` only, all invalid inputs (missing `=`, empty key, missing `-e`) exit 1
- set-log-level: sets `CWAGENT_LOG_LEVEL`, rejects invalid levels
- Stickiness: custom keys survive fetch-config; translator-managed keys regenerated
- Runtime: agent log confirms `I! MY_CUSTOM_VAR is set to ...` after start
- End-to-end: a draft update to the ManageAgent SSM document's MacOS step (separate internal change) drove `set-env` through this ctl successfully with a spaces-containing value

packaging/dependencies/amazon-cloudwatch-agent-ctl
```diff
@@ -448,6 +477,7 @@ main() {
           # helper for rpm+deb uninstallation hooks, not expected to be called manually
      preun) preun_all ;;
      set-log-level) set_log_level_all "${log_level}" ;;
```
jefchien, line 479: nit: Should we have `set-log-level` call `set_env`?

packaging/windows/amazon-cloudwatch-agent-ctl.ps1
```diff
@@ -74,6 +78,12 @@ $UsageString = @"
         -l: log level to set the agent to INFO, DEBUG, WARN, ERROR, or OFF
             this parameter is used for 'set-log-level' only.
 
+        -e: environment variable to set for the agent process, in KEY=VALUE format
```
jefchien, line 81: Did we verify that root permissions are required to run this? The `env-config.json` should be only writable by root.

jefchien (approved):

packaging/windows/amazon-cloudwatch-agent-ctl.ps1
movence: this would be a good check

movence (approved):

packaging/windows/amazon-cloudwatch-agent-ctl.ps1
sky333999: Validated this and it works as expected.

PR #2225, opened 2026-07-31
feat(dbi): Enable PostgreSQL execution plan collection

AshuSrv:
# Description of the issue

PostgreSQL execution plans are never collected for Database Insights. The translator sets the per-interval explain budget to `0`, and the receiver gates explain attempts on `explained < maxExplainEachInterval`, so that comparison is
never true. Every `db.server.top_query` event is emitted with an empty `postgresql.query_plan` attribute.

# Description of changes

Set the per-interval explain budget to `1000` in the PostgreSQL receiver translator so execution plans are collected by default.

```diff
-cfg.MaxExplainEachInterval = 0
+cfg.MaxExplainEachInterval = 1000
```

The value is kept as an explicit assignment rather than removed so it remains easy to adjust.

The expected translator outputs under `translator/tocwconfig/sampleConfig` are updated to match, and a test assertion pins the value so it cannot silently regress to disabled.

`1000` is the upstream default for this setting, so this restores upstream behaviour rather than introducing a value we would have to carry as a divergence.

The number is not the effective cap on work. The explain loop is bounded by `top_n_query`, which is `200`, so at most 200 explains can be attempted in an interval regardless of this setting.

# License

# Tests

## Unit tests

```
go test ./translator/translate/otel/receiver/postgresql/... ./translator/tocwconfig/
ok  	github.com/aws/amazon-cloudwatch-agent/translator/translate/otel/receiver/postgresql
ok  	github.com/aws/amazon-cloudwatch-agent/translator/tocwconfig
```

`make fmt`, `make fmt-sh` and `make lint` are clean.

## Verified on EC2 instance

Amazon Linux 2, PostgreSQL 14.23 with `pg_stat_statements` (`track = all`), a non-superuser role owning the workload tables, and a monitoring user granted `pg_monitor` plus `SELECT` on those tables.

Agent configuration used:

```json
{
  "agent": { "region": "us-west-2" },
  "opentelemetry": {
    "collect": {
      "database_insights": {
        "postgresql": [
          {
            "endpoint": "localhost:5432",
            "instance_name": "explain-plans-test",
            "username": "cw_monitor",
            "password_file": "/opt/databaseinsights/.pgpass",
            "logs": { "file_path": "/var/log/postgresql/postgresql-*.log" }
          }
        ]
      }
    }
  }
}
```

The generated collector configuration now carries the budget:

```yaml
top_query_collection:
    collection_interval: 1m0s
    max_explain_each_interval: 1000
    max_rows_per_query: 5000
    query_plan_cache_size: 1000
    query_plan_cache_ttl: 1h0m0s
    top_n_query: 200
```

Sample `db.server.top_query` record after the change, from
`/aws/self-managed-database-insights/postgresql/raw-events`:

```json
"attributes": {
        "db.system.name": "postgresql",
        "db.namespace": "testdb",
        "db.query.text": "SELECT COALESCE ( ( SELECT extname FROM pg_extension WHERE extname = ? ) )",
        "postgresql.calls": 1,
        "postgresql.rows": 1,
         .
         .
         .
        "postgresql.query_plan": "[{\"Plan\":{\"Node Type\":\"Result\",\"Parallel Aware\":false,\"Async Capable\":false,\"Startup Cost\":1.01,\"Total Cost\":1.02,\"Plan Rows\":1,\"Plan Width\":64,\"Plans\":[{\"Node Type\":\"Seq Scan\",\"Parent Relationship\":\"InitPlan\",\"Subplan Name\":\"InitPlan 1 (returns $0)\",\"Parallel Aware\":false,\"Async Capable\":false,\"Relation Name\":\"pg_extension\",\"Alias\":\"pg_extension\",\"Startup Cost\":0.00,\"Total Cost\":1.01,\"Plan Rows\":1,\"Plan Width\":64,\"Filter\":\"( extname = ? )\"}]}}]"
    },
```

# Requirements

- Run make fmt and make fmt-sh
- Run make lint


EduVencovsky (approved):

translator/tocwconfig/sampleConfig/opentelemetry/combined_v1_v2_eks_config.yaml
```diff
@@ -1863,7 +1863,7 @@ receivers:
             insecure_skip_verify: true
         top_query_collection:
             collection_interval: 1m0s
-            max_explain_each_interval: 0
+            max_explain_each_interval: 1000
```
JayPolanco, line 1866: Is 1000 the agreed upon value they want?

JayPolanco: Do you have an example log?

JayPolanco (approved):

zhihonl (approved):

PR #2231, opened 2026-08-04
Bound per-matrix concurrency in integration test workflows

olowosulu:
## Summary

Adds `max-parallel: 100` to the 13 matrix job definitions in the integration test
workflows, bounding how many jobs from any single matrix run concurrently.

This is defensive hardening rather than a fix for a diagnosed failure. I want to be
upfront about the evidence, because the strength of the argument varies by claim.

## What the change does

```yaml
strategy:
  fail-fast: false
  # Cap concurrent jobs per matrix to limit simultaneous STS AssumeRoleWithWebIdentity calls.
  # 100 is a conservative value, not calibrated against throttle data; account concurrency limits may cap lower.
  max-parallel: 100
  matrix: ${{ fromJson(needs.GenerateTestMatrix.outputs.ec2_linux_matrix) }}
```

Twelve sites in `test-artifacts.yml` and one in `ec2-integration-test.yml`. No other
changes, no deletions.

## The argument I am confident in

The workflow dispatches on the order of 600 jobs. `max-parallel` is scoped per matrix
definition, so today any one matrix can request as much concurrency as it has jobs, and
a single large matrix can consume the account's shared runner pool at the expense of
others.

This is not hypothetical for the largest matrices. Public run history for run
`30584015978` shows the `EC2Linux-0` matrix generating 200 jobs and `EC2Linux-1`
generating 176. Those come from `ec2-integration-test.yml`, which is called nine times
from `test-artifacts.yml`. Capping at 100 measurably constrains those matrices and gives
fairer scheduling across the rest.

For matrices that generate fewer than 100 jobs the setting is inert, so the effect is
partial by design.

## Verification

Both workflow files parse with `yaml.safe_load`, and 13 jobs carry the new key
(12 in `test-artifacts.yml`, 1 in `ec2-integration-test.yml`). The diff touches only
those two files.


.github/workflows/test-artifacts.yml
```diff
@@ -449,6 +449,9 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
+      # Cap concurrent jobs per matrix to limit simultaneous STS AssumeRoleWithWebIdentity calls.
+      # 100 is a conservative value, not calibrated against throttle data; account concurrency limits may cap lower.
```
musa-asad, line 453: nit - "account concurrency limits" reads as an AWS quota here; worth naming the GitHub runner concurrency limit explicitly?

.github/workflows/test-artifacts.yml
```diff
@@ -449,6 +449,9 @@ jobs:
...
+      max-parallel: 100
```
musa-asad, line 454: nit - this matrix resolves to a single row today, so the cap is inert here — worth saying so in the comment?

musa-asad (approved):

PR #2232, opened 2026-08-04
feat(dbi): split postgresql per-resource metrics to a dedicated 60s receiver

EduVencovsky:
# Description of the issue

Database Insights collects per-resource (table/index) metrics and server metrics from a single PostgreSQL receiver at the default 10s interval. Per-resource metrics scale with the number of relations in the monitored database — a database with thousands of tables produces thousands of readings each scrape. Collecting them every 10s is unnecessary for the DBI experience and inflates ingestion volume.

# Description of changes

**Per-resource metrics split to a dedicated receiver at 60s (configurable):**
- The DBI metrics pipeline now runs two `postgresql` receiver instances per monitored database: one for server/per-database metrics at the default 10s, and one for per-resource (table/index) metrics at 60s.
- A new per-instance config field `per_resource_collection_interval` (in seconds) lets customers override the 60s default. If omitted or zero, defaults to 60s.
- Both scopes only ever disable metrics — the set of metrics emitted is identical to before, partitioned by cadence.
- The per-resource instance disables events (query-sample / top-query) so DBLoad and TopSQL are not double-counted.
- A reflection-based helper (`forEachMetric`) walks the receiver's generated metrics config so per-resource metrics are matched by mapstructure name rather than hand-maintained field references.
- `TestTranslator_MetricSets_Disjoint` asserts the two scopes are disjoint and their union equals the default enabled set, guarding against double emission or metric loss.

**Events receiver metrics disabled (`WithMetricsDisabled` / `metricScopeNone`):**
- The events receiver now explicitly disables all metrics, since it feeds a logs pipeline only and the metrics were never collected. The generated config now reflects what the instance actually emits.

**Config shape (unchanged if field omitted):**
```json
{
  "opentelemetry": {
    "collect": {
      "database_insights": {
        "postgresql": [{
          "endpoint": "localhost:5432",
          "username": "cw_monitor",
          "password_file": "/opt/.pgpass",
          "per_resource_collection_interval": 120
        }]
      }
    }
  }
}
```

# License

# Tests

- `TestTranslator_MetricSets_Disjoint` — proves server + per-resource union equals the default set (no drop, no double emission)
- `TestTranslator_ServerMetricsOnly` / `TestTranslator_PerResourceMetricsOnly` — per-field assertions on each scope
- `TestTranslator_MetricsDisabled` — events receiver emits zero metrics, events survive
- `TestParseDbiPostgresqlInstances_PerResourceCollectionInterval` — default (60s), custom (120s), zero-falls-back-to-default
- `TestTranslator_Translate_Defaults_MetricScope` — default scope unchanged
- tocwconfig golden tests (`TestDBIConfigLinux`, `TestCombinedV1V2EC2Config`, `TestCombinedV1V2EKSConfig`) — regenerated and passing
- Local end-to-end: built agent from branch, ran against real PostgreSQL 14 (Docker), confirmed server metrics at 10s (6 samples/min) and per-resource at 60s (1 sample/min) via CloudWatch SampleCount in account 923855483676
- `al2023:otel_collect_database_insights_test` passed in CI (16m58s) on an earlier push of this branch

# Requirements

1. [x] Run `make fmt` and `make fmt-sh`
2. [x] Run `make lint`

### DIFF
--- a/translator/config/schema.json
+++ b/translator/config/schema.json
@@ -1836,6 +1836,11 @@
                       "username": { "type": "string" },
                       "password_file": { "type": "string" },
                       "ca_file": { "type": "string" },
+                      "per_resource_collection_interval": {
+                        "description": "How often per-resource (table/index) metrics are collected, in seconds. Defaults to 60 if omitted.",
+                        "type": "integer",
+                        "minimum": 0
+                      },
                       "logs": {
                         "type": "object",
                         "properties": {
--- a/translator/tocwconfig/sampleConfig/opentelemetry/combined_v1_v2_ec2_config.yaml
+++ b/translator/tocwconfig/sampleConfig/opentelemetry/combined_v1_v2_ec2_config.yaml
@@ -542,9 +542,9 @@ processors:
         ec2_instance_tag_keys:
             - AutoScalingGroupName
         ec2_metadata_tags:
+            - ImageId
             - InstanceId
             - InstanceType
-            - ImageId
         imds_retries: 1
         middleware: agenthealth/statuscode
         refresh_tags_interval: 0s
@@ -1197,59 +1197,59 @@ receivers:
         initial_delay: 1s
         metrics:
             postgresql.backends:
-                enabled: true
+                enabled: false
             postgresql.bgwriter.buffers.allocated:
-                enabled: true
+                enabled: false
             postgresql.bgwriter.buffers.writes:
-                enabled: true
+                enabled: false
             postgresql.bgwriter.checkpoint.count:
-                enabled: true
+                enabled: false
             postgresql.bgwriter.duration:
-                enabled: true
+                enabled: false
             postgresql.bgwriter.maxwritten:
-                enabled: true
+                enabled: false
             postgresql.blks_hit:
                 enabled: false
             postgresql.blks_read:
                 enabled: false
             postgresql.blocks_read:
-                enabled: true
+                enabled: false
             postgresql.commits:
-                enabled: true
+                enabled: false
             postgresql.connection.max:
-                enabled: true
+                enabled: false
             postgresql.database.count:
-                enabled: true
+                enabled: false
             postgresql.database.locks:
                 enabled: false
             postgresql.db_size:
-                enabled: true
+                enabled: false
             postgresql.deadlocks:
                 enabled: false
             postgresql.function.calls:
                 enabled: false
             postgresql.index.scans:
-                enabled: true
+                enabled: false
             postgresql.index.size:
-                enabled: true
+                enabled: false
             postgresql.operations:
-                enabled: true
+                enabled: false
             postgresql.replication.data_delay:
-                enabled: true
+                enabled: false
             postgresql.rollbacks:
-                enabled: true
+                enabled: false
             postgresql.rows:
-                enabled: true
+                enabled: false
             postgresql.sequential_scans:
                 enabled: false
             postgresql.sessions:
                 enabled: false
             postgresql.table.count:
-                enabled: true
+                enabled: false
             postgresql.table.size:
-                enabled: true
+                enabled: false
             postgresql.table.vacuum.count:
-                enabled: true
+                enabled: false
             postgresql.temp.io:
                 enabled: false
             postgresql.temp_files:
@@ -1265,11 +1265,11 @@ receivers:
             postgresql.tup_updated:
                 enabled: false
             postgresql.wal.age:
-                enabled: true
+                enabled: false
             postgresql.wal.delay:
                 enabled: false
             postgresql.wal.lag:
-                enabled: true
+                enabled: false
         passfile: sampleConfig/opentelemetry/testdata/.pgpass
         query_sample_collection:
             collection_interval: 1m0s
@@ -1326,7 +1326,7 @@ receivers:
             postgresql.blks_read:
                 enabled: false
             postgresql.blocks_read:
-                enabled: true
+                enabled: false
             postgresql.commits:
                 enabled: true
             postgresql.connection.max:
@@ -1342,17 +1342,17 @@ receivers:
             postgresql.function.calls:
                 enabled: false
             postgresql.index.scans:
-                enabled: true
+                enabled: false
             postgresql.index.size:
-                enabled: true
+                enabled: false
             postgresql.operations:
-                enabled: true
+                enabled: false
             postgresql.replication.data_delay:
                 enabled: true
             postgresql.rollbacks:
                 enabled: true
             postgresql.rows:
-                enabled: true
+                enabled: false
             postgresql.sequential_scans:
                 enabled: false
             postgresql.sessions:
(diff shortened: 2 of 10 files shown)


JayPolanco:
How would i customer set this up? Will this be in the JSON, do they have to do it int he yaml?

JayPolanco:
Should also have yaml translations for all 3 options probably so we can confirm that the behavior is what we expect per field enabled

translator/translate/otel/pipeline/opentelemetry/databaseinsights/translator.go
```diff
@@ -108,8 +108,20 @@ func (t *dbiTranslator) translateMetrics() (*common.ComponentTranslators, error)
...
+	// Split postgresql metric collection into two receiver instances so per-resource
+	// (table/index) metrics are scraped at 60s while server metrics stay at the default
+	// 10s. This caps per-resource ingestion volume for DBI GA pricing. The server
+	// instance keeps the events (query-sample / top-query) that feed DBLoad and TopSQL;
+	// the per-resource instance disables events so those are not double-counted.
+	serverRx := t.pgReceiver("metrics", postgresql.WithServerMetricsOnly())
+	perResourceRx := t.pgReceiver("metrics_perresource",
+		postgresql.WithPerResourceMetricsOnly(),
+		postgresql.WithCollectionInterval(60*time.Second),
+		postgresql.WithEventsDisabled(),
```
JayPolanco, line 122: Are these disabled by default?

translator/translate/otel/pipeline/opentelemetry/databaseinsights/translator.go
```diff
@@ -108,8 +108,20 @@ func (t *dbiTranslator) translateMetrics() (*common.ComponentTranslators, error)
...
 
+	// Split postgresql metric collection into two receiver instances so per-resource
+	// (table/index) metrics are scraped at 60s while server metrics stay at the default
+	// 10s. This caps per-resource ingestion volume for DBI GA pricing. The server
+	// instance keeps the events (query-sample / top-query) that feed DBLoad and TopSQL;
+	// the per-resource instance disables events so those are not double-counted.
+	serverRx := t.pgReceiver("metrics", postgresql.WithServerMetricsOnly())
+	perResourceRx := t.pgReceiver("metrics_perresource",
+		postgresql.WithPerResourceMetricsOnly(),
+		postgresql.WithCollectionInterval(60*time.Second),
```
zhihonl: Translator is updated when we need to translate a key in the agent JSON to expand into a template YAML config. The changes we've made in this PR doesn't give the customer a way to configure these values in the JSON.

Is this intended to be an opinionated YAML config that shouldn't let customers modify by JSON?

zhihonl:
Can the PR description be updated with more details following the PR template
```
# Description of the issue

# Description of changes

# License

# Tests

# Requirements
2. Run `make lint`

-------
### Integration Tests
```

zhihonl (dismissed):

translator/translate/otel/receiver/postgresql/translator.go
```diff
@@ -90,7 +188,38 @@ func (t *translator) Translate(_ *confmap.Conf) (component.Config, error) {
...
 
 	return cfg, nil
 }
+
+// perResourceMetricNames is the set of per-resource (table/index) metrics, keyed by
+// mapstructure name. These are emitted by the receiver's collectTables / collectIndexes
+// with a postgresql.table.name / postgresql.index.name resource attribute.
+// postgresql.table.count is intentionally excluded: it is a per-database server metric
+// recorded alongside the other server metrics.
+var perResourceMetricNames = map[string]bool{
```
JayPolanco, line 201: Since this is hardcoded, we'll have to keep in mind that upstream could always add more metrics, which would then start being published every 10s like the rest of the metrics. Might be worth having some sort of test in the future

EduVencovsky: Is this even possible? If so how?

JayPolanco: Technically it is. I think it's less likely in the sense that I don't think it's very common for them to implement new metrics as defaults. likely they'd be optional. BUt just a heads up

JayPolanco:
Did we test this on an actual instance? You can copy what our integ test for dbi does to setup and confirm that everything is working as expected

translator/translate/otel/pipeline/opentelemetry/databaseinsights/translators.go
```diff
@@ -48,12 +50,13 @@ func NewTranslators(conf *confmap.Conf) common.PipelineTranslatorMap {
...
-	PasswordFile string `mapstructure:"password_file"`
-	CAFile       string `mapstructure:"ca_file"`
-	InstanceName string `mapstructure:"instance_name"`
-	Logs         struct {
+	Endpoint                      string `mapstructure:"endpoint"`
+	Username                      string `mapstructure:"username"`
+	PasswordFile                  string `mapstructure:"password_file"`
+	CAFile                        string `mapstructure:"ca_file"`
+	InstanceName                  string `mapstructure:"instance_name"`
+	PerResourceCollectionInterval int    `mapstructure:"per_resource_collection_interval"`
```
movence, line 58: should this be added added to `schema.json`?

translator/translate/otel/pipeline/opentelemetry/databaseinsights/translator.go
```diff
@@ -108,8 +108,20 @@ func (t *dbiTranslator) translateMetrics() (*common.ComponentTranslators, error)
 	countConn := count.NewTranslator(common.DbiConnectorDbload)
 	s2mConn := signaltometrics.NewTranslator(common.DbiConnectorTopsql)
 
+	// Split postgresql metric collection into two receiver instances so per-resource
+	// (table/index) metrics are scraped at 60s while server metrics stay at the default
```
movence: the comment could be more specific that it's defaulted at 60s unless configured otherwise

translator/tocwconfig/sampleConfig/opentelemetry/combined_v1_v2_ec2_config.yaml
```diff
@@ -1405,7 +1518,7 @@ receivers:
             insecure_skip_verify: true
         top_query_collection:
             collection_interval: 1m0s
-            max_explain_each_interval: 1000
+            max_explain_each_interval: 0
```
AshuSrv: looks like this is reverting one of my changes.

JayPolanco (dismissed):

translator/config/schema.json
```diff
@@ -1836,6 +1836,11 @@
                       "username": { "type": "string" },
                       "password_file": { "type": "string" },
                       "ca_file": { "type": "string" },
+                      "per_resource_collection_interval": {
+                        "description": "How often per-resource (table/index) metrics are collected, in seconds. Defaults to 60 if omitted.",
+                        "type": "integer",
+                        "minimum": 1
```
movence: this would make configurations like `per_resource_collection_interval: 0` to be invalid which doesn't seem correct

movence (approved):

zhihonl (approved):

JayPolanco (approved):

PR #2233, opened 2026-08-05
feat(clean): reap leaked AMP integ-test workspaces

olowosulu:
## Problem

The integration tests create one AMP (Amazon Managed Service for Prometheus) workspace per `/amp` run (`terraform/ec2/linux/main.tf`, `module "amp"`, `workspace_alias = "cwagent-integ-test-<testing_id>"`) and destroy it with `terraform destroy` at the end of the run.

Workspaces left behind by cancelled/superseded or hard-killed jobs, whose Terraform state dies with the ephemeral runner, are never reclaimed: the `if: cancelled() || failure()` cleanup `terraform destroy` cannot delete a workspace whose state is gone. These orphans accumulate to the per-region **75-workspace service quota**. Once the region is at the cap, every new run fails at `CreateWorkspace`:

```
Error: creating Prometheus Workspace: ... StatusCode: 402,
ServiceQuotaExceededException: Limit exceeded. Maximum workspaces per region: 75
```

`terraform apply` then exits 1 before the Go test binary runs, so `al2:selinux_amp_test` fails with a red check unrelated to the code under review. An audit of the test region found dozens of orphaned `cwagent-integ-test-*` workspaces, the oldest over a year old, none created in the last day (leaked, not in-flight).

## Fix

Every other leaked resource type already has a cleaner under `tool/clean/clean_*` run daily by `clean-aws-resources.yml`; AMP was the gap. This adds:

- **`tool/clean/clean_amp/clean_amp.go`** (modeled on `clean_eks`): lists AMP workspaces and deletes ACTIVE ones whose alias has the `cwagent-integ-test-` prefix and were created longer ago than `clean.KeepDurationOneDay`. Supports `-dry-run`.
- a **`clean-amp-workspaces`** job in `clean-aws-resources.yml`, using the same OIDC role (`vars.TERRAFORM_AWS_ASSUME_ROLE`), region, and daily schedule as the sibling cleaners.

## Verification

- `go build` and `go vet` are clean for `clean_amp` (the only `go vet ./...` finding is pre-existing `unreachable code` in `clean_security_group`, which this PR does not touch).
- `-dry-run` against the test account matches an independent audit: it targets only `cwagent-integ-test-*` workspaces older than a day and spares non-test workspaces.
- Adding `service/amp` pulled an aligned `aws-sdk-go-v2` core bump within the isolated `tool/clean` module; all sibling cleaners still vet.

## Prerequisite

The clean OIDC role (`vars.TERRAFORM_AWS_ASSUME_ROLE`) needs `amp:ListWorkspaces` and `amp:DeleteWorkspace` for the scheduled job to run.


tool/clean/clean_amp/clean_amp.go
```diff
@@ -0,0 +1,107 @@
...
+			}
+			if dryRun {
+				log.Printf("Dry-Run: would delete workspace %s (alias %q, create-date %v)",
+					workspaceID, alias, aws.ToTime(workspace.CreatedAt))
+				continue
+			}
+			log.Printf("Try to delete workspace %s (alias %q, create-date %v)",
+				workspaceID, alias, aws.ToTime(workspace.CreatedAt))
+			if _, err := client.DeleteWorkspace(ctx, &amp.DeleteWorkspaceInput{WorkspaceId: workspace.WorkspaceId}); err != nil {
+				log.Printf("could not delete workspace %s err %v", workspaceID, err)
```
musa-asad, line 109: nit - a dropped delete error keeps the sweep green while the quota fills; would you collect these with errors.Join and return them?

tool/clean/clean_amp/clean_amp.go
```diff
@@ -0,0 +1,107 @@
...
+			workspaceID := aws.ToString(workspace.WorkspaceId)
+			alias := aws.ToString(workspace.Alias)
+
+			if workspace.Status == nil || workspace.Status.StatusCode != types.WorkspaceStatusCodeActive {
+				log.Printf("Ignoring workspace %s (alias %q) since it is not ACTIVE", workspaceID, alias)
+				continue
+			}
+			if workspace.CreatedAt == nil || !expirationDate.After(*workspace.CreatedAt) {
+				log.Printf("Ignoring workspace %s (alias %q) with create-date %v since it was created in the last %s",
+					workspaceID, alias, aws.ToTime(workspace.CreatedAt), clean.KeepDurationOneDay)
```
musa-asad: nit - KeepDurationOneDay is negative, so the skip log renders the window as -24h0m0s; would you negate it before formatting?

tool/clean/clean_amp/clean_amp.go
```diff
@@ -0,0 +1,107 @@
...
+	for paginator.HasMorePages() {
+		page, err := paginator.NextPage(ctx)
+		if err != nil {
+			return err
+		}
+		for _, workspace := range page.Workspaces {
+			workspaceID := aws.ToString(workspace.WorkspaceId)
+			alias := aws.ToString(workspace.Alias)
+
+			if workspace.Status == nil || workspace.Status.StatusCode != types.WorkspaceStatusCodeActive {
```
musa-asad: nit - CREATION_FAILED workspaces are dead but still hold quota, and this ACTIVE-only gate skips them; would you sweep that status too?

tool/clean/clean_amp/clean_amp.go
```diff
@@ -0,0 +1,107 @@
...
+	if err != nil {
+		return err
+	}
+	client := amp.NewFromConfig(defaultConfig)
+	return terminateWorkspaces(ctx, client)
+}
+
+func aliasMatchesWorkspacesToClean(alias string, workspacesToClean []string) bool {
+	for _, workspaceToClean := range workspacesToClean {
+		if strings.HasPrefix(alias, workspaceToClean) {
```
musa-asad, line 65: nit - strings.HasPrefix matches everything on an empty prefix entry, so one future blank string disables the gate; would you skip empty entries?

musa-asad (dismissed):

tool/clean/clean_amp/clean_amp.go
olowosulu: Done.

tool/clean/clean_amp/clean_amp.go
olowosulu: Done.

tool/clean/clean_amp/clean_amp.go
olowosulu: Done.

tool/clean/clean_amp/clean_amp.go
olowosulu: Done.

musa-asad (approved):

lorespiz (approved):

olowosulu:
The failing `PR Build / Build windows-2022` check is unrelated to this change. It is a
timing-sensitive unit test that flakes on slow Windows runners.

Details
- Failing test: `TestConvertOtelMetrics_Dimensions` in
  `plugins/outputs/cloudwatch/convert_otel_test.go`.
- It asserts wall-clock elapsed time under one second:
  `assert.Less(t, time.Since(*d.Timestamp), time.Second)`. The source even notes
  "Assuming unit test does not take more than 1 s."
- On this run the measured value was `1.1038402s`, so the assertion failed. The package
  took about 171s and logged repeated `PutMetricData LimitExceeded` warnings, which shows
  the runner was slow.

Why this PR did not cause it
- This PR changes only `tool/clean/clean_amp/clean_amp.go`, `tool/clean/go.mod`,
  `tool/clean/go.sum`, and `.github/workflows/clean-aws-resources.yml`. It does not touch
  `plugins/` or the root `go.mod`.
- `tool/clean` is a separate Go module, so the root "Unit Test" step never compiles any
  code from this PR.
- The same assertion family fails on unrelated branches. Example: `otel-bump-v0.150.0`
  failed the Windows job on `TestConvertOtelMetrics_NoDimensions` with
  `1.5121874s is not less than 1s`.

Request: please re-run the Windows job; it should pass on a faster runner. A durable fix
(relax or remove the wall-clock assertion) belongs in `plugins/outputs/cloudwatch` and is
out of scope for this PR.

PR #2237, opened 2026-08-06
Fix AKS cloud.resource_id derivation

jefchien:
# Description of the issue
On AKS, `cloud.resource_id` is built from `azure.resourcegroup.name`, which is the node's infrastructure resource group (`MC_<clusterRG>_<cluster>_<region>`). See https://learn.microsoft.com/en-us/azure/aks/faq#why-are-two-resource-groups-created-with-aks-
> **Why are two resource groups created with AKS?**
> AKS builds upon many Azure infrastructure resources, including virtual machine scale sets, virtual networks, and managed disks. These integrations enable you to apply many of the core capabilities of the Azure platform within the managed Kubernetes environment provided by AKS. For example, you can use most Azure VM types directly with AKS, and you can use Azure Reservations to receive discounts on those resources automatically.
>
> To enable this architecture, each AKS deployment spans two resource groups:
>
>    - You create the first resource group. This group contains only the Kubernetes service resource. The AKS resource provider automatically creates the second resource group during deployment. An example of the second resource group is MC_myResourceGroup_myAKSCluster_eastus. For information on how to specify the name of this second resource group, see the next section.
>    - The second resource group, known as the node resource group, contains all of the infrastructure resources associated with the cluster. These resources include the Kubernetes node VMs, virtual networking, and storage. By default, the node resource group has a name like MC_myResourceGroup_myAKSCluster_eastus. AKS automatically deletes the node resource group whenever you delete the cluster. Use this resource group only for resources that share the cluster's lifecycle.

The current ARM ID that gets emitted isn't valid.

# Description of changes
  - Fix AKS `cloud.resource_id` derivation: extracts the cluster resource group via replace_pattern with the cluster name baked in as a regex literal at translate time.
> [!NOTE]
> It is possible to have a resource group for the nodes that doesn't match the pattern, but that feature is in preview and at that point, there isn't anything we can do to detect the cluster resource group.
>
> https://learn.microsoft.com/en-us/azure/aks/faq#can-i-provide-my-own-name-for-the-aks-node-resource-group-
>
>> **Can I provide my own name for the AKS node resource group?**
>> By default, AKS names the node resource group MC_resourcegroupname_clustername_location, but you can provide your own name.
>>
>> To specify your own resource group name, install the [aks-preview](https://learn.microsoft.com/en-us/cli/azure/aks) Azure CLI extension version 0.3.2 or later. When you create an AKS cluster by using the az aks create command, use the --node-resource-group parameter and specify a name for the resource group. If you use an [Azure Resource Manager template](https://learn.microsoft.com/en-us/azure/templates/microsoft.containerservice/2022-09-01/managedclusters) to deploy an AKS cluster, you can define the resource group name by using the nodeResourceGroup property.

  - Add nil guards to all Format arguments in both the AKS and EKS `cloud.resource_id` statements. A nil `cloud.account.id` or `cloud.region` previously rendered <nil> in the emitted ID.
  - Resolve `{instance_id}`/`{hostname}` placeholders in `windows_events` log group and stream names to match the placeholder resolution of `opentelemetry.collect.files`.
  - Drop `role_arn` from the k8s OTel default config. On Kubernetes, the role comes from the pod's environment variables (AKS) or IRSA/Pod Identity (EKS), so a static ${CWAGENT_ROLE_ARN} in the config is unnecessary.

# License

# Tests
Added unit tests and built and deployed agent to AKS cluster with helm chart. Can see `otlp` metrics and logs being emitted with the correct `cloud.resource_id`.

# Requirements
2. Run `make lint`

-------
### Integration Tests


Paamicky:
nit: I think there is more work done than the PR title suggests, like the windows stuffs

translator/translate/otel/processor/transformprocessor/transform_identity_k8s.yaml
```diff
@@ -18,9 +18,13 @@ trace_statements:
...
+      - set(resource.attributes["cloud.resource_id"], Format("arn:aws:eks:%s:%s:cluster/%s", [resource.attributes["cloud.region"], resource.attributes["cloud.account.id"], resource.attributes["k8s.cluster.name"]])) where resource.attributes["cloud.resource_id"] == nil and resource.attributes["cloud.region"] != nil and resource.attributes["cloud.account.id"] != nil and resource.attributes["k8s.cluster.name"] != nil and resource.attributes["cloud.platform"] == "aws_eks"
+      # cloud.resource_id: AKS managed cluster resource ID. azure.resourcegroup.name is the node's RG
+      # (MC_<clusterRG>_<cluster>_<region>); extract the cluster's RG into a temp attribute (cluster
+      # name injected as a regex literal at translate time), build the ID, then drop the temp.
+      - set(resource.attributes["_tmp.azure.resourcegroup.name"], resource.attributes["azure.resourcegroup.name"]) where resource.attributes["cloud.platform"] == "azure_aks" and resource.attributes["azure.resourcegroup.name"] != nil
+      - replace_pattern(resource.attributes["_tmp.azure.resourcegroup.name"], "^MC_(.+)_%CLUSTER_NAME%_[^_]+$", "$$$1") where resource.attributes["cloud.platform"] == "azure_aks"
```
Paamicky, line 26: Should we add a test to document the non-matching behavior?

mitali-salvi, line 26: ~Won't the `$$$$` add an extra `$` after substitution ?~

Nvm, I misread the code. the substitution has the correct number of `$` chars

mitali-salvi (approved):

Paamicky (approved):

PR #2247, opened 2026-08-13
fix(azure): pin subscription from resource ID; fix az tsv parsing and WI detection

EduVencovsky:
# Description of the issue

The Azure setup script (`scripts/azure/setup.sh`) relies on whichever Azure subscription Cloud Shell has as its ambient default. When the target resource lives in a different subscription (common in multi-subscription tenants), every `az` call resolves against the wrong subscription and ARM returns a misleading `AuthorizationFailed` one step downstream of the real problem — making it look like an RBAC gap rather than a subscription mismatch.

Additionally, two pre-existing bugs were found during live testing:
1. Three `az --query "[a, b]" -o tsv` calls emit one value per line (not one tab-separated row), so the `cut -f1/-f2` parsing silently broke. Worst case: the VM identity check read `None\nLinux`, tested truthy, and **silently skipped `az vm identity assign`** while reporting success.
2. The AKS detection keyed only on `oidcIssuerProfile.enabled`. Current AKS enables the OIDC issuer at creation by default (workload identity stays off), so the script logged "OIDC issuer and workload identity enabled" and never ran `--enable-workload-identity` — the webhook that injects OIDC tokens into pods would never be installed, so the agent could not authenticate to AWS via `AssumeRoleWithWebIdentity`.

# Description of changes

**Subscription pinning:**
- New inputs `CWAGENT_AZURE_SUBSCRIPTION` and `CWAGENT_AZURE_RESOURCE_ID`, documented in the header and `usage()`.
- `parse_resource_id()` splits a full ARM resource ID into subscription, resource group, and name, handling both `resourceGroups` and `resourcegroups` spellings.
- Rejects a resource ID whose type does not match `CWAGENT_PLATFORM` (e.g. a VM ID with `azure_aks`) before any mutating call.
- `check_prerequisites()` runs `az account set --subscription` before the existing `az account show`, dying with an actionable message that points at `az account list` and the cross-tenant `az login --tenant` case.

**tsv parsing fix:**
- Changed three JMESPath queries from flat `[a, b]` to nested `[[a, b]]`, which makes `-o tsv` emit a single tab-separated row instead of one value per line.
- Sites: `az account show` (subscription/tenant), `az vm show` (identity + OS type), `az aks show` (OIDC + issuer URL).

**AKS workload identity detection:**
- Now requires both `oidcIssuerProfile.enabled` AND `securityProfile.workloadIdentity.enabled` to be true before skipping `az aks update`. Added `WI_ENABLED` field (issuer URL moves to field 3).

# License

# Tests

**Local (mock az shim):**
- Resource ID happy path: verified call ordering (`az account set` → `account show` → `vm show -g <rg> -n <name>`).
- Wrong subscription: dies immediately with the actionable message, rc=1, downstream calls never fire.
- No subscription set: zero `account set` calls, existing behavior unchanged.
- Platform/type mismatch: VM ID + `azure_aks` rejected; AKS ID + `azure_vm` rejected.
- Invalid/malformed resource IDs: both casings of `resourceGroups` parsed correctly; garbage IDs rejected.

**Live (real Azure subscription, throwaway VM + AKS cluster):**
- VM path: `az account set` pinned the correct subscription, `SystemAssigned` identity actually assigned (verified `principalId` on the VM), idempotent rerun detected existing identity and skipped assign.
- AKS path: OIDC issuer + workload identity both enabled (confirmed both flags `True`; `azure-wi-webhook-controller-manager` pods present), idempotent rerun completed in <2s without touching the cluster.
- Failure paths: bogus subscription dies with the new early error; platform mismatch rejected before any `az` call.

**Linting:**
- `shellcheck` passes with zero warnings.
- `shfmt -i 5` (repo standard) passes with no diff.
- `sh -n` syntax validation passes.

-------
### Integration Tests

### DIFF
--- a/scripts/azure/setup.sh
+++ b/scripts/azure/setup.sh
@@ -42,6 +42,12 @@
 #     CWAGENT_AWS_ROLE_ARN          IAM role ARN, install runs too when set
 #     CWAGENT_AWS_REGION            AWS region telemetry is sent to (required
 #                                   for install)
+#     CWAGENT_AZURE_SUBSCRIPTION    Subscription ID or name the resource lives
+#                                   in (Cloud Shell's ambient default is used
+#                                   when unset)
+#     CWAGENT_AZURE_RESOURCE_ID     Full ARM resource ID of the VM / AKS
+#                                   cluster; supplies the subscription,
+#                                   resource group, and name in one value
 #     CWAGENT_AZURE_RESOURCE_GROUP  Resource group
 #     CWAGENT_EMIT_ENV              When set (1/true/yes/on), print eval-able KEY='value'
 #                                   lines on stdout and route all logging to stderr
@@ -55,6 +61,8 @@ set -eu
 PLATFORM="${CWAGENT_PLATFORM:-}"
 ROLE_ARN="${CWAGENT_AWS_ROLE_ARN:-}"
 REGION="${CWAGENT_AWS_REGION:-}"
+SUBSCRIPTION="${CWAGENT_AZURE_SUBSCRIPTION:-}"
+RESOURCE_ID="${CWAGENT_AZURE_RESOURCE_ID:-}"
 RESOURCE_GROUP="${CWAGENT_AZURE_RESOURCE_GROUP:-}"
 VM_NAME="${CWAGENT_AZURE_VM_NAME:-}"
 CLUSTER_NAME="${CWAGENT_K8S_CLUSTER_NAME:-}"
@@ -127,6 +135,8 @@ Environment variables:
     CWAGENT_PLATFORM              azure_vm | azure_aks
     CWAGENT_AWS_ROLE_ARN          IAM role ARN; when set, install runs too
     CWAGENT_AWS_REGION            AWS region (required for install)
+    CWAGENT_AZURE_SUBSCRIPTION    Subscription ID or name of the resource
+    CWAGENT_AZURE_RESOURCE_ID     Full ARM resource ID (subscription + group + name)
     CWAGENT_AZURE_RESOURCE_GROUP  Resource group
     CWAGENT_EMIT_ENV              Print eval-able KEY='value' lines on stdout
   azure_vm:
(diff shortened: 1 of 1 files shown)


scripts/azure/setup.sh
```diff
@@ -258,19 +277,90 @@ interactive_setup() {
...
      fi
+     # Pin every subsequent az call to the resource's subscription; without this
+     # they all resolve against Cloud Shell's ambient default, and ARM answers
+     # for a resource in another subscription with an AuthorizationFailed that
+     # is indistinguishable from a genuine RBAC gap.
+     if [ -n "${SUBSCRIPTION}" ]; then
+          az_err=$(az account set --subscription "${SUBSCRIPTION}" 2>&1) ||
+               die "cannot select subscription ${SUBSCRIPTION}: ${az_err}
+Check 'az account list'; if it is in another tenant, run: az login --tenant <tenant-id>"
+          log "Subscription: ${SUBSCRIPTION}"
```
jefchien, line 338: Are there any concerns with changing the underlying subscription for the AZ CLI? There's no guarantee that this will be run in a disposable Cloud Shell. Instead of setting it, would it make sense to update all AZ calls to include `--subscription "${SUBSCRIPTION}"` or are we also trying to make sure we have access to the subscription first?

Could we do something like
```
_original_sub=$(az account show --query id -o tsv 2>/dev/null || true)
if [ -n "${_original_sub}" ]; then
    trap 'az account set --subscription "${_original_sub}" 2>/dev/null' EXIT
fi
```
to restore the previous subscription on exit?

scripts/azure/setup.sh
```diff
@@ -433,18 +527,22 @@ setup_azure_aks() {
...
+          --query "[[oidcIssuerProfile.enabled, securityProfile.workloadIdentity.enabled, oidcIssuerProfile.issuerUrl]]" -o tsv 2>/dev/null || true)
      OIDC_ENABLED=$(printf '%s' "${AKS_INFO}" | cut -f1)
-     OIDC_ISSUER=$(printf '%s' "${AKS_INFO}" | cut -f2)
+     WI_ENABLED=$(printf '%s' "${AKS_INFO}" | cut -f2)
+     OIDC_ISSUER=$(printf '%s' "${AKS_INFO}" | cut -f3)
 
      # az -o tsv renders a JSON boolean via Python str(), i.e. "True"/"False",
      # so match case-insensitively rather than against a lowercase "true".
-     if [ "$(printf '%s' "${OIDC_ENABLED}" | tr '[:upper:]' '[:lower:]')" = "true" ]; then
+     if [ "$(printf '%s%s' "${OIDC_ENABLED}" "${WI_ENABLED}" | tr '[:upper:]' '[:lower:]')" = "truetrue" ]; then
```
jefchien, line 545: nit: This works, but is a bit illegible. Consider splitting into two checks.
```
_oidc_lower=$(printf '%s' "${OIDC_ENABLED}" | tr '[:upper:]' '[:lower:]')
_wi_lower=$(printf '%s' "${WI_ENABLED}" | tr '[:upper:]' '[:lower:]')

if [ "${_oidc_lower}" = "true" ] && [ "${_wi_lower}" = "true" ]; then
    log "OIDC issuer and workload identity enabled"
else
    ...
fi
```

scripts/azure/setup.sh
jefchien, line 639: Should skip the `interactive_setup` if `RESOURCE_ID` is provided because it already contains everything and there's no need for the script to prompt for `RESOURCE_GROUP` or `VM_NAME` at that point.

jefchien (approved):

PR #2249, opened 2026-08-13
Fix CWAgent container insights

Paamicky:
### Description of the issue

While bringing OTEL Container Insights (OCI) into the CloudWatch Agent, several pieces of functionality present in the Helm implementation were dropped or omitted, leaving the agent-generated config diverged from the chart. Most notably: OCI logs were never delivered to CloudWatch Logs because the `log-group/log-stream `metadata was missing; cluster-scoped metrics (ap`iserver, kube-state-metrics`) carried the scraper node's identity. Attributes limits of <150 guaranteed to the Zeus endpoint was broken as well as Batching.

### Description of changes
Realigns the agent's OTEL Container Insights pipelines with the Helm implementation and fixes the dropped/omitted functionality:

-  OCI logs delivery:  route CI application/host logs through the shared logs pipeline, which supplies the log-group/log-stream metadata that was missing (logs now reach CloudWatch Logs), including host-log gating (host logs omit
  `service.name/deployment.environment.name`).
- schema_url reordering: move the schema_url clear into the shared metrics + logs pipelines (after the re-stamping
  `resourcedetection/opentelemetry`) and remove the redundant per-CI-pipeline clears.
- Cluster-metric identity suppression: for cluster-scoped metrics (apiserver, kube-state-metrics), suppress the scraper node's `host.*, cloud.availability_zone, and ec2.tag.*` so series stay stable across scraper reschedules.
- Attribute limit: add `awsattributelimit` (150-cap) on the shared metrics pipeline to stay under the CloudWatch OTLP backend limit.
- Batching: set the shared metrics batch to `1000/10s` which affects all OTLP metrics
- resourcedetection: unify the CI resourcedetection definition (detectors: [eks, env, ec2], timeout: 2s) across pipelines.

 All changes are gated to EKS OTEL Container Insights; ECS, host metrics, and non-CI pipelines are unaffected. Goldens regenerated.

### License

### Tests

  - Unit: go build ./... and go test ./translator/... pass (goldens updated); gofmt clean.
  - End-to-end on EKS (cluster mcommey-otelci-final, agent built from this branch, deployed via operator):
  - OCI logs now reach CloudWatch with the correct and verified attributes
  - Cluster metrics (apiserver/KSM via Query Studio): no host.*, no cloud.availability_zone, no ec2.tag.*; node metrics retain node identity.
  - Metrics export cleanly

### Requirements

  2. Run make lint


.github/workflows/e2e-test.yml
```diff
@@ -8,7 +8,7 @@ env:
   ECR_INTEGRATION_TEST_REPO: "cwagent-integration-test"
   CWA_GITHUB_TEST_REPO_NAME: "aws/amazon-cloudwatch-agent-test"
   CWA_GITHUB_TEST_REPO_URL: "https://github.com/aws/amazon-cloudwatch-agent-test.git"
-  CWA_GITHUB_TEST_REPO_BRANCH: "main"
+  CWA_GITHUB_TEST_REPO_BRANCH: "mcommey/containerinsights-test"
```
sky333999, line 11: Dont forget to revert this prior to merging.

translator/tocwconfig/tocwconfig_test.go
```diff
@@ -339,38 +339,48 @@ func TestHostMetricsConfig(t *testing.T) {
...
+
+			var actual interface{}
+			cfg, err := otel.TranslateWithoutValidation(input, context.CurrentContext().Os())
+			require.NoError(t, err)
+			yamlConfig, err := mapstructure.Marshal(cfg)
+			require.NoError(t, err)
+			yamlStr := toyamlconfig.ToYamlConfig(yamlConfig)
+			require.NoError(t, yaml.Unmarshal([]byte(yamlStr), &actual))
+
+			opt := cmpopts.SortSlices(func(x, y interface{}) bool {
```
sky333999, line 378: Should we be comparing with sorted order here? For the processors atleast, seems like the order does matter.

translator/translate/otel/pipeline/opentelemetry/translator_logs_test.go
```diff
@@ -16,6 +16,26 @@ import (
 	"github.com/aws/amazon-cloudwatch-agent/translator/translate/otel/common"
 )
 
+func TestBaseLogsTranslatorActivatedByContainerInsightsLogs(t *testing.T) {
+	agent.Global_Config.Region = "us-west-2"
```
sky333999, line 21: we should clean this up in a t.Cleanup()

translator/translate/otel/pipeline/opentelemetry/translator_metrics.go
```diff
@@ -81,7 +82,30 @@ func (t *baseMetricsTranslator) Translate(conf *confmap.Conf) (*common.Component
...
 	}
 	processors.Set(transformprocessor.NewTranslatorWithName(common.Identity))
-	processors.Set(batchprocessor.NewTranslator(common.WithName("opentelemetry_metrics"), batchprocessor.WithSendBatchSize(common.MaxMetricsPerRequest), batchprocessor.WithSendBatchMaxSize(common.MaxMetricsPerRequest), batchprocessor.WithTimeout(common.BatchTimeout)))
+	// Cluster-scoped CI metrics must not carry the scraper node's identity.
+	// resourcedetection re-adds host.*/AZ post-fanout, so strip them for marked
+	// records, then drop the marker. Gated on container_insights.
+	if conf != nil && conf.IsSet(common.ConfigKey(common.OpenTelemetryKey, common.CollectKey, common.OtelContainerInsightsKey)) {
+		processors.Set(transformprocessor.NewTranslatorWithName("cluster_host_suppress",
+			transformprocessor.WithMetricResourceStatements([]string{
+				`delete_key(resource.attributes, "host.id") where resource.attributes["_tmp.cluster_scoped"] == true`,
```
sky333999: Is it better to do regex based for host.* here as well? Are there any more we remove with that route if we do that?

sky333999:
```
  delete_matching_keys(resource.attributes, "^host\\.")        where _tmp.cluster_scoped == true
  delete_matching_keys(resource.attributes, "^ec2\\.tag\\.")   where _tmp.cluster_scoped == true
  delete_key(resource.attributes, "cloud.availability_zone")   where _tmp.cluster_scoped == true
```

translator/translate/otel/pipeline/opentelemetry/containerinsights/filelog_node.yaml
```diff
@@ -16,30 +16,11 @@ processors:
       statements:
       - set(attributes["k8s.cluster.name"], "{{.ClusterName}}")
       - set(attributes["k8s.node.name"], "{{.NodeName}}")
+      - set(attributes["aws.cloudwatch.log_type"], "host")
```
sky333999, line 19: This looks like its also being used as a marker, similar to the cluster_scoped one.. can we use the same naming convention for both?
So maybe call this one _tmp.<> too?

.github/workflows/e2e-test.yml
Paamicky: ack

translator/translate/otel/pipeline/opentelemetry/containerinsights/filelog_node.yaml
Paamicky: `aws.cloudwatch.log_type` isn't a transient marker, it is consumed downstream (... != "host" gates service.name/deployment.environment enrichment) across the CI, prometheus, and `host_metrics` pipelines, then `delete_key` after use. Renaming to `_tmp.` would break those conditions, so let's keep the name.

translator/translate/otel/pipeline/opentelemetry/translator_metrics.go
Paamicky: Regex catches all host and ec2 tag attributes, done.

translator/translate/otel/pipeline/opentelemetry/translator_logs_test.go
Paamicky: done.

translator/tocwconfig/tocwconfig_test.go
Paamicky: Processor order is already deterministic (`TranslatorMap` is `container/list-backed`), so runtime is fine. `SortSlices` just means the test wouldn't catch an accidental reorder (`processors.Set(...) calls in translator_metrics.go`). We can add ordered comparison for `pipelines.*.processors` as a follow-up to harden it

translator/translate/otel/pipeline/opentelemetry/containerinsights/filelog_node.yaml
sky333999: Right I mean its a marker thats set, and read again to make some decisions before finally being deleted.. similar to the **cluster_scoped** one?
We'd need to apply the rename to the parts where its being read to make a decision as well and the delete.

Basically - how is this diff from cluster_scoped?

.github/workflows/e2e-test.yml
movence: it's better to revert this now and keep it in a stash if needed

sky333999 (approved):

movence (approved):

PR #2253, opened 2026-08-24
fix(state): Make Enqueue slow path non-blocking

lorespiz:
# Description of the issue

`(*rangeManager).Enqueue` in `internal/state/manager.go` uses blocking channel operations in its queue-full slow path, so it can block the calling goroutine. The offset range queue is a lossy ring by design (it drops the oldest entry when full), so `Enqueue` should never block its caller.

# Description of changes

- `Enqueue` now uses only non-blocking channel operations, in a loop: try the send; if the queue is full, try a non-blocking drop of the oldest entry; repeat until the item is enqueued. The fast path and the drop-oldest semantics are unchanged, and the caller's item is always enqueued.
- Added drop accounting: an atomic counter exposed via a `Dropped()` accessor, and drop logging rate-limited to at most one line per 30 seconds instead of one line per drop.

# Tests

- New regression test `Enqueue/NonBlocking`: a producer performs 500k Enqueues against a capacity-1 queue while a competing receiver drains it in a tight loop, with a 30 second deadline. It fails against the previous implementation and passes against this change. It also asserts the drop counter advanced.
- Existing `internal/state` tests pass unchanged, including `Enqueue/QueueOverflow`, which guards the drop-oldest semantics.
- Full package run passes with `-race`.

# License

### DIFF
--- a/internal/state/manager.go
+++ b/internal/state/manager.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"log"
 	"os"
+	"sync/atomic"
 	"time"
 )
 
@@ -15,6 +16,8 @@ const (
 	defaultSaveInterval = 100 * time.Millisecond
 	// defaultQueueSize is the default capacity of the offset queue
 	defaultQueueSize = 2000
+	// dropLogInterval is the minimum duration between offset range drop logs.
+	dropLogInterval = 30 * time.Second
 )
 
 type rangeManager struct {
@@ -24,6 +27,10 @@ type rangeManager struct {
 	saveInterval      time.Duration
 	maxPersistedItems int
 	replaceTrackerCh  chan RangeTracker
+
+	// dropped counts offset ranges discarded because the queue was full. Accessed atomically.
+	dropped     atomic.Uint64
+	lastDropLog atomic.Int64
 }
 
 // FileRangeManager is a state manager that handles the Range.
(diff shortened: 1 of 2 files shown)


internal/state/manager.go
```diff
@@ -53,16 +60,46 @@ func (m *rangeManager) ID() string {
...
+	}
+	// Only the CAS winner logs.
+	if !m.lastDropLog.CompareAndSwap(last, now.UnixNano()) {
+		return
+	}
+	log.Printf("D! Offset range queue is full for %s. Dropped offset range: %s (%d dropped in total)", m.stateFilePath, item, total)
+}
+
+// Dropped returns the number of offset ranges discarded because the queue was full.
+func (m *rangeManager) Dropped() uint64 {
```
sky333999, line 99: nit: Does this need to be exported?

internal/state/manager.go
```diff
@@ -53,16 +60,46 @@ func (m *rangeManager) ID() string {
 }
 
 // Enqueue the Range. Will drop the oldest in the queue if full.
+//
+// Both operations are non-blocking by design: Run drains the queue concurrently, so a queue
+// observed full at the send may already be empty at the receive. Dropping the oldest keeps the
+// freshest offsets, costing duplicate records after a restart rather than lost ones. Each
```
sky333999, line 66: > "Each iteration either enqueues the item or frees a slot, so the loop always terminates"

nit: Like we saw in this case the triggered this bug, it could get drained between, so neither of these could fire - maybe tweak this line slightly to say the next iter's send should always succeed.

sky333999 (approved):

internal/state/manager.go
lorespiz: fair point - it doesn't really need to be exported. 
Following up on it

internal/state/manager.go
lorespiz: Following up on it

EduVencovsky (approved):

PR #2263, opened 2026-08-31
Fix neuron per core multi runtime

spanaik:
# Description of the issue

On a node running more than one Neuron runtime, roughly half of every per-core Neuron
metric is silently replaced by zero. A NeuronCore pinned at 75% utilization reports 0.

`transform/cw_k8s_ci_v0_neuron_promote` runs in `context: datapoint` but writes
`resource.attributes`. Resource attributes are per-`ResourceMetrics`, so with N runtimes
in one RM the statement executes N times and the last write wins. It then deletes
`runtime_tag` from the datapoint — the only attribute distinguishing one runtime's real
reading for a core from another runtime's zero for that same core — so two datapoints
collapse onto one identity with different values.

A debug exporter on the pipeline shows the agent emitting **four** datapoints with only
**two** distinct attribute sets:

```
aws.neuron.core=0  Value: 75.154815   <- real
aws.neuron.core=1  Value:  0.000000
aws.neuron.core=0  Value:  0.000000   <- same identity, shadows the real one
aws.neuron.core=1  Value: 75.363968
```

Impact on a multi-runtime Neuron node: a saturated core renders as idle, a
low-utilization alarm fires falsely, and an idle-reclaim decision could reclaim a busy
core. Confirmed on `neuroncore_utilization_ratio` and all five
`neuroncore_memory_usage_*` families.

Single-runtime nodes are unaffected, which is why this went unnoticed — it needs ≥2
Neuron runtimes on one node to appear.

**`neuron-monitor` is not at fault.** Scraping its own `/metrics` endpoint at the same
instant shows it emitting the full `(core × runtime_tag)` cross-product correctly, with
only the matching pairs non-zero.

# Description of changes

Aggregation is not needed — `groupbyattrs` already does the right thing, it just wasn't
given the runtime dimension.

1. Add `runtime_tag` to the `groupbyattrs/cw_k8s_ci_v0_neuron` keys, so each runtime
   lands in its own `ResourceMetrics` and the datapoints can no longer collide.
2. Reduce `transform/cw_k8s_ci_v0_neuron_promote` to a `context: resource` rename of
   `runtime_tag` → `aws.neuron.runtime.tag`.

The six `set`/`delete_key` statements for `k8s.pod.name`, `k8s.namespace.name` and
`k8s.container.name` are removed as redundant: those keys are already in the
`groupbyattrs` keys, and `groupbyattrs` *moves* its grouping keys — see
`processor/groupbyattrsprocessor/processor.go`:

```go
// These attributes are going to be moved from datapoint to resource level,
// so we can delete those on the datapoint
deleteAttributes(requiredAttributes, attributes)
```

so the promote was re-doing work already done one processor earlier, and its
datapoint-context write was the vehicle for the defect.

Net effect: 11 lines changed in `neuron.yaml` plus the four regenerated golden configs.

### Trade-off

This publishes the full cross-product, so series count grows multiplicatively with
runtime count — 2 → 4 on a 2-core/2-runtime node; a 32-core `trn1` with 4 runtimes goes
32 → 128 series per per-core metric. **This is the cardinality `neuron-monitor` already
emits**; the previous behaviour reduced it only by discarding data.

The alternative is to aggregate the runtime dimension away in-agent (`max` per core),
which keeps the count flat but loses per-runtime attribution. Rejected here because the
PromQL/OTLP surface *can* carry the runtime dimension, and a runtime legitimately spans
multiple cores, so "which runtime is using this core" is a question this surface should
be able to answer.

Reviewers may want to weigh in on whether the cardinality growth is acceptable for large
`trn1` topologies. It has not been measured beyond a 2-core node.

### Scope

**This does not affect the EMF / Container Insights path.** `neuron.yaml`'s pipeline
exports to the `forward/opentelemetry` connector, which is consumed only by
`metrics/opentelemetry` → `otlphttp/metrics`. In the combined v1+v2 config the sole
pipeline touching an EMF exporter is `logs/emf_logs` (receiver `udplog`, one batch
processor, no Neuron processors). The EMF Neuron path is a separate pipeline —
`awscontainerinsightreceiver` → `gpuattributes` → `awsemf/containerinsights` — and shares
no component instance with this one. Its own `max`-per-core reduction in
`gpuattributes/internal/awsneuron_metric_modifier.go` is unchanged and remains correct
for that surface, which does not publish `runtime_tag` as a dimension.

**This alone does not fix EKS add-on clusters.** The `amazon-cloudwatch-observability`
Helm chart carries an independent copy of this same pipeline in
`templates/linux/_otel-container-insights-config.tpl` with the identical defect, and that
is the copy the EKS add-on renders into `AmazonCloudWatchAgent.spec.otelConfig`. This PR
fixes the translator path (config built from `cwagentconfig.json`); the chart fix is
aws-observability/helm-charts#365. Both are needed.

# License

By submitting this pull request, I confirm that you can use, modify, copy, and
redistribute this contribution, under the terms of your choice.

# Tests

**New unit tests** —
`translator/translate/otel/pipeline/opentelemetry/containerinsights/neuron_multiruntime_test.go`.
Instantiates the real `groupbyattrs` and `transform` processors from the `//go:embed`ed
`neuron.yaml`, chains them, and feeds synthetic metrics in the multi-runtime shape
(2 cores × 2 runtimes, non-owning pairs at zero, all four sharing one pod). No cluster or
Neuron hardware required; runs in ~25ms.

| Test | Asserts |
| --- | --- |
| `KeepsEveryCorePerRuntime` | all four `(core, tag, value)` triples survive — both 75.x readings |
| `SeparatesRuntimesIntoResources` | one `ResourceMetrics` per runtime |
| `PromotesPodIdentity` | pod/namespace/container on the resource, absent from datapoints |
| `PreFixConfigLosesData` | negative control — the pre-fix config collapses to one tag and two datapoints per core |

Config is read from the embedded template rather than restated, so a revert fails the
test. **Verified by mutation:** removing `- runtime_tag` from the `groupbyattrs` keys in
`neuron.yaml` fails three of the four.

The negative control matters because it keeps the other three honest — if the collapse
ever stops reproducing, `PreFixConfigLosesData` starts passing and tells us the
assertions have gone vacuous.

**Existing suites** — `go test ./translator/tocwconfig/...` and
`./translator/translate/otel/pipeline/opentelemetry/containerinsights/...` pass. The four
golden configs in this diff are the regenerated snapshots.

**Live cluster** — verified on a 2-core `inf2.xlarge` running two Neuron runtimes:

- Before: 2 series on the PromQL surface, core 0 reading 0 while pinned at ~75%.
- After: 4 correctly-labelled series, core 0 → 75.15 under its own runtime tag, core 1 →
  75.23 under the other.
- Reverting the config on the live agent reproduced the loss (2 series, core 0 back to 0);
  re-applying restored it. So the fix is doing the work, not a caching artifact.
- No attribute regression: 8 existing Neuron integration tests / 26 subtests pass,
  including `TestNeuronPodName`, `TestNeuronNamespace`, `TestNeuronDeviceAttributes` and
  `TestNeuronNoPromotedDatapointKeys` — the last of which specifically asserts
  pod/namespace/container are *not* left on the datapoint, covering the six statements
  removed here.

**Integration test coverage** for the multi-runtime case is a companion change to
`amazon-cloudwatch-agent-test` (adds a second co-located burn deployment so a node runs
two runtimes, plus 8 assertions on the resulting series shape). Worth noting that the
pre-existing integration tests could *not* catch this defect:
`TestNeuronRuntimeTagInResourceScope` passes on the broken code because the collapsed
resource still carries one tag, and `TestNeuronNoDuplicateSeries` passes because the
collision happens in-agent — the surface shows too *few* series rather than duplicated
ones. Cardinality is the signal, not duplication.

https://github.com/aws/amazon-cloudwatch-agent-test/pull/749

# Requirements

- [x] `make fmt` and `make fmt-sh` — no changes produced
- [x] `make lint` — 0 issues (license and import-order checks pass)

### DIFF
--- a/translator/translate/otel/pipeline/opentelemetry/containerinsights/neuron_multiruntime_test.go
+++ b/translator/translate/otel/pipeline/opentelemetry/containerinsights/neuron_multiruntime_test.go
@@ -1,22 +1,19 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: MIT
 
-// Per-core Neuron data loss on multi-runtime nodes.
+// Per-core Neuron attribution on multi-runtime nodes.
 //
-// The shipped neuron.yaml is run for real here -- groupbyattrs then the promote
-// transform, built from the embedded template -- against synthetic metrics in the
-// shape neuron-monitor emits when two Neuron runtimes share a node: every core
-// reported by every runtime, with the non-owning pairs at zero.
+// Invariant: when several Neuron runtimes share a node, every (core, runtime)
+// reading must reach the exporter under its own identity. neuron-monitor reports
+// every core from every runtime -- the owning pair real, the rest zero -- so
+// runtime_tag is the only attribute separating a core's real reading from another
+// runtime's zero for that same core. Stop separating them and two datapoints share
+// an identity, the last write wins, and for one core that is a plausible-looking 0.
 //
-// The defect: the promote ran in `context: datapoint` while writing
-// resource.attributes. Resource attributes are per-ResourceMetrics, so the
-// statement ran once per datapoint and the last write won; it then deleted
-// runtime_tag from the datapoint, the only attribute separating one runtime's real
-// reading for a core from another runtime's zero for that same core. Two datapoints
-// collapsed onto one identity and a core pinned at 75% reported 0.
-//
-// TestNeuronMultiRuntimePreFixConfigLosesData pins that failure mode, so the
-// assertions below are known to discriminate rather than merely pass.
+// These run the shipped neuron.yaml -- groupbyattrs then the promote transform,
+// rendered from the embedded template rather than restated, so a config change that
+// breaks the invariant fails here. TestNeuronMultiRuntimeCollapsedConfigLosesData
+// runs a config that violates it, proving the assertions discriminate.
 
 package containerinsights
 
@@ -54,8 +51,6 @@ const (
 )
 
 // renderedNeuronProcessors returns the processors block of the shipped neuron.yaml.
-// Read from the embedded template rather than restated, so reverting the config
-// fails this test.
 func renderedNeuronProcessors(t *testing.T) map[string]any {
 	t.Helper()
 	tmpl, err := template.New("neuron").Parse(neuronYAML)
@@ -115,8 +110,8 @@ func (componenttestNopHost) GetExtensions() map[component.ID]component.Component
 
 // multiRuntimeMetrics is what neuron-monitor emits for 2 cores x 2 runtimes on one
 // node: the owning pair real, the other zero. Pod identity is identical across all
-// four, which is the collapse-prone case -- one pod holding both cores, as measured
-// on solstice-gpu-test.
+// four -- one pod holding both cores, which occurs in practice and is the
+// collapse-prone case, since pod identity alone cannot separate the datapoints.
 func multiRuntimeMetrics() pmetric.Metrics {
 	md := pmetric.NewMetrics()
 	sm := md.ResourceMetrics().AppendEmpty().ScopeMetrics().AppendEmpty()
@@ -176,8 +171,8 @@ func consume(t *testing.T, chain processor.Metrics, md pmetric.Metrics) {
 	require.NoError(t, chain.ConsumeMetrics(context.Background(), md))
 }
 
-// TestNeuronMultiRuntimeKeepsEveryCorePerRuntime runs the SHIPPED config and
-// requires all four (core, runtime) readings to survive with their real values.
+// TestNeuronMultiRuntimeKeepsEveryCorePerRuntime requires all four (core, runtime)
+// readings to survive the shipped config with their real values.
 func TestNeuronMultiRuntimeKeepsEveryCorePerRuntime(t *testing.T) {
 	processors := renderedNeuronProcessors(t)
 	chain, sink := buildChain(t,
@@ -192,7 +187,7 @@ func TestNeuronMultiRuntimeKeepsEveryCorePerRuntime(t *testing.T) {
 		"core=1 tag=burn-core1 value=75.3",
 	}, observed(t, sink),
 		"both busy cores must survive with their own runtime tag; a missing 75.x "+
-			"reading is the per-core data-loss defect")
+			"reading means one runtime's datapoint was shadowed by another's zero")
 }
 
 // TestNeuronMultiRuntimeSeparatesRuntimesIntoResources pins the mechanism: one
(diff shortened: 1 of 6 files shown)


translator/translate/otel/pipeline/opentelemetry/containerinsights/neuron_multiruntime_test.go
```diff
@@ -0,0 +1,336 @@
...
+//
+// The defect: the promote ran in `context: datapoint` while writing
+// resource.attributes. Resource attributes are per-ResourceMetrics, so the
+// statement ran once per datapoint and the last write won; it then deleted
+// runtime_tag from the datapoint, the only attribute separating one runtime's real
+// reading for a core from another runtime's zero for that same core. Two datapoints
+// collapsed onto one identity and a core pinned at 75% reported 0.
+//
+// TestNeuronMultiRuntimePreFixConfigLosesData pins that failure mode, so the
+// assertions below are known to discriminate rather than merely pass.
```
miconeilaws: This reads like an artifact from the fix you've done rather than future-understandable preface for a set of tests... LEts refactor it to focus the intention of the tests rather than the historical bug

translator/translate/otel/pipeline/opentelemetry/containerinsights/neuron_multiruntime_test.go
```diff
@@ -0,0 +1,336 @@
...
+	testNamespace = "default"
+	testContainer = "burn"
+	tagA          = "burn-core0"
+	tagB          = "burn-core1"
+)
+
+// renderedNeuronProcessors returns the processors block of the shipped neuron.yaml.
+// Read from the embedded template rather than restated, so reverting the config
+// fails this test.
+func renderedNeuronProcessors(t *testing.T) map[string]any {
```
miconeilaws, line 54: Hmm I wonder if this is really a substitute for the integration tests? Do we have other tests like this? Or what are you trying to cover here that you can't cover with an integration test?

translator/translate/otel/pipeline/opentelemetry/containerinsights/neuron_multiruntime_test.go
spanaik: Fair — rewritten to lead with the invariant the tests protect rather than the history of how it broke. The context: datapoint/last-write-wins mechanism moved onto the negative control, which is the one test genuinely about that configuration shape, so it's local to where it's load-bearing.

Also renamed PreFixConfigLosesData → CollapsedConfigLosesData, since "pre-fix" ages the same way the comment did, and gave PromotesPodIdentity the same treatment — it described itself as guarding "the six promote statements removed by the fix" instead of stating that pod identity must land on the resource and not stay on the datapoint.

translator/translate/otel/pipeline/opentelemetry/containerinsights/neuron_multiruntime_test.go
spanaik: Not a substitute — complementary, and there is precedent for the pattern: translator/translate/otel/processor/metricsdecorator/translator_unix_test.go:52 (TestMetricDecoration) does the same thing with the same imports, and its own comment describes it as "using a test TransformProcessor yaml configuration, starting the processor and having it consume test metrics." One deliberate difference: it loads a testdata yaml, whereas this renders the shipped neuron.yaml off the //go:embeded template, so reverting the real config fails this test.
 
Three things the integration test structurally can't do:

1. The negative control can only live here. CollapsedConfigLosesData runs the broken config to prove the assertions discriminate rather than merely pass. You can't deploy a deliberately broken agent in CI.

2. The integration test can't distinguish a collapse from a single-runtime node. I hit this when I reverted the agent on a live cluster to check the integration tests actually fail: the collapse destroys runtime_tag itself, so exactly one tag survives per node — structurally identical to a node that genuinely runs one runtime. The integration test has to bail out with a "check both of these causes" message rather than assert. Here the input is fixed, so it asserts an exact four-series expectation.

3. It fails on a config revert, in this repo's build. The integration test only ever sees whatever config happens to be deployed, and only if the cluster fixture is right.

The integration side is in a companion PR against amazon-cloudwatch-agent-test — that's the layer that proves it on real inf2 silicon.

PR #2274, opened 2026-09-04
feat(agent): enable service.profilesSupport feature gate by default

royrajan13:
# Description of the issue

Issue: https://github.com/aws/amazon-cloudwatch-agent/issues/2273

Configuring a profiles pipeline makes the agent fail to start, because the
collector's profiles signal is gated behind `service.profilesSupport` (alpha, off
by default in collector v0.124.0):

    pipeline "profiles": profiling signal support is at alpha level, gated under the "service.profilesSupport" feature gate

The agent builds the collector's arguments itself and only ever passes
`--config=`, so an operator has no supported way to enable the gate.

# Description of changes

The agent now appends `--feature-gates=+service.profilesSupport` to the arguments
it builds for the collector command, so profiles pipelines start without the
operator needing to know the gate exists.

The flag is only appended after checking the collector's feature-gate registry
for an enableable gate with that ID. `featuregate.Registry.Set` returns an error
for an unknown gate ID, which surfaces as a flag-parse failure and a fatal
startup error. Without the guard, the moment the gate graduates and is deleted
upstream, a routine collector dependency bump would produce an agent that
refuses to start. With it, graduation is a no-op.

The guard also treats `StageDeprecated` as not enableable, because `Set` rejects
enabling a deprecated gate with the same class of error, and deprecated is the
pre-deletion state of a graduating gate. A `StageStable` gate that is still
registered remains enableable, so the flag keeps being passed until the gate is
removed from the registry entirely.

New file `cmd/amazon-cloudwatch-agent/featuregates.go` holds
`collectorFeatureGateArgs` and the registry check. `amazon-cloudwatch-agent.go`
gains one line at the existing `cmd.SetArgs` site. `featuregate` moves from the
indirect to the direct require block in `go.mod` (no version change).

# License

By submitting this pull request, I confirm that you can use, modify, copy, and
redistribute this contribution, under the terms of your choice.

# Tests

New unit tests in `cmd/amazon-cloudwatch-agent/featuregates_test.go` cover all
four registry states: gate registered at alpha (flag passed), gate at stable
(flag still passed), gate absent (no flag, no error), and gate deprecated (no
flag, no error).

Rather than string-comparing the produced arguments, each case parses them
through a real `featuregate` registry flag set, and a separate test asserts the
gate actually flips to enabled -- so the tests validate the flag name and `+`
syntax against the library rather than against a hardcoded expectation.

- `go build ./...` -- pass
[...]

Issue #2273, opened 2026-09-04
Enable service.profilesSupport feature gate so profiles pipelines can start


royrajan13:
**Is your feature request related to a problem? Please describe.**

Configuring a profiles pipeline makes the agent fail to start:

    pipeline "profiles": profiling signal support is at alpha level, gated under the "service.profilesSupport" feature gate

Profiles are gated behind `service.profilesSupport`, which is alpha and off by
default in the collector the agent embeds. The agent builds the collector's
arguments itself and only passes `--config=`, so there is no supported way for an
operator to enable the gate. Profiles are unreachable from the agent today.

**Describe the solution you'd like**

Have the agent enable `service.profilesSupport` by default when it starts the
collector, so profiles pipelines work without operators needing to know the gate
exists.

**Describe alternatives you've considered**

Exposing a general `--feature-gates` passthrough in the agent config, or an
environment-variable opt-in. Both leave profiles off by default and require
operators to know which gate to name.

**Additional context**

I have a change ready and would like to submit a PR for this.

PR #2277, opened 2026-09-07
feat(agent): enable service.profilesSupport feature gate by default

OisinMcMenamin:
# Description of the issue

Issue: #2273
Supersedes #2274 (same change; re-raised from this fork so the work can continue to be driven to completion — the original author's commit and attribution are preserved).

Configuring a profiles pipeline makes the agent fail to start, because the
collector's profiles signal is gated behind `service.profilesSupport` (alpha, off
by default in collector v0.124.0):

```
pipeline "profiles": profiling signal support is at alpha level, gated under the "service.profilesSupport" feature gate
```

The agent builds the collector's arguments itself and only ever passes
`--config=`, so an operator has no supported way to enable the gate.

# Description of changes

The agent now appends `--feature-gates=+service.profilesSupport` to the arguments
it builds for the collector command, so profiles pipelines start without the
operator needing to know the gate exists.

The flag is only appended after checking the collector's feature-gate registry
for an enableable gate with that ID. `featuregate.Registry.Set` returns an error
for an unknown gate ID, which surfaces as a flag-parse failure and a fatal
startup error. Without the guard, the moment the gate graduates and is deleted
upstream, a routine collector dependency bump would produce an agent that
refuses to start. With it, graduation is a no-op.

The guard also treats `StageDeprecated` as not enableable, because `Set` rejects
enabling a deprecated gate with the same class of error, and deprecated is the
pre-deletion state of a graduating gate. A `StageStable` gate that is still
registered remains enableable, so the flag keeps being passed until the gate is
removed from the registry entirely.

New file `cmd/amazon-cloudwatch-agent/featuregates.go` holds
`collectorFeatureGateArgs` and the registry check. `amazon-cloudwatch-agent.go`
gains one line at the existing `cmd.SetArgs` site. `featuregate` moves from the
indirect to the direct require block in `go.mod` (no version change).

# License

By submitting this pull request, I confirm that you can use, modify, copy, and
redistribute this contribution, under the terms of your choice.

# Tests

New unit tests in `cmd/amazon-cloudwatch-agent/featuregates_test.go` cover all
four registry states: gate registered at alpha (flag passed), gate at stable
(flag still passed), gate absent (no flag, no error), and gate deprecated (no
flag, no error).

Rather than string-comparing the produced arguments, each case parses them
through a real `featuregate` registry flag set, and a separate test asserts the
gate actually flips to enabled — so the tests validate the flag name and `+`
syntax against the library rather than against a hardcoded expectation.

- `go build ./...` — pass
- `CGO_ENABLED=0 go test ./...` — pass (full suite)
- `make fmt` / `make fmt-sh` — clean

# Checklist

- [x] Commits are squashed into a logical, reviewable set (one commit for a single change)
- [x] Commits and PR description comply with Amazon internal guidelines
- [x] `make` passes locally (build, unit tests, lint) — `make lint` reports pre-existing repo-wide findings that also fail on unmodified `main`; none are in the new file or on the two modified lines
- [ ] All GitHub Actions checks on the PR are passing (pending CI approval/run)
- [x] Integration test evidence: N/A — the change is behaviorally inert for all currently-expressible configurations: no profiles pipeline can be produced by the config translator yet, so the only observable difference is the extra `--feature-gates` argument, which is fully covered by unit tests. End-to-end profiles evidence lands with the config-translator follow-up PR.
- [x] New or updated integration test coverage: N/A — same reason; integration coverage arrives with the translator PR that makes profiles configurable.
- [x] New functionality has unit tests; bug fixes have a reproducing test
- [x] Config translation changes include updated golden files — N/A, no translator changes
- [x] Breaking or customer-visible changes are called out in the PR description — none; existing configurations see no behavior change

### DIFF
--- a/cmd/amazon-cloudwatch-agent/amazon-cloudwatch-agent.go
+++ b/cmd/amazon-cloudwatch-agent/amazon-cloudwatch-agent.go
@@ -28,6 +28,7 @@ import (
 	"github.com/influxdata/wlog"
 	"github.com/kardianos/service"
 	"go.opentelemetry.io/collector/component"
+	"go.opentelemetry.io/collector/featuregate"
 	"go.opentelemetry.io/collector/otelcol"
 	"go.uber.org/zap"
 
@@ -377,6 +378,7 @@ func runAgent(ctx context.Context,
 	for _, uri := range otelConfigs {
 		e = append(e, "--config="+uri)
 	}
+	e = append(e, collectorFeatureGateArgs(featuregate.GlobalRegistry())...)
 	cmd.SetArgs(e)
 	return cmd.Execute()
 }
--- a/cmd/amazon-cloudwatch-agent/featuregates.go
+++ b/cmd/amazon-cloudwatch-agent/featuregates.go
@@ -1,0 +1,54 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: MIT
+
+package main
+
+import (
+	"log"
+	"strings"
+
+	"go.opentelemetry.io/collector/featuregate"
+)
+
+// profilesSupportGateID is the collector gate that unlocks profiles pipelines. The
+// agent enables it so operators get profiles support without having to pass
+// --feature-gates themselves.
+const profilesSupportGateID = "service.profilesSupport"
+
+// defaultEnabledFeatureGates are the collector feature gates the agent turns on at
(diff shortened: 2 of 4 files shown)


OisinMcMenamin:
Closing in favor of #2274 — the original PR for this change by the commit's author, where review will continue. Same commit, same content; this was raised while we sorted out ownership of the work.

PR #2280, opened 2026-09-08
feat(scripts): make tenant id more prominent and improve logging in setup scripts

EduVencovsky:
# Description of the issue

The onboarding setup scripts (`scripts/`) had two usability gaps in the Azure-to-AWS flow:

1. **Wrong-account failure mode.** The Azure-side setup takes the full IAM role ARN, but `aws/setup.sh` only took a role *name* and trusted whatever account the shell was authenticated against. Running it in the wrong account's credential context created the role and trust there, and the failure only surfaced later — silently — on the assume-role side.

2. **Buried tenant ID.** In the Azure VM flow, the tenant ID (the value the AWS trust step needs) was logged mid-output, after the minutes-long identity assignment, making it easy to miss and hard to copy. A mistyped VM name was also only discovered mid-run, via a raw CLI error from the identity step.

# Description of changes

**`aws/setup.sh`** now also accepts `CWAGENT_AWS_ROLE_ARN`. When set, it validates the ARN format, fails fast when the ARN's partition or account does not match the shell's credentials (`sts get-caller-identity`), and derives the role name from the ARN. A `CWAGENT_AWS_ROLE_NAME` that disagrees with the ARN is rejected; an agreeing or default-valued one is allowed (the dispatcher always exports the default). The interactive prompt now asks for an ARN, still accepting a plain role name or empty input for the default. A warning is printed if the resulting role differs from the provided ARN (e.g. an existing EC2 instance-profile role takes precedence). Validation runs before any IAM mutation.

**`azure/setup.sh`** (azure_vm flow): the `az vm show` probe now fails fast with a clear message when the VM cannot be found (az's own stderr is preserved so permission or throttling errors are not misread as not-found); the tenant ID is emitted before the slow identity and install steps, so the user can start the AWS side in parallel; and it is repeated at the very end for easy copying — in the install-success message ("Agent installed on <VM> in tenant <id>") and above the next-steps text of the identity-only flow.

**`README.md`**: `CWAGENT_AWS_ROLE_ARN` documented in the common environment variable table.

No breaking changes: all existing inputs and flows behave as before when the new variable is unset.

# License

# Tests
1. Ran `make fmt` and `make fmt-sh`
2. Ran `make lint`

### DIFF
--- a/scripts/aws/setup.sh
+++ b/scripts/aws/setup.sh
@@ -318,11 +318,11 @@ check_prerequisites() {
      AWS_ARN=$(printf '%s' "${AWS_IDENTITY}" | cut -f2)
      AWS_ALIAS=$(aws iam list-account-aliases --query 'AccountAliases[0]' --output text 2>/dev/null || true)
      if [ -n "${AWS_ALIAS}" ] && [ "${AWS_ALIAS}" != "None" ]; then
-          log "AWS account: '${AWS_ACCOUNT}' ('${AWS_ALIAS}')"
+          log "AWS account: ${AWS_ACCOUNT} (${AWS_ALIAS})"
      else
-          log "AWS account: '${AWS_ACCOUNT}'"
+          log "AWS account: ${AWS_ACCOUNT}"
      fi
-     log "AWS identity: '${AWS_ARN}'"
+     log "AWS identity: ${AWS_ARN}"
 }
 
 # =============================================================================
@@ -589,7 +589,7 @@ trust_aws_ec2() {
 
           section "Using existing instance profile..."
           log "Instance profile '${PROFILE_NAME}' attached to '${INSTANCE_ID}'"
-          log "Role: '${EXISTING_ROLE}'"
+          log "Role: ${EXISTING_ROLE}"
           if [ -n "${ROLE_ARN_INPUT}" ] && [ "${EXISTING_ROLE}" != "${ROLE_NAME}" ]; then
                logwarn "using this role instead of the provided '${ROLE_ARN_INPUT}'"
           fi
@@ -865,7 +865,7 @@ run_via_ssm() {
                --command-id "${COMMAND_ID}" \
                --instance-id "${INSTANCE_ID}" \
                --region "${REGION}" --query 'StandardErrorContent' --output text >&2
-          die "SSM command finished with status: '${SSM_STATUS_DETAIL}'"
+          die "SSM command finished with status: ${SSM_STATUS_DETAIL}"
      fi
 }
 
@@ -1053,7 +1053,7 @@ main() {
      ROLE_ARN=$(aws iam get-role \
           --role-name "${ROLE_NAME}" \
           --query Role.Arn --output text)
-     log "Role ARN: '${ROLE_ARN}'"
+     log "Role ARN: ${ROLE_ARN}"
 
      ensure_transaction_search
 
--- a/scripts/aws/setup.sh
+++ b/scripts/aws/setup.sh
@@ -293,11 +293,9 @@ interactive_setup() {
      # Skip when a role was already identified through the environment. An ARN
      # is preferred (it pins the account), but a plain name is still accepted:
      # this script often creates the role, so a first run may have no ARN yet.
-     if [ -z "${ROLE_ARN_INPUT}" ] && [ -z "${CWAGENT_AWS_ROLE_NAME:-}" ]; then
-          ask "IAM role ARN or name [create/use ${ROLE_NAME} in this account]:"
+     if [ -z "${CWAGENT_AWS_ROLE_ARN:-}" ] && [ -z "${CWAGENT_AWS_ROLE_NAME:-}" ]; then
+          ask "IAM role ARN or name [create/use '${ROLE_NAME}' in this account]:"
           read -r role_input || die "no input for IAM role"
-          # A plain if, not a trailing && list: an empty answer keeps the
-          # default role name and must not fail the function under set -e.
           if [ -n "${role_input}" ]; then
                case "${role_input}" in
                arn:*) ROLE_ARN_INPUT="${role_input}" ;;
@@ -320,11 +318,11 @@ check_prerequisites() {
      AWS_ARN=$(printf '%s' "${AWS_IDENTITY}" | cut -f2)
      AWS_ALIAS=$(aws iam list-account-aliases --query 'AccountAliases[0]' --output text 2>/dev/null || true)
      if [ -n "${AWS_ALIAS}" ] && [ "${AWS_ALIAS}" != "None" ]; then
-          log "AWS account: ${AWS_ACCOUNT} (${AWS_ALIAS})"
+          log "AWS account: '${AWS_ACCOUNT}' ('${AWS_ALIAS}')"
      else
-          log "AWS account: ${AWS_ACCOUNT}"
+          log "AWS account: '${AWS_ACCOUNT}'"
      fi
-     log "AWS identity: ${AWS_ARN}"
+     log "AWS identity: '${AWS_ARN}'"
 }
 
 # =============================================================================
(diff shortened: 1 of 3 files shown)


MarcoDN (dismissed):

scripts/aws/setup.sh
```diff
@@ -281,11 +290,21 @@ interactive_setup() {
...
      esac
-     # `prompt` returns early on a non-empty current value, and ROLE_NAME always
-     # has one (the built-in default), so ask directly with the default filled in.
-     ask "IAM role name [${ROLE_NAME}]:"
-     read -r role_input || die "no input for IAM role name"
-     [ -n "${role_input}" ] && ROLE_NAME="${role_input}"
+     # Skip when a role was already identified through the environment. An ARN
+     # is preferred (it pins the account), but a plain name is still accepted:
+     # this script often creates the role, so a first run may have no ARN yet.
+     if [ -z "${ROLE_ARN_INPUT}" ] && [ -z "${CWAGENT_AWS_ROLE_NAME:-}" ]; then
```
jefchien: nit: Why do we check `ROLE_ARN_INPUT` and not the `CWAGENT_AWS_ROLE_ARN` itself since we're looking at `CWAGENT_AWS_ROLE_NAME` directly?

scripts/aws/setup.sh
```diff
@@ -281,11 +290,21 @@ interactive_setup() {
...
+          ask "IAM role ARN or name [create/use ${ROLE_NAME} in this account]:"
+          read -r role_input || die "no input for IAM role"
+          # A plain if, not a trailing && list: an empty answer keeps the
+          # default role name and must not fail the function under set -e.
```
jefchien: nit: Since this script is meant to be downloaded by the user, consider trimming the comments to explanations that would be meaningful to that reader. This comment explains why we changed it, but isn't too useful when reading the script fresh IMO.

scripts/aws/setup.sh
```diff
@@ -64,6 +69,7 @@ UPDATE_INSTANCE_ROLE="${CWAGENT_AWS_UPDATE_INSTANCE_ROLE:-}"
 ENABLE_TXN_SEARCH="${CWAGENT_AWS_ENABLE_TRANSACTION_SEARCH:-}"
 CLUSTER_NAME="${CWAGENT_K8S_CLUSTER_NAME:-}"
 ROLE_NAME="${CWAGENT_AWS_ROLE_NAME:-CloudWatchAgentServerRole}"
+ROLE_ARN_INPUT="${CWAGENT_AWS_ROLE_ARN:-}"
```
jefchien: super minor nit: `ROLE_ARN` matches convention of other variables.

scripts/aws/setup.sh
```diff
@@ -308,6 +327,62 @@ check_prerequisites() {
...
+     arn_account=$(printf '%s' "${ROLE_ARN_INPUT}" | cut -d: -f5)
+
+     case "${arn_account}" in
+     [0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9][0-9]) ;;
+     *) die "invalid IAM role ARN: ${ROLE_ARN_INPUT} (the account ID must be 12 digits)" ;;
+     esac
+
+     caller_partition=$(printf '%s' "${AWS_ARN}" | cut -d: -f2)
+     if [ "${arn_partition}" != "${caller_partition}" ]; then
+          die "the role ARN is in partition ${arn_partition} but this shell's credentials are in ${caller_partition}. Rerun with credentials for the account the role lives in"
```
jefchien: nit: The mismatch in this case is the partition not just the account.

```suggestion
          die "the role ARN is in partition ${arn_partition} but this shell's credentials are in ${caller_partition}. Rerun with credentials for the ${arn_partition} partition"
```

scripts/aws/setup.sh
```diff
@@ -970,6 +1046,12 @@ main() {
      ROLE_ARN=$(aws iam get-role \
           --role-name "${ROLE_NAME}" \
           --query Role.Arn --output text)
+     # The role in play can drift from a provided ARN: aws_ec2 swaps in an
+     # existing instance-profile role, and a pathed ARN for a role that did not
+     # exist yet is created at the default path. Flag either, neutrally.
+     if [ -n "${ROLE_ARN_INPUT}" ] && [ "${ROLE_ARN}" != "${ROLE_ARN_INPUT}" ]; then
+          logwarn "using ${ROLE_ARN} instead of the provided ${ROLE_ARN_INPUT}"
+     fi
```
jefchien: It seems like the only place that can reassign `ROLE_NAME` is the `trust_aws_ec2`. Consider moving the warning there and compare it to the existing instance-profile role.

https://github.com/aws/amazon-cloudwatch-agent/blob/7109a156c388f611621399893ac7bb98883cad5e/scripts/aws/setup.sh#L511-L514

If we also add support for pathed IAM roles, then there shouldn't be any further drift and this check can be removed.

scripts/aws/setup.sh
```diff
@@ -308,6 +327,62 @@ check_prerequisites() {
...
+
+     case "${ROLE_ARN_INPUT}" in
+     arn:*:iam::*:role/?*) ;;
+     *) die "invalid IAM role ARN: ${ROLE_ARN_INPUT} (expected arn:<partition>:iam::<account-id>:role/<name>)" ;;
+     esac
+
+     # The role name is the last path segment, so an ARN with a path
+     # (arn:...:role/path/Name) resolves to the plain name the IAM CLI wants.
+     arn_role_name="${ROLE_ARN_INPUT##*/}"
+     [ -n "${arn_role_name}" ] || die "invalid IAM role ARN: ${ROLE_ARN_INPUT} (empty role name)"
```
jefchien, line 349: If we support `ROLE_ARN` as an input, do we need to be able to parse the path and support `--path` in the `create-role` call in `ensure_iam_role`?

scripts/azure/setup.sh
```diff
@@ -446,18 +453,29 @@ run_via_az() {
...
+     # Probing first also verifies the VM exists, so a wrong name or group
+     # fails here, before any setup output or slow identity work. az's own
+     # stderr is not suppressed: a permissions or throttling failure should
+     # show its real error above the message here, not read as not-found.
      VM_INFO=$(az_scoped vm show \
           --resource-group "${RESOURCE_GROUP}" \
           --name "${VM_NAME}" \
-          --query "[[identity.principalId, storageProfile.osDisk.osType]]" -o tsv 2>/dev/null || true)
+          --query "[[identity.principalId, storageProfile.osDisk.osType]]" -o tsv) ||
+          die "cannot find VM ${VM_NAME} in resource group ${RESOURCE_GROUP} (check the names against the subscription logged above)"
```
jefchien: nit: Consider delimiting the variables with single quotes.

```suggestion
          die "cannot find VM '${VM_NAME}' in resource group '${RESOURCE_GROUP}' (check the names against the subscription logged above)"
```

Also, it could be a permission issue as well, so "cannot find" could be inaccurate.

scripts/azure/setup.sh
```diff
@@ -486,14 +501,14 @@ setup_azure_vm() {
...
-               log "Agent installed on ${VM_NAME}"
+               log "Agent installed on ${VM_NAME} in tenant ${TENANT_ID}"
                return
           fi
      else
           install_env="CWAGENT_CLOUD=azure CWAGENT_AWS_ROLE_ARN=${ROLE_ARN} CWAGENT_AWS_REGION=${REGION}"
           if INSTALL_CMD=$(linux_install_cmd "${install_env}"); then
                run_via_az "RunShellScript" "${INSTALL_CMD}"
-               log "Agent installed on ${VM_NAME}"
+               log "Agent installed on ${VM_NAME} in tenant ${TENANT_ID}"
```
jefchien: nit: "in tenant ${TENANT_ID}" is a bit odd to add to this message. If anything, the subscription would make more sense. If we're trying to do a better job of surfacing the tenant ID for use in the `aws/setup.sh`, consider repeating
```sh
log "Tenant ID (for the AWS setup): ${TENANT_ID}"
```
before the install returns instead.

scripts/azure/setup.sh
EduVencovsky: So we must put this everywhere in the script, not only here, right? 

There are many other places that don't follow this pattern

jefchien: Yeah would be nice to be consistent.

jefchien: Maybe that's too much now. Might make copying variables a bit more annoying.

scripts/aws/setup.sh
```diff
@@ -970,7 +1053,7 @@ main() {
      ROLE_ARN=$(aws iam get-role \
           --role-name "${ROLE_NAME}" \
           --query Role.Arn --output text)
-     log "Role ARN: ${ROLE_ARN}"
+     log "Role ARN: '${ROLE_ARN}'"
```
MarcoDN: On the line of previous comments, Role ARN: '${ROLE_ARN}' and OIDC issuer: '${OIDC_ISSUER}' are also of the format <field>: <value> outputs users may copy into the other script, so the quotes get in the way there too.

Same for AWS account: / AWS identity: / Azure subscription:

scripts/azure/setup.sh
```diff
@@ -556,7 +571,7 @@ setup_azure_aks() {
                --query "oidcIssuerProfile.issuerUrl" -o tsv)
      fi
 
-     log "OIDC issuer: ${OIDC_ISSUER}"
+     log "OIDC issuer: '${OIDC_ISSUER}'"
```
jefchien: nit: Consider unquoting to make it easier to copy for further steps (used in `aws/setup.sh`).

jefchien (dismissed):

jefchien (approved):

MarcoDN (approved):

PR #2285, opened 2026-09-14
Add new cache key for nightly build

Paamicky:
### Description of the issue
The `nightly-build`, `soak-test`, and `build-test-artifacts` workflows share one cache key: `cached_container_<sha>`. Nightly doesn't push the `linux-amd64-<sha>` arch image that the other tests need, but its cache entry makes soak and integration skip their own push, so they later fail at `docker manifest create` with `no such manifest. Nightly needsits own cache key.

### Description of changes
Add an optional `CacheKeyOverride` input to `test-build-docker.yml` and have `nightly-build.yml` pass `nightly_${{ github.sha }}`. Nightly now caches under its own namespace and no longer causes `soak/integration `to skip pushing their arch images. Soak and build-test-artifacts are unchanged.

### License
  By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms
  of your choice.

### Tests
  YAML validated locally. Verified nightly resolves to `cached_container_nightly_<sha>` and the untouched workflows still
  resolve to `cached_container_<sha>`.

### Requirements
  2. Run `make lint`

  -------
### Integration Tests


bhavya76 (approved):

.github/workflows/test-build-docker.yml
```diff
@@ -82,7 +92,7 @@ jobs:
         id: cached_container
         uses: actions/cache@6f8efc29b200d32929f49075959781ed54ec270c # v3.5.0
         with:
-          key: "cached_container_${{ github.sha }}"
+          key: "cached_container_${{ inputs.CacheKeyOverride != '' && inputs.CacheKeyOverride || github.sha }}"
```
jefchien, line 95: It seems like we should apply the same combination of fixes (`ArchTagSuffix` + `CacheKeyOverride`) to the Windows build/image as well. The original PR (https://github.com/aws/amazon-cloudwatch-agent/pull/2270) only addressed the linux tag clashing.

jefchien (approved):

.github/workflows/test-build-docker.yml
Paamicky: Sure, I will add in a separate PR.

